PR #18, opened 2016-02-21
Feature: Added RAT Table access to GM tools.

Kurios:
From this you can both roll for rat, and if GM mode is enabled, add
directly into your campaign hanger.

Additionally, GMTools is no longer held in memory when not shown and
some utility functions have been added to the UnitTableData for going
from string to int for the weight, unit, and quality.

image


Akjosch:
It's a useful idea, but ... would be nice if you checked my improvement/fix suggestions first.

Akjosch:
I'm closing this for now. It needs re-writing anyway due to conflicts, and doesn't look like the original creator works on it further anyway.

PR #28, opened 2016-03-07
Updates to the Utilities class

Akjosch:
This PR doesn't change any functionality, just updates the `Utilities` class to be more useful for future development:

`combineString(collection, separator)` is changed so that it can deal with *any* `Collection` class, not just `Vector`. It's also speeded up considerably by using a `StringBuilder` explicitly instead of creating and destroying one implicitly once per loop.

`getRandomItem(collections)` returns a random (unweighted) item from a collection of them. It returns *null* if the collection itself is *null*, has no items, or the picked item is *null*. There is a special optimised version of the method for `List` classes. Where trivially possible, other classes needing it have been changed accordingly.

`lerp(fromValue, toValue, fraction)` is a simple linear interpolation method, in two variants (for integer and double values). Right now unused, it can be useful to create random values from a range (`Utilities.lerp(from, to, rnd.nextDouble)`)  or advancing some values over time until they reach the desired result and similar effects. I'm sure some of the AtB calculations can be vastly simplified with it, but I didn't dive into them yet to know which ones for sure.

The `nonNull(...)` methods return the first non-*null* value in the list of arguments, unless they are all *null*. It's a useful shortcut for implementing the Null Object design pattern (`someValue = Utilities.nonNull(value, SomeValue.NULL);`)  or optional updates to an object's values from another object (`this.val = Utilities.nonNull(other.val, this.val);`). It's also very useful to ensure a consistent and re-playable order of value generation during procedural generation. See method documentation for caveats. From a quick look through the code, the skill system could use the Null Object pattern along with these methods to simplify it, for example.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -34,10 +34,10 @@ import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
-import java.util.Enumeration;
 import java.util.GregorianCalendar;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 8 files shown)


Akjosch:
Update to the PR: Also fixed all cases of missing NON-NLS markings. As with all the patch, it doesn't change any functionality whatsoever, just fixes some obvious oversights.

PR #35, opened 2016-03-22
Adding options to the "Hire in bulk" dialogue

Akjosch:
Along with a general cleanup, this PR adds a few options to the "Hire in bulk" dialogue:

* The ability to hire named Astechs
* The ability to set an age range (GM mode only)

If the age range is set, the person is still generated normally and only if the generation *doesn't* result in them having an age within the range will their age be rerolled. Thus, an age range of 10-99 years literally has no effect at all, one with 10-25 will have some half of the people assigned a new range, and one of 10-15 will generate only young teenagers and throw away all the original generated ages.

Since this option can easily create impossible results (super elite 5 year old BA warriors? Sure!), it doesn't appear if you're not running in GM mode.

Inspired by the discussion here: http://bg.battletech.com/forums/index.php?topic=51039.60

PR for review and comments

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -32,6 +32,7 @@ import java.util.Date;
 import java.util.Enumeration;
 import java.util.GregorianCalendar;
 import java.util.Hashtable;
+import java.util.Map;
 import java.util.StringTokenizer;
 import java.util.UUID;
 import java.util.Vector;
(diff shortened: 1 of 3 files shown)


pheonixstorm:
> The ability to hire named Astechs

Huh? The current dialog already allows this. Did you change the way its handled?

Akjosch:
What I did was adding the missing "None" type, for lowbie Astechs/kids/civilians, down at the bottom. "The ability to hire named Astechs." is probably the wrong way to put it, but IIRC you *can* put them to Astech jobs.

pheonixstorm:
For a role of none you should probably make sure they don't have any skills as the current procreation code does. Since kids and most other dependents would not have any usable skills this would probably be a better way to go.

Akjosch:
Procreation code makes sure babies have no skills?

But yeah, good idea. I'll limit it to "below 12: no skills, 12-13: max skill level 0 (8+), 14: max 1 (7+), 15: max 2 (6+), 16: max 3 (5+), 17: max 4 (4+), 18 and above: no limit". Very few generated people have any skills anyway and with the world being so violent, it makes sense that adolescents would pick up *some* survival skills along the way already.

pheonixstorm:
Yeah, my random marriage code reuses a lot of the procreation code as well as some of the dependent code from AtB. Or will once I rewrite it. The initial code had a few issues adding new dependents so I haven't got around to reworking it.

PR #36, opened 2016-03-23
Cleanup PR for the PersonnelTableMouseAdapter

Akjosch:
This is a big cleanup PR for the PersonnelTableMouseAdapter. It's not ready to pull yet, just started.

First steps:
* Making (almost) all the constant strings actual constants.
* Replacing if/else if chain with a switch/case statement

Next will be finishing up with all the constant strings and moving localisable strings into a property file; then tests.


Akjosch:
Cleanup finished. I'll test it a bit, then pull as-is, unless there are major issues.

PR #37, opened 2016-03-23
Fixing most of the issues with the buttons in "parts in use" tab

Akjosch:
Fixing most of the issues with the buttons in "parts in use" tab:

* The buttons refreshhng themselves, blocking the UI
* General speedup of the parts lookup via caching

It's not a 100% fix yet. In particular, mech heads are cached as `Mech Head (XX tons, (100%, [Sensors, Cockpit]))` and similar and queried as `Mech Head (XX tons, (100%)`. That probably needs additional caching in the `PartStore` to work properly, but I'm out of time ATM.

Will update once I'll find some time, probably tomorrow. Feel free to comment and point to other obvious solutions I missed.

### DIFF
--- a/MekHQ/src/mekhq/MekHQ.java
+++ b/MekHQ/src/mekhq/MekHQ.java
@@ -445,7 +445,7 @@ public class MekHQ implements GameListener {
             campaigngui.filterPersonnel();
             campaigngui.refreshDoctorsList();
             campaigngui.refreshPatientList();
-            campaigngui.refreshReport();
+            campaigngui.initReport();
             campaigngui.changeMission();
             campaigngui.refreshFinancialTransactions();
 
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -172,7 +172,9 @@ import org.w3c.dom.NodeList;
  * @author Taharqa The main campaign class, keeps track of teams and units
  */
 public class Campaign implements Serializable {
-    private static final long serialVersionUID = -6312434701389973056L;
+    private static final String REPORT_LINEBREAK = "<br/><br/>"; //$NON-NLS-1$
+
+	private static final long serialVersionUID = -6312434701389973056L;
 
     // we have three things to track: (1) teams, (2) units, (3) repair tasks
     // we will use the same basic system (borrowed from MegaMek) for tracking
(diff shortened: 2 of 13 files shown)


Akjosch:
Ok, the "buy" buttons have a huge slowdown when the report is full (like, on the order of 1000 lines = a pain). "Add (GM)" is near instant now. Financial tab only updates when it is being shown, else it would add another slowdown for big units (battalion sized and more).

I need to polish this up a bit, but it's fine for making the "Parts In Use" tab actually [i]do[/i] something already.

pheonixstorm:
Have you tested the log refresh yet? How big of an increase is there?

Akjosch:
The log refresh works about three times as fast now.

Which for big logs is still on the order of *seconds*. Too slow.

pheonixstorm:
Better is good though. Found why is runs that slow?

Akjosch:
Parsing the whole document in the context of a running GUI is bad, but replacing huge chunks of it when appending would be enough is *worse*. I hope I fixed that too now.

pheonixstorm:
Will have to test this out first chance I get. I have a campaign with a mech regiment and an RCT setup along the lines of the 2nd ACR. Rather large, the cpnx file itself is 30ish MB

Akjosch:
Have fun. I just added the groundwork for sorting and filtering of the list too.

pheonixstorm:
While your working on this, is there or can you, make sure there is an easy way to calculate total ammunition tonnage? I want to make sure that the IO Pay/Cost calculations are accurate as deep as possible. I would really like it down to individual infantry ammo but thats a little too much detail I think.

For reference the rules are on page 12 http://bg.battletech.com/download/Interstellar%20Operations%20Beta%20-%20Creating%20a%20Force.pdf?e42d90

Pretty sure calculating the cost for spare parts (next paragraph) should already be there.

Akjosch:
I'm not sure it's anywhere near the scope of this PR, but I'll see what I can do.

pheonixstorm:
maybe, maybe not, but I figure (even before these changes) trying to tie into the parts in use code might be easier to get said information than taking another route.

Akjosch:
@pheonixstorm I added a total tonnage used by the stores to the display. Please check if it will be of any help to you.

Akjosch:
Unless there are any major bugs, I think this version is basically usable (... as opposed to the one it's replacing). Any further bug hunting would benefit from play testing by others, too.

pheonixstorm:
Will try and test it out before the weekend and look at the code as well.

pheonixstorm:
@Akjosch adding the stored/used section really helps to see what is ON the unit and what is on hand for repairs. Love it. Though not exactly what I was asking for, it does give me enough to work with that it should not be difficult to add what is needed for the IO peacetime cost rules. Just love how the whole this works now :+1:

PR #44, opened 2016-04-07
Replacing personnel's kill and event log displays with tables

Akjosch:
This replaces the personnel log entries (personal events/history and kills) with tables, to vastly improve its performance. I'm not 100% sure it works with all log entries until now (were there some with HTML code or interactive elements?), thus a PR and not a direct push, but for text-only entries, it works like a charm.


pheonixstorm:
The logs are pure text, no html elements. In fact I think the only place html is used is in the daily log.

Akjosch:
Smells good then.

PR #48, opened 2016-04-12
Refit bug fix for location removal

Akjosch:
Changing Refit class to work properly when removing locations.

This is a work-in-progress and a "dirty" PR, not quite ready for inclusion yet. I'll need to clean it up a bit first, but first and foremost test it with some other reported problems. Possibly write some JUnit test cases too.

Any help and comments are appreciated.

### DIFF
--- a/MekHQ/src/mekhq/campaign/inventory/CampaignWarehouse.java
+++ b/MekHQ/src/mekhq/campaign/inventory/CampaignWarehouse.java
@@ -7,6 +7,8 @@ import java.util.Objects;
 import java.util.Set;
 
 import mekhq.campaign.Campaign;
+import mekhq.campaign.parts.AmmoStorage;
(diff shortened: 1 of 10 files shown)


Akjosch:
Ok, I'm scrapping this PR.

My original plan was to fix all the bugs in the Refit class without changing much in the Part hierarchy. After several days of trying, this would mean one of those two outcomes:

* Making the Refit class way larger and even *less* readable than it already is. Essentially, nobody would be able to understand it afterwards without spending days of reading and trying - not even myself after a few months worth of pause.

* Recreating much of the Part hierarchy and associated classes with a parallel structure, which would need to be maintained in parallel and involve lots of code duplication.

This won't do. I'll try again, this time by fixing all the cruft in Part, Mounted and related classes first.

PR #54, opened 2016-04-20
Planet cleanup

Akjosch:
This PR cleans up the data structures used by the planets.

* In code, they appear in a useful, consistent order
* Everything that's realistically changeable about a planet's data can be changed with generic events
* Static methods are moved to a static helper class
* Planetary data is loaded from additional XML files in the "data/planets" subfolder, which can overwrite, enhance and even delete planets.
* Where easily possible, enums are used instead of magic numbers
* Stellar subclasses can be a floating point number ("K4.5V") and have a wider range of valid, supported classes
* All XML parsing is done via JAXB
* Additional data and "fluff" fields for use with campaigns is available.

There are some points where the data is still missing (in particular: Garrison data field is commented out, since it wasn't used anywhere in the first place) and obvious venues for enhancement (planetary cultures for name generation, points of interest on a planet's surface for missions, images for planets, separate moons, multiple planets per star system), but those are still in design phase. This PR can be used stand-alone as-is, without waiting for all the rest.


Akjosch:
The UI changes are minimal so far.

planets1

Though some days are more newsworthy than others.

planets2

Akjosch:
With things like black holes, pulsars and similar being implemented (and the path finder diligently ignoring them), this should be about as ready to get merged as it's going to get. Extensions will then be handled in future PRs and commits. In particular, there are updates to planets.xml to be made ...

PR #61, opened 2016-04-22
Extra data class for type-safe, small extra properties

Akjosch:
This adds an "ExtraData" class which is like a type-safe "Properties" class with a build-in XML serialiser optimised for not wasting too much space (as much as that's possible with XML) while still being reasonably fast.

It has two a `set(key, value)` methods - one which accepts only values whose type matches that of the key's declared type, and one for String values, which are converted to the key's declared type if possible. The corresponding `get(key)` and `get(key, defaultValue)` method return the correct type according to the key's declared type (and `null` respective the supplied default value if the value doesn't exist).

Multiple keys with the same name but different declared types are considered to be different keys (which can be checked with `key.equals(otherKey)`).

Pre-defined supported types are int/Integer, double/Double, boolean/Boolean, String and (joda time's) DateTime. Adding support for additional types, even at runtime, is made as easy as possible: The keys need to extend `ExtraData.Key<T>` and if writing the data to and reading from XML is required, a custom instance of `StringAdapter<T>` has to be registered via `ExtraData.registerAdapter(...)`. The `Key` subclasses typically amount to simply declaring a fitting constructor and the `StringAdapter` classes consist of implementing a single, usually one-line, method.

Examples for `DateTime`:

    public class DateKey extends Key<DateTime> {
        public DateKey(String name) {
            super(name, DateTime.class);
        }
    }

    ExtraData.registerAdapter(DateTime.class, new StringAdapter<DateTime>() {
        @Override
        public DateTime adapt(String str) { return new DateTime(str); }
    });

An instance of this class is added to "Person", which is the one which could benefit the most from being able to have random custom data attached to from mods and extensions. 

Example test program showing how the data is serialised in a <person> block in campaign files:

    public static void main(String[] args) throws JAXBException, DOMException, ParseException {
        final ExtraData.Key<Integer> INTKEY = new ExtraData.IntKey("int_key");
        final ExtraData.Key<Double> DOUBLEKEY = new ExtraData.DoubleKey("double_key");
        final ExtraData.Key<DateTime> DATEKEY = new ExtraData.DateKey("current date");
        
        Faction.generateFactions();
        Person p = new Person("My Name", new Campaign());
        p.setId(UUID.randomUUID());
        p.getExtraData().set(INTKEY, 75);
        p.getExtraData().set(DOUBLEKEY, 12.5);
        p.getExtraData().set(DATEKEY, new DateTime());
        try(PrintWriter pw = new PrintWriter(System.out)) {
            p.writeToXml(pw, 0);
        }
    }

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -33,7 +33,6 @@ import java.util.Enumeration;
 import java.util.GregorianCalendar;
 import java.util.Hashtable;
 import java.util.Map;
-import java.util.Properties;
 import java.util.StringTokenizer;
 import java.util.UUID;
 import java.util.Vector;
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -33,6 +33,7 @@ import java.util.Enumeration;
 import java.util.GregorianCalendar;
 import java.util.Hashtable;
 import java.util.Map;
+import java.util.Properties;
 import java.util.StringTokenizer;
 import java.util.UUID;
 import java.util.Vector;
@@ -67,6 +68,7 @@ import mekhq.Utilities;
 import mekhq.Version;
 import mekhq.campaign.Campaign;
 import mekhq.campaign.CampaignOptions;
+import mekhq.campaign.ExtraData;
 import mekhq.campaign.LogEntry;
 import mekhq.campaign.parts.Part;
 import mekhq.campaign.unit.Unit;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1329,6 +1337,9 @@ public void writeToXml(PrintWriter pw1, int indent) {
                 + "<acquisitions>"
                 + acquisitions
                 + "</acquisitions>");
+        if(null != extraData) {
```
arlith: Looks like there's a missing space at this line (1340 in `Person`).

Issue #68, opened 2016-05-06
IndexOutOfBoundsException when advancing day


SirMegaV:
MekHQ 0.3.27

Occasionally getting this error when advancing a day after combat, then selling units, then repairs, then removing prisoners or astechs, then assigning doctors. Error will disappear if restarting MekHQ right before the advance day.

Savegame and log attached.

[IndexOutOfBoundsException.txt](https://github.com/MegaMek/mekhq/files/251517/IndexOutOfBoundsException.txt)

[IndexOutOfBoundsException.7z.txt](https://github.com/MegaMek/mekhq/files/251519/IndexOutOfBoundsException.7z.txt)

Akjosch:
It looks like it's the doctors/patients part which makes problems. What are you doing on the day before the error hits, exactly?

SirMegaV:
It was after a battle. I sold unwanted units, then repair units, then GM
removed prisoners and astechs, then assign doctors, then hit next day.
Occasionally getting this after this routine.

dericpage:
Added protection against PatientTableModel being given a bad index when attempting to retrieve an index.  In such an instance, the model will return a NULL value.  Also added null handling to the calling methods.

pheonixstorm:
If this issue is fixed please close.

SirMegaV:
Not seeing it lately... might have been fixed. Pls close, I will post again if I receive the same error.

pheonixstorm:
Its your ticket, you can close it yourself. Faster than waiting on the devs in most cases.

SirMegaV:
Sry, didn't know I can close myself.

pheonixstorm:
No worries, your not the only one.

PR #79, opened 2016-05-18
Issue #68: IndexOutOfBoundsException when advancing day

dericpage:
Added protection against PatientTableModel being given a bad index when attempting to retrieve an index. In such an instance, the model will return a NULL value. Also added null handling to the calling methods.


pheonixstorm:
@dericpage Don't forget to delete this branch now that it is merged.

Issue #91, opened 2016-05-28
Can't Assign patients to doctors


pheonixstorm:
I have one doctor with 4 medics but can't assign my wounded to said Doctor.

Tried to hire a second doctor with the same result. Fired the medical staff and rehired the lot, advanced a day still the same. No errors show up in the log either.

Akjosch:
Well, we'll need a campaign file, I think. At least I couldn't reproduce it.

pheonixstorm:
Ok, will go kill some more infantry with a fresh campaign and post it

pheonixstorm:
[doctorbug.zip](https://github.com/MegaMek/mekhq/files/288117/doctorbug.zip)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/288118/mekhqlog.txt)

Fresh campaign, this time w/o advanced medical and still got it to come up. Still no errors in the log.

Akjosch:
1. Go the the Infirmary
2. Mark (left-click) a doctor.
3. Mark a potential patient.
4. Left-click "Assign" until they are all assigned.
5. In four days, everybody is ready to be thrown back into the grinder (no clue on how realistic or close to the official wound recovery rules that is).

Well, works for me. To be fair, I don't use this tab often, so I couldn't tell if something in the way it works changed. But it *does* work. Weird.

pheonixstorm:
The assign button is grayed out though and clicking it does nothing

pheonixstorm:
I stopped being able to do anything with doctors after the changes made in #79

pheonixstorm:
I did find something odd but unrelated in the campaign file. Before the person endtag there is a <extraData/> tag that doesn't belong. Looks like something from the admin xp fix.

Akjosch:
The `<extraData/>` part is simply a generic marker meaning "no extra data for this person". Given that patients don't reference or use that data part, it has no bearing on the infirmary either.

Akjosch:
Playing a bit with the Infirmary, I can see two things worth fixing:

* "Assigned Patients" scrolls horizontally while "Unassigned Patients" scrolls vertically.
* Changing the selected doctor also changes the selection of the unassigned patients.

However, I think I found the problem (though it's strange that it doesn't hit me personally): the `getSelectedUnassignedPatients()` method pulls from the wrong model (the "assigned patients" one). Bundling the fixes for all of this into a PR right now.

PR #92, opened 2016-05-29
Infirmary cleanup

Akjosch:
Cleans up all the issues and UI design problems mentioned in Bug #91 , as well as adding a separate background image for the Infirmary for additionally more distinctive UI design.


Akjosch:
Not much in the way of visual changes, but a bit more distinctive now.

infirmary_2

pheonixstorm:
Like the image, but make sure it can be removed or at least optional so the image itself doesn't need to be deleted.

Akjosch:
Given that MekHQ has no skin editor nor the capabilities to do such, there's little I can do. At best, I can future-proof it by making the background image file name (programmatically) variable.

In doubt, delete/rename/replace it with something to your liking.

pheonixstorm:
An idea, I just didn't want the thing to crash if someone deletes the image.

Akjosch:
It wouldn't have crashed anyway. It would just display the default (grey) background.

PR #99, opened 2016-06-01
Initial implementation of the fuel manager

Akjosch:
This is a proof-of-concept implementation of a "fuel manager" class (and supporting classes). The idea is to gather all the official rules regarding fuel in a single class, as a reference implementation. Specific campaign modes, plugins or options can then pick and choose those rules, or use them as guidelines how to implement their own.

For now, the work concentrates on creating the part of the framework which doesn't require any existing code to be updated or changed. This includes:
- Querying entities, units and forces for their fuel capacity and current fuel status.
- Managing fuel stockpiles and transferring them to/from fuel tanks and liquid cargo bays.
- Creating `IWork` instances related to refuelling.
- Calculating (and optionally applying) strategic fuel usage.

This PR is just a work-in-progress and not ready to be merged (though it won't hurt). Comments are welcome.

### DIFF
--- a/MekHQ/src/mekhq/adapter/MaterialAdapter.java
+++ b/MekHQ/src/mekhq/adapter/MaterialAdapter.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2016 MegaMek team
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ * 
+ * MekHQ is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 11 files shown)


Akjosch:
I'd like to thank scJazz for his 1h romp through the rules and actual real refuelling procedures which no doubt pissed off the FBI.

Dylan-M:
@Akjosch What's your status with this? Is it ready to be merged in? If so, please do it before Sunday. Thanks ;)

Akjosch:
I put it "on hold" for a while so that the MM work on fuel usage for tanks would be finalised. Otherwise, it's fine, the one tiny conflict just needs to be fixed and it can be included (and then used in other places).

pheonixstorm:
my work on support tank fuel is done. Beyond that add whatever you need for combat related fuel usage. IIRC every scenario is supposed to use up a tanks/indie mechs entire fuel load. Pretty sure this won't break the work i've done for peacetime fuel costs.

Akjosch:
It's not about "breaking" anything, it's about making sure the calculations return identical values everywhere, so we can later refer to `FuelManager` for all fuel-related details (like how much fuel there is in the tank of a given unit, or how much it uses per day).

Dylan-M:
Closing this PR since it seems abandoned.

PR #101, opened 2016-06-05
Feature: Added RAT Table access to GM tools. (#18 Resubmit)

Kurios:
From this you can both roll for rat, and if GM mode is enabled, add
directly into your campaign hanger.

Additionally, GMTools is no longer held in memory when not shown and
some utility functions have been added to the UnitTableData for going
from string to int for the weight, unit, and quality.

image

See: #18


scJazz2:
Is there some stunningly huge reason why this feature hasn't been pulled?

HammerGS:
If I had to guess, none of the coders have had a chance to review it.

Akjosch:
For future PRs: Please, *please* try to have some kind of consistent formatting. Ideally consistent with the rest of the code base.

Issue #113, opened 2016-06-28
Dates are cut off in Personnel log display


Keatonmac:
Hello, developers...

Posting a bug for MekHQ 0.3.29.  When reading a character's Personnel bio, the final number of the date is missing.  You cannot see the year the action took place, just the day and month, and "302" (for 3025, as an example).  An example is this screenshot #1.   

![screen shot i - bio showing missing date](https://cloud.githubusercontent.com/assets/20175361/16398128/2e3d2dfa-3c8e-11e6-86b9-3dafa27c7987.png)

However, if you attempt to Edit Personnel Log, you'll see the date is correctly entered.  It's just not showing on the screen.   

![screen shot ii - edit personnel log](https://cloud.githubusercontent.com/assets/20175361/16398151/4be90c5c-3c8e-11e6-91f0-ffd1536f7bb0.png)

It's not a matter of window size - expanding the window to full size does not expose the missing year digit.  And I cannot find anything in the log file that indicates an issue...all years have four dates that I can see.  Further, this happens if you have developed a new campaign in MekHQ 0.3.29 or imported an existing one from 0.3.21 or older.   A PDF of the campaign file is included. 

[Desolation Angels30250201.pdf](https://github.com/MegaMek/mekhq/files/336063/Desolation.Angels30250201.pdf)

Thank you in advance for all your help...

Keatonmac

Akjosch:
It looks like an Apple-specific styling problem. The default font used for those entries under Windows and Linux is significantly smaller (about the same as the one used in the personnel table on the left side), so they don't have that problem.

Too bad I don't have a system I can test this with. We'll see if I can try to make a fix regardless.

Dylan-M:
@Akjosch I'll try and pull this down for checking today at work. About the only upside to using a Mac for my job ;)

wildj79:
@Dylan-M Looks like @Akjosch fixed this.
![screen shot 2016-11-04 at 6 23 20 pm](https://cloud.githubusercontent.com/assets/2126857/20025395/eed2fc32-a2bb-11e6-8994-d34fe9b1b2d8.png)

PR #127, opened 2016-07-29
Fix for Issue #113 - MekHQ 0.3.29 Dates are cut off in Personnel log …

wildj79:
…display

This is a fix for Issue #113.  This only seemed to be affecting Mac OSX. The default font, Lucida Grande in this case, seems to render wider on OSX than it does on other OS's. The solution is to widen the column slightly to accomodate the wider font.  Will need to test that this doesn't have an adverse effect on other OS's.

### DIFF
--- a/MekHQ/src/mekhq/gui/model/PersonnelEventLogModel.java
+++ b/MekHQ/src/mekhq/gui/model/PersonnelEventLogModel.java
@@ -104,7 +104,7 @@ public class PersonnelEventLogModel extends DataTableModel {
             case COL_DATE:
                 return 90;
             case COL_TEXT:
-                return 290;
+                return 300;
             default:
                 return 100;
         }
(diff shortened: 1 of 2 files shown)


wildj79:
screen shot 2016-07-29 at 12 10 59 am

pheonixstorm:
You may want to shrink the log body by 10 points or make sure the kill log table and the section above the log is the same size (I don't know if it was changed to a table as well). If the size is the same for all four sections there should be no issue. The fourth I forgot to mention above is for the persons bio.

wildj79:
I actually thought about that after I issued the pull request :wink:. I'll make the change and push it back into my fork.

wildj79:
I've been able to test this on my Windows box, it doesn't seem to have any adverse effects. I'll try testing it on my Debian box later to see if these changes have any adverse effects on Linux.

wildj79:
Closing this so I can resubmit it the correct way: create a new branch and open a pull request with that, so I can keep master in sync with the upstream repo 😑. Live and learn.

PR #143, opened 2016-08-23
Fix for Issue #113 Dates are cut off in Personnel log

wildj79:
Resubmitting my previous pull request.  I'm still new to how things work on Github, so bear with me.  Anyway, this will fix an issue with MekHQ on OS X where the font on the Personnel log for events are getting cutoff.  I've tested this on other platforms, and it does not seem to have an adverse affect on them.

### DIFF
--- a/MekHQ/src/mekhq/gui/model/PersonnelEventLogModel.java
+++ b/MekHQ/src/mekhq/gui/model/PersonnelEventLogModel.java
@@ -104,7 +104,7 @@ public class PersonnelEventLogModel extends DataTableModel {
             case COL_DATE:
                 return 90;
             case COL_TEXT:
-                return 290;
+                return 300;
             default:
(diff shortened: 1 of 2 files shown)


pheonixstorm:
One thing on this, is there a way to test if the OS is mac or not? Seems better to do that test instead of changing the values for linux and windows. I know its possible in C++ but not sure how that might work for Java since it compiles all three from whatever OS eclipse is on.

Dylan-M:
Yes, you can test for OS within the code. In fact, MHQ already does this to set the Look and Feel for OSX.

wildj79:
Looks like @Akjosch fixed this in https://github.com/MegaMek/mekhq/commit/6ec524d3f9ee8ac0231f4d586d574ac170bdbc8b.

PR #202, opened 2016-10-17
Mass Repair/Salvage

kipstafoo:
Mass Repair/Salvage

I created a way to do configurable mass repair/salvage. Having used
MekHQ for over 200 campaign fights, the time to perform repairs is to
most time consuming part. Looking to improve that, I've added this small
feature.

In essense, it creates adds a menu item when right clicking on a unit in
the repair tab. The menu item allows to perform all possible (as
configured) repairs or salvages.

Within the campaign configuration is additional configurations that
allow a user to determine what things are elligible for mass action. For
each of the groups (see below for groups), you can configure if it's
active, what the min skill level to handle it is and the min/max BTHs.

The groups are:
- Armor
- Ammo
- Weapons
- Mek/Tank Locations
- Gyros
- Engines
- Actuators
- Electronics (Cockpit, life support, sensors)
- Other

Additionally, you can configure if the system will:
- use extra time lower the BTH for a tech to the desired target
- use 'rush job' to increase the BTH on easy jobs (i.e., an elite
repairing armor)
- allow carry-over to the next day(s), i.e., allowing a tech to perform
a task that will tke more than their currently available tech minutes

The system, when activated, will find all parts needing repair/salvage
and all techs that could work on it based upon skill type. It then
determines, for each tech, if they can perform the task based upon
configured minumim skill level and target BTHs. If allowed, the system
will attempt to alter the BTH by using extra time/rush job. Once the
list of valid techs is found, the system will pick the one with the
least skill (to aid in increasing techs skills).

I can supply screen shots, etc...


HammerGS:
Would be good to see screen shots

kipstafoo:
Try this DB link guys:
https://www.dropbox.com/sh/nq1xctqeokdhcmo/AADW3oq8HTN2wfdutHdjEsdfa?dl=0

There's 5 files in there:
- configuration dialog addition.png -> This is the changes to the campaign
configurations dialog. I updated the layout of this page. One could argue
that it could be made scrollable and the and spinners made smaller, changes
I've glad to make if the team so desires/requires.

- repairable mech.png -> A repairable mech. Shows the initial menu addition

- repairable mech - sub options.png -> Same repairable mech, but with the
sub options. This shows that you can use the first option to simple use the
configuration you have (i.e., only act on the items that you have marked as
'active' in the configuration), or to override the 'active' state and act
on a single item type. For example, you may salvage a tank and really only
care about getting the armor off of it, if you choose the 'Only Armor'
option, it'll just strip the armor off and leave everything else as it was.

- salvageable mech.png -> Like a repairable mech, but with a menu that is
appropriate for the state of the unit (Mass Salvage vs Mass Repair)

- salvageable mech - post action -- see log.png -> This is the result of
the mass salvage option on the selected LNX. It does all the repairs then
adds a log entry stating that it is done and how many actions it performed.

The mass repair/salvage will keep going until it has nothing left to do. So
if there are chained repairs, i.e., you need to attach a location before
the other items can be done, it'll sweep through the list of repairs,
complete what it can (based upon menu selection/configuration), and then
sweep through again and again until there's nothing left for it to do. You
could, in theory, have a mech which is simply a center torso but if you had
enough techs (and some good rolls :) ), you could repair the thing with one
click of the menu.

Feel free to let me know if you have more questions. I'll admit that I
haven't done any swing work since I last did work on MegaMek. I was tempted
to get one of the engineers that work for me to code review it, but it's
pretty darn stable and I think it'll help speed up the repair cycles and
get people back to smashing mechs together. ;)

  -C-

scJazz2:
Kip join the Slack!
In addition to the new domain, we have begun using Slack for internal communication. Discussions there lead to the creation of a public slack for the community as well. In order to join the public slack, you must be invited. In order to facilitate self-service, you may invite yourself by visiting http://megamek.org:3000/

Dylan-M:
While this is amazing, the options should be on a pop-up dialog that appears when you select to start using configuration. Rather than on the campaign wide options. I'm not against both of them existing, and in fact it's probably a good idea. But, I was to be able to adjust them on the fly as it were.\

Anyway, just thoughts. Nice work.

mchapman1970:
love this feature  yet another bucket list item checked for the HQ/MM being the perfect BT

HammerGS:
Wanted to note, I've been using this for the last week and haven't had weirdness.  I would agree with the pop-up dialog as I have gone in and tweaked stuff on the fly.  One thing that I think also should be added is column headers to explain what the columns do.

image

But really love this and adds a lot to MekHQ.

BLOODWOLF333:
How does it handle hip and shoulder crits?

HammerGS:
Knock on wood....haven't had any in the games I've playerd

kipstafoo:
BW: I am actually looking to improve that. Right now, it's stupid... it simply ignores them and fixes everything else. Sub-optimal, I agree. What I'm thinking is this:

- If it's a salvage, who cares... you're salvaging
- If it's a repair, salvage all possible parts from the location and remove the location, then re-attach location and replace parts.

I need to explore this because right now I'm simply tapping into the main repair commands, which take the state of the unit. I'll need to flip the state of the unit from repair to salvage to do the work on the limb in question then switch back for everything else.

kipstafoo:
Hammer/Dylan, I agree on the pop-up. So, since you two brought it up, how would you envision it working? Would you have all the options available -- min skill level, min & max BTHs -- or just the switch saying what item types are available. If you changed something, would that affect the campaign-wide options? If not, would it remember your settings the next time you popped open the dialog during this session of MekHQ? For all future sessions until you change it?

Dylan-M:
@kipstafoo Regarding the salvaging of components within a limb when it needs to be re-attached, you don't need to flip the state. It's already flipped for _those items only_.

Regarding options on the dialog, I'd like to have it display them all (Perhaps on a tab?). I would not think that it would save them at all. Once the work based on that instance of the dialog is finished, it should forget the changes. Perhaps offer a checkbox for making the changes permanent.

kipstafoo:
Thanks for all the comments. I'm going to make some changes to this based upon a lot of the discussions I've had with people and then I'll post new code and screen shots for round 2.

kipstafoo:
Ok, we're done(ish?). I've totally overhauled the MRMS feature.

On the UI side, the I ripped all the settings out from the campaign options dialog. Everything now lives in a new dialog which is initiated from the the taskList pop-up menu:



One launched, the new dialog allows you to select multiple units to act on at one time. The upper part shows the valid units and lets you select directly from the table, select units that are assigned to pilots, select units that are not assigned and clear the selection.

The lower part are the settings we had before (with better explanations) as well as an additional setting to limit the max skill level. So, if you want to deal with repairs by elites yourself, set the max skill level to Veteran.

Finally, you can kick off the MRMS from the button. Additionally, you can tell the system to save the current settings as the new default which means the next time the dialog is launched, these will be the default settings.



On the backend, I did a lot of cleanup and made sure it worked smoother. The biggest backend addition is how the system handles meks w/ shoulder or hip crits. These issues take priority over all other repairs. If there is a damaged limb with a bad shoulder or hip and you're trying to REPAIR the unit, the system will flip the unit to salvage mode and remove all existing parts from the limb(s) in question. It will NOT use extra time to do, so all repairs must be completed in a day and not overlap. The reason is that the tasklist UI is none-to-friendly if you have a unit in 'repair' mode but it has 'salvage' actions happening... i.e., it doesn't show them. Getting back to the flow, once all parts are removed from the limb(s), the system will then scrap the limb(s), rebuild the list of repairs to do and continue on its merry way.

Now, because I know it's a pain in everyone's butt, I lumped an additional change in here... mostly because it's built off the code for identifying what type of part everything else. Bloodwolf and I worked on a change for the repair list to show different icons for different parts. BW did the images, I did the backend. Again, this is built off some of the MRMS code, which is why it's here, but it's very encapsulated and can be turned off.



Currently, the different icons shown are:

- Armor
- Ammo
- Actuator
- Engine
- Electronics (Life Support/Sensors/Cockpit)
- Heatsink
- Weapons -- 4 different images for energy, missile, ballistic and artillery
- Locations (Mek only)
- Melee weapons
- Everything else

I think that about sums everything up. Questions, comments, flames... all are welcome.

kipstafoo:
Ok, final thing... I hope. :)

- Added a popup message after the action is complete to tell how many repairs were completed/scheduled
- Changed the text of the 'Cancel' button to 'Done'
- Merged in the latest master and resolved the conflict with CampaignOptions

PR #241, opened 2016-11-07
Tech ghost assignments

kipstafoo:
Sometimes techs are assigned to units that no longer exist. Due to the way unassigning works, it can never be cleared because the system clears from the point of view of the unit. I've yet to figure out HOW it happens, but I've added a simple fix for when it does happen. Essentially, if you unassign a tech from all units, after processing all the existing units, set the techs' unit list to null which will make sure they are all clear.

And, yes, that silly Air Assault file keeps coming up, even after a hard reset.


pheonixstorm:
That silly air assault file is a windows issue. It happened with another set of files awhile back due to changing from lower case to upper case in the file name. It does not show up on linux to my knowledge

PR #248, opened 2016-11-14
Added meaningfull names for new AtB contracts.

aunoor:
Replace default "New contract" with some meaningfull description for contracts from Contract Market.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/ContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/ContractMarket.java
@@ -409,6 +409,9 @@ public class ContractMarket implements Serializable {
 
         contract.initContractDetails(campaign);
         contract.calculateContract(campaign);
+        
+        contract.setName(Faction.getFaction(employer).getFullName(campaign.getCalendar().get(Calendar.YEAR)) + " - " + AtBContract.missionTypeNames[contract.getMissionType()]);
+        
 		return contract;
 	}
 


pheonixstorm:
You might want to include more as this could lead to contracts containing the same name, something which HQ does not allow. The majority of people rename the contract before accepting it though.

Also, if I am reading this right it looks like things will be jumbled together as well. Might want to add a separator so it would look like FedSun-year-missiontype, though I would suggest Faction-mmyy-missiontype (FedSuns-0325-Objective Raid). Optionally you could add the duration to make tracking the end date a bit easier. ie Faction-mmyy-mission-duration (FedSuns-0325-Raid-03)

Dylan-M:
Actually the ideal format would be:
Federated_Suns_30650317_Raid_03

In other words, full ISO standard date format and underscores for separators.

purpledragon2:
I would love to see this.  For Unit historical sake I would include the name of the planet in there somehow.

pheonixstorm:
The name of the planet is already in the contract itself, I don't think it would be useful personally but that's up to ralgith.

aunoor:
HQ allow to contracts have same names. Hungreds of contracts with "New contract" name is user problem only. :)
Underscore as separators is bad idea, I think. From other variants, variant from @pheonixstorm is more likely for me.
But, I don't vew method to shorten names of factions and mission types. Then generated contract name may be very long, ex.: "Federated Suns-306503-Objective Raid-03"

pheonixstorm:
Then the only contracts able to do so are atb contracts which should be fixed. At least for ACCEPTED contracts. For the contract generator I don't think it matters what the name is. Yes the name might be a bit long but there isn't much to shorten it. For naming my contracts I use the 2-letter faction code, though the format is a bit different. ie FS-Garrison-18-0325

aunoor:
May be use hardcoded to source faction code instead full name? "FRR" instead "Free Rasalhague Republic" maybe will be much better?

pheonixstorm:
There should be a function to call the short name for the factions, though I have never looked so no clue where it might be.

aunoor:
Please close this request as obsolete.

Akjosch:
You can close your own PRs and tickets ...

PR #249, opened 2016-11-15
Autogenerate AtB contract name in Contract market

aunoor:
Autogenerate AtB contract name in Contract market with "Faction_short_name-yyyymm-mission_type-length" format.
Ex: "CC-306701-Recon Raid-3"


Dylan-M:
I have one final request: I'd like to see the target in there as well. Maybe at the end?
CC-306701-Recon Raid-FS-3

aunoor:
Added enemy faction code to contract name.

Dylan-M:
Strangely, I forgot all about this and made a version that generates a random ID for the contract instead. Those 2 implementations are going to conflict, so an option will have to be put in for them.

Issue #259, opened 2016-11-27
Inconsistent date/time formats


gloowa:
Hi, i am a long term fan of MekHQ and for the longest time i had but one issue: date format.
It's not a big issue, but it's one of those very annoying small things.

In some places, MekHQ uses middle endian format of month/day/year.

In others (daily log for instance) it uses system locale, with full day and month names, STILL in middle endian order: 
środa, października 22 3045 (wednesday, October 22nd 3045) - so still month, day, year.

I realise that all MekHQ devs are most likely from United States, but please give us the option to use standard big or little endinan formats of dd/mm/yyyy or yyyy/mm/dd, uniformly, everywhere. Or at least option for all date/time fields to be in ISO 8601 format would be appreciated. https://en.wikipedia.org/wiki/ISO_8601

The mm/dd/yyyy date format is only used by two countries on the planet: United States and Federated States of Micronesia, for the rest of us it extremely confusing.

I hope i didn't come off as condesending or aggressive, my english is not perfect.

Dylan-M:
To make this worse, some places do use ISO 8601, which is my preferred date format despite being from the USA ;)

Also, we have devs from all over the world: US, CA, and DE for sure.

gloowa:
Sorry about writing here but github removed messaging system a while back - I decided to weigh in and forked the repo to address this by myself. Unfortunately i'm having issues with importing the project into eclipse - when i import the forked repo using EGit, it downloads everything, then (i presume?) unpacks it and then hangs for around 10 minutes and then Eclipse crashes. Tried increasing eclipse memory allowance but it didn't make a difference. After i restart Eclipse, it looks like it only imported .jar's in the lib directory.

Is that something anyone encountered?

Dylan-M:
It is not. There shouldn't be any "unpacking", since it isn't packed in any way shape or form. EGit is just a java based, Eclipse integrated, frontend for git. I posted on the CBT forums awhile back a step-by-step on how to bring the projects in via EGit, but of course the staff there ignored my request to sticky.

gloowa:
Ok, i've just found a bug report on eclipse page which might explain what's happening to me - basically import existing projects option appears to be broken, and one needs to use a workaround.

Gonna post if i manage to work anything out.

gloowa:
Hey, i managed past the Eclipse issues.

I created a branch with my changes
https://github.com/gloowa/mekhq/tree/fix_for_date_formats_%23259

[...]

PR #260, opened 2016-11-30
Date format changes - fix for #259

gloowa:
This is a fix for date format inconsistencies.

Changes:
- new class MekHQOptions that stores settings that are relevant to application itself, rather then campaign
- MekHQOptions is a singleton
- currently only stores selected date format
- defaults to ISO format (both short and long date representations 
- new Dialog to accompany MekHQOptions, accessible from startup GUI and menu bar (File->MekHQ Options)
- saving/loading MekHQOptions to/from .xml
- replaced all instances of locally formatted dates with MekHQOptions.getInstance().getDateFormat<Short/Long>() for display
- replaced all instances of locally formatted dates withMekHQOptions.getInstance().getDateFormatDataStorage() when saving/loading data
- all dates saved in ISO format ( yyyy-MM-dd HH:mm:ss )
- ensured previously saved files will load properly by attempting to parse date with old pattern if new pattern failed

**Issues:**
- **Daily report log is being saved as a set of html strings. because of that, changing date format in options has no effect on currently displayed report, or on report in saved campaign. Only new reports (advance day) will display date in new format.**
- ~~**I was unsure as to what is the role of mekhq.adapter.DateAdapter so i left it untouched. There is an instance of locally created date format there.**~~

Other info
Because MekHQOptions can be accessed anywhere via getInstance(), there should be no more problems with date formats - whenever someone needs to format a date, appropriate format is avaliable via
- getDateFormatDataStorage() for ISO date/time format when saving/loading data
- getDateFormatShort() for short date representation (year number, month number and day number)
- getDateFormatLong() for long date representation (year number, month full name, day number and day of the week name)

Tested on several saved campaigns and a new campaign and didn't encounter any errors.

### DIFF
--- a/MekHQ/.gitignore
+++ b/MekHQ/.gitignore
@@ -17,3 +17,4 @@ MekHQ.jar
 /*.tar.bz2
 /*.blame
 /.classpath
+/mekhqOptions.xml
--- a/MekHQ/src/mekhq/MekHQOptions.java
+++ b/MekHQ/src/mekhq/MekHQOptions.java
@@ -24,6 +24,12 @@ public class MekHQOptions {
 
 	public static final String DATE_PATTERN_ISO_LONG = "yyyy-MM-dd";
 	public static final String DATE_PATTERN_ISO_SHORT = "yyyy-MM-dd";
+	
+	public static final String DATE_PATTERN_BIG_ENDIAN_LONG = "yyyy MMMM dd, EEEE";
+	public static final String DATE_PATTERN_BIG_ENDIAN_SHORT = "yyyy-MM-dd";
+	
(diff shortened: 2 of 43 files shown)


Dylan-M:
This branch has quite a few conflicts at the moment, and I also feel like this is an incorrect way to handle this. We _already_ have an option for this as part of the MM options. MHQ should just obey those options instead of having a duplicate option.

gloowa:
No problem mate, i admit to just skimming over the relevant parts of the code without digging to see if there already are some facilities elsewhere.

If you had time to lay down how you want this executed (in general) i could take another shot at this subject.

PR #357, opened 2017-03-26
Allow setting repair modes/scrapping of multiple tasks/parts at once

luiges90:
If you select multiple tasks/parts and change mode (Extra Time / Rush Job) / scrap, all of them should change. I am tired of setting it one by one...


neoancient:
I've been working on implementing reduced salvage and replacement times for pod-mounted equipment in omnis. I made some changes when merging this into the omni branch, since the reduced time for pod replacement is set at 30 minutes per location and can't be modified by mode, so if you are salvaging or replacing a missing part the mode change doesn't appear on the menu unless all the selected parts have a fixed mounting.

While I was doing that I also expanded this to apply GM > Complete Task to all selected parts (as long as all selected repairs are possible).

PR #392, opened 2017-04-26
Newrecruitdialog cleanup

neoancient:
This class is responsible for three of the five Codebeat quick fix suggestions.


Dylan-M:
If you look at the codebeat analysis of the branch (click 'Show all checks'), it shows that 12 were fixed, but 6 were introduced with the fixing. This will be a very long uphill climb I feel.

neoancient:
The six that were introduced are much lower severity, and after looking at them I'm not sure that further changes to meet the criteria would be an improvement.

Dylan-M:
I tend to agree, just observing that raising the score in general will be an uphill battle.

PR #394, opened 2017-05-01
Fix custom spas

Dylan-M:
A long time ago, Tah started the ability to add custom SPAs by moving SPA definitions to an XML file. However, the code to actually make proper use of that file was never updated, and thus everything worked off of the PilotOptions from MegaMek still. This is, I hope, a finish to that implementation with the single exception of having it read from a player defined SPAs file. I don't plan on adding that until we add plugins, otherwise there will be no way to define what the SPA does.

I've included an unofficial SPA called clan_tech_knowledge to represent techs who've earned enough XP to understand clan equipment.


scJazz2:
tap tap reminding that this is still a problem

PR #397, opened 2017-05-22
AtBScenario refactor

kipstafoo:
Completely refactored AtBScenarios, keeping existing working mostly intact. The main idea here is to make creation of scenarios much easier in the future. Currently adding new scenario types was difficult given the logic around handling special missions and big battles. Additionally, all the code was in large switch statements in AtBScenario and has now been moved to allow easier understanding of what a specific scenario does beyond the default scenario behavior.

- All scenario-specific code was ripped out of AtBScenario.java and each type of scenario was given its own class file. AtBScenario was moved to an abstract class and scenarios are created via the new AtBScenarioFactory singleton.

- The concept of 'Special Mission' and "Big Battle' was removed as a range of defined scenario types and moved as boolean overrides to the new scenario implementations.

- Creating a new scenario is as simple as the following:
o Create a new class in mekhq.campaign.mission.atb.scenario which extends mekhq.campaign.mission.AtBScenario
o Implement getScenarioType() to tell it which of the 21 scenarios types is represents
o Implement getScenarioTypeDescription() to set the short name/description of the scenario
o Implement getResourceKey() so that the proper resources can be pulled from AtBScenarioBuiltIn.properties
o [Optional, yet suggested] Override setExtraMissionForces() to customize how the enemy/ally forces are generated
o [Optional] Override other methods for map size, what can be re-rolled, etc.
o [Optional] To mark a scenario as a 'special mission' or 'big battle', simply override either isSpecialMission() or isBigBattle(), returning true.
o Add new class to static scenario registry in AtBScenarioFactory
o Add appropriate resource strings to AtBScenarioBuiltIn.properties (these properties were originally in AtBScenarioViewPanel.properties but have been extracted out)

- When a new scenario is requested by type from the scenario factory, it references the list of scenarios, grouped by type. If there is only one, it returns it. If there is more than one of a given type, it randomly returns one. Future improvement: Allow weighting of scenarios via code or XML configuration.

- Additional changes:
o A button was added to the briefing tab to allow [GM] generation of missions for the current week. It will not generate a mission for a lance that already has a pending mission for that week. This generate code was moved from Campaign.java to AtBScenarioFactory.java
o Minor updates to the 'Extraction' scenario to make it a bit more player friendly. The civvies now attempt to extract to the players side of the map instead of simply running around and engaging the enemy units.


purpledragon2:
thank you for your work.

PR #404, opened 2017-06-20
Implementation of pirate-centric "contract" generation mechanism

NickAragua:
As per AtB pirate rules.

Added two new classes, PirateContractMarket and AtBPirateContract, which override ContractMarket and AtBContract respectively and handle all the pirate-specific "contract" logic. Also made minor changes to Campaign.java and ContractMarket.java to accomodate this new functionality. LanceAssignmentView adjusted to handle the possibility of 0 lances being required.


NickAragua:
As a note, I didn't come up with any new contract types, I shoe-horned the existing ones as AtB does. Also, I haven't implemented loot/kidnapping rolls or mutinies or anything like that.

kipstafoo:
I'd like to look at this deeper, but I have a few initial comments:
- I'd like to see the private instance variables stay private. If we need access to them in a derived class, my preference would be to add proper accessors.
- IMO, campaign.setFactionCode() is not the proper place to initialize the contractMarket. The setters should be kept for setting the specified data without additional side effects. I would prefer to see some sort of initializeContractMarket() method which is called at the proper place and time.
- There's an argument to be made to change ContractMarket.writeToXml() and ContractMarket.generateInstanceFromXML() write and then read the classname of the actual contract market instance instead of having a special case within generateInstanceFromXML() to determine which sort of contract market to instantiate (using reflection). If there is no class defined (existing save games), then it defaults to the base ContractMarket.
- Would it be worth while create an IContractMarket interface which both markets implement? Do we foresee additional specialized markets being created? Possibly house-specific markets?
- Building on the above, instead of returning AtbContract in PirateContractMarket.java, should we introduce an IAtBContract interface and change the various method signatures to return the interface allowing future expansion.
- We should add @Override annotations to methods that are being overridden in the subclasses.
- I'd consider updating the documentation in docs/AtB Stuff to talk about how a pirate campaign works
- The history.txt file needs to be updated

NickAragua:
Thank you for your feedback. I'll update the code changes to address the issues you've raised. 

Do we know of any place *other* than the campaign option screen where the player's current faction can be updated? Also, somewhat related, I'm currently making the assumption that the player switching factions clears the contract market.

As far as additional contract markets, I can imagine house-specific and clan markets (house-specific rules are already in AtB, and I've seen folks discussing clan-specific rules on the battletech forums), so an interface may be a reasonable approach. The main issue I had is that the pirate market is capable of generating standard mercenary contracts as well, in addition to using some of the standard mercenary contract generation functionality, but I don't necessarily want to duplicate that functionality - so I used inheritance for easy access to it. What kind of options would we consider for dealing with that? Other than keeping a reference to a standard ContractMarket within the PirateMarket class.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -486,8 +486,18 @@ public ContractMarket getContractMarket() {
     	return contractMarket;
     }
 
-    public void generateNewContractMarket() {
-    	contractMarket.generateContractOffers(this);
+    // This function clears the campaign's existing contract market
+    // and generates a brand new instance based on the campaign's current faction
+    // use with caution :)
+    public void initializeNewContractMarket() {
```
NickAragua: Function was completely unused.

Dylan-M: Maybe at the moment it is, but keep in mind that sometimes that's done on purpose for future expansion/planning or to be accessed when the program is used as a library

Dylan-M:
Looks like you have conflicts that you need to resolve. Also, you introduced 9 "issues" that CodeBeat doesn't like. Please review those as well.

neoancient:
Though ContractMarket only generates AtB contracts currently, it was written to support multiple generation methods just like PersonnelMarket. The selected generation method is stored in the `method` field. This is checked in `generateContractOffers` to determine how the contracts should be generated.

This can be improved both to simplify `ContractMarket` and to support plugins by the interface that @kipstafoo suggested. Something like this:
```
public interface IContractGenerator {
    String getContractGeneratorName();
    String getContractGeneratorDescription();
    List<Contract> generateContractOffers(Campaign c, boolean newCampaign);
    List<Contract> removeContractOffers(List<Contract> currentOffers);
}
```
The type of `ContractMarket.method` would be `IContractGenerator` instead of `int` and in response to a new day event the contract market would add any offers returned by `generateContractOffers` and remove any returned by `removeContractOffers` (though probably in the other order). The methods that return a name and a description would be used by the campaign options dialog.

NickAragua:
I'm *reasonably* sure that I merged the changes I made with the master branch, but it'd be superb if someone could confirm that.

I think I'd like to leave the interface extraction for another day though, if you folks don't mind.

NickAragua:
Ok, think I figured the merge out. 

Looking at the codebeat stuff:

ContractMarket.java - it's complaining about generateInstanceFromXML() being too long, and the "assignment branch condition" being too high in writeToXml(). Not really sure what I'm expected to do about these, since these function existed before and I only made minor additions to them.

PirateContractMarket.java, generateContractOffers() and generateAtBContract() also complains about "assignment branch condition". I guess I'm not really sure I understand why having assignments and logic branches in a function is a bad thing. As for pickPlanetAndEnemy(), it's a pretty atomic operation, and I've documented the interior pretty thoroughly. Not sure that splitting it up would serve much purpose.

arlith:
codebeat is something we started testing out somewhat recently.  You don't have to strictly adhere to everything it says, although it can be good to keep some of those things in mind.

Dylan-M:
And as a follow on to what Arlith said, I only asked you to _look_ at them ;) Often there isn't anything you can do. Codebeat isn't really going to like much of what we do because we have some very long methods and lots of branching. It likes neither of those things.

NickAragua:
Gotcha. Well, you can't say I *didn't* look at them. :)

Anyway, I'll work on extracting an IContractMarket interface since there seems to be some consensus that it's a good idea.

neoancient:
I'm not talking about extracting an interface from `ContractMarket`. I'm talking about an interface to be implemented by methods that generate contracts. Maybe `IContractGenerator` or `IContractFactory` would be a better name. `ContractMarket` is written to support multiple generation methods, and there should be no need to extend it to add a different method of generating contracts.

NickAragua:
Ok, I think I see what you're suggesting. That ContractMarket contain an instance of of IContractGenerator, and we pull the functions you mentioned into specific implementations of that interface. Currently: AtBMercContractGenerator and AtBPirateContractGenerator. That way, code outside the ContractMarket is isolated from what kind of market it is.

Took me a while to figure it out. One of the drawbacks of remote dev work, I guess. This would have been a five minute conversation at work. :)

NickAragua:
I think I'm going to back this one out. If I'm going to mess around with the contract market to the desired degree, this particular set of changes won't be at all relevant.

PR #409, opened 2017-07-12
Fix for incorrect treatment of tornadoes

NickAragua:
Some bullet proofing. Whenever we get ICE_STORM weather in a scenario, the PlanetaryConditions weather name resolver does not appreciate it, as it has an off by one error in it. 

Also, noticed that gales and tornadoes were resolving as downpour and light snow.


pheonixstorm:
The issue with weather was fixed on the MegaMek side by PR 533. There was a missing comma in the names which caused many fun little issues since several new weather types were added.

NickAragua:
Ok. I've rolled back the bulletproofing, but the tornado issue still needs fixing.

pheonixstorm:
Yeah, never noticed that. Don't think anyone else ever did either since no bug report was ever posted.

AtB was written by neoancient so you might want to ask him if he is going to update it for the new weather or if he is too busy you could give it a shot ;) I know he has a lot on his plate at the moment.

Issue #363, opened 2017-04-01
Feature Request: Order from Warehouse


ThomasMarik:
In the warehouse tab I would find it very useful to be able to right click on an item, such as ammo, and order from teh right click menu.  Could the option to order and order in bulk be added to that context menu?

PR #412, opened 2017-07-20
#363 Implementation of requested "order from warehouse" feature

NickAragua:
Just what it says. Followed existing menu pattern, re-use existing "add shopping item" functionality.


kipstafoo:
Can you include screen shots, please? This may be overlapping with my new 'Parts In Use' overhaul. There's some funkiness with our parts management in general which I want to deal with.

NickAragua:
orderfromwarehouse

UI change looks like this. Based on what I saw in your 'parts in use' changes, we shouldn't have any overlaps. I invoke campaign.getShoppingList().addShoppingItem(), which is the highest level way that I'm aware of to add an item to the shopping list.

Issue #402, opened 2017-06-19
[0.43.2-git : Bug] mekhqlog.txt Size Causing Performance Issue


SirMegaV:
Load campaign, mekhqlog.txt = 28,208 KB. 

Go to 'Repair Bay', make 1 armor repair with 'Rush Job (1/2)' to e.g. Moonsoon with, mekhqlog.txt = 129,842 KB.

No errors in the log file, just information for debugging(?) such as checking <crew name> and adding <unit name> to unit counts.

Make another armor repair with the same setting, mekhqlog.txt = 206,180 KB.

Basically almost everything that I do, I will have to wait for quite a few seconds before it will take effect.

Attached campaign and custom file for testing.

SirMegaV:
[mekhqlog.txt Size Causing Performance Issue.7z.txt](https://github.com/MegaMek/mekhq/files/1083885/mekhqlog.txt.Size.Causing.Performance.Issue.7z.txt)

NickAragua:
Missing a unit called "Command Van (Standard)".

NickAragua:
That is an impressively large save game file, though. How many game years have you run this thing through?

SirMegaV:
Command Van missing? Will check it out.

Starter at 2400 now 2505.. 105 years? :)

SirMegaV:
@NickAragua, what do you mean by missing unit "Command Van (Standard)"? It can be found in the "purchase unit" window.

HammerGS:
Might me a custom unit for you.

NickAragua:
Correct, the Command Van (Standard) appears to be a custom unit.

SirMegaV:
The blk file is located under unofficial.zip/vehicles/Others/Command Van.blk

neoancient:
@NickAragua The MekHQ packaging utility builds the distribution using the data files from the MegaMek repository. While we occasionally update the HQ repository, it is usually several months behind MM, and changes in the unit files, RATs, unit sprites, etc. may not be available when you run it in Eclipse.

[Command Van.blk.zip](https://github.com/MegaMek/mekhq/files/1165848/Command.Van.blk.zip)

NickAragua:
Ah, thanks. Did not know that.

NickAragua:
Three months later, it's fixed! (in the upcoming nightly builds only)

NickAragua:
Should be fixed with PR #415

PR #415, opened 2017-07-25
#402 - Unit rating performance fix

NickAragua:
This change prevents the re-initialization and subsequent re-evaluation (in about 99% of the cases) of the unit rating mechanism. This drastically improves performance for large save games and also cuts down the clutter in the log file. 

We still re-calculate the unit rating every day, whenever we do something that affects the unit rating, etc, as well as whenever we re-initialize the campaign object for whatever reason. However, we only do it once, as opposed to 200 times. :)


neoancient:
I've requested review from @dericpage because he has done a lot of work on the unit ratings code.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8541,4 +8542,22 @@ public boolean checkYearlyRetirements() {
...
     }
+
+	/**
+	 * Returns the type of rating method as selected in the Campaign Options dialog. Lazy-loaded for performance.
+	 */
+	public IUnitRating getUnitRating() {
+		if(unitRating == null) {
+			UnitRatingMethod method = getCampaignOptions().getUnitRatingMethod();
+			
+		    if (method.equals(UnitRatingMethod.FLD_MAN_MERCS_REV)) {
```
dericpage: This should be flipped around against the off-chance that _method_ somehow ends up being null.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8541,4 +8542,22 @@ public boolean checkYearlyRetirements() {
         }
         return false;
     }
+
+	/**
+	 * Returns the type of rating method as selected in the Campaign Options dialog. Lazy-loaded for performance.
+	 */
```
dericpage: JavaDoc should probably mention that the default return value is a CampaignOpsReputation object.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -2821,8 +2821,11 @@ private void refreshFunds() {
...
-            String text = "<html><b>Dragoons Rating:</b> "
-                    + getCampaign().getUnitRating() + "</html>";
+        	// this is the one situation where we do want to refresh the rating,
+        	// as it means something has happened to influence it
+        	getCampaign().getUnitRating().reInitialize();
+        	
+            String text = "<html><b>Dragoons Rating:</b> " + getCampaign().getUnitRatingText() + "</html>";
             lblRating.setText(text);
         } else {
             lblRating.setText("");
```
dericpage: I believe that when this method was written, FMMr was the only options for a unit rating, which is why the rating label only ever gets a value when that is selected.  It seems that, now that we have COR as an alternative, we probably should put in an _else if_ condition to populate the label with it's unitRatingText value.

NickAragua:
Added text update. Also fixed a bug where the unit rating method wouldn't change if you switched unit rating methods on the campaign options interface.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+        int cost = 0;
+
+        cost += getPayRoll(getCampaignOptions().useInfantryDontCount());
+        cost += getMonthlySpareParts();
+        cost += getMonthlyFuel();
+        cost += getMonthlyAmmo();
+        return cost;
+    }
+
+    public long getMonthlySpareParts() {
```
dericpage: Can be private.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+        for (Unit u : getUnits()) {
+            if (u.isMothballed()) {
+                continue;
+            }
+            partsCost += u.getSparePartsCost();
+        }
+        return partsCost;
+    }
+
+    public long getMonthlyFuel() {
```
dericpage: Can be private.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+        for (Unit u : getUnits()) {
+            if (u.isMothballed()) {
+                continue;
+            }
+            fuelCost += u.getFuelCost();
+        }
+        return fuelCost;
+    }
+
+    public long getMonthlyAmmo() {
```
dericpage: Can be private.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+            if (u.isMothballed()) {
+                continue;
+            }
+            ammoCost += u.getAmmoCost();
+        }
+        return ammoCost;
+    }
+
+    public boolean isFactionComstar() {
+        if (getFactionCode().equals("CS")) {
```
dericpage: This should be flipped to put the constant on the right in case getFactionCode returns null.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+            case "CSL":
+            case "CWI":
+            case "CW":
+            case "CWE":
+            case "CWIE":
+            case "CWOV":
+            case "CEI":
+            case "RA":
+                return true;
+            default: return false;
```
dericpage: Should be on the following line.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8682,4 +8689,90 @@ public IUnitRating getUnitRating() {
...
+
+    public boolean isFactionComstar() {
+        if (getFactionCode().equals("CS")) {
+            return true;
+        }
+        return false;
+    }
+
+    public boolean isFactionClan() {
+        switch (getFactionCode()) {
```
dericpage: This has the potential to throw a NPE.  You should do a null check before the switch statement.

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
```diff
@@ -413,6 +413,8 @@ protected BigDecimal getNumberUnits() {
      */
     protected abstract int calculateUnitRatingScore();
 
+    public abstract UnitRatingMethod getUnitRatingMethod();
```
dericpage: Since you have this in the interface, you shouldn't put it in the abstract unless you're adding logic to it.

MekHQ/src/mekhq/campaign/Campaign.java
pheonixstorm: That was part of the peacetime costs feature

MekHQ/src/mekhq/campaign/Campaign.java
pheonixstorm: Same

MekHQ/src/mekhq/campaign/Campaign.java
pheonixstorm: Same

MekHQ/src/mekhq/campaign/Campaign.java
pheonixstorm: Same

MekHQ/src/mekhq/campaign/Campaign.java
pheonixstorm: Same, though honestly if faction returned null then there is an even bigger bug somewhere else. the faction code should never return null in this case. See line 340 of this file.

dericpage: While the faction code is not supposed to return a null, it's possible that future code could change this.  In any case, it's a best practice to do a null-check before invoking a switch statement.  Better to detect and gracefully handle the null along with logging a sensible error than allowing the switch statement to blow up with a generic NPE.

NickAragua: I decided to re-use the Faction.java mechanism for identifying whether a faction is clan, to save on duplicated logic. The campaign's current faction is always set to something by default (based on the campaign's constructor), so I don't think a null check is really worthwhile.

NickAragua:
Made requested changes.

MekHQ/src/mekhq/campaign/Campaign.java
dericpage: Unfortunately, setFactionCode doesn't contain null protection, so it is possible to set the factionCode null.  Technically, the better solution would, indeed, be to make sure that factionCode is only ever set through setFactionCode and add a null check there.

Issue #416, opened 2017-07-26
[43.2-git][AtB][bug] Base Attack scenarios fail to generate turrets


NickAragua:
When using AtB and the game generates a Base Attack scenario while on contract, the slots that are supposed to be turrets instead say "no random assignment table found for faction". Saving and loading the game empties those slots out completely, so no save game can be provided. However, it's trivial to simulate in code by changing the MekHQ.Campaign.Force.Lance.checkForBattle() method to just return AtBScenarioFactory.createScenario(c, this, AtBScenario.BASEATTACK, true, getBattleDate(c.getCalendar()));, then cycling the day past sunday.

It appears that the calls to getEntityByName(randomGunEmplacement(), [snip]); fail to return a proper entity. 

I'm planning on working on this myself, but entered this issue to keep track of it.

pheonixstorm:
I wonder when this issue started happening. Last I played AtB it worked, but that was before all the changes were made to the scenarios and maybe the rat tables

NickAragua:
Fixed with PR #417

PR #417, opened 2017-08-01
Issue #416: AtB turret generation and base attack scenario fix

NickAragua:
A somewhat crude implementation of a turret generation mechanism for base attack scenarios (and now usable for any other scenarios in the future which may want turrets).  I'm not aware of a way that turrets can be generated from the standard RAT generator, so I chose to imitate the mechanism used to add civilian units instead.

Notes: the AtB rules state that the civilian unit count on base attack missions should be 8, so 8 it is.

Also: This fix is dependent on pull request #415 , as it uses the updated unitGenerator.getUnitRatingName() function.


neoancient:
AtB used to generate turrets using a crude hard-coded method based tables provided by Makinus, who wrote the AtB rules. I've requested code review by kipstafoo since he has be reworking the AtB scenario generation code.

kipstafoo:
So, two comments:
- It looks like the turrets have been changed to be part of the civilian force, instead of the defending NPC force. Was this a conscious change? If so, why? I don't see a problem with it, I'm just interested as to why.
- It looks like the randomGunEmplacement() method is now unnecessary. Is there a reason to keep it in the scenario class?

NickAragua:
For #1, the AtB rules state that the "base force" during a base attack gets 8 civilian units and 6 turrets. So, I'm following the text. It doesn't really matter either way, as the pilots are generated at the opfor skill and tech level.

For #2, no I don't think there is a reason to keep it around. I'll get rid of it.

NickAragua:
I think I'd like to keep the randomGunEmplacement() method around but commented out for historical purposes.

kipstafoo:
Ok, I'm good with it being part of the civs. Like I said, it really doesn't matter as long as the turrets target the right 'enemy'.

For the randomGunEmplacement(), let's nuke it please instead of commenting it out. Git maintains history for us, so no sense in muddying the code with unreachable/unused code. I'm a stickler for clean code. ;) And that you for removing the unused imports earlier... that drives me utterly bonkers.

NickAragua:
No idea why the diff looks so weird, but commented out code removed.

NickAragua:
Using my amazing powers of prediction (wooooooooo), I see a future where Bloodwolf has created more turret RATs, rendering the original fix obsolete. The new fix accounts for the possibility of many different combinations of turret RATs, just as long as they follow the pattern of "Turret YYYY Q", where YYYY is a year and Q is the single-letter quality level.

BLOODWOLF333:
K, Yeah when I get around to grinding through the clan turret RATs they will all be in the same format.

PR #430, opened 2017-08-31
Mrms improvements

kipstafoo:
Refactor MR/MS/MRW and move much of the code to a service level class from the dialog. 
 o Fix multiple bugs related to 'impossible' fixes such as damaged hip/shoulders, missing parts, etc...
 o Make the process much more efficient by not making checks that aren't necessary over and over.
 o Add additional color-coded messaging when parts can't be fixed due to configuration or
   existing techs.


kipstafoo:
@neoancient I commented out your 'PodSpace' changes. They were blocking other repairs from happening and I don't think they were doing what you wanted them to do. Feel free to talk to me and we can re-enable them.

Issue #437, opened 2017-09-07
ATB ambush mission generates the incorrect amount of enemy units


Question2:
By it's description, its supposed to generate 3 enemy units (and used to) but it just generated 11 enemy units for me.

[#battletech.zip](https://github.com/MegaMek/mekhq/files/1283682/battletech.zip)

Bonepart:
I ran into this myself, I just forgot what the scenario type was, so when I did some testing I didn't see any problems. I'll look at the Ambush scenario when I get home tonight.

Bonepart:
A quick look at the obvious trouble spots doesn't present any problems. The Ambush scenario is flagged as a special mission and generates special mission forces, so the refresh should change the opFor to match the unit deployed. I'll have to actually run it in debug when I get home to try to see where it goes wrong. I'll try to test the other special mission scenarios as well to make sure none of the others have issues.

HammerGS:
Closed as merges.

Issue #410, opened 2017-07-17
Allied traitors scenario no longer working


kipstafoo:
With the redone scenario code, the Allied Traitors scenario is no longer working. It is not properly selecting the appropriate enemies for the assigned unit and is instead selecting all weight class enemies.

I would imagine that this applies to all 'special mission' type of scenarios.

Bonepart:
https://github.com/MegaMek/mekhq/issues/437 I suspect these are related. I know at least one of the special missions works, as when I did my testing I let the system randomly pick one and selected for the unit deployed okay.

HammerGS:
Closed as merged.

PR #438, opened 2017-09-08
Fix for issue #437 and #410

Bonepart:
Fix for https://github.com/MegaMek/mekhq/issues/437 & https://github.com/MegaMek/mekhq/issues/410

reinitialized enemyEntities in between loops


Bonepart:
Went through and tested each special mission scenario, and they now all load the appropriate forces and change them for the unit deployed.

PR #439, opened 2017-09-08
Calculate DropShip rental costs based on untransported units only

jslater89:
This fixes a minor bug in the drop ship requirement calculations. Prior to this commit, the raw count of mechs, vehicles, and ASFs was used to determine DropShip requirements. This commit uses the count of untransported units, which is already calculated earlier in the method.

The attached saved game demonstrates the problem and verifies the solution. It contains a company of 12 mechs with a few attached ASFs, and has a Colossus and a JumpShip. The transport costs in the released dev snapshot are nonzero, despite the unit fitting entirely in its own transport assets. The transport costs in the version built with this commit are 0, as expected.

https://drive.google.com/file/d/0B4_Fe7AvNpDLTEdvSmpoelN5WTQ/view?usp=sharing

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/Contract.java
+++ b/MekHQ/src/mekhq/campaign/mission/Contract.java
@@ -408,6 +408,15 @@ public class Contract extends Mission implements Serializable, MekHqXmlSerializa
         return getLength() + travelMonths;
     }
 
+    /**
+     * Get the estimated total profit for this contract. The total profit is the total contract
+     * payment including fees and bonuses, minus overhead, maintenance, payroll, spare parts,
+     * and other monthly expenses. The duration used for monthly expenses is the contract duration
+     * plus the travel time from the unit's current world to the contract world and back.
+     *
+     * @param c The campaign with which this contract is associated.
+     * @return The estimated profit in C-bills.
+     */
     public long getEstimatedTotalProfit(Campaign c) {
         long profit = getTotalAmountPlusFeesAndBonuses();
         profit -= c.getOverheadExpenses() * getLengthPlusTravel(c);
@@ -426,6 +435,13 @@ public class Contract extends Mission implements Serializable, MekHqXmlSerializa
 		return profit;
 	}
 
+    /**
+     * Get the number of months left in this contract after the given date. Partial months are counted as
+     * full months.
+     *
+     * @param date
+     * @return
+     */
 	public int getMonthsLeft(Date date) {
 		GregorianCalendar cal = new GregorianCalendar();
 		cal.setTime(date);
@@ -445,50 +461,50 @@ public class Contract extends Mission implements Serializable, MekHqXmlSerializa
 	 * the ink is signed, which is a no-no.
 	 * @param c
 	 */
-	public void calculateContract(Campaign c) {
+    public void calculateContract(Campaign c) {
 
-		//calculate base amount
-		baseAmount = (long)(paymentMultiplier * getLength() * c.getContractBase());
+        //calculate base amount
+        baseAmount = (long)(paymentMultiplier * getLength() * c.getContractBase());
 
-		//calculate overhead
-		switch(overheadComp) {
-		case OH_HALF:
-			overheadAmount = (long)(0.5 * c.getOverheadExpenses() * getLength());
-			break;
-		case OH_FULL:
-			overheadAmount = (long)(1 * c.getOverheadExpenses() * getLength());
-			break;
-		default:
-			overheadAmount = 0;
-		}
+        //calculate overhead
+        switch(overheadComp) {
+        case OH_HALF:
+            overheadAmount = (long)(0.5 * c.getOverheadExpenses() * getLength());
+            break;
+        case OH_FULL:
+            overheadAmount = (long)(1 * c.getOverheadExpenses() * getLength());
+            break;
+        default:
+            overheadAmount = 0;
+        }
 
-		//calculate support amount
-		if (c.getCampaignOptions().usePeacetimeCost()
-		        && c.getCampaignOptions().getUnitRatingMethod().equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
-		    supportAmount = (long)((straightSupport/100.0) * c.getPeacetimeCost() * getLength());
-		} else {
-		    long maintCosts = 0;
-	        for (Unit u : c.getUnits()) {
-	            if (u.getEntity() instanceof Infantry && !(u.getEntity() instanceof BattleArmor)) {
-	                continue;
-	            }
-	            maintCosts += u.getWeeklyMaintenanceCost();
-	        }
-	        maintCosts *= 4;
-	        supportAmount = (long)((straightSupport/100.0) * maintCosts * getLength());
-		}
+        //calculate support amount
+        if (c.getCampaignOptions().usePeacetimeCost()
+                && c.getCampaignOptions().getUnitRatingMethod().equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
+            supportAmount = (long)((straightSupport/100.0) * c.getPeacetimeCost() * getLength());
+        } else {
+            long maintCosts = 0;
+            for (Unit u : c.getUnits()) {
+                if (u.getEntity() instanceof Infantry && !(u.getEntity() instanceof BattleArmor)) {
+                    continue;
+                }
+                maintCosts += u.getWeeklyMaintenanceCost();
+            }
+            maintCosts *= 4;
+            supportAmount = (long)((straightSupport/100.0) * maintCosts * getLength());
+        }
 
-		//calculate transportation costs
-		if(null != Planets.getInstance().getPlanetById(planetName)) {
+        //calculate transportation costs
+        if(null != Planets.getInstance().getPlanetById(planetName)) {
             JumpPath jumpPath = c.calculateJumpPath(c.getCurrentPlanet(), getPlanet());
 
-			// FM:Mercs transport payments take into account owned transports and do not use CampaignOps dropship costs.
-			// CampaignOps doesn't care about owned transports and does use its own dropship costs.
-			boolean campaignOps = c.getCampaignOptions().useEquipmentContractBase();
-			transportAmount = (long)((transportComp/100.0) * 2 * c.calculateCostPerJump(campaignOps, campaignOps) * jumpPath.getJumps());
-		}
+            // FM:Mercs transport payments take into account owned transports and do not use CampaignOps dropship costs.
+            // CampaignOps doesn't care about owned transports and does use its own dropship costs.
+            boolean campaignOps = c.getCampaignOptions().useEquipmentContractBase();
+            transportAmount = (long)((transportComp/100.0) * 2 * c.calculateCostPerJump(campaignOps, campaignOps) * jumpPath.getJumps());
+        }
 
-		//calculate transit amount for CO
+        //calculate transit amount for CO
         if (c.getCampaignOptions().usePeacetimeCost()
                 && c.getCampaignOptions().getUnitRatingMethod().equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
             //contract base * transport period * reputation * employer modifier
--- a/MekHQ/src/mekhq/campaign/mission/Contract.java
+++ b/MekHQ/src/mekhq/campaign/mission/Contract.java
@@ -20,6 +20,21 @@
  */
 package mekhq.campaign.mission;
 
+import megamek.common.BattleArmor;
+import megamek.common.Infantry;
+import mekhq.MekHqXmlSerializable;
+import mekhq.MekHqXmlUtil;
+import mekhq.Utilities;
+import mekhq.campaign.Campaign;
+import mekhq.campaign.JumpPath;
+import mekhq.campaign.unit.Unit;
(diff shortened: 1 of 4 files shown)


jslater89:
That's a fair point, and one I hadn't considered, but as it stands, there's no incentive to supply your own transport, either. Obtaining a DropShip with the current rules will always make your transport significantly more expensive—you pay for the calculated number of collars, then for any dropships you actually own. That seems counter-intuitive to me, a worst-of-both-worlds sort of situation.

I'd be willing to do the work to make transport terms cover spacecraft maintenance for the duration of the contract, or in the case of JumpShips, maybe just the duration of the trip. (JumpShips seem like an extreme outlier to me, though, either way.) Then again, owned spacecraft are also covered by straight support, and by the percentage of purchase cost base payment if that option is enabled.

The TLDR version is that this is an area I'm willing to put in the work to improve, so if we can come to some kind of consensus, I'm happy to code it up.

HammerGS:
Their are updated rules for this in Campaign Ops. Pg 43. I think wherever possible we should evolve to those rules.

jslater89:
That's a good place to kick off for rented transport, but doesn't (seem) to cover company-owned transport. (It does seem reasonable to update the travel costs with those assumptions in mind.)

The Base Payment section on p41 remarks that the base monthly payment should be 75% of the peacetime operating cost (maintenance, salaries, overhead?), plus 5% of total cost of combat units and 1% of total cost of large spacecraft exposed to combat conditions (generally DropShips only).

Would it be reasonable to add that as a new option under the 'Mercenary' tab of the campaign options for payment calculation? That way the base payment will be more or less fair for mercenary companies with and without transport.

pheonixstorm:
The 1% option for dropship contract combat payment is already finished but has not been submitted as well as all the other peacetime operating cost rules.

As for transport reimbursement. CO does not reimburse for transports owned by the unit, only for what you have to pay out for. This was a change from the old FMMercs rules which should still use that rule.

kipstafoo:
There are a lot of things I like about CO... this is one of the things I don't. I forgot about that and that explains why my TT campaign still uses better/logical payments.

jslater89:
So at present, the following seems to be true:

1. CO-type payments (75% operating costs plus percentages of unit cost) are finished. CO payment rules don't require any consideration for owned-transport travel costs, because the operating cost is counted in the base payment.
2. Per FM:Mercs, employers  _should_ reimburse for travel costs in owned transport, but doesn't appear to do so now. (With the fix in this commit, travel costs are given as 0, and no reimbursement is provided; without the fix in this commit, you pay for upkeep for your own DropShip as well as jump ship collar fees, as well as for the calculated DropShip rental cost and however many collars the calculated DropShips are assumed to take.)

So, where does that leave these changes? Should I do nothing? Change the travel costs to match the CO rules more closely? Add large spacecraft maintenance to travel costs to match the FM:Mercs rules, which can then be superseded by CO rules if that option is selected?

On a similar note, the percentage-based maintenance costs for large spacecraft are pretty staggering. It might be worth revisiting those if it's desired to make it feasible to play without GM help or crazy large contracts to support company-owned transport. Or maybe look at implementing long-term contract auxiliaries per the CO rules.

pheonixstorm:
Percentage based maintenance is unofficial and should not be part of the PR as it is something completely different. But, back to the rest of it.

1. And this is also for you kip, the contracts for CO may not include transport cost the way FMMercs did but they add something different. You get paid for the transport time. So a 4 month trip from one location to another you get paid for the 8 months (roundtrip) of travel time. The transport rate is a little different from the main contract rate, but is close to what you might see for rented transport. The rules in FMMercs for owned transport never made sense to me though I did like the extra C-bills

2. I will have to reread that section of the rules, but I do know that the code you are changing was never really finished and was a pure hack job (as stated in its comments). 

For these changes, take all the feedback you can and... I don't know. At least beyond not touching anything outside of that function. So keep whatever you change to just calculateCostPerJump. I believe that the reimbursement and cost for rented transport is the same, or nearly the same, between the two rulesets. You could add a new variable to pass check is someone is using the CO or FMM rating system and make changes accordingly.

pheonixstorm:
Having thought about it a bit more now would be a good time to go ahead and actually fix some of the issues with this such as taking into account for owned transports and doing a correct calculation for those transports and then use the hack forumla for whatever the unit cannot transport. If a unit has a cargo transport such as a mule you could then push whatever is left into that based on available cargo space. I think CO and FMM calculate the rate the same still. Again, would need to check both sets of rules to be sure.

NickAragua:
One thing that I'd like to see changed in general (may be good to do for this PR while you're at it) is that jump costs don't seem to count cargo (either mothballed units or the boatloads of ammo and spare parts a unit transports). If you look at the code, the "mothballedAsCargo" variable is ignored completely, and the "cargoSpace" variable is declared/initialized, but not calculated.

jslater89:
@NickAragua I may as well do that while I'm here, yeah.

I think my plan, then, is this:

1. Improve the calculateCostPerJump function somewhat, to take into account owned transport and cargo. Also add a flag to swap between FM:Mercs and CO costs for the hacky method.
2. Add an option under the FM:Mercs payment option in settings for transport rights covering owned transport. (Separate pull request?)
3. Take a look at dropship and jumpship percent maintenance and try to bring them a little more in line with known transport costs. (Separate pull request.)

pheonixstorm:
1. that would be fine
2. No need, this function should process the boolean excludeOwnTransports but doesn't do anything with it. In the contract code and newDay() code you can (should) pass true or fasle into this function for actual payment vs contract reimbursement so there *shouldn't be a reason to add extra functions or more code beyond what can be done in this function
3. Will have to comment more on this later after I get back home.

jslater89:
Oh, I totally missed the parameter. That does make it simpler.

Something would have to be arranged elsewhere in the codebase for FM:Mercs' broader transport rights, though, wouldn't it? This change does have the side-effect of making transport rights a lot less useful, especially if you own a JumpShip.

pheonixstorm:
calculateCostPerJump is called from several places. First is newDay() also in Campaign.java on line 169(ish) which sets excludeOwnTransports as true. The next two are in Contract.java calculateContract() line 445 and getEstimatedTotalProfit() line 391.

The final spot is JumpPathViewPanel.java line 271 but as I never use this option I haven't a clue, but best guess is it shows the cost in the map section.

So, this function should calculate this as exclude owned transport for CO rating with 100,000 c-bill per collar per jump cost and include owned transport with 50,000 c-bill per collar per jump for FMMercs rating. That should take care of the contract reimbursement issue and still allow for actual expenses in newDay() to continue as normal. There are other aspects that will need to be addressed eventually such as cargo unless you want to try to get it all out of the way at once. You might want to talk to whoever wrote the percentage maintenance code before messing with it. I have no clue why it was written the way it was, but there could be a reason it is as high as it is.

jslater89:
Getting close to done with this.

Reading through the FM:Mercs manual, the example calculation for transit costs on p. 166 suggests that the DropShip cost is a one-time expense per trip, rather than a cost per jump. Similarly, Campaign Ops gives DropShip cost as a cost per month rather than a cost per jump. Is that best handled elsewhere? Or should calculateCostPerJump take parameters for number of jumps and number of months and handle all the transport cost stuff, since it currently contains the code for 'number of roughly average dropships to use'?

jslater89:
The new changes implement Campaign Ops travel costs (0.5% of dropship purchase price per month for dropships, 100,000 per jumpship collar), as well as travel rights reimbursements for owned transport for FM:Mercs rules (by way of correctly implementing the excludeOwnTransports rule).

With percent maintenance and owned transport reimbursement, the contract generator can roll some absurdly high-value contracts for mercenaries with their own jumpships, on the order of hundreds of millions of C-bills, because of the interaction between straight support and transport rights. Given that the percent maintenance rules are unofficial and mercenaries with jumpships are something of a rarity anyway, I don't know how much to be concerned about it.

pheonixstorm:
Maintenance should not be included in this. That would be part of the Straight Support section of contracts for both FMMercs as well as CO

jslater89:
Basing the cost of owned transport on the cost of leased transport makes for more reasonable contracts.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
+			overheadAmount = 0;
+		}
+
+		//calculate support amount
+		if (c.getCampaignOptions().usePeacetimeCost()
+		        && c.getCampaignOptions().getUnitRatingMethod().equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
+		    supportAmount = (long)((straightSupport/100.0) * c.getPeacetimeCost() * getLength());
+		} else {
+		    long maintCosts = 0;
+	        for (Unit u : c.getUnits()) {
```
NickAragua: Please fix the indentation.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
+	            }
+	            maintCosts += u.getWeeklyMaintenanceCost();
+	        }
+	        maintCosts *= 4;
+	        supportAmount = (long)((straightSupport/100.0) * maintCosts * getLength());
+		}
+
+		//calculate transportation costs
+		if(null != Planets.getInstance().getPlanetById(planetName)) {
+            JumpPath jumpPath = c.calculateJumpPath(c.getCurrentPlanet(), getPlanet());
```
NickAragua: Indentation.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
+
+			// FM:Mercs transport payments take into account owned transports and do not use CampaignOps dropship costs.
+			// CampaignOps doesn't care about owned transports and does use its own dropship costs.
+			boolean campaignOps = c.getCampaignOptions().useEquipmentContractBase();
+			transportAmount = (long)((transportComp/100.0) * 2 * c.calculateCostPerJump(campaignOps, campaignOps) * jumpPath.getJumps());
+		}
+
+		//calculate transit amount for CO
+        if (c.getCampaignOptions().usePeacetimeCost()
```
NickAragua: Indentation.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
+		if(null != c.getPlanet(planetName) && c.getCampaignOptions().payForTransport()) {
+			JumpPath jumpPath = c.calculateJumpPath(c.getCurrentPlanet(), getPlanet());
+
+			boolean campaignOps = c.getCampaignOptions().useEquipmentContractBase();
+			profit -= 2 * c.calculateCostPerJump(campaignOps, campaignOps) * jumpPath.getJumps();
+		}
+		return profit;
+	}
+
+	public int getMonthsLeft(Date date) {
```
NickAragua: Let's make a javadoc-style comment that explains what this function does.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
+    private int getLengthPlusTravel(Campaign c) {
+        int travelMonths = (int) Math.ceil(getTravelDays(c) / 30.0) * 2;
+        return getLength() + travelMonths;
+    }
 
-        // calculate support amount
-        if (c.getCampaignOptions().usePeacetimeCost() && c.getCampaignOptions().getUnitRatingMethod()
-                .equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
-            supportAmount = (long) ((straightSupport / 100.0) * c.getPeacetimeCost() * getLength());
+    public long getEstimatedTotalProfit(Campaign c) {
```
NickAragua: Public method needs javadoc comment.

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -391,88 +392,107 @@ public boolean test(int codePoint) {
...
-                .equals(mekhq.campaign.rating.UnitRatingMethod.CAMPAIGN_OPS)) {
-            // contract base * transport period * reputation * employer modifier
-            transitAmount = (long) (c.getContractBase()
-                    * (((c.calculateJumpPath(c.getCurrentPlanet(), getPlanet()).getJumps()) * 2) / 4)
-                    * (c.getUnitRatingMod() * .2 + .5) * 1.2);
+		if(null != c.getPlanet(planetName) && c.getCampaignOptions().payForTransport()) {
+			JumpPath jumpPath = c.calculateJumpPath(c.getCurrentPlanet(), getPlanet());
+
+			boolean campaignOps = c.getCampaignOptions().useEquipmentContractBase();
+			profit -= 2 * c.calculateCostPerJump(campaignOps, campaignOps) * jumpPath.getJumps();
```
NickAragua: Why multiply by 2?

jslater89: To follow the pattern elsewhere in Contract. I assume the reasoning there is that the transport payment pays for travel from a company's homeworld to the contract world, and from the contract world back to the homeworld.

That could well be wrong. If so, I can change that in this pull request, or it can be a follow-on pull request.

jslater89:
Changes made.

Issue #429, opened 2017-08-31
Switching ammo does not update the UI


pheonixstorm:
This happens in both the hangar and repair tabs. There are no errors in the log and reloading the campaign will fix it.

![ammo](https://user-images.githubusercontent.com/1787879/29905210-399fb470-8ddb-11e7-9629-96b76dcbfa00.png)

jslater89:
I poked at this and saw the following behavior:

1. Swap ammo generates a repair task.
2. Until the repair task is completed, the old ammo type displays with the full number of shots. 
3. If the game is reloaded, the new ammo type displays with zero shots.

So it's inconsistent, but I'm not sure in which direction to fix it.

NickAragua:
Fixed with PR #448

PR #448, opened 2017-09-11
Fix for issue #429: change ammo type and unload on changeMunition action

jslater89:


neoancient:
If you unload the bins automatically when the ammo is changed it does not take any tech time. According to StratOps, p. 186 the amount of time required to swap ammo is twice the time it takes to reload. In fact `AmmoBin` already accounts for this in `getBaseTime()` and handles the unloading of the old ammo and load of the new in `loadBin()`, which is called by `fix()`. It looks to me like the only thing lacking is a check to see if the munition type of the bin is the same as the munition type of the `Entity`in `needsFixing()` to make it show up in the list of tech tasks.

jslater89:
I'll switch that around, then.

Note that leaves @pheonixstorm's bug unfixed; it's just that the bug is the other way around from what I thought. On saving and restoring a campaign, a unit with an ammo swap scheduled loads with the old ammo unloaded and the new ammo type selected but not currently loaded. Swap ammo then an immediate 'load ammo' maintenance task is correctly 30 minutes; swap ammo, save, reload, and load ammo is only 15 minutes.

pheonixstorm:
The bug is graphics related as the UI doesn't get updated. Never checked the issue on swap reload times so if that is bugged as well then that would be a separate issue.

jslater89:
I don't think the current functionality represents a bug as such. Changing the ammo type in the technical readout before it's unloaded seems wrong to me--it's no longer an accurate depiction of the mech. I bet I could make the ammo line yellow, though, to signify that there's work to be done there.

pheonixstorm:
I think that would be more ideal, so long as the UI does a proper update. I think several of the campaign reloadable bugs are like that rather than functionality bugs.

neoancient:
I was working from my phone yesterday and I assumed that the failure to update was that a repair task did not appear, but in testing I see it does. So assume (and this conversation would seem to confirm it) that the expected behavior is that the change is not reflected in the unit status. The reason for that is that nothing is actually wrong with the unit, though I can see why it would be beneficial to indicate a scheduled replacement of the munition type. As for saving and restoring, I'm sure that comes from a call to `updateConditionFromPart()` which sets the munition on the `Entity` from what was saved in the `AmmoBin`, effecting a removal of the previous ammo without a tech action. I won't swear to this, but I think it will work if we remove the line that assigns the ammo type. I looked through all the places that call that method and I didn't see anything I thought would cause a problem.

jslater89:
I've just about got the save/load bug sorted, but the original display issue is nontrivial to fix, it looks like. Since the technical readout is generated over in Megamek, there's no straightforward way to base it on the AmmoBin, and it seems prudent to avoid editing the Entity until the reload has actually taken place.

MekHQ/src/mekhq/campaign/parts/equipment/AmmoBin.java
```diff
@@ -474,7 +491,7 @@ public void updateConditionFromPart() {
 				mounted.setHit(false);
 		        mounted.setDestroyed(false);
 		        mounted.setRepairable(true);
-		        mounted.changeAmmoType((AmmoType) type);
+		        //mounted.changeAmmoType((AmmoType) type);
```
NickAragua: Let's get rid of the commented-out code, it's just clutter and we can look it up in source control if we need it later.

jslater89:
Change made.

PR #451, opened 2017-09-13
IO Tech

neoancient:
Replaces hard-coded tech base, availability, and tech level values for `Part`s with the new MM tech advancement system.

New campaign options:
Variable tech level (uses MM game option): use the production state of the part/unit in the campaign year to determine its tech level.
Use faction intro dates: if checked, will determine availability and tech level of parts based on the player's faction. For units this requires recalculating the `CompositeTechLevel` of each unit. This is done as a background task any time the campaign's faction code is changed, which is normally only once at startup.


MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -1275,12 +1283,34 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
...
+        gridBagConstraints.anchor = java.awt.GridBagConstraints.NORTHWEST;
+        panTech.add(variableTechLevelBox, gridBagConstraints);
+
+        factionIntroDateBox.setText(resourceMap.getString("factionIntroDateBox.text")); // NOI18N
+        factionIntroDateBox.setToolTipText(resourceMap.getString("factionIntroDateBox.toolTipText")); // NOI18N
+        factionIntroDateBox.setName("factionIntroDateBox"); // NOI18N
+        gridBagConstraints = new java.awt.GridBagConstraints();
+        gridBagConstraints.gridx = 0;
+        gridBagConstraints.gridy = 8;
+        gridBagConstraints.gridwidth = 2;
```
dericpage: The previous control was at gridy 7 with a width of 2 and this is on gridy 8 also with a width of 2.  Shouldn't this one be on gridy 9 to avoid overlap?

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -1275,12 +1283,34 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
...
+        gridBagConstraints.anchor = java.awt.GridBagConstraints.NORTHWEST;
+        panTech.add(factionIntroDateBox, gridBagConstraints);
+
         useAmmoByTypeBox.setText(resourceMap.getString("useAmmoByTypeBox.text")); // NOI18N
         useAmmoByTypeBox.setToolTipText(resourceMap.getString("useAmmoByTypeBox.toolTipText")); // NOI18N
         useAmmoByTypeBox.setName("useAmmoByTypeBox"); // NOI18N
         gridBagConstraints = new java.awt.GridBagConstraints();
         gridBagConstraints.gridx = 0;
-        gridBagConstraints.gridy = 7;
+        gridBagConstraints.gridy = 9;
```
dericpage: Same here.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8570,4 +8601,87 @@ public boolean checkYearlyRetirements() {
...
     }
+
+    @Override
+    public int getTechIntroYear() {
+        if (campaignOptions.limitByYear()) {
+            return calendar.get(Calendar.YEAR);
+        } else {
+            return 0;
+        }
+    }
```
dericpage: I don't see this method being used anywhere, so this may not matter, but should we use a named constant rather than just 0 if limitByYear is turned off?

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
neoancient: width affects horizontal spacing. gridy is vertical positioning.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
neoancient: ditto

MekHQ/src/mekhq/campaign/Campaign.java
neoancient: Now that I look at this I see I got the logic backwards. The zero is there because it's a literal zero. Any tech that is introduced at a year >= 0 is legal. But it's actually the other way around, and it's only legal if 0 >= the intro year, which is not going to work.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
dericpage: Gah, you're right.

MekHQ/src/mekhq/campaign/Campaign.java
neoancient: Yup, that's what I did. Changed to Integer.MAX_VALUE, which will work until the in-game universe gets a bit past the year two billion.

PR #453, opened 2017-09-15
MRMS overhaul

kipstafoo:
- Completely overhaul how messaging is done in MRMS to make significantly less 'noisy' and to give a better post-action summary
- Overhaul how parts and techs are determined to make system much faster
- Fix several bugs that made the process stop prematurely.
- Rush jobs apply a reduction on the experience on the tech. This results in the MRMS system rushing jobs with low BTHs but then making it too hard for the tech to do. The system now better handles when it moves a job to a 'rush' mode.
- Fix bug in MRW where it was building the list of available MR types incorrectly and therefore not applying configuration properly but also not allow many parts to be fixed via MRW
- Fix newly introduced bug (from last changeset) that resets the 'mode' of a part to 'normal' before calculating which techs can handle it.
- Added ability to find salvageble parts/parts needing fixing that aren't currently being worked on.
- [Unrelated] Add name of unit to repair/scrap messages to help understanding which 'Left Arm' didn't get repaired.


MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -179,85 +210,218 @@ public int compare(Unit o1, Unit o2) {
...
+		// Filter our tech list to only our techs that can work on this unit
+		for (int i = techs.size() - 1; i >= 0; i--) {
+			Person tech = techs.get(i);
 
-			for (int i = 0; i < mroList.size(); i++) {
-				mroByTypeMap.put(mroList.get(i).getType(), mroList.get(i));
+			if (!tech.canTech(unit.getEntity())) {
+				techs.remove(i);
 			}
+		}
```
dericpage: It's generally a bad idea to modify a `Collection` while you're iterating over it.  Better options would be either
1. Create an `availableTechs` `Collection` to contain the techs you want to use or
2. Create a `removeTechs` `Collection` and then call `techs.removeAll(removeTechs)` after the loop.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -274,46 +438,41 @@ private static int performUnitMassTechAction(CampaignGUI campaignGUI, Unit unit,
...
+			//
+			// if (m != null && m.isReplacementAvailable()) {
+			// continue;
+			// }
+			// }
+			//
+			// temp.add(p);
+			// }
+			//
+			// parts = temp;
```
dericpage: If we're no longer using this code, we should probably just remove it and allow the repository's history track what used to be here.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -673,36 +854,89 @@ private static boolean repairPart(CampaignGUI campaignGUI, IPartWork partWork, U
...
 			}
-			
-			if (partWork instanceof PodSpace) {
-				continue;
-			}
-			
+
+			/*
+			 * if (partWork instanceof PodSpace) { continue; }
+			 */
```
dericpage: Again, should probably just be removed.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -780,61 +1013,131 @@ private static boolean checkArmorSupply(IPartWork part) {
...
 			}
 		}
 
 		return new WorkTimeCalculation();
 	}
 
+	private static void debugLog(String msg, Object... replacements) {
+		if (!debugMode) {
+			return;
+		}
```
dericpage: Instead of using a separate debugMode variable, you can use `if (!MehKQ.getLogger().willLog(getClass(), LogLevel.DEBUG) {` for the same purpose.  This way, if the _log4j.xml_ config gets updated to change the logging level on this class, you're already accounted for.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -780,61 +1013,131 @@ private static boolean checkArmorSupply(IPartWork part) {
...
+
+		if ((null != replacements) && (replacements.length > 0)) {
+			msg = String.format(msg, replacements);
+		}
+
+		MekHQ.getLogger().log(MassRepairService.class, "[DEBUG]", LogLevel.DEBUG, msg);
```
dericpage: Where you're passing `"[DEBUG]"`, you should be passing the name of the originating method.  The logger will already include the `LogLevel` in the logging statement written out to the file.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -889,6 +1192,243 @@ public int compare(Person tech1, Person tech2) {
 		}
 	}
 
+	public static class MassRepairPartAction {
+		public interface STATUS {
+			public static final int REPAIRED = 0;
+			public static final int MAX_SKILL_REACHED = 1;
+			public static final int MRO_DISABLED = 2;
+			public static final int NO_TECHS = 3;
+		}
```
dericpage: Seems like this should be an `enum`.  That way you can limit the `status` argument in your constructor to just these values.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -889,6 +1192,243 @@ public int compare(Person tech1, Person tech2) {
...
+
+	public static class MassRepairUnitAction {
+		public interface STATUS {
+			public static final int NO_ACTIONS = 0;
+			public static final int ACTIONS_PERFORMED = 1;
+			public static final int NO_TECHS = 2;
+			public static final int UNFIXABLE_LIMB = 3;
+			public static final int NO_PARTS = 4;
+			public static final int ALL_PARTS_IN_PROCESS = 5;
+		}
```
dericpage: Looks like another `enum` candidate.

MekHQ/src/mekhq/campaign/parts/Part.java
```diff
@@ -1058,8 +1072,16 @@ public boolean canNeverScrap() {
...
+		
+		if (null == getUnit()) {
+			msg = getName() + " scrapped.";
+		} else {
+			msg = getName() + " on " + unit.getName() + " scrapped.";
+		}
+		
 		remove(false);
-		return getName() + " scrapped.";
+		return msg;
```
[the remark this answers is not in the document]
kipstafoo: Actually, that won't work. The call to remove() is done last, after the msg is built, because the remove() call sets the unit to null, thus you'll never have the unit's name in the message.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -556,29 +556,37 @@ public void runDiagnostic(boolean checkForDestruction) {
...
         ArrayList<IPartWork> salvageParts = new ArrayList<IPartWork>();
         for(Part part: parts) {
-            if(part.isSalvaging()) {
+            if(part.isSalvaging() && (!onlyNotBeingWorkedOn || (onlyNotBeingWorkedOn && !part.isBeingWorkedOn()))) {
                 salvageParts.add(part);
             }
         }
         for (PodSpace pod : podSpace) {
-            if (pod.hasSalvageableParts()) {
+            if (pod.hasSalvageableParts() && (!onlyNotBeingWorkedOn || (onlyNotBeingWorkedOn && !pod.isBeingWorkedOn()))) {
```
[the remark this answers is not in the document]
kipstafoo: Fair enough... change made.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -86,39 +88,39 @@ public static int performWarehouseMassRepair(List<IPartWork> selectedParts, List
...
-						totalActionsPerformed++;
-					}
+					partSet.addPartAction(
+							repairPart(campaignGUI, part, null, validTechs, mroByTypeMap, configuredOptions, true));
 				}
 			}
 		}
 
-		return totalActionsPerformed;
+		return null;
```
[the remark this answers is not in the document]
kipstafoo: Actually, this is a bug. It shouldn't return null as that'll throw an NPE downstream. This method is being re-factored.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -129,27 +131,56 @@ public static void performSingleUnitMassRepairOrSalvage(CampaignGUI campaignGUI,
...
+		case MassRepairUnitAction.STATUS.NO_TECHS:
+			msg += " No actions were performed because there are currently no valid techs.";
+			break;
+
+		case MassRepairUnitAction.STATUS.UNFIXABLE_LIMB:
+			msg += " No actions were performed because there is at least one unfixable limb and configured settings do not allow location repairs.";
+			break;
+		}
 
 		campaignGUI.getCampaign().addReport(msg);
```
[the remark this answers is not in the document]
kipstafoo: I disagree. There is, at most, 2 concats here (ignoring the calls to String.format() which the StringBuilder won't help with anyway. StringBuilders come into their own when append in a loop.

Where a StringBuilder would be useful (and is now in use) is in the massRepairSalvageUnits() method.

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -179,85 +210,218 @@ public int compare(Unit o1, Unit o2) {
...
+
+		generateCampaignLogForUnitStatus(unitActionsByStatus, MassRepairUnitAction.STATUS.NO_PARTS,
+				"Units with no valid parts:", campaignGUI);
+		generateCampaignLogForUnitStatus(unitActionsByStatus, MassRepairUnitAction.STATUS.NO_TECHS,
+				"Units with no valid techs:", campaignGUI);
+		generateCampaignLogForUnitStatus(unitActionsByStatus, MassRepairUnitAction.STATUS.UNFIXABLE_LIMB,
+				"Units with unfixable limbs:", campaignGUI);
+
+		if (!unitActionsByStatus.isEmpty()) {
+			ArrayList<Person> techs = campaignGUI.getCampaign().getTechs(false);
```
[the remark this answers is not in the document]
kipstafoo: Agreed totally. Fixed several instances of poor interface vs concrete class definition.

MekHQ/src/mekhq/service/MassRepairService.java
kipstafoo: I disagree. For an ArrayList (which this is) arrayCopy is very very fast. It's a very basic level function. If you iterate from the back of the list and remove those that you don't need, it'll be extremely fast and you don't end up with index collisions. Additionally, creating a list of techs to remove is minimally impactful but possibly unnecessary if we don't even need to remove any of the techs. We're creating a list for no reason and doing additionally checks on it for the removeAll().

MekHQ/src/mekhq/service/MassRepairService.java
kipstafoo: Sure. We can turn these into enums. Mostly we only gain type-safety and protection from the random data, but I'm fine to change it.

MekHQ/src/mekhq/service/MassRepairService.java
kipstafoo: Sure. That's fine. As long as someone doesn't turn on DEBUG as the root. The MRMS debug logging is pretty spammy.

MekHQ/src/mekhq/service/MassRepairService.java
kipstafoo: See above comment

Issue #455, opened 2017-09-16
Repair bay uses the wrong ammo bin


Question2:
In this save I have a damaged Vulture in the repair bay whch uses ATMs, and two missing ATM ammo bins, one of which is a normal ATM ammo bin and the other is a ATM HE ammo bin. I only have ATM HE Ammo bins in my warehouse, but the repair bay thinks i have normal ATM ammo bins available as well. If i go ahead with the repair, the repair bay will replace the ATM ammo bin with a ATM HE ammo bin.

Im pretty sure this wasnt the case in a previous version, if you didnt have the exact ammo bin, the repair bay would tell you and you could order the part manually. The way it works now, users will often do mass repairs and not have any idea that the ammo bin replaced is the wrong one unless they manually check every unit after every repair.

[#battletech.zip](https://github.com/MegaMek/mekhq/files/1308732/battletech.zip)

Bonepart:
ATM and ATM HE ammo bins are compatible, they are essentially the same thing, just getting loaded with a different variant of the same missiles. As such I made a fix that checks the munitions type of the replacement part and will change it before replacing it. In this case it will change the ATM HE bin to a standard ATM bin so that after it's replaced you are reloading the ammo you expect.

I should be able to put up the pull request by tomorrow.

PR #460, opened 2017-09-18
Check muntions type before replacing missing ammo bin

Bonepart:
Fix for https://github.com/MegaMek/mekhq/issues/455

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/equipment/MissingAmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/MissingAmmoBin.java
@@ -95,7 +95,16 @@ public class MissingAmmoBin extends MissingEquipmentPart {
 	public void fix() {
 		Part replacement = findReplacement(false);
 		if(null != replacement) {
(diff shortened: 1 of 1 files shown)


neoancient:
I haven't looked at the code for this one yet but I had given it some thought based on what I remember of the class. Would it work to set the ammo type on the cloned replacement? I don't guess it matters much in terms of efficiency but I think the code would be cleaner.

Bonepart:
I tried that first, and it causing it to reload the correct ammo, but in the repair task it still shows as the original ammo type. It basically changed the munition type without changing anything else in the class that usually gets set when it is constructed. So since we are building a new part anyway I just told it to make a new ammo bin instead of cloning the part it found for replacement.

neoancient:
Right. Didn't think of that. This is what comes of figuring out how to do it before you actually look at the code.

Bonepart:
Code wise it doesn't add a whole lot, just an If statement to check type, if they match it clones, if not it generates a new ammo bin

Issue #452, opened 2017-09-13
AtB contract search radius doesn't do much


jslater89:
Looking at the code, it appears that it's only used to limit employers—any faction with a controlled world within contract-search-radius light-years is included in the possible employers list, and any border planet controlled by an employer is a possible contract destination.

It seems reasonable to also limit destination worlds by the search radius.

purpledragon2:
I would say, if possible, have the destination worlds on a separate search radius.  Maybe have the search radius be from the employer's "hiring" planet.

NickAragua:
Fixed with PR #465

PR #465, opened 2017-09-21
Issue #452: limit contract search by AtB search radius

jslater89:
These changes filter destination worlds for contracts based on
the search radius (in addition to the pre-existing filtering based
on employers with worlds in range). Some loosening of constraints
is included, adding 10% of the search radius per retry, in case
of difficult locations.

Tested by generating contracts in the near Periphery, Point Barrow in the middle of the Federated Suns, and near Terra. Doesn't seem to slow down contract generation.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/ContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/ContractMarket.java
@@ -368,14 +368,17 @@ public class ContractMarket implements Serializable {
 			}
 		}
 
+		int searchRadius = campaign.getCampaignOptions().getSearchRadius();
+		searchRadius += (int)(searchRadius * 0.10 * retries);
+
 		boolean isAttacker = (contract.getMissionType() == AtBContract.MT_PLANETARYASSAULT ||
 				contract.getMissionType() >= AtBContract.MT_PLANETARYASSAULT ||
 				(contract.getMissionType() == AtBContract.MT_RELIEFDUTY && Compute.d6() < 4) ||
 				contract.getEnemyCode().equals("REB"));
 		if (isAttacker) {
-			contract.setPlanetName(RandomFactionGenerator.getInstance().getMissionTarget(contract.getEmployerCode(), contract.getEnemyCode(), campaign.getDate()));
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/market/ContractMarket.java
```diff
@@ -368,14 +368,17 @@ else if (contract.getMissionType() == AtBContract.MT_RIOTDUTY)
 			}
 		}
 
+		int searchRadius = campaign.getCampaignOptions().getSearchRadius();
+		searchRadius += (int)(searchRadius * 0.10 * retries);
```
neoancient: Why multiply the fudge factor by the number of remaining retries? This increases the search radius by 30% the first time, then 20% the second retry, then 10%.
Also, if we're to limit the contract location as well as the employer to the selected search radius, why increase it beyond what the user has set?

MekHQ/src/mekhq/campaign/universe/RandomFactionGenerator.java
```diff
@@ -734,7 +735,9 @@ public String getMissionTarget(String attacker, String defender, Date date) {
 						if (f.getShortName().equals(defender) ||
 								defender.equals("PIR") ||
 								(f.getShortName().equals(attacker) && defender.equals("REB"))) {
-							planetList.add(planetKey);
+							if(filterByDistance && currentPlanet.getDistanceTo(planetKey) <= searchRadius) {
```
neoancient: This can be combined with the previous if statement rather than adding another level of nesting.

MekHQ/src/mekhq/campaign/market/ContractMarket.java
jslater89: Doh, that's my misreading of the code.

The intent was to start at the user-specified radius, then cheat upward if no contract locations could be found.

MekHQ/src/mekhq/campaign/universe/RandomFactionGenerator.java
jslater89: I'll go ahead and make that change.

PR #466, opened 2017-09-22
Fixed typo

NickAragua:
It appears that elite mechwarriors were not actually very highly valued. Maybe it's because they're so snobby.


HammerGS:
Seems easy enough.

PR #467, opened 2017-09-23
news and planetevent update

LaserEye32:
Just a few news articles (Placeholders really) 
and a planet event file that will make changes to the planets at set times
and Updated Planet.xml by adding line breaks and also Changed some dates to better fit canon
I also started transferring date sensitive information into 0002_planetevents.xml


HammerGS:
Merged

PR #479, opened 2017-09-30
Data serialization project

NickAragua:
Code to support the ability to import planets from a specifically formatted .tsv file, and to allow the user to export their planet list to xml format.

Contains a lot of hacks due to some ... quirks ... of the SUCS-supplied data.


NickAragua:
The TSV file is the subset of the SUCS-supplied data after I removed a few rows and columns. Not sure where the document process should go - wiki or docs folder or what.

HammerGS:
I'd suggest adding them to the Docs folder.

PR #489, opened 2017-10-06
Fix two tech team-related bugs in Campaign Ops reputation calculations

jslater89:
1. The technicians count was never reset. updatePersonnelCounts()
adds to that variable, and is called when advancing the day and
(possibly) when making other changes, so the number of astechs
required would increase every day.
2. In getSupportDetails(), the code used getAstechPool() (which
only gets the number of temp astechs) instead of getNumberAstechs()
(which takes into account personnel with the role 'astech').
(Only the display was wrong; the calculation used to actually
determine the rating was correct.)

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
+++ b/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
@@ -323,6 +323,8 @@ public class CampaignOpsReputation extends AbstractUnitRating {
 
     private void updatePersonnelCounts() {
         setNonAdminPersonnelCount(0);
+        technicians = 0;
+
         List<Person> personnelList =
                 new ArrayList<>(getCampaign().getPersonnel());
         for (Person p : personnelList) {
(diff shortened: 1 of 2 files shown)


pheonixstorm:
The calculations for the astechs are correct as far as the actual score is concerned. The other one... never noticed as the visual in the rating report (dialog and panel in overview) both show the correct number, though the panel should refresh more often.

jslater89:
In one of my AtB campaigns, I had 73 temp astechs and 17 personnel with the role astech, thanks to the 'dependents' AtB special event. Prior to this change request, the report showed 73/90, but no penalty for techs.

As far as the astech incrementing, I wonder if that was introduced along with the unit reputation performance fixes pull request? The always-incrementing astechs happens when the unit reputation object is reused without being re-instantiated.

pheonixstorm:
Haven't seen it, but there is a bug where hiring temp astechs will also add additional tech astechs to cover a ships engineer. So if you have six techs you should only need 36 astechs. If you have a large craft on the other hand, hiring new temps will count each large craft so that 36 might become 42 or more depending on the number of craft.

jslater89:
So it does. The comment for Campaign.getTechs() says it gets all people who qualify as techs per Person.isTech(), but the code then special-cases engineers in (who are not techs, according to Person.isTech()).

MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
```diff
@@ -323,6 +323,8 @@ private void calcNeededTechs() {
 
     private void updatePersonnelCounts() {
         setNonAdminPersonnelCount(0);
+        technicians = 0;
```
dericpage: This should probably be added to the initValues() logic as well.

jslater89: initValues() calls updatePersonnelCounts().

dericpage: You're right.  Forgot about that.

Issue #488, opened 2017-10-05
[43.4] Select bombs throws "index out of bounds" errors in particular campaign file


NickAragua:
[Goldenrod Menace.zip](https://github.com/MegaMek/mekhq/files/1360869/Goldenrod.Menace.zip)

.cpnx and log file attached.

It's in this state after I resolved a battle where I used up a bunch of bombs. Possibly related to the other 'select bombs' issue.

To reproduce:
Load up the attached save file. Go to Hangar. Try to "Select Bombs" on the Lucifer not being mothballed. The expected pop-up doesn't appear, and the log indicates an error.

NickAragua:
As an update, simply loading the save file doesn't do the trick, as the error appears to self-correct at that point.  So you'd need to start up a battle with a bomb-laden fighter, use up all the bombs, then come back to MekHQ.

Bonepart:
There is definitely something funny going on. I deployed two ASF. One of them dropped all their bombs, the other only half. After resolving neither showed bombs in their inventory in the hanger and I got the index out of bounds error in the log.

Bonepart:
Digging into this the problem is because when HQ is resolving the scenario the ASF's bombChoices is returning all 0's. bombChoices is supposed to hold the number of bombs that were loaded, per type. Because it is 0 when it unloads the bombs that were mounted instead of, for example, subtracting 4 bombs used from the 5 bombs loaded and leaving an inventory of 1 bomb, it's subtracting 4 from 0, leaving a negative number. And that is really bad.

We will have to figure out where and why bombChoices is being set to 0.

neoancient:
It gets set to 0 in the entity exchange phase in MM when the bomb choices are converted into bomb equipment.

Bonepart:
Was that a change? It didn't work that way before. 

Maybe that was how it was meant to work and I 'fixed it' in the wrong direction. When I did my bomb work I saw it wasn't correctly modifying bomb inventory when coming back from a scenario. Fixing it was a simple sign change, changing += to -=. If bombChoices is meant to be 0 then you would indeed want to += for each non-expended bomb.

I had assumed bombChoices was meant to hold the total loaded bombs, because that is how it was working at the time.

neoancient:
That's how it was supposed to work before, and it was working by accident on individual fighters but it broke squadrons.

Bonepart:
That fixes it. I tested it and bombs unmount correctly. That prevents this error.

arlith:
[...]

PR #493, opened 2017-10-07
Change Campaign.clearGameData(Entity) to fix Bomb unmounting

Bonepart:
Was previously expecting Aero.bombChoices to keep total bombs loaded on
ASF, bombChoices is now set to 0, corrected bomb unmounting code for
that.

This is the fix for https://github.com/MegaMek/mekhq/issues/488


Bonepart:
2nd commit for cycle efficiency. There is no need to get and set bombChoices for an Aero that does not have bombs mounted.

Bonepart:
Fixed a small error I discovered while fixing previous issue

Issue #495, opened 2017-10-08
CamOPs Not identifying Techs properly


LaserEye32:
I have set up 4 techs, Aero&Mech&Mechanic&BA.
I have also Mass Trained all four techs to be masters of each of the Tech Fields.
CamOps now says that all 4 are MechTechs instead of identifying the job they are assigned and/or identifying the vehicle they are working on. Thus giving me a -5 Support.

This was set up on mekhq 0.43.4

[Bug Test- techs error30000101.zip](https://github.com/MegaMek/mekhq/files/1366930/Bug.Test-.techs.error30000101.zip)

jslater89:
Thanks for the test case. PR497 fixes it.

NickAragua:
Should be fixed with #497

PR #497, opened 2017-10-09
Issue #495: tech roles determine multi-skilled tech team type

jslater89:
i.e., if you have someone with Tech/Mech and Tech/Aero skills, the CamOps reputation method will use their selected role to decide what kind of tech team they have. They still have to have some ranks in the skill to count as a tech of a given type.


jslater89:
D'oh, good point. Will fix on my lunch break.

jslater89:
I went ahead and made the change to allow secondary techs to lead tech teams.

Issue #321, opened 2017-02-13
[feature request] ability to randomly assign portraits to personnel with out portraits


plutonick:
instead of having to assign portraits manually to all personnel that have no portraits, i would like to request an option to randomly "mass assign portraits" (from the portrait folder)

Faolind:
Would an option for doing the same to OpFor/Ally personnel be too much to ask? (not as a right click, but as an automatic in the campaign options menu)

Qwertronix:
Currently taking a shot at this, just adding a right-click context menu option to randomize portraits of all selected personnel in the Personnel tab. 

@Faolind That's doable too but should probably be its own separate request.

NickAragua:
Implemented in PR #527

PR #527, opened 2017-11-16
New context menu option to randomize portraits for selected personnel.

Qwertronix:
^title, in Personnel tab. 

Does not check if the person already had a portrait; if it did, it simply runs the generator anyway which will override the previous portrait. 

As-is this works for me, but we could also add the option of having it only generate portraits for selected personnel who didn't have one.


Qwertronix:
Per feedback on slack, making it ignore personnel who already have a portrait by default.

Qwertronix:
Oh, if accepted this should also resolve #321.

MekHQ/src/mekhq/resources/PersonnelTableMouseAdapter.properties
```diff
@@ -21,6 +21,7 @@ changeCallsign.text=Change Callsign...
 changeMDClass.text=Change Manei Domini Class
 changeMDRank.text=Change Manei Domini Rank
 changePortrait.text=Change Portrait...
+randomizePortrait.text=Randomize Portrait...
```
NickAragua: Usually, we use "..." to indicate that there's going to be some kind of popup. So, I think we should get rid of the ...

Qwertronix:
Doh! Will do.

Qwertronix:
@NickAragua Done.

PR #530, opened 2017-11-18
Greater opfor variety

NickAragua:
Adds the capability that an opfor will also have some extra aircraft and/or infantry/turrets/battle armor. 

If the opfor is one of the planet owners, then we'll be seeing conventional aircraft at the user specified odds.

If the opfor is not one of the planet owners, then we'll be seeing aerotech units at half the user specified odds.

If the opfor is one of the planet owners, and the terrain is some kind of urban, we'll be seeing turrets at the user specified odds.

If the opfor is one of the planet owners, and the terrain is some kind of urban or woods, we'll be seeing conventional infantry at the user specified odds.

If the opfor is not one of the planet owners and the terrain is some kind of urban or woods, we'll be seeing battle armor at half the user specified odds, except that if the faction doesn't have battle armor then we substitute conventional infantry.

number of extra units is always 1 per "pip" of difficulty (1 at ultra-green, 5 at elite)
Infantry/turrets only show up if the player is the "attacker".
Extra units shouldn't show up in "special missions" (star league caches and the like), but may show up in big battles.

Can always be turned off by users in the campaign options -> atb tab.


NickAragua:
campaignoptions
scenariowithextradudes

PR #533, opened 2017-11-19
Founders never retire

Qwertronix:
Another small one from the "made it for me, sharing if anyone else wants" shelf. Pretty self-evident: If the option is checked, personnel marked as founding members will never get a retirement roll, just like Prisoners, Bondsmen, etc.


MekHQ/src/mekhq/campaign/personnel/RetirementDefectionTracker.java
```diff
@@ -188,7 +188,8 @@ public static long getShareValue(Campaign campaign) {
     	for (Person p : campaign.getPersonnel()) {
     		if (!p.isActive() || p.isDependent()
     				|| p.isPrisoner() || p.isBondsman()
-    				|| p.isDeployed()) {
+    				|| p.isDeployed() ||
+					p.isFounder() && campaign.getCampaignOptions().getFoundersNeverRetire()) {
```
NickAragua: Hate to nitpick, but let's make the || formatting consistent here.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -3467,65 +3469,72 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         gridBagConstraints.gridy = 5;
         panSubAtBAdmin.add(chkCustomRetirementMods, gridBagConstraints);
 
+        chkFoundersNeverRetire.setText(resourceMap.getString("chkFoundersNeverRetire.text"));
+        chkFoundersNeverRetire.setToolTipText(resourceMap.getString("chkFoundersNeverRetire.toolTipText"));
+        chkFoundersNeverRetire.setSelected(options.getFoundersNeverRetire());
+        gridBagConstraints.gridx = 0;
+        gridBagConstraints.gridy = 6;
```
NickAragua: If you go back to where the gridbagConstraints.gridy is first used for this column and declare an int that starts at 0, then, instead of having to update every gridy after any item you added, you can just use "gridBagConstraints.gridy = intvalue++".

MekHQ/src/mekhq/resources/PersonnelTableMouseAdapter.properties
```diff
@@ -2,7 +2,8 @@ kiaQ.text=KIA?
 abilityDesc.format=%s %s
 addPregnancy.text=Add Pregnancy
 addSingleLogEntry.text=Add Single Log Entry...
-addXP.text=Add XP
+addXP.text=Add XP...
+add1XP.text=Add 1 XP
```
NickAragua: This keeps popping up in all your reviews. It doesn't seem like a problem, but what I'd suggest is to use a separate branch for each pull request so that your code doesn't mix between PRs.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
NickAragua: Now that I think about it, you could just set gridBagConstraints.gridy to 0 at the beginning of the column, then call ++ on it for each subsequent row.

Qwertronix: Yeah I like this, would save us the trouble each time this gets changed.

MekHQ/src/mekhq/resources/PersonnelTableMouseAdapter.properties
Qwertronix: Oops. I based the branches off of the fix then forgot about it.

MekHQ/src/mekhq/campaign/personnel/RetirementDefectionTracker.java
Qwertronix: No worries, I would have nitpicked on it too if I saw it 👍

Qwertronix:
Updated with the changes 😄

Issue #280, opened 2016-12-26
Feature Request: Discrete Portait Folders


mnmib1:
Discreet portrait folders.  Some of my pictures (a lot of the ink effect) are perfect for soldiers.  I have a lot of people in suits (doctors/admins).  Others are good for MechWarriors.  And my battle armor pictures are only good for battle armor.  Ralgith should know what I'm talking about.  He had some ideas about how to implement this years ago.

Kendricke:
I would like to +1 this request.  I have literally hundreds of custom portraits I've created for use in specific roles which I'm currently manually assigning each time I bring on a new recruit.

Kendricke:
Related:  https://github.com/MegaMek/mekhq/issues/471

Bonepart:
Seems like a reasonable request and it's been around for awhile. I don't think it should be TOO hard to implement. I'll try to look into it in the near future.

Qwertronix:
Nice to see this is in the 0.43.5 code! If I could make just one suggestion though, any chance we can get a similar single folder for all "Tech" personnel like was done for "Admin"? Not a big deal, but I don't think most people will want separate sets of portraits for each different tech role.

Kendricke:
I actually do have separate portraits I use for different types of techs and admin.  Is it possible to just run a check to see if different subfolders exist?  If not, use the base tech or admin folder?

Qwertronix:
Sure, I think a layered fallback would work. Search for folder with the exact name first, then group name, then if still not found then the gender one.

Kendricke:
For purposes of troubleshooting (and actually better using the feature), can we get a list of the folder structure we should be using moving forward?  I can't find this in documentation or notes and some of the roles dont' seem to be working and the layered fallback mentioned in #534 doesn't seem to be working, either.

Qwertronix:
I'll update my files to the dev snapshot and take a look at what's not working. 

The basic structure is that first it looks for a folder with the exact name of the primary role. If that folder doesn't exist, is empty, or contains only pictures which are already assigned, it is skipped. Next it checks for "group" folders, if it applies to the role. Those are `Admin/`, `Tech/`, `Medical/` and `Vessel Crew/`. Same here, if empty, etc., skip. Next it looks for the `Combat/` or `Support/` folders. If it still hasn't found any possible portraits, it finally goes back to the regular gendered `Male/` and `Female/` folders. 

[...]

PR #534, opened 2017-11-19
Expand discrete portraits feature

Qwertronix:
As per #280, expanding on PR#481 this adds a layered fallback approach to random portraits: first look into a folder matching the person's primary role, then if no valid unused pictures exist check the "group" folder (Admin/, Tech/, etc.), then Combat/ or Support/, and then just the gendered folders.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -7157,6 +7157,9 @@ public class Campaign implements Serializable, ITechManager {
         if (searchCat_Role.startsWith("Admin/")) {
             searchCat_Role = "Admin/";
         }
+        if (searchCat_Role.endsWith("Tech")) {
+        	searchCat_Role = "Mechanic/";
+        }
         
         while (categories.hasNext()) {
             String category = categories.next();
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -7211,7 +7211,7 @@ public class Campaign implements Serializable, ITechManager {
 
         possiblePortraits = getPossibleRandomPortraits(portraits, existingPortraits, searchCat_Gender + searchCat_Role);
 
-        if (possiblePortraits.isEmpty()) {
+        if (possiblePortraits.isEmpty() && !searchCat_RoleGroup.isEmpty()) {
             possiblePortraits = getPossibleRandomPortraits(portraits, existingPortraits, searchCat_Gender + searchCat_RoleGroup);
         }
         if (possiblePortraits.isEmpty()) {
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -7146,7 +7146,10 @@ public class Campaign implements Serializable, ITechManager {
         }
         ArrayList<String> possiblePortraits = new ArrayList<String>();
         Iterator<String> categories = portraits.getCategoryNames();
-        
+
+        // Will search for portraits in the /gender/primaryrole folder first,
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7172,6 +7192,41 @@ public void assignRandomPortraitFor(Person p) {
                 }
             }
         }
+
+        if (possiblePortraits.isEmpty()) {
```
NickAragua: This if statement block seems like a prime candidate to split out into a separate function, as it recurs three times and is mostly identical.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7146,18 +7146,38 @@ public void assignRandomPortraitFor(Person p) {
...
         }
         String searchCat_Role = Person.getRoleDesc(p.getPrimaryRole(), false) + "/";
+        String searchCat_RoleGroup = "";
+        String searchCat_CombatSupport = "";
         if (searchCat_Role.startsWith("Admin/")) {
-            searchCat_Role = "Admin/";
+            searchCat_RoleGroup = "Admin/";
         }
-        
+        if (searchCat_Role.endsWith("Tech/") || searchCat_Role.equals("Mechanic/")) {
```
NickAragua: Are these role names defined as constants anywhere (like in the Person class)? If so, you should make use of those constants, rather than having magic strings.

Qwertronix: I believe so. I just went with the same type of hack that was used for `Admin/`, but we could easily replace this with a proper complete enumeration of the matching roles using constants.

Qwertronix: Ugh, the relevant place to look at is `Person.getRoleDesc()` which takes an int enum and returns a hard-coded string via select case, so each comparison would have to look like `if (searchCat_Role.equals(Person.getRoleDesc(Person.T_MECH_TECH))) { // do; }`. 

Considering this, I think it might be more robust to grab the current person's role int to compare to the int enum, e.g. `if (person.getPrimaryRole() == Person.T_MECH_TECH) { // do; }`.

NickAragua: Make it so, #1

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7126,6 +7126,25 @@ public void addLogEntry(Person p, LogEntry entry) {
         p.addLogEntry(entry);
     }
 
+    public ArrayList<String> getPossibleRandomPortraits (DirectoryItems portraits, ArrayList<String> existingPortraits, String subDir ) {
```
NickAragua: Public method should have javadoc comment

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7146,48 +7165,60 @@ public void assignRandomPortraitFor(Person p) {
...
         } else {
             searchCat_Gender += "Male/";
         }
         String searchCat_Role = Person.getRoleDesc(p.getPrimaryRole(), false) + "/";
-        if (searchCat_Role.startsWith("Admin/")) {
-            searchCat_Role = "Admin/";
+        String searchCat_RoleGroup = "";
+        String searchCat_CombatSupport = "";
+//        if (searchCat_Role.startsWith("Admin/")) {
+        if (p.getPrimaryRole() == Person.T_ADMIN_COM
```
NickAragua: Let's get rid of the commented-out code here and for the other checks.

NickAragua:
Once the commented out code is cleaned up, this is ready to be merged in.

Qwertronix:
Oh! I can probably make time to at least get that done tonight. I didn't realize that's all I had left to do on this one.

Issue #526, opened 2017-11-16
RFE: Select Portrait when promoting personnel


scJazz2:
If, because of options, a portrait is not selected and then you purchase skills and change Primary Role that promote a Person into a group that does use portraits... please select a portrait.

Qwertronix:
By "Select" here you mean "have MHQ randomly select one for me" right? That's what I understood based on our discussion in Slack at least.

scJazz2:
yes

Qwertronix:
If I can figure out how the Campaign Options UI works and how to add the option in there, I can probably do this sometime this weekend. The actual portrait assignation process itself should be just an extra 3-4 lines since all the code to check for a portrait and randomly assign one is already available. Making it a selectable option in the menus is actually the more difficult part for me :rofl:

Windchild292:
This has been implemented and can be closed

Windchild292:
Reopening and reclosing to see if that fixes a github state issue.

PR #535, opened 2017-11-19
Implements #526: option to auto-randomize portrait on job change

Qwertronix:
When the new option is enabled in the Name and Portraits tab, any personnel with no current portrait whose primary role is changed to one that is selected for random portraits in the same options will automatically gain a random portrait. Discussed with scJazz and this solve #526.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
@@ -201,6 +201,7 @@ public class CampaignOptionsDialog extends javax.swing.JDialog {
     private JCheckBox usePeacetimeCostBox;
     private JCheckBox useExtendedPartsModifierBox;
     private JCheckBox showPeacetimeCostBox;
+    private JCheckBox chkAssignPortraitOnRoleChange;
 
     private JTextField[] txtSalaryBase;
     private JSpinner[] spnSalaryXp;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -3049,6 +3052,16 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         gridBagConstraints.anchor = java.awt.GridBagConstraints.NORTHWEST;
         panNameGen.add(panRandomPortrait, gridBagConstraints);
 
+        chkAssignPortraitOnRoleChange.setText(resourceMap.getString("chkAssignPortraitOnRoleChange.text"));
+        chkAssignPortraitOnRoleChange.setToolTipText(resourceMap.getString("chkAssignPortraitOnRoleChange.toolTipText"));
+        chkAssignPortraitOnRoleChange.setName("chkAssignPortraitOnRoleChange");
+        gridBagConstraints = new java.awt.GridBagConstraints();
+        gridBagConstraints.gridx = 0;
+        gridBagConstraints.gridy = 4;
```
NickAragua: Let's implement a variable to keep track of current gridy (or you could even just do a ++ on the gridy, now that I think about it).

Qwertronix: Since this section frequently reinitializes `gridBagConstraints` to reset the parameters as each control is different, I went with a variable to increment the `gridy` here.

Qwertronix:
Updated & tested

Issue #519, opened 2017-11-06
Rename Overview Tab: Dragoons Rating


scJazz2:
The Overview Tab > Dragoons Rating text should be changed to "Unit Rating" which syncs it to the Reports Menu text which is "Unit Rating"

NickAragua:
Fixed with PR #538

Issue #510, opened 2017-10-24
Dragoons rating: Medical support includes KIA/MIA/Retired Personnel


MattDroz:
I'm working on recreating Team Banzai to run through MekHQ through the Clan Invasion and Civil War. As I'm building the "history" of the Merc Unit before I begin, I've got a lot of data to get the Dragoon's Rating up. I've built it so they have 576 Active Combat Personnel, 193 Support, 73 Dependents, 1 prisoner, 113 MIA, 170 KIA, and 11 Retired. Going by the FM:Mercs formula, that means the Medical Support hours should be:

840/7 = 120 * 4 = 480 
3 + 3/5 = 3.6
484 Man-Hours.

But the Dragoons Rating is showing I need 652 man-hours, which is what it would be if the doctors are maintaining non-active personnel:

1134/7 = 162 * 4 = 648
3 + 3/5 = 3.6
652  Man-hours

[Team Banzai30250101.zip](https://github.com/MegaMek/mekhq/files/1412064/Team.Banzai30250101.zip)

NickAragua:
Fixed with PR #538

PR #538, opened 2017-11-20
Unit rating fixup

Qwertronix:
Fix for #510, making Dragoons Rating medical support requirement count only active personnel. 

Also makes CamOps reputation admin support requirement count only active personnel.

Also renames the "Dragoons Rating" panel in the Overview tab to "Unit Rating" for consistency, as requested in #519.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -1176,6 +1176,10 @@ public class Campaign implements Serializable, ITechManager {
         return personnel;
     }
 
+    /**
+     * Provides a filtered list of personnel including only active Persons.
+     * @return ArrayList<Person>
+     */
     public ArrayList<Person> getActivePersonnel() {
         ArrayList<Person> activePersonnel = new ArrayList<Person>();
         for (Person p : getPersonnel()) {
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -1176,6 +1176,14 @@ public class Campaign implements Serializable, ITechManager {
         return personnel;
     }
 
+    public ArrayList<Person> getActivePersonnel() {
+        ArrayList<Person> activePersonnel = new ArrayList<Person>();
+        for (Person p : getPersonnel()) {
+            if (p.isActive()) {activePersonnel.add(p);}
+        }
+        return activePersonnel;
+    }
+
     public ArrayList<Ancestors> getAncestors() {
         return ancestors;
     }
--- a/MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
+++ b/MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
@@ -321,7 +321,7 @@ public class FieldManualMercRevDragoonsRating extends AbstractUnitRating {
         int numSquads = new BigDecimal(activePersonnel.size())
                 .divide(new BigDecimal(7), 0,
                         RoundingMode.DOWN).intValue();
-        int leftOver = getCampaign().getPersonnel().size() - (numSquads * 7);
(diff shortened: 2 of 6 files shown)


pheonixstorm:
Is there an open bug report that needs the change made to the camops rating or did you test if it needed the change? I haven't noticed an issue (with prisoners at least) that would cause the count to be off.

Qwertronix:
It counted inactive personnel. Even in the unit test, the mock crew had 200 personnel on record, only 98 of which were set to be "active" in the tests, which should result in a required admin support of 10, right? But it expected 20. 

Now that I realized this though, I'm going through everything again in both as they report totals including KIA, MIA and retired personnel in every category, not just CamOps admin and FM Medical.

Qwertronix:
I'm still looking around and it seems there are a couple more places that use getPersonnel() and implicitly expected to get only active personnel. I'm wondering if it might be better to just add getActivePersonnel() and replace all those cases with the new method that returns only active staff.

Qwertronix:
Just to clarify, as of 73076f7 this seems to work for me and I split the getPersonnel() issue I mentioned into its own PR.

MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
```diff
@@ -314,10 +315,13 @@ private void updateDropJumpShipSupportNeeds(Entity en) {
     //   3 + (4/5) = 3 + 0.8 = 3.8 = 4 hours.
     //   total = 16 hours.
     private void calcMedicalSupportHoursNeeded() {
-        int numSquads = new BigDecimal(getCampaign().getPersonnel().size())
+        ArrayList<Person> activePersonnel = new ArrayList<>();
+        // Only count active personnel. Dead or absent people don't need medical support. >.>
+        for (Person p : getCampaign().getPersonnel()) { if (p.isActive()) { activePersonnel.add(p); } }
```
NickAragua: I think we're doing this "isActive" check often enough that Campaign probably warrants a "getActivePersonnel()" method that carries the work out internally.

Qwertronix:
Updated and mostly-tested. 8b23464 shows up both here and in #539 as they both depend on it now.

MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
```diff
@@ -314,10 +307,10 @@ private void updateDropJumpShipSupportNeeds(Entity en) {
     //   3 + (4/5) = 3 + 0.8 = 3.8 = 4 hours.
     //   total = 16 hours.
     private void calcMedicalSupportHoursNeeded() {
-        int numSquads = new BigDecimal(getCampaign().getPersonnel().size())
+        int numSquads = new BigDecimal(getCampaign().getActivePersonnel().size())
```
NickAragua: So, one more comment: since "getActivePersonnel()" can get expensive for a lot of personnel, you should consider storing that count locally rather than calling it multiple times.

The alternatives solution is to store the number of active personnel locally within Campaign, but then you'd have to update it when the personnel count changes.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1176,6 +1176,14 @@ public Date getDate() {
         return personnel;
     }
 
+    public ArrayList<Person> getActivePersonnel() {
```
NickAragua: Public method needs javadoc comment.

Qwertronix: Whelp. Googling "how to javadoc", brb. 😛

MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
Qwertronix: Hmm I'd prefer option 2 really, but need to make sure Person is firing events for any method that changes their status first.

PR #539, opened 2017-11-20
getPersonnel() usage cleanup pass

Qwertronix:
This goes over a few places where Campaign.getPersonnel() was used and the calling method implicitly assumed it was querying active personnel. I've added checks to make sure those places are only counting active persons.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -1176,6 +1176,10 @@ public class Campaign implements Serializable, ITechManager {
         return personnel;
     }
 
+    /**
+     * Provides a filtered list of personnel including only active Persons.
+     * @return ArrayList<Person>
+     */
     public ArrayList<Person> getActivePersonnel() {
         ArrayList<Person> activePersonnel = new ArrayList<Person>();
         for (Person p : getPersonnel()) {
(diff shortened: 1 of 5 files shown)


Qwertronix:
Going to go over this later to do another pass and put `getActivePersonnel()` calls instead, as discussed in PR #538.

Qwertronix:
Updated and mostly-tested. 8b23464 shows up here and in #538 as they both depend on it.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1176,6 +1176,18 @@ public Date getDate() {
         return personnel;
     }
 
+    /**
+     * Provides a filtered list of personnel including only active Persons.
+     * @return ArrayList<Person>
+     */
+    public ArrayList<Person> getActivePersonnel() {
+        ArrayList<Person> activePersonnel = new ArrayList<Person>();
```
NickAragua: Well, at the very least, this "getActivePersonnel()" function is now in a single location, so we can optimize it later on.

Qwertronix: I imagine the most straightforward way to do that would be to ensure `PersonAddedEvent`, `PersonRemovedEvent` and `PersonChangedEvent` (or perhaps a subclass thereof only for status changes) get fired diligently, and subscribe an updater method to maintain a private activePersonnel list?

NickAragua: Yeah, I think that's probably the best approach.

Qwertronix: I'll add that on my todo list for its own later PR, if no one else gets to it before then.

PR #543, opened 2017-11-22
New startup/load screen art. Auto-resize startup/load screen.

NickAragua:
Auto-resizes startup and load screens based on user's resolution. Introduces Spooky's new art for MekHQ startup and load screens.


Qwertronix:
Yay! New eye candy!

Issue #464, opened 2017-09-20
KIA personnel that have been set to 0 hits in the battle resolution screen are still moved to the KIA roster


Question2:
Personnel killed in a scenario and then set to 0 hits (or anything other than "dead") are still incorrectly moved to the KIA roster after the post battle resolution screen. They should only be moved to KIA if they are at 6 hits/dead.

[#battletech.zip](https://github.com/MegaMek/mekhq/files/1318905/battletech.zip)

If you load the autosave file for the first base attack scenario and let the BA get killed, you will end up with some dead BA pilots in the battle resolution screen afterwards. Setting them to 0 hits will still result in them being moved to the KIA roster at 0 hits.

pheonixstorm:
I noticed this earlier in my own game. Will take a look at it later and see if I can find the cause unless a dev beats me to it.

Qwertronix:
Noticed this as well, and it might be related to another report (which I can't seem to find at the moment) of units "disappearing" on their deployment phase in MM if they had previously come up as KIA and been put back to active status in this way. 

Also probably the same issue as #324.

Dylan-M:
Closing as duplicate.

Duplicates #324

Issue #324, opened 2017-02-18
BUG: Personnel keep trying to fake their death


ThomasMarik:
when resolving a battle in mekHQ after doing the megamek match you have the opportunity to review and alter damage taken by your personnel. If someone has taken 6 points of damage ad you reduce it to 5 the game still considers him dead and sets him KIA. This is incorrect as a person with 5 hits is not dead unless a Lost head result is rolled.

Dylan-M:
I cannot replicate this on the current code. I'm pretty sure this has been resolved.

PR #552, opened 2017-12-03
Display KIAs on scenario resolution UI

Qwertronix:
Throwing this up here for discussion and suggestions. 

There are a number of cases where a pilot or crewmember or other person might die on the battlefield or during scenario resolution that bypass the usual "6+ hits = dead" rule. Some examples are random rolls when multi-crew mech cockpits are destroyed or when calculating infantry casualties. Generally these would set the hits count of the person to 6 anyway, but also set a `forceDead` flag. 

This `forceDead` thing was not shown to the player and was never modified again later after the ResolveScenarioWizardDialog is launched. This meant that a player modifying the hits slider from 6 to 3 would still find that person sent to KIA rosters, unassigned, logged as dead, and processed through all other things that happen when McCoy shakes his head. 

I've considered the option of simply setting this to false whenever the user modifies the hits slider to something below 6, but what if the slider is changed to another value different than 6 by some other competing calculation before the wizard dialog is displayed? We then end up with a situation where a person has 2 hits shown (which seems fine to the user), remains unchanged, but ends up KIA still after the report. 

The two alternatives seem to be to either drop this whole `forceDead` thing altogether, and replace it with `setHits(HITS_DEAD)`, or show it to the user and give them control. The changes in this PR are for the latter just in case this should be kept around for something else I'm unaware of. Deleting `forceDead` instead would be fairly trivial. 

If we keep the display/UI option, a sub-question would be: What's a good general term to use there? I've put "Forced Death" but I don't like it.


Qwertronix:
As an aside, this is something I found while trying to fix #464 and #324, but nothing proposes here would actually fix the root cause of those issues.

NickAragua:
Can you post a screenshot of what this looks like?

Qwertronix:
Sure, looks like this: 

image

Qwertronix:
Following some discussion, changed it to be a general KIA checkbox that simply controls whether the person is dead or not, making it no longer controlled by the hits slider. 

Looks like this now: 

image

So far this is only the UI changes with no cleanup of the data handling in the background yet.

Issue #212, opened 2016-10-26
Campaign Option To Use Lowest Skilled Admin When Ordering Parts


BLOODWOLF333:
Currently HQ always uses your highest skilled admin to order parts first before moving on to the next highest skilled.  Asking for an option to set an order of precedence for what personnel to use first when ordering parts.

Use lowest skilled first <what I need
Use highest skilled first
[any other combinations we can come up with?]

I treat my unit's motor/mech-pool like a platoon with a platoon commander (2Lt with scrounge as the skill to acquire parts) and then all the techs as enlisted.  So when that 2LT gets to CPT and moves to the battalion motor-pool I dont need them ordering parts for the lance anymore since I have a new butterbar come in but, then he never gets any xp.  Would also be nice to assign an admin to each force and that admin would order parts for that force first before going to the Captain or the next admin.  Kind like how techs are done with the units having an assigned tech to each unit.

Dylan-M:
Better would be to do what we do for refits: dialog showing available personnel.

PR #599, opened 2018-01-08
Acquisition skill priority ordering option

Qwertronix:
Naive implementation of #212.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
@@ -298,6 +298,7 @@ public class CampaignOptionsDialog extends javax.swing.JDialog {
     private JCheckBox chkSupportStaffOnly;
     private JSpinner spnAcquireWaitingPeriod;
     private JComboBox<String> choiceAcquireSkill;
+    private JComboBox<String> choiceAcquireSortPriority;
     private JSpinner spnAcquireClanPenalty;
     private JSpinner spnAcquireIsPenalty;
     private JCheckBox chkCapturePrisoners;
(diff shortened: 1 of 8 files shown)


MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -1072,11 +1073,31 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
...
 
+        DefaultComboBoxModel<String> acquisitionSortPriorityModel = new DefaultComboBoxModel<String>();
+        acquisitionSortPriorityModel.addElement(CampaignOptions.S_BESTFIRST);
+        acquisitionSortPriorityModel.addElement(CampaignOptions.S_WORSTFIRST);
+        acquisitionSortPriorityModel.setSelectedItem(options.getAcquisitionSortPriority());
+        choiceAcquireSortPriority = new JComboBox<String>(acquisitionSortPriorityModel);
+
+        gridBagConstraints = new java.awt.GridBagConstraints();
+        gridBagConstraints.gridx = 0;
+        gridBagConstraints.gridy = 3;
```
NickAragua: Let's make this gridy self-incrementing (except for those situations where two items are on the same row of course), rather than with these static numbers.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -1072,11 +1073,31 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
...
+        gridBagConstraints.fill = java.awt.GridBagConstraints.NONE;
+        gridBagConstraints.anchor = java.awt.GridBagConstraints.WEST;
+        panSubAcquire.add(new JLabel("Buyer Skill Priority:"), gridBagConstraints);
```
NickAragua: I'd like to see new text turned into text resources instead of being hard-coded. 

Ideally, the other ones too, but let's not get too carried away.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
NickAragua: Also, can we space the Acquisition Skill and Buyer Skill Priority dropdowns out a little better? Right now, at least with my "standard windows theme", they're right on top of each other.

giorgiga:
@NickAragua I went dow a similar path as this PR while investigating Bloodwolf issue with junior procurers never seeing action and thus never getting XPs (I didn't know of this PR at the time – I just noticed it now) and IMHO there may more bad than good down that path...

The trouble is, `getLogisticsPerson()` (which currently returns the best free procurer) is not only used when performing the actual acquisition checks, but also to display the TN for acquisition in the UI: if the policy of `getLogisticsPerson()` is inverted all displayed acquisition TNs will be based on the worst procurer rather than the best one, which I think might greatly confuse the players (personally, the fact that TNs change when I buy something -because my best procurer gets exhausted- already confuses me).

sixlettervariables:
Closed as stale.

PR #603, opened 2018-01-13
Edge Trigger toggle tweaks

### LINKED ISSUE #590: [43.6] Cannot set mass edge triggers
I'm not able to set edge triggers for multiple personnel.  I can ONLY set the "off" status, but not the "on" status.  I can't recall if this was ever actually implemented, but based on https://github.com/MegaMek/mekhq/commit/04ba4ee7141af79833850d87c80b5b6c55766fd3, this seemed to be the case.

Qwertronix:
Fixes #590 and an unreported bug where support edge triggers would show in the multi-select options even if that unofficial setting was disabled. 

Also always display all enabled edge triggers for single-select context menus, with triggers not "valid" for the person's current role grayed out. This part isn't a "fix" per-se but I found the behavior unintuitive and it _felt_ bugged when I'd go mouse over the "Set Edge Triggers" option and just not get any submenu pop up. 

This also allows me to set edge triggers even when a person isn't _currently_ on a role that can use it, which I find desirable too.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
@@ -1,5 +1,6 @@
 package mekhq.gui.adapter;
 
+import java.awt.Color;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.MouseEvent;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -1904,81 +1905,95 @@ private void maybeShowPopup(MouseEvent e) {
...
-                        cbMenuItem = new JCheckBoxMenuItem(resourceMap.getString("edgeTriggerMASCFailure.text")); //$NON-NLS-1$
-                        cbMenuItem.setSelected(person.getOptions()
-                                .booleanOption(OPT_EDGE_MASC_FAILURE));
-                        cbMenuItem.setActionCommand(makeCommand(CMD_EDGE_TRIGGER, OPT_EDGE_MASC_FAILURE));
-                        cbMenuItem.addActionListener(this);
-                        menu.add(cbMenuItem);
-                    //Doctors    
-                    } else if (person.getPrimaryRole() == Person.T_DOCTOR && gui.getCampaign().getCampaignOptions().useSupportEdge()) {
+                    //Start of Edge reroll options
+                    cbMenuItem = new JCheckBoxMenuItem(resourceMap.getString("edgeTriggerHeadHits.text")); //$NON-NLS-1$
```
NickAragua: This five-liner block really wants to be turned into a helper function that takes some parameters (the text, the edge trigger, list of roles) and returns a JCheckBoxMenuItem ready for addition to the menu.

Qwertronix: It's practically begging, but I had my headset on with some loud music so I didn't hear it at first.

Qwertronix:
Currently, this thing only looks at the primary role to determine whether someone is considered to have a particular edge trigger "usable", and you need to enter the GM edit dialog to set them otherwise or change the person's role, set them, change back. 

I like to set up my edge triggers in advance whenever someone first gets an edge point, and I tend to forget about them afterwards, even through role changes. 

It's also relevant for the rare case where someone has a combat role as their secondary role, as they can go into battle and use those edge triggers but you're currently unable to set those edge triggers without the extra steps. 

All of this could also be superseded and better handled by a set of configurable "default edge triggers", but implementing that in a forward-compatible way with any future edge triggers is slightly trickier.

PR #604, opened 2018-01-13
Fix GM "New Contract" payment modifier spinner

### LINKED ISSUE #308: Contract Payment Multiplier cannot be changed
Reported on Slack MegaMek-public by scjazz and verified by me, both on 0.42.0:

In GM mode, going to the briefing tab and adding a new mission/contract, you have a bunch of options including a payment multiplier. This multiplier is fixed to 1.2 (for me, for scjazz 1.125), using any method of editing there does not change it (neither arrows nor textual input, after the attempted change it changes back to 1.2). scjazz also reported that changing PaymentMultiplier in the CPNX does not affect the campaign.

Qwertronix:
This allows editing of the "Payment Multiplier" spinnerbox in new AtB-style contracts being added in GM mode, in most situations. It was previously being recalculated in the event handler for _any_ changes to any of the controls in the dialog, including this one. 

It should now allow editing but get recalculated automatically if the planet, factions or contract type are changed. 

Fixes #308

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/NewAtBContractDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/NewAtBContractDialog.java
@@ -588,7 +588,9 @@ public class NewAtBContractDialog extends NewContractDialog {
     		contract.setEnemyQuality(cbEnemyQuality.getSelectedIndex());
     	}
     	
-    	if (needUpdatePayment) { updatePaymentMultiplier(); }
+    	if (needUpdatePayment) {
+            updatePaymentMultiplier();
+        }
     	super.doUpdateContract(source);
     	
     	addAllListeners();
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/NewAtBContractDialog.java
```diff
@@ -577,8 +582,8 @@ protected void doUpdateContract(Object source) {
     		contract.setEnemyQuality(cbEnemyQuality.getSelectedIndex());
     	}
     	
-   		updatePaymentMultiplier();
-   		super.doUpdateContract(source);
+    	if (needUpdatePayment) { updatePaymentMultiplier(); }
```
Dylan-M: This if statement also should be changed to reflect our coding style standards:
    if (conditional) {
        code_here
    }

Rather than the one line version it currently has.

Qwertronix: Will do and update.

PR #640, opened 2018-02-05
Aero System Repairs Errata

mkerensky:
This adds a test of an errata suggestion for aero systems that have multiple hits, allowing the systems to be repaired/replaced based on the number of hits taken rather than rolling a (10+) as per 'other' equipment hit in combat.  An errata entry has been sent to CGL for review - this will allow us to test it while they mull it over.

Also, it breaks out the published repair times for ASF systems from the December 2017 errata.


MekHQ/src/mekhq/campaign/parts/FireControlSystem.java
```diff
@@ -72,15 +72,53 @@ public void updateConditionFromEntity(boolean checkForDestruction) {
...
 	
 	@Override 
 	public int getBaseTime() {
+        if (campaign.getCampaignOptions().useAeroSystemHits()) {
+            //Test of proposed errata for repair times
+            Entity e = unit.getEntity();
+            if (isSalvaging()) {
+                if (e.hasETypeFlag(Entity.ETYPE_DROPSHIP) || e.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
+                    if (e.hasNavalC3()) {
+                        return 8640;
```
Dylan-M: Instead of a fixed number, we should have the base and then have a multiplier for Naval C3.

Dylan-M: In fact, it looks like Naval C3 is always double? If so, I would have this code look like this (Everywhere Naval C3 has an impact):
    int time = 240;
    if (e.hasNavalC3()) {
        time *= 2;
    }
    return time;

MekHQ/src/mekhq/campaign/parts/FireControlSystem.java
```diff
@@ -72,15 +72,53 @@ public void updateConditionFromEntity(boolean checkForDestruction) {
...
+                    return 240;
+                } else {
+                    return 120;
+                }
+            } 
+            if (hits == 2) {
+                if (e.hasETypeFlag(Entity.ETYPE_DROPSHIP) || e.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
+                    if (e.hasNavalC3()) {
+                        return 960;
+                    }
```
Dylan-M: I also see a lot of inconsistency in repair times. Here, the base repair time for AeroSpace is 240, the larger time for DS/JS/WS is 480 (double), and the Naval C3 doubles that again. Yet in other places, the DS/JS/WS value is **much** more than double... it would be nice to have a consistent formula, and then just apply said formula at the end of the method after setting the base value. This is feedback on the proposed errata to have a formula like that, rather than pulling values out of thin air.

MekHQ/src/mekhq/campaign/parts/MissingAeroSensor.java
```diff
@@ -57,7 +57,12 @@ public MissingAeroSensor(int tonnage, boolean drop, Campaign c) {
     
     @Override 
 	public int getBaseTime() {
-		return 1200;
+        //Published errata for replacement times of small aero vs large craft
+        Entity e = unit.getEntity();
+        if (e.hasETypeFlag(Entity.ETYPE_DROPSHIP) || e.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
+            return 1200;
+        }
```
Dylan-M: This is an example of what I meant in the comment above, 1200 for large space vessels is ~4.62* the small craft value. Not consistent at all. We should be proposing a more consistent and simplified errata I think :/

MekHQ/src/mekhq/campaign/parts/MissingAvionics.java
```diff
@@ -53,6 +53,15 @@ public MissingAvionics(int tonnage, Campaign c) {
     
     @Override 
 	public int getBaseTime() {
+        if (campaign.getCampaignOptions().useAeroSystemHits()) {
+            //Test of proposed errata for repair times
+            Entity e = unit.getEntity();
+            if (e.hasETypeFlag(Entity.ETYPE_DROPSHIP) || e.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
+                return 4800;
```
Dylan-M: One final note on the previous 2 inline comments' remarks about consistency: here the large craft is 8 times the small craft time. Major WTF factor ;)

MekHQ/src/mekhq/resources/CampaignOptionsDialog.properties
```diff
@@ -34,6 +34,8 @@ resetToFirstTechCheckBox.text=After repair reset to the technician at the top of
 resetToFirstTechCheckBox.toolTipText=After any repair/salvage task, always selects the technician at the top of the list\n(Can be combined with option to place tech assigned to unit at top of list)
 useDamageMargin.text=Damage/destroy parts by margin of failure
 useDamageMargin.toolTipText=Instead of destroying/damaging parts only by elite techs, parts are damaged/destroyed when the tech roll\nfails by a given margin.
+useAeroSystemHits.text=(Unofficial) Damage/destroy Aero system parts by number of hits taken
```
Dylan-M: Maybe replace (Unofficial) with (Proposed Errata) to make it clear to people what is going on here

MekHQ/src/mekhq/campaign/parts/FireControlSystem.java
mkerensky: There is a rhyme and reason for my times. I left the large craft times as published in SO. I happen to agree with you heartily on their being absurd in some cases, but it was my understanding that RAW is RAW and those get left alone. 
December 2017 errata added the small craft section to aero repairs, which includes sensors and life support but not FCS or avionics, so I put those into this. 
SO as published has repair times for all of these systems, which typically don't get used in MM/MHQ because the system gets destroyed and requires replacement on a roll of 2-9 by default. I took these times into account and tried to make them realistic increments for 1 hit/2hits/3hits where the system is not destroyed. In a lot of cases they're really small, so I referenced Mech repair times in order to make the numbers for small craft less than those for large craft. 
I was thinking that repair time should go up when a system takes multiple hits, but maybe a different option would be for the difficulty modifier to go up. Isn't that how ground unit repairs work, for things like IS and Engines?
If we want to change this to be an unofficial 'sanity check' on large craft repair times, I'd be more than happy to reduce some of the absurd numbers like 6720 and 4320.  I think 5-10x the small craft value is fair. WWII repair times DID happen with 24 hour work in shifts, lots of extra people and a shipyard. SO covers all those things, once the planets project identifies systems with shipyards/factories/stations with repair facilities and I get docking implemented.

Dylan-M: Thats fair, but even a warships damage control parties are more than one man, and more than one party for that matter. Hence my comments that only getting the equivalent of 1 crew member worth of time is... dumb... and that the time should be based on number of crew. Would also give an advantage to carrying more crewmen than the minimum required

mkerensky:
I think I've made all requested changes here.

MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -884,6 +887,18 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
                 }
             }
         });
+        
+        useAeroSystemHitsBox.setText(resourceMap.getString("useAeroSystemHits.text")); // NOI18N
```
NickAragua: One more small change: can we move this above the "Damage/Destroy Parts by Margin of Failure" checkbox?

NickAragua:
Looks good other than the one minor requested change.

mkerensky:
That should take care of it.

Issue #639, opened 2018-02-04
BUG: CO calculates Infantry Skill Level Incorrectly


scJazz2:
Infantry do not have "Piloting" skill they have "Anti-Mech"

Pursuant to this question posted on BT Game Rules...
http://bg.battletech.com/forums/index.php?topic=60067.0

Quoting Slack chat...
hammer-developer [12:28 PM]
Going with lowest of either gunnery or AM skill

scjazz [12:29 PM]
right, which is why the proposed change will be to use gunnery for Fatigue, and Gunnery or Anti-Mek whichever is lower for CO <<<< this?

FIX: Use the Infantry "Gunnery" (aka Small Arms) to determine Infantry Skill Level... Elite, Veteran, Regular, Green.

PR #670, opened 2018-03-15
Adjusted unit rating calculation and infantry counts

NickAragua:
Two sets of changes:

One spurred by issue #639, on line 171 in CampaignOpsReputation.java. This emulates the behavior of the dragoons rating code, which, if no anti-mek skill is present, assumes an anti-mek skill of "small arms + 1".

The rest is because the average skill level wasn't computing correctly due to always dividing the number of infantry by 28, even though infantry units may have man counts of not 28. That also led me to mess around with transport capacity-related code.

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
+++ b/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
@@ -497,110 +497,32 @@ public class CampaignOpsReputation extends AbstractUnitRating {
 
         int totalValue = 0;
 
-        // todo Superheavys.
-        // Find out how short of transport bays we are.
-        boolean doubleExcessCapacity = true;
-        boolean fullCapacity = true;
-        boolean excessCapacity = true;
-        int heavyVeeBays = getHeavyVeeBayCount();
-
-        if (getMechBayCount() == getMechCount()) {
-            excessCapacity = false;
-            doubleExcessCapacity = false;
-        } else if (getMechBayCount() < getMechCount()) {
-            fullCapacity = false;
-            excessCapacity = false;
-            doubleExcessCapacity = false;
-        } else if (getMechBayCount() < getMechCount() * 2) {
-            fullCapacity = false;
-            doubleExcessCapacity = false;
+        TransportCapacityIndicators tci = new TransportCapacityIndicators();
+        tci.updateCapacityIndicators(getMechBayCount(), getMechCount());
+        tci.updateCapacityIndicators(getProtoBayCount(), getProtoCount());
+        tci.updateCapacityIndicators(getHeavyVeeBayCount(), getHeavyVeeCount());
+        
+        // vehicles are the only units that can share heavy/light bays so we have some special logic
+        // we've stuffed all possible heavy vehicles into heavy vehicle bays.
+        // if we have some heavy vehicle bays left over, add them to the light vehicle bay count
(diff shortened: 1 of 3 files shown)


NickAragua:
Better hold off on this one for a bit, looks like it breaks a couple of unit tests.

dericpage:
As I recall Campaign Ops explicitly states to divide by 28; no provision is given for varied unit sizes unless there's an eratta.

NickAragua:
Huh. The more I work on this the more it occurs to me that it's pretty much impossible to accurately calculate transport capacity vs desired transport capacity due to varying infantry platoon/squad weights and sizes.

Dylan-M:
This is where you just calmly tell yourself that the rules writers are idiots... and then make an unofficial rule to make it non-stupid. But, you do that post-stable. ;)

MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
```diff
@@ -494,110 +497,32 @@ public int getTransportValue() {
...
-            excessCapacity = false;
-            doubleExcessCapacity = false;
-        } else if (getSmallCraftBayCount() < getSmallCraftCount()) {
-            fullCapacity = false;
-            excessCapacity = false;
-            doubleExcessCapacity = false;
-        } else if (getSmallCraftBayCount() < (getSmallCraftCount() * 2)) {
-            fullCapacity = false;
-            doubleExcessCapacity = false;
+        TransportCapacityIndicators tci = new TransportCapacityIndicators();
```
NickAragua: Refactored this code (not sure it was really working properly prior anyway) to unify the "sufficient/excess/double" transport capacity computation. This also had the effects of fixing the unit tests.

NickAragua:
As per one of PheonixStorm's comments yesterday, I went back and changed the "excess capacity" logic. Basically, the idea is that, to have "excess" capacity, we only need to have more bay space than we need for one unit type, as long as we have sufficient capacity for everything else. At least, that's how the english language definition of the word "excess" works.

The other logic remains the same.

pheonixstorm:
There might be another issue... If a bay type is not being used it doesn't count.

ie Light Vee bays but only Heavy Vees/Bays. The light bays don't count toward the total.

NickAragua:
That's the case for most bay types. Existing rules allow light vehicles to go into heavy vehicle bays, thus the special case for those.

Dylan-M:
@NickAragua But, they should use light bays first ;)

NickAragua:
I think they do. The code loads all heavy vehicles into heavy vehicle bays first, then assigns the remaining heavy vehicle bays to the light vehicle bay count.

Dylan-M:
@NickAragua That's the opposite of what I said ;)

Logic should be thus:
Fill all light vee bays with light vees
Fill all heavy vee bays with heavy vees
Fill any remaining heavy vee bays with any extra light vees that didn't fit in light bays.

Rather than:
Heavy bays with heavies
Extra heavy bays with lights
Light bays with lights

Which is how I've done it for the transport overview. Unless someone has monkeyed with it.

NickAragua:
I don't think it matters very much for the purposes of the "sufficient, excess, double" calculation. If we were actually placing vehicles, sure, but this is just an estimate.

Dylan-M:
I guess I'm just saying I already wrote the calculations, and I thought I had them in their own re-usable methods so doing it again another way is just reinventing the wheel ;)

NickAragua:
I'll take a look tonight to see if it can be reused without too much crowbar.

PR #673, opened 2018-03-20
fix for inability to acquire ammo in parts acquisition dialog via gm …

NickAragua:
…mode

A random bug I noticed while working on another issue.


NickAragua:
It looks like the file wasn't using 4-space tabs to begin with. I hit it with eclipse's auto-indenter and it looks a little better now.

Issue #677, opened 2018-03-23
Unable to open Campaignoptions


BRDiger:
Hey guys, found another one. Though is suspect that it´s more related to my game and me tinkering than mhq in general.

### Environment
0.43.9 RC3
OS:  Windows 7 + 10
J: V8 U161

### Description
For some builts i experienced problems while trying to open my campaignoptions:
I click it, it and nothing happens.

`Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: (minimum <= value <= maximum) is false
	at javax.swing.SpinnerNumberModel.<init>(Unknown Source)
	at javax.swing.SpinnerNumberModel.<init>(Unknown Source)
	at mekhq.gui.dialog.CampaignOptionsDialog.initComponents(CampaignOptionsDialog.java:1973)
	at mekhq.gui.dialog.CampaignOptionsDialog.<init>(CampaignOptionsDialog.java:463)
	at mekhq.gui.CampaignGUI.menuOptionsActionPerformed(CampaignGUI.java:1674)
	at mekhq.gui.CampaignGUI.access$1000(CampaignGUI.java:163)
	at mekhq.gui.CampaignGUI$12.actionPerformed(CampaignGUI.java:719)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.AbstractButton.doClick(Unknown Source)
	at javax.swing.plaf.basic.BasicMenuItemUI.doClick(Unknown Source)
	at javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.processMouseEvent(Unknown Source)
[...]

PR #683, opened 2018-03-29
Campaign options bulletproofing

NickAragua:
Added some upper-bounds checking when loading some data that goes into value-limited spinners.

Ignore the AcquisitionsDialog.java for the purposes of code reviews, that's a different PR (I forgot to switch to master branch before creating the new branch).

This is for mekhq issue #677


NickAragua:
I agree, I don't see a reason why we should have an upper limit. Camops specifies that 5%/1%/1%/1% are the defaults, but there's no reason the player shouldn't be able to change those values to what they want. If there are no objections, I'll up the max percentage to 100%, leaving the default at 5%.

The relevant spinners do reference the constants in question (see changes to CampaignOptionsDialog.java)

Issue #688, opened 2018-03-30
Cannot slot Endo Steel crits


Linerunner99:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.

### Environment
What version of MekHQ does your issue apply to? 

[Steele Brigade30510222.zip](https://github.com/MegaMek/mekhq/files/1864993/Steele.Brigade30510222.zip)
-0.43.9-RC3
What operating system are you using?  
-Windows
What java version are you using?  
-Java Update 161 (64-bits)

### Description
Describe the issue and provide screenshots ingame if able. 
-When customizing any mech, switching from normal to endo steel structure produces 14 crits to be slotted as it should. However, these crits cannot be slotted anywhere on the mech. Drag & drop, right click and choose a location and even auto allocate functions all fail. Every other component can be slotted just fine, only ES is broken. This also happens if the mech already has ES and you remove any of those crits from the mech. They become unable to slot.

And provide the steps you have taken to reproduce the issue in the campaign file if applicable. 
-It happens all the time with no special steps taken.

### Files
Provide the .cpnx campaign file that this issue occurs in.
Campaign save is attached. Nothing custom in it. No errors are logged when encountering this bug.

neoancient:
What is happening here is that the tabs are refreshed after assigning the slot, the unit is being updated with the values on the structure tab, which removes the assigned internal structure slots as if the IS type is changed. This is not happening in MML, so it's something involving the way MHQ handles refreshes that I haven't been able to find yet.

PR #689, opened 2018-03-31
Fix for #688: Cannot slot Endo Steel crits.

neoancient:
Pretty straighforward.

The call to refreshSummary() is supposed to refresh just the weight/slot
breakdown view on the structure tab, but it was calling refresh()
instead. This was causing a listener to get added a second time, so when
the listener was removed to refresh data without updating the Entity it
was updating anyway. This was causing any critical slots to assigned to
internal structure to be unallocated.

I double-checked the panels for the other unit types, and `MekPanel` was the only one with the typo.


Dylan-M:
Fixes #688: Cannot slot Endo Steel crits

Issue #668, opened 2018-03-13
0.43.8 - RC2   Empty heat sink/double heat sink omnipods not available for Purchase Parts


mkerensky:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.

### Environment
0.43.8 Dev release / Current Eclipse build 3-13-18. 
Windows
Java 8.161  x64

### Description
I am unable to purchase empty Omnipod heat sink parts for mechs.  I can purchase empty omnipods for Aero HS, Aero DHS and Radical HS, but no other types.  I can purchase those heat sink types in omnipod versions, just not empty omnipods to convert fixed sinks with. 
I have tried testing with different campaign dates, Tech Limits options and campaign files and haven't found any combinations where the parts are available. I have also tried different parts list filters, including 'all parts', so I don't think these are misfiled. 

The OmniPod parts filter is cluttered with available empty omnipods for all sorts of built-in large spacecraft parts and structure modifications that should probably not exist. 

### Files
This issue can be reproduced with the stock Fist and Falcon or Wolf and Blake campaigns that ship with MHQ. It's not just my campaign file.

PR #690, opened 2018-03-31
Fix for #668: Empty heat sink/double heat sink omnipods not available for Purchase Parts

neoancient:
Pods were only being added in the general case scenario and not where they required special handling (heat sinks, jump jets, MASC/supercharger, variable weight).

In addition to adding the missing empty pods to the store I changed the filter in the dialog to match either the name or the details so you can actually find the pod you want. I also had the parts store exclude any equipment without the appropriate MiscType or WeaponType flag for mech, vee, or asf equipment.


Dylan-M:
Fixes #668: Empty heat sink/double heat sink omnipods not available for Purchase Parts

Issue #108, opened 2016-06-16
Import from MUL Option Does Not Seem to Load Saved Abilities


mkerensky:
I have a bunch of MUL files with pilot abilities and custom mech settings saved into them.  Imported pilots are not getting the abilities, but they are getting their skills and the mech settings (such as reduced MG ammo) are importing correctly.  I have attached one of the MUL files for testing. 
[337th_2nd_Titan.zip](https://github.com/MegaMek/mekhq/files/319146/337th_2nd_Titan.zip)

MekHQ version is 0.3.29.
These MUL files were created with previous versions of MegaMek. The last save was probably under 0.41.15.

pheonixstorm:
I don't think the MUL import was designed to handle this though I could be wrong. Were you able to do this under a previous version of HQ?

mkerensky:
This is the first time I've used HQ. I saw that HQ can handle SPAs for pilots and crews as well as MUL imports and made a first-timer's assumption that HQ would import the SPAs if the MUL file contained them.

pheonixstorm:
There is an import for personnel to transfer from one campaign to another is this is what you are trying to do. If this is an import from something you did in MM then it will probably have to be a feature request to do so. ralgith would know more about it, though with RL claiming his free time it might be awhile before you get an answer.

mkerensky:
I was porting a campaign into MekHQ from saved MM files and a word document. I just edited all of the HQ versions of my pilots and gave them their SPAs back.

Dylan-M:
Fixed by #692

PR #692, opened 2018-04-01
Fix for #108 - implement SPA import from MUL

NickAragua:
In most cases, this should allow SPAs to properly come in from a MUL file.


NickAragua:
In case of multi-person crews, I opted to give everyone the SPAs as MULs don't really have the granularity to give SPAs to individual crew members.

Dylan-M:
Fixes #108

PR #699, opened 2018-04-10
New day cleanup

### LINKED ISSUE #698: Income/Expenses Processed in wrong order
0.43.8-RC2-windows

When advancing to a new day, expenses are paid prior to any income: this creates a problem if you do not have enough cash on hand without the payment.

Example: on the first of the month you usually receive contract payments and pay salaries and overhead. If you have 750k salary expenses, 500k cash on hand, and 2 million c-bill contract pay you will be told you have a shortfall. The Salary will attempt to pay prior to receiving the contract payments; even though they're received the same day.

Income should always be processed prior to expenses, assuming they're the same day.

Dylan-M:
Should fix #698, and if it doesn't we'll have to dig deeper.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2689,8 +2689,9 @@ public class Campaign implements Serializable, ITechManager {
 
         processNewDayUnits();
 
-        // check for anything else in finances
+        // check for anything in finances
         finances.newDay(this);
+
         MekHQ.triggerEvent(new NewDayEvent(this));
         return true;
     }
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -226,10 +226,10 @@ import mekhq.gui.utilities.PortraitFileFactory;
 public class Campaign implements Serializable, ITechManager {
     private static final String REPORT_LINEBREAK = "<br/><br/>"; //$NON-NLS-1$
 
-	private static final long serialVersionUID = -6312434701389973056L;
+    private static final long serialVersionUID = -6312434701389973056L;
 
-	private UUID id;
-	
+    private UUID id;
+
     // we have three things to track: (1) teams, (2) units, (3) repair tasks
     // we will use the same basic system (borrowed from MegaMek) for tracking
     // all three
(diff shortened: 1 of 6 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -226,10 +226,10 @@
 public class Campaign implements Serializable, ITechManager {
     private static final String REPORT_LINEBREAK = "<br/><br/>"; //$NON-NLS-1$
 
-	private static final long serialVersionUID = -6312434701389973056L;
```
NickAragua: These Game.java changes are just whitespace formatting, right?

MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -231,6 +234,122 @@ public void addLoan(Loan loan) {
...
+                    if (debit(shares, Transaction.C_SALARY, "Shares payments for " + contract.getName(),
+                            calendar.getTime())) {
+                        campaign.addReport(formatter.format(shares) + " C-bills have been distributed as shares.");
+                    } else {
+                        /*
+                         * This should not happen, as the shares payment is less than the contract
+                         * payment that has just been made.
+                         */
+                        campaign.addReport(
+                                "<font color='red'><b>You cannot afford to pay shares!</b></font> Lucky for you that personnel morale is not yet implemented.");
```
NickAragua: Can we tokenize the text in this method?

MekHQ/src/mekhq/campaign/Campaign.java
Dylan-M: Well... that isn't Game.java, but no. Review at the commit level rather than the full change level. I did the whitespace as a separate commit so that the real changes could be seen by looking at commits.

MekHQ/src/mekhq/campaign/finances/Finances.java
Dylan-M: Sure I guess, though it could also wait until the code that matters is actually implemented. This was just a refactor to move code out of Campaign.java and into Finances.java where it belongs (and also into smaller methods). But adding this text to the resource bundle is easy.

NickAragua: Right, I picked a poor line for the comment. We can probably ignore text for "placeholder" mechanisms. I was thinking about the other text for features that have been implemented.

Dylan-M: I've added done the resource bundle work here. Another case of review by commit instead of the final result since I did another whitespace format commit and a couple of merge conflict commits too.

PR #701, opened 2018-04-15
Update Mission.planetName to Mission.planetId for consistency.

### LINKED ISSUE #328: If you accept a contract on Taussen, the game teleports you back to Terra once you arrive
To replicate this generate and then pick a contract with a low travel time. Edit the contract so that Taussen is the location, then use the GM tools to move your troops there. Once you are on Taussen advance days until the first mission pops up, this should ensure that the contract is active. Then all you have to do is save and reload the game and you should be on Terra.

I discovered this bug when I took a security contract from Comstar on Taussen using MekHQ v42.0. I replicated it in other campaigns and other versions of mekhq.

Dylan-M:
It was used as both name and ID, which is not correct as they're 2 different fields.
I chose to make it the ID since that is what it should always have been.
This is a dangerous merge and needs thorough testing prior to stable.
This is likely to delay stable by another month.

Fixes #328 If you accept a contract on Taussen, the game teleports you back to Terra once you arrive

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2296,7 +2296,7 @@ public class Campaign implements Serializable, ITechManager {
              * system for transport or splitting the unit, etc.
              */
             if (!getLocation().isOnPlanet() && // null != getLocation().getJumpPath() &&
-                    getLocation().getJumpPath().getLastPlanet().getId().equals(m.getPlanetName())) {
+                    getLocation().getJumpPath().getLastPlanet().getId().equals(m.getPlanetId(null))) {
                 /*
                  * transitTime is measured in days; round up to the next whole day, then convert
                  * to milliseconds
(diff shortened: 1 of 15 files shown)


Dylan-M:
@neoancient They should always come out the same if the current date is correct, but I agree. It is... annoying.

What are your thoughts on improvements?

neoancient:
If two planets have the same display name in the current campaign year it will match the first one in the iteration, which may not be the one that was selected. For the most part those planets have deliberately been given modified display names, such as `Foo (Clan)` when there is a world in Clan space that was named after an IS world. I don't actually know whether the data currently has such potential name collisions currently, but there is nothing that keeps them from occurring. We don't even currently have a way to check against two planets having the same id, which should at least be logged in `Planets.addPlanet()`.

As for `JSuggestField`, my thoughts are to parameterize it and add a field for `Function<T, String>` that would handle conversion from `Planet` to the current display name, but I don't have time to work on it today.

jayof9s:
There should not be any duplicate names now or in the future.

That said, we used to ONLY have the Name, which is why it was also the ID. Things just didn't get updated after that.

Also, I see this is in terms of Contracts, is this also going to affect the 'current location' for a save file? Which can be completely independent of contracts/campaigns/etc.

Dylan-M:
This PR only addressed Missions (and by extension Contracts). I'm pretty sure current location already uses the ID.

jayof9s:
When I tested it, I didn't look at contracts. I literally just traveled to the Taussen (which worked okay) but then saved/reloaded and was back on Terra. So I don't think it is working correctly.

PR #707, opened 2018-04-19
Added MissingOmniPod class needed for acquisition.

neoancient:
The `OmniPod` class represents a part that does not have the equipment installed in it, which can only exist in the warehouse (an intact pod is indicated by a boolean field on the `EquipmentPart`). For the purposes of acquisition, it requires a corresponding `MissingPart`

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MissingOmniPod.java
+++ b/MekHQ/src/mekhq/campaign/parts/MissingOmniPod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 MegaMek team
+ * Copyright (C) 2018 MegaMek team
  * 
  * This file is part of MekHQ.
  * 
@@ -29,7 +29,6 @@ import megamek.common.Entity;
 import megamek.common.EquipmentType;
 import megamek.common.MiscType;
 import megamek.common.TechAdvancement;
-import megamek.common.TechConstants;
 import megamek.common.logging.LogLevel;
 import mekhq.MekHQ;
 import mekhq.MekHqXmlUtil;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/MissingOmniPod.java
```diff
@@ -0,0 +1,231 @@
...
+                        if (et instanceof MiscType
+                                && (et.hasFlag(MiscType.F_HEAT_SINK)
+                                        || et.hasFlag(MiscType.F_DOUBLE_HEAT_SINK)
+                                        || et.hasFlag(MiscType.F_IS_DOUBLE_HEAT_SINK_PROTOTYPE))) {
+                            partType = new HeatSink(0, et, -1, false, campaign);
+                        } else if (et instanceof MiscType && et.hasFlag(MiscType.F_JUMP_JET)) {
+                            partType = new JumpJet(tonnage, et, -1, false, campaign);
+                        } else if (et instanceof MiscType
+                                && et.hasFlag(MiscType.F_MASC)
+                                && (et.getSubType() & MiscType.S_SUPERCHARGER) == 0) {
```
mkerensky: Per TM, MASC can't be placed in an omnipod.  It looks like a Supercharger can be.  We should probably also allow other variable weight equipment to write to / load from XML, like hatchets, targeting computers and such. I don't see anything that says those items can't be omnipods.

Dylan-M: @mkerensky Re-read that check, it **is** checking for a Supercharger. Code defines that as a subtype of MASC. TCs can definitely be pod mounted, almost all the Clan ones are. I don't know about physical weapons, but maybe.

neoancient: This has nothing to do with determining whether a part can be podded or not. It deals with tracking what type of part the pod is built for writing whatever data is needed to restore the `EquipmentPart` to the campaign file.

Issue #706, opened 2018-04-19
ArrayIndexOutOfBoundsException in BaseAttackBuiltInScenario


kipstafoo:
The change added on 11/27 for orthogonal deployment introduced a possible bug:

int attackerStartIndex = Compute.randomInt(4); <-- This can be a zero, which is fine
int angleChange = Compute.randomInt(2) - 1; <-- This can be a -1, which is 'mostly' fine
int secondAttackerForceStart = startPos[(attackerStartIndex + angleChange) % 4]; <-- Combined together, things go poorly

Issue #695, opened 2018-04-02
Base Attack (Defense) mission generate with no Allies or OpFor in AtB


BlueThing00:
### Environment
MekHQ: 0.43.9
OS: Windows 10
Java: 1.8.0_161-b12

### Description
A Base Attack (Defense) mission was generated with no Allies or Opfor.

This exception is in mekhqlog.txt:
```
Exception in thread "AWT-EventQueue-0" java.lang.ArrayIndexOutOfBoundsException: -1
	at mekhq.campaign.mission.atb.scenario.BaseAttackBuiltInScenario.setExtraMissionForces(BaseAttackBuiltInScenario.java:83)
	at mekhq.campaign.mission.AtBScenario.setStandardMissionForces(AtBScenario.java:722)
	at mekhq.campaign.mission.AtBScenario.setForces(AtBScenario.java:655)
	at mekhq.campaign.mission.atb.AtBScenarioFactory.createScenariosForNewWeek(AtBScenarioFactory.java:286)
	at mekhq.campaign.Campaign.newDay(Campaign.java:2415)
	at mekhq.gui.CampaignGUI.advanceDay(CampaignGUI.java:1406)
	at mekhq.gui.CampaignGUI.access$2100(CampaignGUI.java:163)
	at mekhq.gui.CampaignGUI$48.actionPerformed(CampaignGUI.java:1224)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(Unknown Source)
	at java.awt.AWTEventMulticaster.mouseReleased(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
[...]

Issue #674, opened 2018-03-21
AtB fails to generate opfor


bczretribution:
### Environment
mekhq 43.7 rc 2

### Description
AtB occasionally generates new battles and fails to generate an opfor.  shown image is the second time this contract.  note how there are no folders listing terran hegemony or LC units for this base defense
![glitch](https://user-images.githubusercontent.com/34722287/37734516-d0c42832-2d21-11e8-8285-3294d56b17e6.jpg)

### Files
[glitch.zip](https://github.com/MegaMek/mekhq/files/1835067/glitch.zip)

NickAragua:
Next time you see this, could you post the mekhqlog file as well?

bczretribution:
Just had it happen earlier today,  ill grab it on the next one.

bczretribution:
[glitch.zip](https://github.com/MegaMek/mekhq/files/1862028/glitch.zip)
here you go, one campaign save with mekhq log file.  this one also contains a bonus potentially related bug where its assigning units to missions that were not deployed, as well as the no opfor bug.

NickAragua:
Missing Clint CLNT-2-3b, Overlord (2775) Combined Arms customs.

bczretribution:
Sorry about that, I haven't figured out yet how to tell which customs are required and which ones aren't.  this raises another interesting question, however, as there is no custom for the overlord.  I've scoured my folders and could not find one, I then loaded up the campaign file and double checked the stats of the overlord there with one opened from the cache in MML and confirmed that they were identical.  I've added the clint into the zip, but I am not sure what to do about the overlord.  If you want me to zip the entire mechfiles folder and upload a google drive link, let me know.

[glitch.zip](https://github.com/MegaMek/mekhq/files/1866112/glitch.zip)

NickAragua:
Generally speaking, anything in mechfiles/customs should be supplied.

NickAragua:
I think I see the problem. The Terran Hegemony doesn't have a RAT entry for the year you're playing in. Until the issue is addressed, you can always manually generate the allied forces (probably using a more generic RAT).

bczretribution:
[...]

PR #712, opened 2018-04-24
fix potential array out of bounds error

NickAragua:
Now, we're guaranteed to get a number between 0 and 3 for the "secondary start location". Addresses #706, and probably #695 and #674


MekHQ/src/mekhq/campaign/mission/atb/scenario/BaseAttackBuiltInScenario.java
```diff
@@ -80,7 +80,7 @@ public void setExtraMissionForces(Campaign campaign, ArrayList<Entity> allyEntit
 
 		// the "second" force will be deployed (orthogonally) between 90 degrees clockwise and counterclockwise from the "primary force".
 		int angleChange = Compute.randomInt(2) - 1;
```
neoancient: I'm not sure the original code is correct here. `Compute.randomInt(2)` will produce either 0 or 1, so `angleChange` is -1 or 0, and the second attacker is given a start position that is either 90 anticlockwise or the same as the first. The change shifts it to clockwise instead if the first attacker position is north instead of throwing an exception.

NickAragua:
Ok, I see what I did. I forgot I had the other pull request still pending.

The changes to Mission.java are 99% whitespace. I added legacyPlanetName on line 69, and a method getPlanetName() on line 134. Also, logic to handle "legacy" planet names when saving/ loading on lines 231/281.

"Legacy" meaning a planet name that's specified for a contract, but, for whatever reason, does not show up in our planets "database". Rather than going berserk, the UI will output the old planet name.

Issue #715, opened 2018-04-25
[0.43.9-RC3 : Bug] Unit.isCommander Error


SirMegaV:
### Environment

0.43.9-RC3

### Description

Screen corrupted with Unit.isCommander(Unit.java:3730) error. This has been going on for years, only now reporting.

To replicate, load campaign file attached.
1. At Personnel tab, sort by Unit Assignment.
2. At TO&E tab, go to 12th Combined Armored Infantry Company.
3. Select Sergeant Lizbeth Minet, right click, select Go to Pilot/Commander in Personnel.
4. Select **ALL** personnel in the same unit, right click, assign to "Motorized Platoon (SRM) ## 12th Combined Armored Infantry Company"
5. Click TO&E tab. Nothing will happen. Error "Unit.isCommander(Unit.java:3730)" found in the log. Log attached. 
6. Although the screen fail to refresh, if you scroll up and down, it will reappear.
7. Basically this will almost always happen when you remove all personnel from a unit that is currently in the TO&E.

### Files

[Customs - SirMegaV - Mechfiles.zip](https://github.com/MegaMek/mekhq/files/1945335/Customs.-.SirMegaV.-.Mechfiles.zip)
[Unit.isCommanderError.cpnx.zip](https://github.com/MegaMek/mekhq/files/1945336/Unit.isCommanderError.cpnx.zip)
[Unit.isCommanderError.txt](https://github.com/MegaMek/mekhq/files/1945337/Unit.isCommanderError.txt)

NickAragua:
Fixed by PR #718

PR #718, opened 2018-04-27
Fix for #715: Unit.isCommander Error

neoancient:
isCommander() does not check the return value of getCommander() for null, which is possible when the unit has no assigned crew. Added a null check and also added javadoc comments to both methods.

If following the steps to recreate given in #715, you can find the infantry unit in question in Order of the Scale Combined Arms Regiment > 1st Combined Armored Battalion > 12th Combined Armored Battle Company > 12th Combined Armored Infantry Company.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -3747,14 +3747,14 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
      * 
      * @param person A <code>Person</code> in the campaign. The person need not be assigned to the unit as
      *               crew, in which case the return value will be false.
-     * @return       Whether the person is considered the unit commander. A unit with no crew has no
-     *               commander and this method will return false.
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3726,8 +3742,19 @@ public boolean isGunner(Person person) {
...
+     *               crew, in which case the return value will be false.
+     * @return       Whether the person is considered the unit commander. A unit with no crew has no
+     *               commander and this method will return false.
+     * 
+     * @see {@link #getCommander()}
+     */
     public boolean isCommander(Person person) {
-        return person.getId().equals(getCommander().getId());
+        Person commander = getCommander();
+        return (null != commander) && person.getId().equals(commander.getId());
```
dericpage: Probably should add a null-check for "person" just to be on the safe side.

neoancient: Agreed

PR #725, opened 2018-05-11
minor adjustment to dragoons rating

NickAragua:
Dragoons rating counted unmanned units for the purposes of "average skill" calculations. getCrew() never returns null, but can return a 0-length list, so the code was adjusted to account for that.

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
+++ b/MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
@@ -106,7 +106,7 @@ public class FieldManualMercRevDragoonsRating extends AbstractUnitRating {
                                 " is mothballed.  Skipping.");
                 continue;
             }
-
+            
             updateUnitCounts(u);
             BigDecimal value = getUnitValue(u);
             getLogger().log(getClass(), METHOD_NAME, LogLevel.DEBUG,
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
```diff
@@ -106,7 +106,7 @@ protected void initValues() {
                                 " is mothballed.  Skipping.");
                 continue;
             }
-
+
```
neoancient: Nitpicking: addition of extraneous whitespace.

MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
```diff
@@ -428,7 +428,7 @@ private void updateSkillLevel(Unit u, BigDecimal value) {
         final String METHOD_NAME = "updateSkillLevel(Unit, BigDecimal)";
 
         //Make sure this is a combat unit.
-        if ((null == u.getEntity()) || (null == u.getEntity().getCrew())) {
+        if ((null == u.getEntity()) || (u.getCrew().size() == 0)) {
```
neoancient: We could check for both null and empty, but I don't think there's any situation where crew can be null, so this should be fine.

NickAragua: Yeah, first thing the getCrew() method does is initialize a new collection.

Issue #748, opened 2018-06-09
[0.43.11-RC5 : Bug] Fatal error when "Restore Unit" Black Eagle


SirMegaV:
### Environment

0.43.11-RC5

### Description

Fatal error when "Restore Unit" Black Eagle. Game will be frozen. Nothing in log. Might be related to case #745 and #747. To replicate:

1. Load campaign
2. Go to hanger
3. Right click, GM Mode, Restore Unit
4. Game frozen

![image](https://user-images.githubusercontent.com/17064665/41187058-eb6b8288-6bd4-11e8-9d30-1b81b99d3231.png)

### Files

[Fatal error when Restore Unit Black Eagle.zip](https://github.com/MegaMek/mekhq/files/2086588/Fatal.error.when.Restore.Unit.Black.Eagle.zip)

neoancient:
The InvocationTargetException in NativeMethodAccessorImpl.invoke0 is a wrapper around a different exception thrown in the process of dispatching an event. The cause is the exception thrown by the event listener, which in this case is a NullPointerException when attempting to update the daily log. The stack trace from the original exception is truncated, but this is what is in the log:

```Caused by: java.lang.NullPointerException
	at mekhq.gui.DailyReportLogPanel.appendLog(DailyReportLogPanel.java:112)
	at mekhq.gui.CampaignGUI.refreshReport(CampaignGUI.java:2969)
	at mekhq.gui.CampaignGUI.handle(CampaignGUI.java:3029)
	... 48 more```

I was unable to replicate it and there is nothing at that point in the code that suggests a cause to me.

SirMegaV:
I tested once again, using the attached campaign file. Same error. The screen frozen, no log messages. 

What I did.
1. Download and unzip attached campaign
2. Run RC5, load campaign
3. Go to hanger tab
4. Right click Black Eagle, GM Mode, Restore Unit
5. Screen frozen

Log attached again.

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/2092928/mekhqlog.txt)

NickAragua:
Got it. Looks like some kind of endless loop.

NickAragua:
Fixed with #751

PR #751, opened 2018-06-13
Fix for endless loop causing issue #748

NickAragua:
This was an amazing example of an endless loop that didn't eventually run out of memory.

Basically, the loop was:
a) Not actually replacing the missing parts.
b) Re-initializing the parts list from somewhere so that the missing part would pop back up even if it was replaced. 

Changed it to actually replace the missing part. Also fixed an issue where "restore unit" wouldn't actually reload ammo bins.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
@@ -510,6 +510,7 @@ public class UnitTableMouseAdapter extends MouseInputAdapter implements
                             part.setTeamId(null);
                             part.cancelReservation();
                             part.remove(false);
+                            needsCheck = true;
                         } else {
                             if(part.needsFixing()) {
                                 needsCheck = true;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -526,7 +520,25 @@ public void actionPerformed(ActionEvent action) {
...
+                        if(part instanceof Armor) {
+                            final Armor armor = (Armor) part;
+                            armor.setAmount(armor.getTotalAmount());
+                        } else if(part instanceof AmmoBin) {
+                            final AmmoBin ammoBin = (AmmoBin) part;
+                            
+                            // we magically find the ammo we need, then load the bin
+                            // we only want to get the amount of ammo the bin actually needs                            
+                            if(ammoBin.getShotsNeeded() > 0) {
+                                ammoBin.getAcquisitionWork(ammoBin.getShotsNeeded()).find(0);
```
neoancient: `AmmoStorage.find(int)` calls Campaign.buyPart(Part, int)`, which debits the cost of the ammo from the finances (unless paying for parts is disabled). This is not what I would expect with a GM restore.

I tested with this and it appeared successful:
```                            
if (ammoBin.getShotsNeeded() > 0) {
    ammoBin.setShotsNeeded(0);
    ammoBin.updateConditionFromPart();
}
```

NickAragua:
Made the suggested change. Also, put back the "needsCheck = true" in the part replacement code, as sometimes the part being replaced is an omnipod.

PR #760, opened 2018-06-19
Added arbitrary contract score modifier

Qwertronix:
Adds a simple spinner found in the "Edit Mission" dialog for the player to set a custom, arbitrary contract score modifier, mostly as a "crutch" to handle conditions not currently covered by MekHQ for now. 

Looks like this, in a new GM-created Contract that started at the usual contract score of 1: 
image

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -148,6 +148,7 @@ public class AtBContract extends Contract implements Serializable {
 	
 	protected int playerMinorBreaches;
 	protected int employerMinorBreaches;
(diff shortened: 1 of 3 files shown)


NickAragua:
Why not just give the ability to directly set the contract score instead?

Qwertronix:
The contract score is calculated on every `getScore()` call based on scenario results, contract breaches, etc. rather than being incremented or decremented by the events. So it would take a bit more rework that might not be functionally identical to what the code does currently in order to have the user "set" the score directly, and then some more to have subsequent events/scenarios continue to adjust that score after the user has changed it. 

An additive/subtractive modifier seemed altogether the path of least resistance with the current contract implementation.

PR #764, opened 2018-06-21
make link on the Help->About page a clickable URL link

firefly2442:
Tested on Linux and Windows.


neoancient:
Why a button? I think hypertext would fit the look of the dialog better. And I agree with @NickAragua. It's butt-ugly. It could at least use some padding around the text.

firefly2442:
Changed to JLabel to mimic a hyperlink per feedback.

PR #766, opened 2018-06-22
Add graphs to Finance tab to show information about transactions

firefly2442:
Adds a daily balance line graph and a monthly revenue and expenditures barchart to the Finances tab through the [JFreeChart library](https://github.com/jfree/jfreechart).


firefly2442:
I believe the JSplitPane is working correctly now.  This is ready for review.

Issue #121, opened 2016-07-08
Purchase Parts En-Masse


VampireSeraphin:
For large units in MekHq (anything over company sized) the current system for parts acquisition is extremely tedious. Only one type of part can be purchase at a time (for example, six 30ton upper arm actuators).

A system should be implemented where multiple different parts can be selected and purchased at once. For example, the player should be able to highlight 30 ton mech head, 35 ton mech head, & 40 ton mech head all at once, then right click and have the option to purchase one, or more, of each, in the parts dialog window.

Likewise, the player should be able to highlight some, or all, of the parts in use in the parts in use window, and purchase/gm-add X number of each highlighted part.

This will remove the need to create bone yard files for very large units every time you want to make a new one.

BLOODWOLF333:
I second this request and for everyone else on the forums who posted on the IO thread.  http://bg.battletech.com/forums/index.php?topic=51039.0

scJazz2:
I would add turn off acquisition rolls and XP when doing it

HammerGS:
Another suggest to this is have an Order all parts for that unit option.

So if I have an Atlas I can have a one button option that buys me an Atlas but as parts only.

VampireSeraphin:
I like Akjosh's suggestion as a secondary add in addition to this large individual parts select. 

Right click on the unit(s) -> Order > Replacement body parts, replacement actuators, replacement engine, replacement gyro, replacement systems, replacement for all parts, ammo and fuel

molemanBT:
I would like a reminder of how many parts of the item I just used I have remaining, i.e. if I replace a lower arm conduit, I'd like to know I have 1 left and so need to order a few more. Would save the mad scramble when you run out of armour all of a sudden!

BLOODWOLF333:
We kinda already have that.  In the overview tab/parts in use it lists how many parts you have in use and then how many you have stored in the warehouse back when AJ updated it.

VampireSeraphin:
Has anyone looked into this recently? For players of large campaigns it is one of the remaining very large quality of life improvements yet to be done.

PR #768, opened 2018-06-23
allow purchasing and adding of parts en-masse, addresses issue: #121

firefly2442:
Use CTRL + CLICK to select multiple rows.

If someone could double-check the logic, especially all of those booleans that would be great.  I had to refactor a bit for the bulk case.  I believe it should all work for purchasing (single and bulk) as well as adding (GM mode).  I didn't test the bonus and left that as-is single selection since it has to decrement the bonus amount.  I also don't have a campaign to test that condition.


NickAragua:
It looks like you'll need to resolve a conflict before this can be merged.

firefly2442:
Resolved and rebased.

PR #769, opened 2018-06-23
add weight tooltip to tonnage JLabels on the force view

firefly2442:


pheonixstorm:
Outside of AtB why is this needed?

firefly2442:
It's not, should I add a check to see if AtB is enabled?

pheonixstorm:
That would probably be the best way to do it. I don't think any of the current rules have a set tonnage limit for different lances. Even the formation building rules in CO only sets a limit on the number of X weight class for a given lance which in most cases would push a lances weight well beyond what is in AtB.

firefly2442:
OK thanks.  Yeah I was just thinking it would help people setting up their forces before flipping to the Briefing tab and seeing that what they expected was going to be a light lance ended up being a medium lance for example.

firefly2442:
Tweaked the commit to check if AtB is enabled per request.

Issue #775, opened 2018-06-29
Newsfeed as seperate drop down


VampireSeraphin:
Can we get the news feed broken off of the daily report and put into its own separate drop down. I have lost track of how many times I've accidentally clicked past a day with a news story that I would have liked to have read.

A drop down that opens a new panel with all the news stories from say, the last 30-45 days would be perfect.

PR #792, opened 2018-07-08
adds an in-memory daily log history that is reset at each game launch

firefly2442:
addresses issue: #775

This one was kinda tricky.  I had expected that the daily log was saved but after discussion in Slack it turns out this is not saved to the campaign file.  Given the potential size, it was recommended that this not be added.  The daily log utilizes the `DailyReportLogPanel.java` on the right hand side.  However, in adding entries, many of them are interspersed with HTML tags.  I was able to re-use the `LogEntry` class which is used for Personnel logs which I think is good.  The logs are cleared at each game re-launch.  There was discussion about exporting or saving the logs in a separate file but I thought this would at least tide us over for now.  Because of the intermingling of the date and the HTML in the log panel, I had to do some checks on the date to identify if it was a true entry to add or not.  The alternative I believe is refactoring `addReport(String r)` to something like `addReport(Date date, String r)`.  This however would involve a ton of work across many files to change.

If you have comments or suggestions on a better way to achieve this, please let me know.


sixlettervariables:
Reviewing the usages of `addReport` which bring over an HTML encoded date, show that only two places in the code using it, [Campaign.java](https://github.com/MegaMek/mekhq/blob/097b7bd3ff5d4325056f461408a173571ce4ab5d/MekHQ/src/mekhq/campaign/Campaign.java#L2670):
```java
        addReport("<b>" + getDateAsString() + "</b>");
```
And, [DataLoadingDialog.java](https://github.com/MegaMek/mekhq/blob/85821b0b835eb8235b11ba982c8086a65df4523a/MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java#L270):
```java
        			campaign.addReport("<b>" + campaign.getDateAsString() + "</b>");
```
I may have missed one or two, but that's not a lot of value for a regular expression replacement plus a string parse, with exceptions being the norm.

Reviewing those two usages, they both begin a daily report. Consider creating a new method `beginReport` and refactoring out the shared behavior into a new method `addReportInternal`:
```java
    public void beginReport(String r) {
        inMemoryLogHistory.add(new LogEntry(getDate(), ""));

        addReportInternal(r);
    }

    public void addReport(String r) {
        String htmlStripped = r.replaceAll("<[^>]*>",""); // remove HTML tags
        inMemoryLogHistory.add(new LogEntry(getDate(), htmlStripped));

        addReportInternal(r);
    }

    private void addReportInternal(String r) {
        currentReport.add(r);
        if( currentReportHTML.length() > 0 ) {
            currentReportHTML = currentReportHTML + REPORT_LINEBREAK + r;
            newReports.add(REPORT_LINEBREAK);
            newReports.add(r);
        } else {
            currentReportHTML = r;
            newReports.add(r);
        }
        MekHQ.triggerEvent(new ReportEvent(this, r));
    }
```
This avoids using exceptions as control flow and still maintains the intent of the majority of usages of `addReport`.

firefly2442:
Thanks for the review.  That was very helpful.  I was able to refactor the code per your suggestion.  The regex is still there but the exception for control flow is now gone.  The `addReport` is actually used in a ton of places.  If you just add a parameter to it and try to compile you can see how extensive it is.

```
$ grep -roh "addReport(" . | wc -w
177
```

Many of the calls send back HTML formatted strings.

sixlettervariables:
Indeed, you would only need to switch out the two calls that send a date (unless I missed a few more) with beginReport. The rest could probably stay the same.

firefly2442:
I added the ability to enable/disable this feature given the concern on Slack over memory usage and people running MekHQ on lower-end hardware.  I also limited the history to 120 days.  I used this commit: https://github.com/MegaMek/mekhq/pull/799/commits/89250ac91d94a11bc407fcb94d7018a1f204867d as an example of storing a new option in the Campaign Options.

MekHQ/src/mekhq/gui/dialog/HistoricalDailyReportDialog.java
```diff
@@ -0,0 +1,168 @@
...
+import mekhq.gui.CampaignGUI;
+import mekhq.gui.DailyReportLogPanel;
+
+
+
+public class HistoricalDailyReportDialog extends JDialog {
+
+    private static final long serialVersionUID = -4373796917722483042L;
+    
+    public static final int MAX_DAYS_HISTORY = 120;
```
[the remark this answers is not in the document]
firefly2442: Ah thanks, adjusted the code to account for days not entries in `Campaign.java`.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3160,7 +3165,37 @@ public void setRetainerEmployerCode(String code) {
         retainerEmployerCode = code;
     }
 
+    private void addInMemoryLogHistory(LogEntry le) {
+        if (inMemoryLogHistory.size() != 0) {
+            long diff = le.getDate().getTime() - inMemoryLogHistory.get(0).getDate().getTime();
+            while ((diff / (1000 * 60 * 60 * 24)) > HistoricalDailyReportDialog.MAX_DAYS_HISTORY) {
```
sixlettervariables: (these are just my general notes, as a random internet person, and not something important to do)

Constantly removing from the head and adding to the tail of a `List<>` data structure is not terribly efficient, due to all of the copies performed (practical viewpoint: who cares, this isn't going to be a huge list, so O(n) is lost in the jitter).

If the data structure was instead a `TreeMap<Integer, List<LogEntry>>` the prune operation becomes easier (although you'd likely want to ditch the `LogEntry` in favor of a new `DailyLog` object which you could reuse and append log entries to).
```java
// assumes: TreeMap<Long, xxx> inMemoryLogHistory;
Long currentTime = le.getDate().getTime();
Long pruneBefore = currentTime - (HistoricalDailyReportDialog.MAX_DAYS_HISTORY * 1000 * 60 * 60 * 24);
inMemoryLogHistory.headMap(pruneBefore).clear();
xxx dailyLog = inMemoryLogHistory.get(currentTime);
if (dailyLog == null) {
  // ...add a new daily log
}
else {
  // ...append to the daily log
}
```
Taking it one step further you'd abstract this out entirely into a `CampaignReport` class which would be used for all campaign reporting and campaign history, hiding the machinery from `Campaign` (promoting reuse while decreasing internal coupling). This would also promote further enhancements to reporting and historical logging.

Again, I'm just some internet person who dropped by: zero requirement to follow any of this.

NickAragua: From a pure computer science point of view, you want a Queue. You probably want a LinkedList, as that implements the Queue interface.

MekHQ/src/mekhq/resources/CampaignOptionsDialog.properties
```diff
@@ -307,3 +308,5 @@ chkCapturePrisoners.text=Capture Prisoners in Scenarios
 chkCapturePrisoners.toolTipText=If checked, you will have the opportunity to capture prisoners when resolving scenarios
 chkAssignPortraitOnRoleChange.text=Automatically assign portrait on role change
 chkAssignPortraitOnRoleChange.toolTipText=With this enabled, a person without a portrait will automatically gain a random portrait when their primary role is switched to one of those selected below
+chkShowHistoricalDailyReport.text=Show the historical daily report for each gaming session
```
NickAragua: Seems a little verbose. Maybe "Temporarily retain daily log history", or something like that?

firefly2442:
Per feedback:
* switched to LinkedList
* added some comments to the public methods and constants
* adjusted `chkShowHistoricalDailyReport.text` to not be so verbose

Let me know if I missed anything.

MekHQ/src/mekhq/gui/dialog/HistoricalDailyReportDialog.java
```diff
@@ -0,0 +1,174 @@
...
+            JLabel notice = new JLabel(resourceMap.getString("enableInCampaignOptions.text"));
+            getContentPane().add(notice, gridBag);
+        }
+    }
+
+    private void updateLogPanel(Integer days) {
+        logPanel.clearLogPanel();
+        Date trackDay = null;
+        for (int i = gui.getCampaign().inMemoryLogHistory.size()-1; i >= 0; i--) {
+            long diff = gui.getCampaign().getDate().getTime() - gui.getCampaign().inMemoryLogHistory.get(i).getDate().getTime();
```
sixlettervariables: Avoid indexers like `get(n)` on a linked list, as these must traverse every index less than `n` to get there. Switch to a `for` loop over the elements instead.

sixlettervariables: Hoist `getCampaign().getDate().getTime()` into a local variable outside of the loop.

firefly2442:
Thanks.

* switch to for loop over elements
* move constant getTime() long outside loop

PR #795, opened 2018-07-09
add Java vendor and version to the About dialog

firefly2442:


neoancient:
The About box should contain information about the application itself. Information about the environment should be added to the log file, but is unnecessary clutter in the About box.

firefly2442:
This same information is in the MegaMek about dialog.  I agree it should also be in the log file but it seems reasonable to me to put it here as well.  If I was a user, this is the first place I would look for this information.

neoancient:
Closing as unnecessary.

Issue #547, opened 2017-11-25
Feature Request: Awards for personnel


hell-fish:
Hi

This may sound pretty minor, but I'm playing a non-mercenary campaign at the moment, and I wish I had the ability to grant the characters in my unit medals and awards. Doesn't have to be terribly complex, just go into the personnel tab menu and do something like right-click to grant an award to a character from a list. Sorta like how you change their rank.

On right click:

Award:

Award 1
Award 2
Award 3
...

And maybe somewhere in the campaign options menu we can chose a pre-made list of awards (Davion, Kurita, etc.) or customize them with images (like portrait pics). Thus, when my unit deploys to Carver V for a 3-month mission, when it's all said and done, I can give them all a Carver V campaign medal, and then other medals for the characters who did important stuff, like held off a Blakist assault in a half-destroyed Locust or rescued the civilians in one of those special missions.

Like I said, minor, but it could go a way to helping me humanize my little puppet soldiers.

BLOODWOLF333:
@Dylan-M what ever happened with that?

pheonixstorm:
He said there were issues that still needed to be worked out iirc.

BLOODWOLF333:
[Awards.zip](https://github.com/MegaMek/mekhq/files/1679192/Awards.zip)

My images for ranks, service stripes, and campaign ribbons.  Along with just the rank portion of my scenario generator spreadsheet since this keeps getting brought up now and then.

bandildo:
I got a working version going on, with some work still need to be done. Here's how it looks:

![awards1](https://user-images.githubusercontent.com/12190053/42338757-ca8301b4-8082-11e8-90d2-b7c3497448a2.png)
![awards2](https://user-images.githubusercontent.com/12190053/42338759-cab227c8-8082-11e8-8ea5-7ccda4887422.png)

So, a couple of questions I'd like to debate:

- Where would be the best place to put the ribbons? I was thinking vertically, on a column to the right side of the person panel view, shown in the attached files. But maybe it will be too cluttered of medals and ribbons?

- I currently have each award to be defined by:

```
	combatCommendation.description=Company Kills 72
	combatCommendation.medal=CombatCommendationM.png
	combatCommendation.ribbon=CombatCommendation.png
	combatCommendation.text=Combat Commendation
	combatCommendation.xp=2
```
[...]

PR #801, opened 2018-07-12
Personnel awards

bandildo:
Implementation of #547 

Original award data kindly provided @BLOODWOLF333 

Maybe an improved selection of the award set should be discussed to include as the default set.

The awards can have 3 different components: ribbon, medal and misc.  **NO MISC HAS BEEN INCLUDED** in the default award set since copyright was mentioned as a possible issue.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -277,6 +279,11 @@
     private int rankSystem = -1;
     private Ranks ranks;
 
+    // AwardNames
+    protected AwardsFactory awardsFactory;
```
neoancient: There's no need to add a field to store an instance of a singleton class.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -277,6 +279,11 @@
...
+    protected ArrayList<Award> awards;
```
neoancient: Use the more generic interface type (List) as the field type rather than a specific implementation.
Is there some reason this needs to be protected rather than private?

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -387,6 +395,8 @@ public Person(String name, Campaign c, String factionCode) {
         bloodname = "";
         primaryDesignator = DESIG_NONE;
         secondaryDesignator = DESIG_NONE;
+        awardsFactory = AwardsFactory.getInstance();
```
neoancient: See note above about storing references to a singleton class in every instance of another class.

MekHQ/src/mekhq/campaign/personnel/AwardsFactory.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public static List<String> getAllSetNames(){
+        return new ArrayList<>(awardsMap.keySet());
+    }
+
+    public static List<Award> getAllAwardsForSet(String setName){
+        return new ArrayList<>(awardsMap.get(setName).values());
+    }
+
+    public static Award GenerateNew(String setName, String awardName, Date date){
```
neoancient: Method names should begin with a lower-case letter.

MekHQ/src/mekhq/gui/utilities/WrapLayout.java
```diff
@@ -0,0 +1,192 @@
+package mekhq.gui.utilities;
+
+import java.awt.*;
+import javax.swing.JScrollPane;
+import javax.swing.SwingUtilities;
+
+/**
```
neoancient: If you copy someone else's work, be sure to attribute it to the original author and provide a reference and all pertinent information about terms of use.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -227,9 +297,100 @@ private void initComponents() {
         gridBagConstraints.fill = GridBagConstraints.BOTH;
         gridBagConstraints.anchor = GridBagConstraints.NORTHWEST;
         add(txtFiller, gridBagConstraints);
+    }
+
+    private void setRibbons() {
+    	int maxRibbonsPerRow = 4;
```
neoancient: This should at least be declared as a constant at the beginning of the class definition to make it easier to change. It would probably be better as a value in CampaignOptions, but I think user configuration of the layout might better be left for a followup.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -227,9 +297,100 @@ private void initComponents() {
...
 
+    	ArrayList<Award> awards = person.getAwards().stream().filter(a -> a.getRibbonFileName() != null)
```
neoancient: There's no need for a specific implementation of the List interface here. I would declare the variable as `List<Award>` and use the `Collectors.toList()` method.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -227,9 +297,100 @@ private void initComponents() {
...
+        for(Box box : rowRibbonsBoxes){
+            boxRibbons.add(box);
+        }
     }
 
+    private void setMedals(){
+
+        ArrayList<Award> awards = person.getAwards().stream().filter(a -> a.getMedalFileName() != null)
+                .collect(Collectors.toCollection(ArrayList::new));
+        Collections.sort(awards);
```
neoancient: You can sort the stream before collecting to the list using `sorted()`

neoancient:
Two other things that I missed as I rushed to finish the review in the time I had earlier:
1. The classes that handle awards are divided between `mekhq.campain` and `mekhq.campaign.personnel`. I would put all of them in the personnel package.

2. All public methods should have javadoc comments detailing their usage, no matter how simple. It is also a good practice for all but the most trivial private/package/protected methods.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
bandildo: By changing this to
```java
List<Award> awards = person.getAwards().stream().filter(a -> a.getMedalFileName() != null).sorted().collect(Collectors.toList());
```

The following error appears:

Incompatible types. Required List<Award> but 'collect' was inferred to R: no instace(s) of type variable(s) T exist so that List<T> conforms to List.

bandildo:
@neoancient thank you for the feedback! I think I got most of it, the only issue was with the comment I replied. It's always good to have someone look at your code.

Let me know if something is still amiss.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
neoancient: This is because Collectors::toList returns a value of type java.util.List which it's trying to assign to a variable that is of type java.awt.List because of the wildcard import of the java.awt package. You could either use the fully qualified name of the class (java.util.List<Award> awards = ....), remove the wildcard import and import the eight classes individually, or go back to what you had before. Developing the habit of using the interface that it makes the code more flexible, but in this case it's used entirely in a local context and there's no particular advantage to the abstraction.

neoancient:
I just looked over the rest of the code changes and everything else looks in order.

Issue #796, opened 2018-07-10
LAM Pilot shows as Support in Personnel Breakdown


ndrwstn:
### Environment
MekHQ 0.44

### Description
Under "Overview" -> "Personnel Breakdown" : "LAM Pilot" shows up under Support Personnel instead of Combat.

PR #803, opened 2018-07-14
Fix for #796: LAM Pilot shows as Support in Personnel Breakdown

neoancient:
Goes beyond a simple fix and pulls the logic for deciding whether a role is combat or support to a central location in Person.


neoancient:
That was broken by PR #726

PR #804, opened 2018-07-14
Two fixes to GM mode finances -> edit transaction

firefly2442:
There were two issues:
* Upon editing a transaction and clicking cancel, it would still add a report to the daily log on the right.
* Upon making an edit, it would show the new entry for both transaction reports in the daily log.


firefly2442:
Change to using Eclipse auto-generated .equals() method.

PR #805, opened 2018-07-14
string identity comparison to .equals()

firefly2442:
Found via [static code analysis tool SpotBugs](https://spotbugs.readthedocs.io/).

> ES: Comparison of String objects using == or != (ES_COMPARING_STRINGS_WITH_EQ)
> 
> This code compares java.lang.String objects for reference equality using the == or != operators. Unless both strings are either constants in a source file, or have been interned using the String.intern() method, the same string value may be represented by two different String objects. Consider using the equals(Object) method instead.

This should have no visible impact to the user.


neoancient:
Some of that code was written by people who should know bettter.

Only one of them was mine.

PR #806, opened 2018-07-15
Refactors part inventories from a string array to a concrete class

sixlettervariables:
- Implements 2016 TODO from Cord Awtry (kipstafoo)
- Hoists out some common formatting of in transit and on order counts

(apologies if some of this is not idiomatic Java, I'm not active in Java development anymore)


Qwertronix:
I like it a lot. This should make it easier to understand what's going on with the code and work with this stuff later for sure. FWIW I didn't spot any traps and style seems to conform at least in the lines directly changed.

PR #807, opened 2018-07-17
Contract profit fix

bandildo:
This fixes the following issues:
- the transportation costs was trying to fetch a planet using its name in using a method that required its ID, returning null and messing up the total profit calculation;
- the method in Contract.java `getLenghtPlusTravel(Campaign c)`, in order to calculate the 2-way duration, was multiplying by 2 after applying Mathf.Ceil(), which would result in an overestimation of the travel duration. Now it is only ceiling the final result;
- campaign options regarding payment of maintenance, payroll and overhead are now taken into account when calculating the estimated profit.

On my way to debug these issues, I found the data presented by the UI a bit unreliable. I took the liberty to re-organize the contract payment summary, and refactored the code. It is working for both the Contract Market and Contract Creation.

contractpaymentsummary


pheonixstorm:
Your MRBC image proves nothing. RAW the fee is taken out of the gross contract amount. If the total payout is 5 million C-bills you subtract the MRBC fee to get your net payout. Once you have the net payout you then divide it up to get the monthly amount.

The only thing that your image proves is that the MRBC fee should not be added to the code that handles the estimated profit.

bandildo:
I'll look into it, for sure. The number X is the number of months, for each of the monthly costs.Any suggestion on how to make it more perceptible? 

I'm currently with very limited internet connection, so it might take a while to address these issues.

NickAragua:
Ah, I see. You're including the travel times in the computation for those costs. We can leave it as is for now, I think, let's just deal with the renegotiate buttons.

bandildo:
Now it looks like this:

image

sixlettervariables:
To make the Nx $###,### more discoverable you could change it slightly:
```
75% of Contract Pay:    4 months @ 3,571,923
Total Overhead (est.):    8 units @ -1,568
Total Maint (est.): 8 units @ -20,772
Total Payroll (est.): 8 personnel @ -31,360
```

bandildo:
@sixlettervariables the units are all months. The contract pay is paid for 4 months, which is the duration of the contract (Contract Lenght). Total overhead, maintenance and payroll are expenses that occur for the duration of the contract + transport time.

I totally agree that the Nx ###,### is not very friendly...

MekHQ/src/mekhq/gui/view/ContractPaymentBreakdown.java
```diff
@@ -589,4 +588,10 @@ private void setLblTransportationExpenses2(){
     private void setLblEstimatedProfit2(){
         lblEstimatedProfit2.setText(formatter.format(contract.getEstimatedTotalProfit(campaign)));
     }
+
+    private String generateMonthlyHeader(int lenght){
```
sixlettervariables: spelling: `length`

PR #811, opened 2018-07-19
Feature: Mass mothball/reactivate

NickAragua:
Implements the ability for the user to mothball and reactivate large numbers of units of different types at a time. All units selected must be either active & available or mothballed for the menu item to show up.

- Tells you which techs are doing maintenance currently
- Tells you approximate time the operations will take
- Schedule as many mothballs as you want at a time
- Reactivating a unit will do its best to put the crew/tech back and put the unit back into the force it came from. 
- Mothball information persists across saves so you can mothball, save, then reactivate

image


pheonixstorm:
Visual looks good.

Issue #834, opened 2018-07-30
Ammunition Reload issues (still?)


Jesterr:
### Environment
mekhq-0.44.0-windows 
Windows 10
C:\Users\rjoiner>java -version
java version "1.8.0_171"
Java(TM) SE Runtime Environment (build 1.8.0_171-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.171-b11, mixed mode)

### Description
Similar to "Ammunition Reload Issues #583" marked closed.  Commented on that thread too. but couldn't "reopen."

Meks with artemis ammo or gauss ammo will infinitely consume ammunition in "repair bay" without ever "fixing" the repair.  Ie, ammunition is consumed, the repair takes time to complete, there is a roll, but the ammunition counter on the vee/mek is not incremented up, so the vee/mek never clears out of the repair queue.   Mass Repair in group mode, Mass repair selected on an individual mek, individually selecting the repair task, and GM Mode -> complete task, all act the same way described above.  (Routine use of mass repair will drain ammo stores in the warehouse _quickly_.)

This week(end) I upgraded my mekhq from 0.42.2 to 0.44.0, and transported my current campaign into the new version when I noted the problems (several fights in when I wondered where my ammo stockpiles went.)  Upgrade consisted of me of unzipping a the new version into its own directory, and copying over camo, custom meks, portraits, and ./campaigns/*
I started the new mekhq, and loaded the last campaign file saved explicitly to import.  had a complaint about an infantry unit.  Deleted that in 0.42.2, re-imported campaign save without offending infantry, game loaded fine.  When scenarios are launched, megamek gives a complaint about the unit, but I've deleted it.  This info is included for full disclosure report.

The problem is not consistent across meks, in that some that utilize artemis or gauss ammo, can still be reloaded, but once a mek/vee is 'stuck' in this mode, there is no clearing it.

 Each mek, as it comes out of a mission, becomes impossible to reload, but only for meks with Artemis or "advanced" gear like a gauss rifle. Standard gear only meks/vees seem to work fine. (Ie, my Von Luck tank, with SRM6 and SRM4s reload fine. My Battlemaster customer with SRM6 and artemis, will not reload the SRM6 ammo. My Highlander HGN-732b won't reload its LRM20 (std not artemis). )

So far, the actions have persisted across restarts of mekhq. 

Screenshots:  (Custom battlemaster has 2 tons of ammo for srm/6, but 3(5) reloads hasn't worked):

[...]

PR #839, opened 2018-08-02
Better logging on equipment mismatch

sixlettervariables:
The first two commits partially address #834. It won't deduct your ammo at least, but your tech will lose the time. A log warning will be printed as well. When a unit is loaded that is likely to cause this issue, it will have a warning printed with the equipment numbers and parts that are suspect.


sixlettervariables:
@NickAragua check out the first two commits, the whitespace is only in the third commit.

Issue #475, opened 2017-09-28
Sort Personnel by Name in Choose Spouse (Mate) list


MattDroz:
Currently, the list to choose a spouse/mate is coming up unsorted, which makes it harder to choose a match when trying to play matchmaker between two characters.

pheonixstorm:
Sorting by name doesn't really help all that much. I think sorting by age would be better. That is how I would want it. Someone else may want it another way completely.

HammerGS:
Age and Name.

PR #843, opened 2018-08-04
Sorts Choose Spouse (Mate) and Special Ability lists

bandildo:
Implements #475 
Also added scroll pane to award list.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
@@ -1769,8 +1769,8 @@ public class PersonnelTableMouseAdapter extends MouseInputAdapter implements
                             menu.add(spouseMenu);
                         }
                     }
-                    if (menu.getItemCount() > 30) {
-                        MenuScroller.setScrollerFor(menu, 20);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -1814,6 +1820,10 @@ private void maybeShowPopup(MouseEvent e) {
                         menuItem.addActionListener(this);
 
                         setAwardMenu.add(menuItem);
+
+                        if (setAwardMenu.getItemCount() > 30) {
```
sixlettervariables: Should this be smaller? We've had requests by folks with small vertical resolutions elsewhere.

bandildo: I'll change it to the 20 items like the spouse one, assuming that value is ok.

sixlettervariables: I'll double check on my 13" MacBook today.

sixlettervariables: You still left the `if` test at 30.

bandildo:
Sorry, brainfart here. Hopefully all is okay now.

PR #870, opened 2018-08-16
Fix for #858 plus code cleanup

giorgiga:
I'm sending this because I think it solves #858 (ie: links work for me, but I really didn't see the issue in the first place).

On top of that, you'll see A LOT more changes. They are quite simple though (they are mostly automatic edits with eclipse code cleanup tool) so  I guess it's not a big deal if they create conflicts?
If conflicts are an issue (I have no idea what the project etiquette says about it), let me know and I'll look up how to cherry pick changes in pull requests.

PR #872, opened 2018-08-16
Fix #858

### LINKED ISSUE #858: Allow HTML formatted news in the historical daily log
The recent [addition of the Historical Daily Log](https://github.com/MegaMek/mekhq/pull/792) [strips out the HTML](https://github.com/MegaMek/mekhq/pull/792/commits/e385a39e81c9204a8734a8adf47b9db107dc6950#diff-7147823e21e3a3306a81d504c995465cR3182).  However, this removes potentially important information such as news events and other items.  Try to add back in the ability to have clickable links.

giorgiga:
Escaped XML was not correctly unescaped in
LogEntry.generateInstanceFromXML(); this removes the need for the
workaround of stripping HTML tags in Campaign

### DIFF
--- a/MekHQ/src/mekhq/campaign/LogEntry.java
+++ b/MekHQ/src/mekhq/campaign/LogEntry.java
@@ -39,9 +39,10 @@ import mekhq.MekHqXmlUtil;
  */
 public class LogEntry implements Cloneable, MekHqXmlSerializable {
 
-    private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss"); //$NON-NLS-1$
-    // LATER SimpleDateFormat is not thread-safe: use throw-away instances or switch to java.time
-    // private static final DateTimeFormatter DTF = DateTimeFormatter.ofPattern("", Locale.ENGLISH);
+    private static final SimpleDateFormat dateFormat() {
+        // LATER centralise date formatting so that every class doesn't have its own format and - possibly - switch to java.time
+        return new SimpleDateFormat("yyyy-MM-dd hh:mm:ss"); //$NON-NLS-1$
+    }
 
     public LogEntry(Date date, String desc) {
         this(date, desc, null);
@@ -85,7 +86,7 @@ public class LogEntry implements Cloneable, MekHqXmlSerializable {
     public void writeToXml(PrintWriter pw, int indent) {
         StringBuilder sb = new StringBuilder();
         sb.append(MekHqXmlUtil.indentStr(indent)).append("<logEntry>"); //$NON-NLS-1$
-        if (date != null)    sb.append("<date>").append(DATE_FORMAT.format(date)).append("</date>"); //$NON-NLS-1$ //$NON-NLS-2$
+        if (date != null)    sb.append("<date>").append(dateFormat().format(date)).append("</date>"); //$NON-NLS-1$ //$NON-NLS-2$
         assert desc != null; sb.append("<desc>").append(MekHqXmlUtil.escape(desc)).append("</desc>"); //$NON-NLS-1$ //$NON-NLS-2$
         if (type != null)    sb.append("<type>").append(MekHqXmlUtil.escape(type)).append("</type>");  //$NON-NLS-1$//$NON-NLS-2$
         sb.append("</logEntry>"); //$NON-NLS-1$
@@ -109,7 +110,7 @@ public class LogEntry implements Cloneable, MekHqXmlSerializable {
                 } else if (nname.equals("type")) { //$NON-NLS-1$
                     type = MekHqXmlUtil.unEscape(node.getTextContent());
                 } else if (nname.equals("date")) { //$NON-NLS-1$
-                    date = DATE_FORMAT.parse(node.getTextContent().trim());
+                    date = dateFormat().parse(node.getTextContent().trim());
                 }
             }
         } catch (Exception ex) {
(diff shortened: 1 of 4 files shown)


giorgiga:
Note: this is a follow-up to PR https://github.com/MegaMek/mekhq/pull/870 (closed)

MekHQ/src/mekhq/campaign/LogEntry.java
```diff
@@ -35,116 +35,124 @@
...
-        return sb.toString();
-    }
-    
-    @Override
-    public LogEntry clone() {
-        return new LogEntry(getDate(), getDesc(), getType());
-    }
-}
\ No newline at end of file
+public class LogEntry implements Cloneable, MekHqXmlSerializable {
```
sixlettervariables: Switch to 4 spaces instead of tabs.

MekHQ/unittests/mekhq/campaign/LogEntryTest.java
```diff
@@ -0,0 +1,45 @@
+package mekhq.campaign;
```
sixlettervariables: 4 spaces instead of tabs.

MekHQ/src/mekhq/campaign/LogEntry.java
```diff
@@ -35,116 +35,124 @@
...
+		Date   date = null;
+		String desc = null;
+		String type = null;
+
+		try {
+			NodeList nl = wn.getChildNodes();
+			for (int x = 0; x < nl.getLength(); x++) {
+				Node   node = nl.item(x);
+				String nname = node.getNodeName();
+				if (nname.equalsIgnoreCase("desc")) { //$NON-NLS-1$
```
sixlettervariables: Minor nit: XML element names are case sensitive, so the `equalsIgnoreCase` could be just an `equals`. (I realize the rest of MekHQ probably uses `equalsIgnoreCase`)

MekHQ/src/mekhq/campaign/LogEntry.java
```diff
@@ -35,116 +35,124 @@
...
+	@Override
+	public LogEntry clone() {
+		return new LogEntry(date, desc, type);
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((date == null) ? 0 : date.hashCode());
```
sixlettervariables: Consider `Objects.hashCode(object)` instead of the ternary (but only consider, do not consider this a normative statement).

MekHQ/unittests/mekhq/campaign/LogEntryTest.java
```diff
@@ -0,0 +1,45 @@
...
+
```
sixlettervariables: Mixed tabs and spaces, missing a copyright header.

MekHQ/src/mekhq/campaign/LogEntry.java
```diff
@@ -35,116 +35,124 @@
...
-    @Override
-    public LogEntry clone() {
-        return new LogEntry(getDate(), getDesc(), getType());
-    }
-}
\ No newline at end of file
+public class LogEntry implements Cloneable, MekHqXmlSerializable {
+
+	private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss"); //$NON-NLS-1$
+	// LATER SimpleDateFormat is not thread-safe: use throw-away instances or switch to java.time
```
sixlettervariables: You could go ahead and implement this now, transitioning `DATE_FORMAT` to a `String` and building the `SimpleDateFormat` on the fly below.

MekHQ/src/mekhq/campaign/LogEntry.java
giorgiga: Good catch!

I've seen a mix of `equals(s)`, `equalsIgnoreCase(s)` and `contains(s)` around the codebase, but didn't think of double checking the very class I was modifying :-)

MekHQ/unittests/mekhq/campaign/LogEntryTest.java
giorgiga: Good point, but – now that I look at it – most files have missing or outdated license headers (besides the © year, some are for GPL 2 while the project is GPL 3).

Do you know if there are plans to add license-header management to the build system? (maybe to the new one @neoancient ?)

I'm by no means a license-savvy man, so I'll just add a GPL3 header for https://github.com/licenses/license-templates/blob/master/templates/gpl3-header.txt

giorgiga:
@sixlettervariables all concerns should be addressed in the new changes - please make sure to double-check the GPL header!

sixlettervariables:
@giorgiga you should use the copyright header from https://github.com/MegaMek/megamek/wiki/MegaMek-Coding-Style-Guide

giorgiga:
Thanks @sixlettervariables – I will change the header as directed.

Please be aware that the [license included in MekHQ](https://github.com/MegaMek/mekhq/blob/master/MekHQ/license.txt) is GPL v3 while the license header references v2 of the GPL (may or may not be an issue - IDK, IANAL)

sixlettervariables:
I misunderstood your reply earlier (I read it in the opposite direction!), you are correct. The following is the appropriate header, and I'll update the Wiki for MHQ:
```java
/*
 * Copyright (c) 2018 The MegaMek Team. All rights reserved.
 *
 * This file is part of MekHQ.
 *
 * MekHQ is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * MekHQ is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with MekHQ.  If not, see <http://www.gnu.org/licenses/>.
 */
```

giorgiga:
No problem @sixlettervariables – it only takes a sec to update a file and push it :-)

PR #882, opened 2018-08-19
Feature: Adds gzipped Campaign save file support

sixlettervariables:
This PR adds support for gzipped campaign save files.

1. Campaign XML parsing is moved to `CampaignXmlParser`.
2. Campaigns are loaded via a factory which detects the format of the file.
3. Saving the campaign checks to see if it was loaded from a `.cpnx.gz` and will try and honor that.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -231,6 +231,9 @@ public class Campaign implements Serializable, ITechManager {
     private int astechPoolOvertime;
     private int medicPool;
 
+    // indicates whether or not the campaign should be gzipped, if possible.
+    private Boolean preferGzippedOutput;
+
     // I need to put a basic game object in campaign so that I can
     // assign it to the entities, otherwise some entity methods may get NPE
     // if they try to call up game options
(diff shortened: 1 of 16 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -253,11 +230,8 @@
     private int astechPoolOvertime;
     private int medicPool;
 
-    private int lastTeamId;
-    private int lastPartId;
-    private int lastForceId;
-    private int lastMissionId;
-    private int lastScenarioId;
+    // indicates whether or not the campaign should be gzipped, if possible.
+    private Boolean preferGzippedOutput;
```
neoancient: This is not initialized until you load a campaign file. If you start a new campaign it has the value of null and throws an NPE when you attempt to save it. I don't see a reason to use an Object wrapper here and would use a simple primitive `boolean` instead.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -738,19 +814,22 @@ public News getNews() {
         return news;
     }
 
+    private int getLastForceId() {
+        return forceIds.size() > 0 ? forceIds.lastKey() : 0;
```
neoancient: Does this really offer an advantage? Instead of tracking the last index and incrementing each time something is added, you're requiring a recursive search of the map each time to find the last entry (best case log2 n, worst case n). Tracking the next legal value and incrementing it each time is a fairly common practice. The same goes for parts, missions, and scenarios.

sixlettervariables: This code should not have come over as part of this work.

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: My editor appears to prefer `Boolean` over `boolean`, don't know why. Fixed.

sixlettervariables:
I'm going to hold on this as it won't actually accept a CPNX saved with a BOM.

sixlettervariables:
Ok, my update will accept BOM XML...by virtue of only using inference for GZIP detection. I've noticed hand edited CPNX files may have it added which broke the (naive) XML inference in place.

sixlettervariables:
Ok @NickAragua @neoancient, I've rebased this against `master`. All of my test campaigns still read/write successfully as both `.cpnx` and `.cpnx.gz`.

PR #890, opened 2018-08-21
Fixed issue with date format of LogEntry

giorgiga:


sixlettervariables:
This appears to not have been noticed, as the time portion is always `12:00` on serialization.

giorgiga:
XML files currently contain `... 12:00:00`, which is read as midnight.

With this change, _all_ `... 12:00:00` will be read as midday instead.

`2750-01-01 12:00:00` is read as:

- Sun 01 Jan 2750 00:00:00 with `yyyy-MM-dd hh:mm:ss`
- Sun 01 Jan 2750 12:00:00 with `yyyy-MM-dd HH:mm:ss`

Will close this PR until it is decided if times can be altogether ignored.

PR #892, opened 2018-08-22
convert some system.out.println()s to standard logging format

firefly2442:
Should be no visible change from the point of view of the user.


firefly2442:
Ah right, keep forgetting.  How's this?

PR #893, opened 2018-08-22
File dialogs update

giorgiga:
This changes all `JFileChooser`s to `FileDialog`s, which are much more similar to the native platform look&feel.

On MacOS, this changes the dialogs from these:

load

save

to these:

load-new
save-new


giorgiga:
I _think_ I checked all the dialogs and to me they _seem_ to behave reasonably (I'm not familiar with all the imports/exports)

Issue #836, opened 2018-07-31
Replace Java-based Load Campaign dialog with OS native dialog


ndrwstn:
### Environment
MekHQ 0.44.0
OS X 10.13.5
Java 8 update 171

### Description
File -> Load Campaign (and the matching start screen dialog) uses a Java-based open dialog; it would be preferable to use OS Native dialogs.

PR #906, opened 2018-08-29
Feature: File dialogs update + remember last used folder

giorgiga:
_If you think you've already seen this but something doesn't add up, it might be because this same changes had already been submitted as PR #893, which I then closed after screwing it up._

----

This PR changes, by default, all `JFileChooser`s to `FileDialog`s, which are much more similar to the native platform look&feel.

To continue using `JFileChooser`s,  users may add this to `mmconf/mekhq.properties`:

```
# Determines which open/save file dialogs MekHQ uses.
#
# Supported values:
#  awt:   use awt FileDialog (default)
#  swing: use swing JFileChooser
#
gui.file.dialogs=swing
```

Additionally, this PR makes so that MekHQ remembers what directory was last used (whether for opening or saving a file) and opens subsequent dialogs pointing to that directory.

----

On MacOS, the "old" swing `JFileChooser`s look like:

load

save

The "new" awt `FileDialog`s look like:

load-new
save-new


giorgiga:
Note: the "native dialogs" part of these changes have been requested as an enhancement with #836

PR #922, opened 2018-09-08
Adopt Log4j2

giorgiga:
_This is part of a set of coordinated PRs - please see https://github.com/MegaMek/megamek/pull/1126_

### DIFF
--- a/MekHQ/build.gradle
+++ b/MekHQ/build.gradle
@@ -37,7 +37,7 @@ dependencies {
         exclude group: 'org.python', module: 'jython'
         exclude group: 'org.mozilla', module: 'rhino'
     }
-    
+
     // We have been using a modified version of svgSalamander in MML, which we'll keep around until
     // the conversion to batik is complete.
     implementation files('lib/svgSalamander.jar')
@@ -48,7 +48,6 @@ dependencies {
     implementation 'org.apache.commons:commons-csv:1.4'
     implementation 'javax.vecmath:vecmath:1.5.2'
 
-    implementation 'log4j:log4j:1.2.17'
     implementation 'javax.xml.bind:jaxb-api:2.3.0'
     runtimeOnly 'org.glassfish.jaxb:jaxb-runtime:2.3.0'
     runtimeOnly 'org.glassfish.jaxb:jaxb-core:2.3.0'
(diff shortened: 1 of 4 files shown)


MekHQ/build.gradle
```diff
@@ -554,7 +553,7 @@ task javadocJar(type: Jar) {
 publishing {
     publications {
         publishMMLibrary(MavenPublication) {
-            artifactId = "mekhq${mmBranchTag}"
+            artifactId = "mekhq${hqBranchTag.replaceAll('[^A-Za-z0-9_\\-.]','-')}"
```
giorgiga: Some characters (notably `/`) are allowed for branch names but not for maven artifact names (I changed this in all three projects - no need to check).

MekHQ/build.gradle
```diff
@@ -48,15 +48,14 @@ dependencies {
     implementation 'org.apache.commons:commons-csv:1.4'
     implementation 'javax.vecmath:vecmath:1.5.2'
 
-    implementation 'log4j:log4j:1.2.17'
```
giorgiga: It's best to piggy-back on the logging libraries from megamek (fetched via transitive dependencies) - this way we won't have versioning conflicts.

MekHQ/src/mekhq/gui/MekLabTab.java
```diff
@@ -601,10 +601,10 @@ void setTechFaction(int techFaction) {
         private static final long serialVersionUID = 4348862352101110686L;
         
         private SmallCraft entity;
-        private megameklab.com.ui.Dropship.tabs.DropshipStructureTab structureTab;
+        private megameklab.com.ui.aerospace.DropshipStructureTab structureTab;
         private megameklab.com.ui.Aero.tabs.EquipmentTab equipmentTab;
         private megameklab.com.ui.Aero.tabs.BuildTab buildTab;
-        private megameklab.com.ui.Dropship.tabs.TransportTab transportTab;
+        private megameklab.com.ui.aerospace.TransportTab transportTab;
```
giorgiga: `DropshipStructureTab` and `DropshipStructureTab` have been moved in MML master: this is so that this PR compiles against MML master.

PR #923, opened 2018-09-09
Tech spa

AaronGullickson:
Some simple SPAs for techs that allow them to specialize (a -1 bonus) when working with particular part types. I also wanted to add a non-weapon equipment specialization but that is not specifically tracked with the repair part types currently.


MekHQ/docs/history.txt
```diff
@@ -4,6 +4,8 @@ v0.45.2-SNAPSHOT
 + Switched build system from ant to gradle.
 + Issue #917: Hire Full Compliment Regularly Hires Bondsmen
 + Issue #919: Dropship Crewmembers/techs displayed as "Nobody" In Repair Bay
++ added tech specialist SPAs
```
sixlettervariables: We'll modify history.txt after the change lands, still you can revert this change.

AaronGullickson:
Ahh, I was wondering how you were handling the history log with the pull request system. I have removed that entry.

PR #927, opened 2018-09-11
Map Colors for Socio-Industrial Level

AaronGullickson:
This feature adds in alternate coloration of the map based on socio-industrial level. This is selectable by the user from the map options panel. Here is an example using tech level:

screen shot 2018-09-11 at 11 42 34 am

Of course, a lot of systems are grey because of lack of information, but thats more incentive to get those values imputed. Here is a rough idea of what it would look like with no missing values by randomly imputing values:

screen shot 2018-09-11 at 11 41 25 am

Of course, the values make no sense because they are random. In actuality there should be a lot more dark color at the core and lighter colors at the periphery. 

The shading colors here come from the great [ColorBrewer website](http://colorbrewer2.org). All shades should be color-blind safe.  

All of the magic happens in `InterstellarMapPanel.getPlanetColors`, and it is extendable to other kinds of planetary data.

### DIFF
--- a/MekHQ/src/mekhq/gui/InterstellarMapPanel.java
+++ b/MekHQ/src/mekhq/gui/InterstellarMapPanel.java
@@ -80,6 +80,7 @@ import mekhq.campaign.JumpPath;
 import mekhq.campaign.universe.Faction;
 import mekhq.campaign.universe.Faction.Tag;
 import mekhq.campaign.universe.Planet;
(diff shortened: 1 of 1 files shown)


AaronGullickson:
yes, I thought about a key but maybe best as a later feature enhancement. Part of the reason I want this now is I am thinking about methods to impute SIC levels where we dont have them and wanted an easy way to visualize it.

PR #932, opened 2018-09-16
Scenario template editor

NickAragua:
Here we go, the first part of the AtB upgrade. 

A pretty powerful scenario template editor. I'll be writing some documentation up shortly. Basically, allows you to specify an arbitrary combination of player/attached/allied bot/enemy bot forces and parameters for them, as well as some map parameters. Accessible via Manage Campaign menu -> Scenario Template Editor.

Also hooked it up to the campaign so you can write up your own scenario templates and use them to generate scenarios. To do that, you'll want to be in an AtBContract, preferably on-world (but probably not required). Click 'Add Scenario', then 'Generate from Template' at the bottom. Assign a lance (or however many player aligned / player supplied forces you specified in the template). Then edit scenario -> finalize. It'll generate the opfor and update map size. Now you can play. Whee!

Looking for feedback on (among other things) configuration combinations that seem nonsensical.


neoancient:
From a little bit of playing around:
1. I set up a scenario where the enemy force size is based on my own and I hit finalize before I had deployed my force, giving me an empty bot force, which remained. Perhaps finalize should clear old forces if I decide I want to change the force I deploy.
2. Selecting an opposing force based on the number of units works, but I tried doing one on BV first at 1.5x then at 2x and both times it generated a force with 1/3 of my BV.
3. It's often helpful to suggest a file extension in the save dialog.

NickAragua:
I've addressed the BV generation issue. Finalize clears out old forces as requested. We now suggest a file name as well. Also fixed a raft of other bugs and introduced a "force generation ordering" mechanism, so that the order in which the forces are generated is explicit, so that forces generated per each "generation order" bucket are dependent on the forces generated from previous buckets.

NickAragua:
Fixed one more bug, and added an 'edit' feature for existing forces within a template.

A screenshot of what this thing currently looks like:
https://imgur.com/SvTfazm

Issue #865, opened 2018-08-12
Changing Award Date Does Not Update Award Tooltip


BLOODWOLF333:
### Environment
MekHQ 0.45.0
Windows 8.1
Java 1.8

### Description
Changed the bronze star award date to Mar and observed that the award tooltip does not update.  (might update with a HQ restart but I would request that it does that after changing the award date)
![image](https://user-images.githubusercontent.com/16950832/44004176-abb0eeb8-9e23-11e8-9bde-1bf66743246d.png)

### Files
[BloodwolfCampaign.zip](https://github.com/MegaMek/mekhq/files/2281119/BloodwolfCampaign.zip)

bandildo:
Did you change the award date via the Edit Personnel Log? I could "fix" this, but a bit reluctantly, since giving the Log control over the awards seems a bit non-programmer-friendly: the log should be only the reflex of the awards, not the opposite. But I do understand that right now this is the simplest solution.

MattDroz:
I noticed it to. The way I got around it was changing the Campaign Date in Options, then awarding the personnel to reflect the date. But, as BloodWolf put it, this can be time-consuming when you just want to give them awards (or rank insignias too, based on when they ranked up).

PR #934, opened 2018-09-16
Log entry refactoring

bandildo:
This PR aims to cleanup and refactor the personal log entries.

It divides the log entries into 5 different types:
```
public enum LogEntryType {
    MEDICAL,
    AWARD,
    SERVICE,
    PERSONAL,
    CUSTOM,
    HISTORICAL, //Used for historical daily log
}
```

It creates specific wrapper classes that inherit from `LogEntry`. All log entries should be made from them, since `LogEntry` constructors have now been made protected. This intends to avoid logging entries without a type.

A lot of string resources have been created. All log entry related strings are now centralized in a single resource file. The access to these resources has also been centralized in the various Logger classes created for each of the types.

Backward compatibility was taken into account. Any log entries without a type will be parsed through various regex in order to determine their type. The type is saved in the campaign, as it was done with medical, allowing for this backward compatibility process to run only once.

A small refactoring had to be made in order to allow changing the dates of the awards via the personal log editor. However instead of doing something half-assed that would serve the purpose of fixing the award dates, I've decided to make a full log entry refactoring.


MekHQ/src/mekhq/campaign/log/LogEntryController.java
```diff
@@ -0,0 +1,231 @@
+/*
```
sixlettervariables: Whitespace needs normalization.

bandildo:
Fixed merging issues.

bandildo:
Took the liberty to include the fix to #865 in this PR. The whole refactoring was motivated by fixing this issue, which is highly coupled with the refactoring.

MekHQ/src/mekhq/campaign/log/AwardLogEntry.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class AwardLogEntry extends LogEntry {
+
+    public AwardLogEntry(Date date, String desc){
+        super(date, desc, LogEntryType.AWARD);
+    }
+
+    @Override
+    public void onLogEntryEdited(Date originalDate, Date newDate, String originalDesc, String newDesc, @Nullable Person person) {
+        if(person == null) return;
```
neoancient: Why is null treated as valid value here? Passing a null does nothing, and I can't find an execution path that would get here with a null Person. While I'm in favor of defensive programming, an unexpected state should not be silently ignored. At the very minimum I would expect a log entry.

MekHQ/src/mekhq/campaign/log/LogEntryController.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This class is responsible of some general control of log entries.
+ * @author Miguel Azevedo
+ */
+public class LogEntryController {
+
+    private static LogEntryController logEntryController = null;
```
neoancient: See comments on AwardLogger class.

MekHQ/src/mekhq/campaign/log/ServiceLogger.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.text.MessageFormat;
+import java.util.Date;
+import java.util.ResourceBundle;
+
+/**
+ * This class is responsible to control the logging of Service Log Entries.
+ * @author Miguel Azevedo
+ */
+public class ServiceLogger {
```
neoancient: See comments on LogEntryController class as singleton.

MekHQ/src/mekhq/campaign/log/AwardLogger.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Date;
+import java.util.ResourceBundle;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This class is responsible to control the logging of Award Log Entries.
+ * @author Miguel Azevedo
+ */
+public class AwardLogger {
```
neoancient: This is effectively a stateless class, and I see no reason to create an instance. Every method can be made static. Resource bundles are caches, so the lookup carries very little overhead and I don't see much advantage to saving that as a class variable, but if you think that's a concern it could be made a static field initialized lazily.

bandildo:
@neoancient thank you for the feedback.

I've made all the Logger methods static as requested.

I cannot recall why I had the @Nullable and why I was defending against it. It seems too much an effort to be doing it blindly, so I believe something might have happened. I've removed that, so we will soon find out if we can get a null person in that spot.

PR #935, opened 2018-09-16
Reduced declared checked exceptions in MekHqXmlUtil

giorgiga:
This simplifies the `MekHqXmlUtil` API so that it doesn't declare useless checked exceptions
(SAX already does that enough).

As an opportunity change, this also removes a static `SimpleDateFormat`
from AwardsFactory and normalises whitespace in `MekHqXmlUtil` from tabs to spaces.

**Note to reviewers**
There are two commits in this PR: you might want to only look at the first one to check out the actual code (the second commit changes whitespace, and so basically every line is modified)

**EDIT:**
I added a third commit to this PR, which removes some unnecessary complexity in how entities were parsed out of embedded mul-formatted tags and, in theme with the PR, also removes the need for a few other declared checked exceptions.


NickAragua:
Doesn't crash when loading campaign files, so that's pretty good.

PR #937, opened 2018-09-17
Add findbugs to build.gradle and start addressing its warnings

giorgiga:
This PR adds findbugs to the MekHQ build (it's executed during `check` and doesn't fail the build no matter what) and also starts addresses most of the high-priority warnings produced by findbugs (the easier/most obvious ones).

For information on Findbugs, see:
https://docs.gradle.org/current/userguide/findbugs_plugin.html
http://findbugs.sourceforge.net/

**Note to reviewers**
This PR is probably best reviewed one commit at a time.
Also, the notes below can help speeding up the process for the first commits (where the same changes have been applied over and over).

### Commit 1. Add findbugs

This only adds findbugs to the gradle build.

### Commit 2. Renamed a bunch of clone() methods to copy()

The contract of `Object.clone()` is very specific and... the whole cloning mechanism is practically deprecated (not really deprecated, but it's in a state similar to `Vector`, `Hashtable` or the serialization machinarium: basically it's all old stuff that still works and will continue to do so, but that is considered surpassed nowadays).

Many classes (most notably, `Part` and its subclasses) implemented `clone()` in a way that is semantically correct, but different from what the `Object.clone()` contract prescribes (basically "implement `Cloneable`" and "obtain clone via `super.clone()`", ...).

Since it seemed to make little sense to fix the implementations to make them compliant with a surpassed specification, all the `clone()` methods have been renamed to `copy()`.

While I was at it, I also `grep -nr 'clone()' src/` and reviewed all the places where `clone()` was mentioned - basically, it's now only used for `Date`, `GregorianCalendar` and once for `Path2D`.

### Commit 3. Removed Serializable from a bunch of classes

Findbugs was (correctly) complaining that some classes (most notable: `Campaign` and `Person`) contained non-transient non-serializable fields. Since we are not using serialization anyways, I removed `Serializable` from the list of implemented interfaces in those classes.

### Commit 4. Findbugs "Correctness Warnings" addressed

- `Illegal format string "%s%\_ExportedPersonnel.prsx" in mekhq.gui.FileDialogs.savePersonnel(JFrame, Campaign)`
- `java.util.UUID is incompatible with expected argument type megamek.common.Entity in mekhq.campaign.market.PersonnelMarket.removePersonnelForDay(Campaign)`
- `Integral value cast to double and then passed to Math.ceil in mekhq.campaign.parts.equipment.EquipmentPart.resolveVariableCost(boolean)`
- `Integral value cast to double and then passed to Math.ceil in mekhq.campaign.parts.MissingQuadVeeGear.getTonnage()`
- `Field LanceAssignmentTableModel.columnNames masks field in superclass mekhq.gui.model.DataTableModel`
- `Field RequiredLancesTableModel.columnNames masks field in superclass mekhq.gui.model.DataTableModel`
- `Non-virtual method call in new mekhq.campaign.parts.PodSpace() passes null for non-null parameter of new PodSpace(int, Unit)`
- `Possible null pointer dereference of en in mekhq.campaign.unit.UnitTechProgression.calcTechProgression(MechSummary, int)`
- `Possible null pointer dereference of null in mekhq.gui.CampaignGUI.loadPersonFile()`
- `Possible null pointer dereference of en in mekhq.gui.model.PersonnelTableModel$VisualRenderer.getTableCellRendererComponent(JTable, Object, boolean, boolean, int, int)`
- `Nullcheck of luminosity at line 449 of value previously dereferenced in mekhq.campaign.universe.StarUtil.getSpectralType(Integer, Double, String)`
- `Repeated conditional test in mekhq.campaign.personnel.Person.isCombat()`
- `Return value of String.replaceAll(String, String) ignored in mekhq.campaign.mission.Loot.writeToXml(PrintWriter, int)`
- `Invocation of toString on String.split(String) in mekhq.campaign.universe.StarUtil.<static initializer for StarUtil>()`
- `Invocation of toString on Utilities.romanNumerals in mekhq.Utilities.getArabicNumberFromRomanNumerals(String)`

### Commit 5. Findbugs "Dodgy code Warnings" addressed

- `Dead store to gbc in mekhq.gui.dialog.UnitMarketDialog.initComponents()`
- `Result of integer multiplication cast to long in mekhq.campaign.unit.Unit.getSellValue()`
- `Potentially dangerous use of non-short-circuit logic in mekhq.campaign.personnel.Person.getEligibleSPAs(boolean)`
- `Potentially dangerous use of non-short-circuit logic in mekhq.campaign.ResolveScenarioTracker.loadUnitsAndPilots(File)`
- `Potentially dangerous use of non-short-circuit logic in mekhq.gui.dialog.EditSpecialAbilityDialog.editSkillPrereq(int)`
- `Potentially dangerous use of non-short-circuit logic in mekhq.gui.dialog.EditSpecialAbilityDialog$6.actionPerformed(ActionEvent)`
- `Potentially dangerous use of non-short-circuit logic in mekhq.gui.utilities.JSuggestField$5.keyReleased(KeyEvent)`
- `Return value of SkillType.getName() ignored, but method has no side effect`
- `Return value of getTotalCombatUnits() ignored, but method has no side effect`
- `Double assignment of field CampaignOptions.usePlanetaryConditions in new mekhq.campaign.CampaignOptions()`

This one is reported by findbugs, but seems to actually be ok:
- `Test for floating point equality in mekhq.campaign.universe.RegionPerimeter.isInsideRegion(double, double, List)` (line 115)

I don't know how this one should be fixed:
- `Test for floating point equality in mekhq.gui.MekLabTab.refreshRefitSummary()` (line 350)

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/PersonnelMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/PersonnelMarket.java
@@ -124,7 +124,7 @@ public class PersonnelMarket {
 		}
 	}
 
-    /*
+    /**
      * Remove personnel from market on a new day
      * The better they are, the faster they disappear
      */
@@ -133,9 +133,7 @@ public class PersonnelMarket {
             List<Person> toRemove = method.removePersonnelForDay(c, personnel);
             if (null != toRemove) {
                 for (Person p : toRemove) {
-                	if (attachedEntities.contains(p.getId())) {
-                		attachedEntities.remove(p.getId());
-                	}
+                    attachedEntities.remove(p.getId());
                 }
                 personnel.removeAll(toRemove);
             }
--- a/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
+++ b/MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
@@ -247,9 +247,6 @@ public class CampaignOpsReputation extends AbstractUnitRating {
         int totalBattleArmor = 0;
         int forceTotal;
 
-        // Count total units for transport
-        getTotalCombatUnits();
-
         List<Unit> unitList = getCampaign().getCopyOfUnits();
         for (Unit u : unitList) {
             if (u == null) {
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1173,20 +1173,20 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
                 }
 
                 // fuel tanks
-                partsValue += 200 * js.getFuel() / js.getFuelPerTon();
+                partsValue += 200L * js.getFuel() / js.getFuelPerTon();
 
                 // armor
                 partsValue += js.getArmorWeight(js.locations()) * EquipmentType.getArmorCost(js.getArmorType(0));
 
                 // heat sinks
-                int sinkCost = 2000 + 4000 * js.getHeatType();// == HEAT_DOUBLE ? 6000:
+                long sinkCost = 2000 + 4000l * js.getHeatType();// == HEAT_DOUBLE ? 6000:
                                                            // 2000;
                 partsValue += sinkCost * js.getHeatSinks();
 
                 // grav deck
-                partsValue += 5000000 * js.getGravDeck();
-                partsValue += 10000000 * js.getGravDeckLarge();
-                partsValue += 40000000 * js.getGravDeckHuge();
+                partsValue += 5000000l  * js.getGravDeck();
+                partsValue += 10000000l * js.getGravDeckLarge();
+                partsValue += 40000000l * js.getGravDeckHuge();
 
                 // get bays
                 int baydoors = 0;
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1179,14 +1179,14 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
                 partsValue += js.getArmorWeight(js.locations()) * EquipmentType.getArmorCost(js.getArmorType(0));
 
                 // heat sinks
-                long sinkCost = 2000 + 4000l * js.getHeatType();// == HEAT_DOUBLE ? 6000:
+                long sinkCost = 2000L + 4000L * js.getHeatType();// == HEAT_DOUBLE ? 6000:
                                                            // 2000;
                 partsValue += sinkCost * js.getHeatSinks();
 
                 // grav deck
-                partsValue += 5000000l  * js.getGravDeck();
-                partsValue += 10000000l * js.getGravDeckLarge();
-                partsValue += 40000000l * js.getGravDeckHuge();
+                partsValue += 5000000L  * js.getGravDeck();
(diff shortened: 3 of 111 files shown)


MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
```diff
@@ -247,9 +247,6 @@ private int getTotalForceUnits() {
         int totalBattleArmor = 0;
         int forceTotal;
 
-        // Count total units for transport
-        getTotalCombatUnits();
```
sixlettervariables: Do we know if each of the methods called by `getTotalCombatUnits` is side-effect free?

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1173,20 +1173,20 @@ public long getSellValue() {
...
                 // fuel tanks
-                partsValue += 200 * js.getFuel() / js.getFuelPerTon();
+                partsValue += 200L * js.getFuel() / js.getFuelPerTon();
 
                 // armor
                 partsValue += js.getArmorWeight(js.locations()) * EquipmentType.getArmorCost(js.getArmorType(0));
 
                 // heat sinks
-                int sinkCost = 2000 + 4000 * js.getHeatType();// == HEAT_DOUBLE ? 6000:
+                long sinkCost = 2000 + 4000l * js.getHeatType();// == HEAT_DOUBLE ? 6000:
```
sixlettervariables: prefer `4000L` over `l` (and might as well add an L to 2000).

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1173,20 +1173,20 @@ public long getSellValue() {
...
                                                            // 2000;
                 partsValue += sinkCost * js.getHeatSinks();
 
                 // grav deck
-                partsValue += 5000000 * js.getGravDeck();
-                partsValue += 10000000 * js.getGravDeckLarge();
-                partsValue += 40000000 * js.getGravDeckHuge();
+                partsValue += 5000000l  * js.getGravDeck();
```
sixlettervariables: `L`

MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
giorgiga: They happen to be (just checked), but... nice catch!
I had deleted the call out of blind trust in findbugs and didn't think of checking all the methods called in `getTotalCombatUnits()` (which turn out to be simple getters, implemented only in `AbstractUnitRating`)

MekHQ/src/mekhq/campaign/market/PersonnelMarket.java
```diff
@@ -133,9 +133,7 @@ public void removePersonnelForDay(Campaign c) {
             List<Person> toRemove = method.removePersonnelForDay(c, personnel);
             if (null != toRemove) {
                 for (Person p : toRemove) {
-                	if (attachedEntities.contains(p.getId())) {
```
giorgiga: `Hashmap.contains()` is the same as `Map.containsValue()`.
Since here it's called with a key, the condition is always false.
(this was reported by findbugs)

Calling `remove()` with a non-contained key is a nop, so I just removed the `if` altogether.

PR #930, opened 2018-09-14
New icon/portrait rendering code

giorgiga:
This PR started as a way to push some new code to render unit icons preserving their color and ended up with some extra unrelated change too (see code comments).

Four different ways to blend unit icons with camo/background are implemented; in the image below, the first row is the current MekHQ strategy (take the blue channel and use it to darken the camo) and the third row is the one that is active in this PR (it preserves all pixels that are not perfectly grayscale, and use luminance in lieu of the blue channel in to darken the camo.

In case you are interested, the screenshot is from a window opened by running the main method in `megamek.util.ImagesTest`and the four strategies are implemented in `megamek.util.Images`.

camos

PR #948, opened 2018-09-25
Person review

giorgiga:
This is the first of what will possibly become a  series of PR concerning the review of `Person` and related classes.

What this PR achieves is:
1. Information on Manei Domini classes removed from `Person` and made into an enumeration
2. Information on Manei Domini ranks removed from `Rank` and made into an enumeration
3. Added a `DomProcessor` class to aid in parsing DOM trees and used it `Person.generateInstanceFromXML`

**Note to reviewers** the `DomProcessor` included here is a successor of the one included with #930


giorgiga:
Bug squashed.

The test I've done on this is:

1. checkout master, load some old campaign and save it (to have the newest `.cpnx` format)
2. checkout the PR, load the campaign and save it under a different name
3. diff the two files and verify changes: the only expected change (in <person>) is that the manei domini tags disappear when unset (there is also some difference in skill ordering - that's because they are in an hash table)

Issue #856, opened 2018-08-07
Unable to Edit Mission during AtB campagin (0.45.0)


servo01:
Pressing the Edit Mission Tab from the Briefing Room screen produced no response. 

I went into the Campaign Settings and unchecked the "use against the bot campaign rules" which fixes the issue, but there seems to be no other way to tweak mission settings or leave notes with the current mission during an AtB campaign.

HammerGS:
We would need to have a CPNX file.

servo01:
This is a fresh install of 0.45 with no edits or modifications to the default campaign setting. 
I turned on the AtB rules, generated a mission from the marketplace and accepted it. 

Unable to edit the current mission unless I turn the AtB rules to off. 

[TestGame.cpnx.zip](https://github.com/MegaMek/mekhq/files/2312546/TestGame.cpnx.zip)

NickAragua:
Looks like an issue with the number of "contract required lances".

PR #950, opened 2018-09-26
Fix for issue #856 - added bulletproofing to customize atb contract dialog

NickAragua:
This seems to have been caused by errors in the "required lances" calculation logic. That has since been fixed, but older save games may have bad data, so we introduce logic into the Customize AtB Contract dialog to make it more fault-tolerant.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CustomizeAtBContractDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CustomizeAtBContractDialog.java
@@ -188,8 +188,10 @@ public class CustomizeAtBContractDialog extends JDialog {
     	btnEnemyCamo = new JButton();
         JLabel lblEnemyRating = new JLabel();
     	JLabel lblRequiredLances = new JLabel();
-    	spnRequiredLances = new JSpinner(new SpinnerNumberModel(contract.getRequiredLances(), 1,
-    			null, 1));
+    	
+    	int requiredLances = contract.getRequiredLances() > 0 ? contract.getRequiredLances() : 1; 
+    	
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/CustomizeAtBContractDialog.java
```diff
@@ -188,8 +188,10 @@ private void initComponents() {
     	btnEnemyCamo = new JButton();
         JLabel lblEnemyRating = new JLabel();
     	JLabel lblRequiredLances = new JLabel();
-    	spnRequiredLances = new JSpinner(new SpinnerNumberModel(contract.getRequiredLances(), 1,
-    			null, 1));
+    	
+    	int requiredLances = contract.getRequiredLances() > 0 ? contract.getRequiredLances() : 1;
```
giorgiga: You may want to call `mekhq.campaign.mission.AtBContract.calculatePaymentMultiplier(Campaign)` instead of hardcoding a minimum of 1 lance.

For reference, this is the commit that solved the issue for the contract market dialog (which does call `calculatePaymentMultiplier()` in its `initComponents()`)  https://github.com/MegaMek/mekhq/commit/3337b64006efd2a71f79d53d75cbe8395c3a8e9b

NickAragua: I see your point, but I'm not sure that I want to re-calculate all the contract details in this situation. For better or worse, the contract was signed based on a particular state of the player force, and it may not be a pleasant experience to suddenly have the contract details (such as payment) change suddenly after it is signed.

giorgiga: Good point.

PR #952, opened 2018-09-27
Copypasted code when generating bloodnames

giorgiga:
This consolidates some copypasted code into a new `Bloodname` method.
As a bonus, it also fixed a minor UI issue.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
@@ -238,6 +238,7 @@ public class CustomizePersonDialog extends javax.swing.JDialog implements Dialog
             textBloodname.setText(person.getBloodname());
             panDemog.add(textBloodname, gridBagConstraints);
 
+            btnRandomBloodname.setEnabled(campaign.getFaction().isClan());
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
```diff
@@ -238,6 +238,7 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
             textBloodname.setText(person.getBloodname());
             panDemog.add(textBloodname, gridBagConstraints);
 
+            btnRandomBloodname.setEnabled(campaign.getFaction().isClan());
```
NickAragua: I'm not sure this is entirely correct. The ability to give a blood name should probably depend on whether the person is designated as a clanner, as well as the current campaign faction.

giorgiga: TL;DR: I disabled the button because it currently doesn't work (ie: throws an exception) unless your faction is a clan (wrote this in the commit comment - I should have put this somewhere more prominent, sorry)

My thought was, I should always be able to generate a bloodname for a clanner, but the generator needs the name name of a clan and we don't have an "origin faction" for people (we just have the "clanner" flag). I tried feeding the generator the generic "CLAN" faction, but that doesn't work either so I just disabled the button.

PR #961, opened 2018-10-07
Do not use spare parts during GM restore

sixlettervariables:
- Adds a bool to Part::fix indicating if infinite resources are available for the fix
- Fixes bug where Armor work mode was sticky after a fix (was not restored to Normal)


sixlettervariables:
I was never satisfied with this fix.

PR #969, opened 2018-10-18
System refactor

AaronGullickson:
This is the beginning of a refactoring of the way we pull in and use planetary system data based on the creation of full non-canon planetary system data (including social data) that we are currently building [here](https://github.com/AaronGullickson/system_generation).

The new format will wrap all planetary information inside of a `<system>` node. System-level data (e.g. stellar type, recharge stations) will be kept within this node, while all planetary information will be kept in children `planet` nodes. The system data will identify a system position id for the primary planet and this information will be used to pull planetary information quickly to the system level. 

To do this, I have created a new `PlanetarySystem` object that now contains some of the information formerly contained within the `Planet` object as well as a list of `Planet` objects that are contained within the system. I have also created a `Systems` object that will ultimately replace the `Planets` object for reading in the planetary data. 

This is still very much a WIP but we are now at the stage with the system generation project, that we can begin to think about how to incorporate the new data back into MHQ. Ultimately, I would like to include the ability to generate a planetary system map when you double-click a system on the interstellar map. However, that will likely be a subsequent PR once this one is fully integrated and functioning. Another future RFE that would be nice would be an InterstellarMapEditor that would open separately from the main MHQ that would make future corrections and changes to the interstellar data possible through a GUI interface.


AaronGullickson:
Yeah, this is not yet ready for deployment. Just putting it up here as a placeholder ATM. The goal is to give other devs and contributors a branch to work off of to look at the newly generated systems.

AaronGullickson:
I have removed all conflicts with the master branch and will continue working on this. We need to finish random name generation and knock off a few other details and then it should be good to go.

AaronGullickson:
Okay that attempt to address conflicts was a total disaster apparently, so I have rewound everything and I will try it again.

PR #971, opened 2018-10-22
Allows grouping personnel by their unit on the Personnel tab

sixlettervariables:
This makes the personnel tab easier to review with large numbers of vehicles or infantry.

screen shot 2018-10-21 at 6 17 03 pm

screen shot 2018-10-21 at 6 19 45 pm

Not accomplished:
- Context menu does not apply any actions to the grouped persons.
- You cannot expand an individual (all or nothing).


sixlettervariables:
I can work on that.

sixlettervariables:
Updated:

image

PR #997, opened 2018-12-14
Weekly Support Edge Refresh

mkerensky:
Refreshes the edge points available to support personnel each week. Combat personnel already get their edge points back at the start of each MM battle in which they are deployed. Available support edge points are displayed on the PersonViewPanel.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2877,6 +2877,13 @@ public class Campaign implements Serializable, ITechManager {
                     u.resetPilotAndEntity();
                 }
             }
+            
+            // Reset edge points to the purchased value each week. This should only
+            // apply for support personnel - combat troops reset with each new mm game
+            if ((p.isAdmin() || p.isDoctor() || p.isEngineer() || p.isTech())
+                    && calendar.get(Calendar.DAY_OF_WEEK) == Calendar.MONDAY) {
+                p.resetCurrentEdge();
+            }
 
             if (getCampaignOptions().getIdleXP() > 0 && calendar.get(Calendar.DAY_OF_MONTH) == 1 && p.isActive()
                     && !p.isPrisoner()) { // Prisoners no
@@ -2892,7 +2899,6 @@ public class Campaign implements Serializable, ITechManager {
                     p.setIdleMonths(0);
                 }
             }
-
         }
 
         for (Person baby : babies) {
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -1890,8 +1890,8 @@ public class Campaign implements Serializable, ITechManager {
         //If we get a natural 2 that isn't an automatic success, reroll if Edge is available and in use.
         if (getCampaignOptions().useSupportEdge()
                 && (doctor.getOptions().booleanOption(PersonnelOptions.EDGE_MEDICAL))) {
-            if (roll == 2  && doctor.getEdge() > 0 && target.getValue() != TargetRoll.AUTOMATIC_SUCCESS) {
-                doctor.setEdge(doctor.getEdge() - 1);
+            if (roll == 2  && doctor.getCurrentEdge() > 0 && target.getValue() != TargetRoll.AUTOMATIC_SUCCESS) {
+                doctor.setCurrentEdge(doctor.getCurrentEdge() - 1);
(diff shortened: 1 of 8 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1443,6 +1446,13 @@ public void writeToXml(PrintWriter pw1, int indent) {
                         + "<edge>"
                         + String.valueOf(getOptionList("::", PilotOptions.EDGE_ADVANTAGES))
                         + "</edge>");
+            // For support personnel, write an available edge value
+            if (isSupport() || isEngineer()) {
```
sixlettervariables: Should this check for isAdmin etc too, like above?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -2877,6 +2877,13 @@ public void processNewDayPersonnel() {
                     u.resetPilotAndEntity();
                 }
             }
+            
+            // Reset edge points to the purchased value each week. This should only
+            // apply for support personnel - combat troops reset with each new mm game
+            if ((p.isAdmin() || p.isDoctor() || p.isEngineer() || p.isTech())
```
sixlettervariables: In Person you're checking for isSupport, but not here. Should those two places be consistent?

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1660,6 +1670,8 @@ public static Person generateInstanceFromXML(Node wn, Campaign c, Version versio
                     advantages = wn2.getTextContent();
                 } else if (wn2.getNodeName().equalsIgnoreCase("edge")) {
                     edge = wn2.getTextContent();
+                } else if (wn2.getNodeName().equalsIgnoreCase("edgeAvailable")) {
+                    retVal.currentEdge = Integer.parseInt(wn2.getTextContent());
```
sixlettervariables: Does `retVal.currentEdge` need to start as -1 and get set to edge at the end if it stays -1? I'm thinking of the case where you load a campaign and the people have not been updated yet.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -725,6 +729,30 @@ private void fillStats() {
             gridBagConstraints.fill = GridBagConstraints.NONE;
             gridBagConstraints.anchor = GridBagConstraints.NORTHWEST;
             pnlStats.add(lblEdge2, gridBagConstraints);
+            
+            if (person.isAdmin() || person.isDoctor() || person.isTech()) {
```
sixlettervariables: Again, should this be consistent with the other checks? (Support, engineer, etc)

MekHQ/src/mekhq/campaign/personnel/Person.java
mkerensky: Doesn't need to. isSupport() covers astechs and medtechs not covered by isTech() and isDoctor() respectively.  Neither can use edge currently, but I didn't mind them having the value.

MekHQ/src/mekhq/campaign/Campaign.java
mkerensky: See above about the differences between isSupport and isAdmin/Doctor/Tech.  In part I was experimenting with the different ways to handle this, and in the other part, I didn't want to reset edge for Astechs and Medtechs - I have a lot of Astech/Vehicle Crews and Soldier/Medtechs in my campaigns that I wanted to handle as combat personnel edgewise.    If we need consistency or if I get some compelling reasons from people who play differently I can change it.

MekHQ/src/mekhq/campaign/personnel/Person.java
mkerensky: I'm not quite sure I understand the case you're describing.  I want to be able to save and load campaigns where currentEdge has been spent for the week, and that seems to work as I planned in testing.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
mkerensky: This is very intentional. I don't want the person view for Astechs and Medtechs to show available edge that they can't use.  The Ship's Engineer is a composite crewman that doesn't actually have a person view -  individual Vessel Crewmembers are what shows up here, and they're technically combat personnel that shouldn't show the new value either, in my mind.

PR #1007, opened 2018-12-22
Cleans up and (potentially) fixes AtB initialization

sixlettervariables:
This consolidates the multiple ways in which AtB is initialized in `Campaign`. It also (potentially) fixes a bug where the `AtBEventProcessor` is not hooked up for brand new AtB campaigns.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -3459,17 +3459,21 @@ public class Campaign implements Serializable, ITechManager {
         shoppingList.restore();
 
         if (getCampaignOptions().getUseAtB()) {
+            RandomNameGenerator.initialize();
             RandomFactionGenerator.getInstance().startup(this);
-            while (!RandomUnitGenerator.getInstance().isInitialized()) {
-                //Sleep for up to one second.
+
+            int loops = 0;
+            while (!RandomUnitGenerator.getInstance().isInitialized()
+                || !RandomNameGenerator.getInstance().isInitialized()) {
                 try {
                     Thread.sleep(50);
+                    if (++loops > 20) {
+                        // Wait for up to a second
+                        break;
+                    }
                 } catch (InterruptedException ignore) {
-
                 }
             }
-            RandomNameGenerator.getInstance();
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7906,79 +7910,86 @@ public void initTimeInService() {
...
-        Date founding = null;
-        for (Person p : getPersonnel()) {
-            for (LogEntry e : p.getPersonnelLog()) {
-                if (null == founding || e.getDate().before(founding)) {
-                    founding = e.getDate();
+        if (!newCampaign) {
+            /*
+            * Switch all contracts to AtBContract's
+            */
+            for (Map.Entry<Integer, Mission> me : missions.entrySet()) {
```
NickAragua: I'm not sure that this approach is correct, as a user may have a mixture of AtB and non-AtB missions.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7906,79 +7910,86 @@ public void initTimeInService() {
...
-                                /*
-                                 * We're only guessing anyway, so we use this hack to avoid actually loading the
-                                 * entity to check for IS2
-                                 */
-                                p.setOriginalUnitTech(1);
-                            }
-                            if (null != p.getUnitId() && null != units.get(p.getUnitId())
-                                    && ms.getName().equals(units.get(p.getUnitId()).getEntity().getShortNameRaw())) {
-                                p.setOriginalUnitId(p.getUnitId());
+                    if (e.getDesc().startsWith("Joined ")) {
```
NickAragua: I know there's not really a way to address this, but :(

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: That was the original setup for when a user switched to AtB. I can revise that so it doesn't if you think that's best.

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: Definitely not for old campaigns. I'll noodle on it some.

PR #1015, opened 2018-12-27
Bug: Can't export personnel

### LINKED ISSUE #985: Cant export personnel at all
What version of MekHQ does your issue apply to?  45.1 and 45.2 (tried both)
What operating system are you using?  Windows, Mac, or Linux?  Windows 10
What java version are you using?  If unsure it will state your java version at the top of the mekhqlog.txt file.    Java version 1.8.0_161

Right clicking any personnel to export them results in nothing happening.  No dialog or anything.  I have tried this on 45.1 and 45.2.  To reproduce this issue (at least on my end): 

Start a new campaign (or load one)
Go to personnel tab
Select some people and right click them.
Choose Export Personnel and poof nothing happens.

Log attached. 
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/2609712/mekhqlog.txt)

sixlettervariables:
Corrects string format statement in export personnel, fixes #985.

### DIFF
--- a/MekHQ/src/mekhq/gui/FileDialogs.java
+++ b/MekHQ/src/mekhq/gui/FileDialogs.java
@@ -56,7 +56,7 @@ public class FileDialogs {
      */
     public static Optional<File> savePersonnel(JFrame frame, Campaign campaign) {
 
-        String fileName = String.format( "%s%_ExportedPersonnel.prsx", //$NON-NLS-1$
+        String fileName = String.format( "%s%s_ExportedPersonnel.prsx", //$NON-NLS-1$
                                          campaign.getName(),
                                          campaign.getShortDateAsString() );
 
@@ -196,5 +196,4 @@ public class FileDialogs {
                                    new File(MekHQ.CAMPAIGN_DIRECTORY, fileName),
                                    FileType.CPNX );
     }
-
 }


Felderburg:
I'm not sure if this is the appropriate place to say this, but I think I'm supposed to comment on previous bugs instead of starting a one, even if they're closed. I am having this issue, same as #985 and this forum post: https://forum.megamek.org/thread-2608.html If I try to export personnel to an XML, I get a blank error popup, and no XML file. This happens regardless of whether it's my campaign or one of the pre-loaded campaigns. This happens regardless of where I try to save the file.

I am using 46.1. I'm not sure if my mekhqlog is applicable, but it's attached (apparently it's supposed to tell me what version of java I have, but I don't see that).
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/5468334/mekhqlog.txt)

HammerGS:
This should be fixed in the current Dev cycle releases as indicated in the notes for #985. Try it in 47.11 Dev.

Felderburg:
I'm new to github, so all I saw was that 985 was marked as closed, and I don't see anything relating to version numbers in either issue, just the dates that things are done (2018 for both, which led me to believe it should've been fixed in the version I have). Am I missing something?

HammerGS:
When you see merged it indicates we've pushed code to fix the issues. When you check the bug report of #985 you'll see we fixed it using this commit.

image

Felderburg:
Ok, thanks for the reply. I still don't see anything in this thing that indicates it would be going into 47.11, rather than 46.1, but I assume I just need to learn github.

HammerGS:
Left you a comment on Slack

PR #1037, opened 2019-01-06
Make startup GUI show all four buttons at lower res

sixlettervariables:
This "fixes" something I saw somebody note, which I never really noticed myself on my MacBook.

image

If the splash screen image will be less than 840 (buttons plus gutters) it switches to "small" buttons and text.

### DIFF
--- a/MekHQ/src/mekhq/gui/StartUpGUI.java
+++ b/MekHQ/src/mekhq/gui/StartUpGUI.java
@@ -56,49 +56,6 @@ public class StartUpGUI extends javax.swing.JPanel {
         
         ResourceBundle resourceMap = ResourceBundle.getBundle("mekhq.resources.StartUpDialog", new EncodeControl()); //$NON-NLS-1$
         
-        btnNewGame = new javax.swing.JButton(resourceMap.getString("btnNewGame.text"));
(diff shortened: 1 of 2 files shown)


PhoenyxCullen:
Is there a way to get HTML/CSS -like stylings in Java. I know how to do this in those. Base the buttons on window size and have the window resizable. Just a thought.

sixlettervariables:
Not that I'm aware of in Swing.

MekHQ/src/mekhq/gui/StartUpGUI.java
```diff
@@ -56,19 +56,41 @@ private void initComponents() {
...
-        btnNewGame.setMaximumSize(new Dimension(200, 25));
+        // initialize splash image
+        double maxWidth = app.calculateMaxScreenWidth();
+        imgSplash = getToolkit().getImage(app.getIconPackage().getStartupScreenImage((int) maxWidth));
+
+        // wait for splash image to load completely
+        MediaTracker tracker = new MediaTracker(frame);
+        tracker.addImage(imgSplash, 0);
+        try {
+            tracker.waitForID(0);
```
NickAragua: "Multi-threaded" programming at its finest.

Issue #1036, opened 2019-01-06
Allow 0.0% intensity in AtB Settings


rogueranger1993:
I like to use MekHQ to track my IRL tabletop games, and the AtB markets for units and contracts are very useful. Rather than ask for those two markets to be made available outside of AtB mode, I thought about a second option to give a similar result - allowing the AtB intensity setting to be set to zero. This would allow me to use the AtB markets, but would eliminate the ATB-generated scenarios, thereby allowing me to generate contract scenarios for my games in any way I please with minimal hassle, and it seems like it wouldn't need too much effort to implement such a change on the part of the Devs, either.

sixlettervariables:
Implemented in #1038.

PR #1038, opened 2019-01-06
Enable zero intensity level for AtB Scenario Generation

sixlettervariables:
This implements enhancement request #1036 where 0% can be chosen for AtB scenario intensity levels.

image

This also removes a deployment nag if you have no possible scenarios to deploy into (this isn't perfect, but handles the most likely case).

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2675,8 +2675,20 @@ public class Campaign implements Serializable, ITechManager {
     public int getDeploymentDeficit(AtBContract contract) {
         int total = -contract.getRequiredLances();
         int role = -Math.max(1, contract.getRequiredLances() / 2);
+        boolean hasScenarios = contract.getScenarios().size() > 0;
+
         for (Lance l : lances.values()) {
-            if (l.getMissionId() == contract.getId() && l.getRole() != Lance.ROLE_UNASSIGNED) {
+            if (hasScenarios) {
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -2675,8 +2675,20 @@ public void readNews() {
...
         for (Lance l : lances.values()) {
-            if (l.getMissionId() == contract.getId() && l.getRole() != Lance.ROLE_UNASSIGNED) {
+            if (hasScenarios) {
+                if (l.getMissionId() == contract.getId() && l.getRole() != Lance.ROLE_UNASSIGNED) {
+                    total++;
+                    if (l.getRole() == contract.getRequiredLanceType()) {
+                        role++;
+                    }
+                }
+            } else if (l.getMissionId() == Lance.NO_MISSION && l.getRole() != Lance.ROLE_UNASSIGNED) {
```
NickAragua: I'm not sure I understand the purpose of this addition. getMissionID checks the *contract* to which a lance is assigned, rather than the scenario. A lance not assigned to a contract shouldn't be getting counted for this purpose.

sixlettervariables: Apparently you are only assigned to a contract when you get assigned to a scenario. I can look into cleaning that up.

sixlettervariables:
Ok this fixes the deployment nag by having it report 0 deficits if a contract:
1. Is inactive.
2. Has yet to begin.

PR #1045, opened 2019-01-16
Crewed Unit Options Handling

mkerensky:
3 changes in this one

1) PilotOptions handling - where Edge values, SPAs, edge triggers and the like used to be set for all units from the Commander's value, now infantry, battle armor, vehicles, tripod mechs and large spacecraft will only gain an ability if > 50% of the crewmembers have it. This is RAW for several cyber implants.

2) Adds the ability to set new aerospace edge triggers for MM to relevant personnel

3) Adds an errata fix for the reload time of Capital missile ammo. This is now 1 hour per missile instead of several days per the basic formula. 

Thanks to @sixlettervariables for all the help with collecting and sorting PilotOptions for multiple crewmembers.


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -2942,11 +2942,16 @@ public void resetPilotAndEntity() {
             //This double enumeration is annoying to work with for crew-served units.
             //Get the option names while we enumerate so they can be used later
             List<String> optionNames = new ArrayList<String>();
+            List<String> cyberOptionNames = new ArrayList<String>();
```
sixlettervariables: Make this `Set<String> cyberOptionNames = new HashSet<String>();`. Running `List<T>::contains` is a much more expensive operation than `HashSet<T>::contains`.

mkerensky: Fixed.

PR #1052, opened 2019-01-27
Static analysis cleanup

VicenteCartas:
Cleaning most of the errors reported by lgtm (https://lgtm.com/projects/g/MegaMek/mekhq/alerts/?mode=list&severity=error)

Some seem real errors or possible bugs. The ones that are about type checking for subclasses could be argued if they make the code more or less clear.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -62,6 +62,7 @@ import java.util.stream.Collectors;
 import javax.swing.JTable;
 import javax.swing.table.TableModel;
 
+import mekhq.campaign.parts.equipment.*;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVPrinter;
 import org.joda.time.DateTime;
@@ -99,10 +100,6 @@ import megamek.common.util.EncodeControl;
 import mekhq.campaign.Campaign;
 import mekhq.campaign.CampaignOptions;
 import mekhq.campaign.parts.Part;
-import mekhq.campaign.parts.equipment.AmmoBin;
-import mekhq.campaign.parts.equipment.BattleArmorEquipmentPart;
-import mekhq.campaign.parts.equipment.EquipmentPart;
-import mekhq.campaign.parts.equipment.MissingEquipmentPart;
 import mekhq.campaign.personnel.Person;
 import mekhq.campaign.personnel.SkillType;
 import mekhq.campaign.unit.CrewType;
@@ -1154,7 +1151,7 @@ public class Utilities {
                     for(int equipNum : equipNums) {
                         i++;
                         Mounted m = unit.getEntity().getEquipment(equipNum);
-                        if (!allMunitions
+                        if (!allMunitions && (part instanceof MissingAmmoBinAmmoBin)
                                 && (!(m.getType() instanceof AmmoType)
                                         || (((AmmoType) epart.getType()).getMunitionType()
                                                 != ((AmmoType) m.getType()).getMunitionType()))) {
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -1154,7 +1154,7 @@ public class Utilities {
                     for(int equipNum : equipNums) {
                         i++;
                         Mounted m = unit.getEntity().getEquipment(equipNum);
-                        if (!allMunitions && (part instanceof AmmoBin)
+                        if (!allMunitions
                                 && (!(m.getType() instanceof AmmoType)
                                         || (((AmmoType) epart.getType()).getMunitionType()
                                                 != ((AmmoType) m.getType()).getMunitionType()))) {
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -1094,7 +1094,6 @@ public class Utilities {
             equipNums.add(unit.getEntity().getEquipmentNum(m));
         }
         List<Part> remaining = new ArrayList<>();
-        List<Part> notFound = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


MekHQ/src/mekhq/Utilities.java
```diff
@@ -1154,7 +1154,7 @@ public static void unscrambleEquipmentNumbers(Unit unit) {
                     for(int equipNum : equipNums) {
                         i++;
                         Mounted m = unit.getEntity().getEquipment(equipNum);
-                        if (!allMunitions && (part instanceof AmmoBin)
```
VicenteCartas: On this branch, we are checking for MissingParts, so it can't be an AmmoBin.

sixlettervariables: Should be MissingAmmoBin.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -6767,8 +6755,7 @@ public PartInventory getPartInventory(Part part) {
         inventory.setOrdered(nOrdered);
 
         String countModifier = "";
-        if (part instanceof Armor || part instanceof ProtomekArmor
-                || part instanceof BaArmor) {
+        if (part instanceof Armor) {
```
sixlettervariables: Can you add a comment that says this includes protomek and BA?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -7636,7 +7623,7 @@ public String getCombatPersonnelDetails() {
                 countTotal);
         sb.append(buffer);
 
-        for (int i = 0; i <= Person.T_NUM; i++) {
+        for (int i = 0; i < Person.T_NUM; i++) {
```
sixlettervariables: Can you add a comment in Person.java that explains that T_NUM should be increased if you add a role? A comment exists which says where to put new roles, but doesn't tie it into T_NUM.

MekHQ/src/mekhq/campaign/market/PartsStore.java
```diff
@@ -143,7 +143,7 @@ public void stock(Campaign c) {
 			p.setBrandNew(true);
 			sb.setLength(0);
 			sb.append(p.getName());
-			if(!(p instanceof Armor || p instanceof BaArmor || p instanceof ProtomekArmor)) {
+			if(!(p instanceof Armor)) {
```
sixlettervariables: Comment indicating this covers all three

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -883,7 +883,7 @@ public void calculate() {
                 time = 40320;
             } else if (newEntity instanceof megamek.common.Dropship || newEntity instanceof megamek.common.Jumpship) {
                 time = 13440;
-            } else if (newEntity instanceof Mech || newEntity instanceof megamek.common.Aero || newEntity instanceof megamek.common.ConvFighter || newEntity instanceof megamek.common.SmallCraft) {
+            } else if (newEntity instanceof Mech || newEntity instanceof megamek.common.Aero) {
```
sixlettervariables: Comment indicating it covers those two b/c Aero does.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1902,18 +1896,14 @@ public void initializeParts(boolean addParts) {
                 locations[((MissingBattleArmorSuit)part).getTrooper()] = part;
             } else if(part instanceof Armor) {
```
sixlettervariables: Comment indicating it covers protomek and BA.

MekHQ/src/mekhq/gui/WarehouseTab.java
```diff
@@ -485,7 +485,7 @@ public boolean include(Entry<? extends PartsTableModel, ? extends Integer> entry
                 if (nGroup == SG_ALL) {
                     inGroup = true;
                 } else if (nGroup == SG_ARMOR) {
-                    inGroup = (part instanceof Armor || part instanceof ProtomekArmor || part instanceof BaArmor);
+                    inGroup = (part instanceof Armor);
```
sixlettervariables: Same comment as above.

MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
```diff
@@ -385,7 +385,7 @@ public boolean include(Entry<? extends PartsTableModel, ? extends Integer> entry
         		if(nGroup == SG_ALL) {
         			return true;
         		} else if(nGroup == SG_ARMOR) {
-        			return part instanceof Armor || part instanceof ProtomekArmor || part instanceof BaArmor;
+        			return part instanceof Armor;
```
sixlettervariables: Same.

MekHQ/src/mekhq/campaign/parts/Refit.java
NickAragua: Honestly, I'd use the Entity.ETYPE_ flags here rather than type comparisons. Unless I'm operating under the false assumption that a type comparison is slower than an int comparison, in which case disregard.

sixlettervariables: I believe instanceof is quite cheap.

MekHQ/src/mekhq/Utilities.java
VicenteCartas: @sixlettervariables allMunitions is always false when we get to that point of the code. That looks like a bug no? We set it to false in line 1119 and then of that for loop we never change the value again. Any idea what's the correct logic? (I can't figure out).

VicenteCartas:
Addressed all PR comments from @sixlettervariables

PR #1054, opened 2019-01-27
Don't let peacetime costs overflow

sixlettervariables:
We were stuffing long's into int's. This fixes that (found via: https://lgtm.com/projects/g/MegaMek/mekhq/snapshot/82b2780e05bf73dec210033fc277da6a6e2cecf9/files/MekHQ/src/mekhq/campaign/Campaign.java?sort=name&dir=ASC&mode=heatmap#x110f0bc0b025a7ab:1)


sixlettervariables:
Also fixes some widening issues that may have resulted in overflows (albeit unlikely).

PR #1056, opened 2019-01-30
Styling files

VicenteCartas:
Adding editorconfig and Intellij default settings following MegaMek Code Style Guide.

Read more about using this on Intellij here: https://www.jetbrains.com/help/idea/configuring-code-style.html

### DIFF
--- a/Intellij-Java-Default.xml
+++ b/Intellij-Java-Default.xml
@@ -1,0 +1,11 @@
+<code_scheme name="Default" version="173">
+  <codeStyleSettings language="JAVA">
+    <option name="KEEP_LINE_BREAKS" value="false" />
+    <option name="KEEP_FIRST_COLUMN_COMMENT" value="false" />
+    <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
+    <option name="IF_BRACE_FORCE" value="3" />
+    <option name="DOWHILE_BRACE_FORCE" value="3" />
+    <option name="WHILE_BRACE_FORCE" value="3" />
+    <option name="FOR_BRACE_FORCE" value="3" />
+  </codeStyleSettings>
+</code_scheme>
--- a/MekHQ/.editorconfig
+++ b/MekHQ/.editorconfig
@@ -1,0 +1,9 @@
+# EditorConfig: https://EditorConfig.org
+
+[*]
+end_of_line = crlf
+
+[*.java]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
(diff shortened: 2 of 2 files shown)


Intellij-Java-Default.xml
```diff
@@ -0,0 +1,11 @@
+<code_scheme name="Default" version="173">
+  <codeStyleSettings language="JAVA">
+    <option name="KEEP_LINE_BREAKS" value="false" />
+    <option name="KEEP_FIRST_COLUMN_COMMENT" value="false" />
+    <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
+    <option name="IF_BRACE_FORCE" value="3" />
```
VicenteCartas: These settings indicate force braces even if the expression is a single line.

MekHQ/.editorconfig
```diff
@@ -0,0 +1,9 @@
+# EditorConfig: https://EditorConfig.org
+
+[*]
+end_of_line = crlf
```
sixlettervariables: Three things:
1. I don't know that the project uses `crlf`. I know at least a few us use non-Windows. git should be controlling this for us anyways with core.crlf auto mode. I'd wager these files are actually `lf`.
2. Add `charset = utf-8`
3. Add `insert_final_newline = true`

MekHQ/.editorconfig
```diff
@@ -0,0 +1,9 @@
...
+
+[*.java]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
sixlettervariables: I suggest adding:
1. `continuation_indent_size = 8`

Also:
```editorconfig
[*.xml]
indent_style = space
indent_size = 4
trim_trailing_whitespace = true
```

MekHQ/.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+insert_final_newline = true
+continuation_indent_size = 8
+
+[*.xml]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
VicenteCartas: Addressing PR comments.

MekHQ/.editorconfig
```diff
@@ -0,0 +1,16 @@
+# EditorConfig: https://EditorConfig.org
+
+[*]
+charset = utf-8
+
+[*.java]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
```
sixlettervariables: `insert_final_newline` and `trim_trailing_whitespace` should ideally be in `[*]` rather than in `[*.java]` and `[*.xml]`.

VicenteCartas: Good point, addressed!

PR #1059, opened 2019-01-31
Lgtm cleaning

VicenteCartas:
Cleans several errors and warnings reported from LGTM:

- Not closed files
- Unreachable casts
- Narrowing conversions (double to long/int)
- int math overflow

Still a lot of warnings pending, trying to break this into smaller PRs.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioTemplate.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioTemplate.java
@@ -19,6 +19,7 @@ import javax.xml.namespace.QName;
 import javax.xml.transform.Source;
 import javax.xml.transform.stream.StreamSource;
 
+import org.apache.commons.io.IOUtils;
 import org.w3c.dom.Node;
 
 import mekhq.MekHQ;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1113,7 +1113,7 @@ else if (heatSinkType == MiscType.F_HEAT_SINK)
     }
 
     public long getSellValue() {
-        long partsValue = 0;
+        double partsValue = 0;
```
sixlettervariables: I wish Java had a native decimal type we could use for currency. I'm not 100% sure `BigDecimal` is the right type either, perhaps Joda Money's `BigMoney` type? We've had a few bugs reported regarding costs being "off", which mostly seem like rounding issues.

sixlettervariables: I think I'd rather the cost related work be broken out into a separate PR so we can perhaps explore better options (if any exist).

MekHQ/src/mekhq/campaign/mission/ScenarioTemplate.java
```diff
@@ -110,15 +111,19 @@ public void Serialize(PrintWriter pw) {
...
             Unmarshaller um = context.createUnmarshaller();
-            Source inputSource = MekHqXmlUtil.createSafeXmlSource(new FileInputStream(inputFile));
+            file = new FileInputStream(inputFile);
+            Source inputSource = MekHqXmlUtil.createSafeXmlSource(file);
             JAXBElement<ScenarioTemplate> templateElement = um.unmarshal(inputSource, ScenarioTemplate.class);
             resultingTemplate = templateElement.getValue();
         } catch(Exception e) {
             MekHQ.getLogger().error(ScenarioTemplate.class, "Deserialize", "Error Deserializing Scenario Template", e);
+        } finally {
+            IOUtils.closeQuietly(file);
```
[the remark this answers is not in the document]
VicenteCartas: Nice! I searched for best practices with streams and I found a website recommending the IOUtils thing. I'll change to the try-with.

MekHQ/src/mekhq/campaign/unit/Unit.java
VicenteCartas: I think this first pass is a good approach. Now we use double all the way in the code and only truncate at the end. The main issue right now is that we are using things like are doubles (like weights) with costs in long/int and some methods are casting to long in every single operation, introducing a lot of rounding issues.

With this at least the method signature is the same (long return), but internally we don't accumulate any rounding errors.

HammerGS:
This pull request **fixes 45 alerts** when merging 3d9a0c8e2b9753a944eea38050e0b7929c81e477 into 789cc5f416852bb4cd1b1b0689fc4610c5d14c38 - [view on LGTM.com](https://lgtm.com/projects/g/MegaMek/mekhq/rev/pr-fb55647e10403b78bea850d0d9876e300e78044d)

**fixed alerts:**

* 28 for Implicit narrowing conversion in compound assignment
* 10 for Result of multiplication cast to wider type
* 3 for Spurious Javadoc @param tags
* 3 for Potential input resource leak
* 1 for Contradictory type checks

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #1060, opened 2019-01-31
Fix bug preventing addition of new force templates

NickAragua:
Apparently you need to initialize an array list before calling Collections.copy to put items into it.


MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
```diff
@@ -260,6 +260,7 @@ public ScenarioForceTemplate(int forceAlignment, int generationMethod, double fo
         this.destinationZone = destinationZone;
         this.retreatThreshold = retreatThreshold;
         this.allowedUnitType = allowedUnitType;
+        deploymentZones = new ArrayList<>();
```
sixlettervariables: why not: `this.deploymentZones = new ArrayList<>(deploymentZones);`

NickAragua: Originally, I did the Collections.copy in response to a code review comment, but I can do it this way too, as I don't see a practical difference.

Issue #155, opened 2016-09-08
[Request] Advance Day from Within Personnel Market+choose default sort order.


VampireSeraphin:
Simple one this time. Can we get a button to advance the day from within the personnel market. Its a pain to close and reopen it dozens of times for the recruitment methods that refresh daily.

Likewise, can we make it so MekHq remembers what sort column you specified last time so you don't have to resort every time you open the personnel market.

Windchild292:
Actually, #1546 was the second part that needed to be implemented. This can be closed as both parts have been implemented in MekHQ

HammerGS:
Implemented

PR #1081, opened 2019-02-14
Add more user settings for mekhq

### LINKED ISSUE #230: Remember last campaign save dir
It would be great if MHQ remembered what the last directory used to save/load campaign files was. I have an external dir that I attach to my google drive. I could symlink it to the 'campaigns' directory within the install but I prefer to keep it external. MHQ automatically looks in the 'campaigns' dir every time, but I think it would be good if we simply looked at the last used dir.

VicenteCartas:
- Fixes #230 
- Fixes the settings part of #155 
- Adds settings to the unit market dialog.

This PR is to discuss a new way of storing the state of the UI as user settings so it's remembered the next time the user launches the app.

Right now, we are storing the theme on the mmconf/mekhq.properties file, the new code expands what we save there, and builds a new class called MekHqPreferences that wraps around java.util.Properties so we have typed access to properties instead of having to use everything as strings.

This type of saving UI state could be added to other parts (for example, Personnel, Hangar and Warehouse tabs), but before doing that work, making sure we agree on the approach for this.


HammerGS:
This pull request **fixes 2 alerts** when merging 8ca7c2aa89a2c4170b880c22f637273d60ef8cd4 into 812497c6291bbc8c5a866b53cda82f3c043ba8cf - [view on LGTM.com](https://lgtm.com/projects/g/MegaMek/mekhq/rev/pr-56781bcda43248bf4baa7acfa7ccf8bda5ee6a6e)

**fixed alerts:**

* 1 for Potential output resource leak
* 1 for Potential input resource leak

---

*Comment posted by [LGTM.com](https://lgtm.com)*

Issue #1057, opened 2019-01-30
45.2 : exception when using GM mode to edit VTOL damage


Question2:
[a.zip](https://github.com/MegaMek/mekhq/files/2813166/a.zip)

Exception error when trying to use GM -> Edit damage for the Harpy VTOL that is in salvage mode, see log file for error.

neoancient:
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: (minimum <= value <= maximum) is false
	at javax.swing.SpinnerNumberModel.<init>(Unknown Source)
	at javax.swing.SpinnerNumberModel.<init>(Unknown Source)
	at megamek.client.ui.swing.UnitEditorDialog.initArmorPanel(UnitEditorDialog.java:235)
	at megamek.client.ui.swing.UnitEditorDialog.initComponents(UnitEditorDialog.java:135)
	at megamek.client.ui.swing.UnitEditorDialog.<init>(UnitEditorDialog.java:120)
	at mekhq.gui.adapter.UnitTableMouseAdapter.actionPerformed(UnitTableMouseAdapter.java:494)

sixlettervariables:
Somehow that VTOL has more internal listed than it should in its rotor.

sixlettervariables:
If you search in your CPNX for:
```xml
<armor points="4" type="Internal"/>
```
Change points to 3 and save it, that should clear that up.

Not sure how that happened.

sixlettervariables:
Ok that was weird.
```java
    @Override
    public void updateConditionFromPart() {
        if(null != unit && unit.getEntity() instanceof VTOL) {
            unit.getEntity().setInternal(unit.getEntity().getOInternal(VTOL.LOC_ROTOR) - damage, VTOL.LOC_ROTOR);
        }
    }
```
Damage is `-1`, so after loading the unit it actually adds one to the internal.

sixlettervariables:
Apparently damage is `-1` because:
```java
    @Override
    public void updateConditionFromEntity(boolean checkForDestruction) {
        if(null != unit) {
            if(IArmorState.ARMOR_DESTROYED == unit.getEntity().getInternal(loc)) {
                remove(false);
            } else {
                damage = unit.getEntity().getOInternal(loc) - unit.getEntity().getInternal(loc);
                if(unit.isLocationBreached(loc)) {
                    breached = true;
                }
            }
        }
    }
```

sixlettervariables:
[...]

PR #1088, opened 2019-02-16
Don't allow damage to go negative or use negative damages #1057

sixlettervariables:
This is a partial fix for #1057, keeps damages for rotors from rolling negative or using negative damages. Not 100% sure of the root cause, seems either `Rotor::fix` got called twice, or some calculation for the internals on the entity was incorrect in MM.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/TankLocation.java
+++ b/MekHQ/src/mekhq/campaign/parts/TankLocation.java
@@ -207,7 +207,8 @@ public class TankLocation extends Part {
             if(IArmorState.ARMOR_DESTROYED == unit.getEntity().getInternal(loc)) {
                 remove(false);
             } else {
-                damage = unit.getEntity().getOInternal(loc) - unit.getEntity().getInternal(loc);
+                int internal = unit.getEntity().getInternal(loc);
+                damage = unit.getEntity().getOInternal(loc) - Math.max(internal, 0);
                 if(unit.isLocationBreached(loc)) {
                     breached = true;
                 }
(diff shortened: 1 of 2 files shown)


neoancient:
The root cause may have started here in TankLocation#updateConditionFromEntity:
```damage = unit.getEntity().getOInternal(loc) - unit.getEntity().getInternal(loc);```
If a location is destroyed, `getInternal(loc)` returns IArmorState.ARMOR_DESTROYED`, which has a value of -3.
This would give a damage value greater than the original internal. If the location was not truly destroyed and fully repaired, it would have had an internal higher than the original, then when the condition is updated from the Entity the damage is computed as a negative value.

File: MekHQ/src/mekhq/campaign/parts/TankLocation.java [hunk 1]
```diff
@@ -207,7 +207,7 @@ public void updateConditionFromEntity(boolean checkForDestruction) {
             if(IArmorState.ARMOR_DESTROYED == unit.getEntity().getInternal(loc)) {
                 remove(false);
             } else {
-                damage = unit.getEntity().getOInternal(loc) - unit.getEntity().getInternal(loc);
+                damage = Math.max(unit.getEntity().getOInternal(loc) - unit.getEntity().getInternal(loc), 0);
```
neoancient: If getInternal(loc) is `IArmorState.ARMOR_DESTROYED`, `damage` will be original internal + 3.

sixlettervariables: Agreed, but isn't that checked above?

neoancient: So it is. I was only looking at the line that was changed on not the surrounding context.

sixlettervariables: I've updated it to cap damage at original internals.

PR #1108, opened 2019-02-24
Npeinbot

### LINKED ISSUE #1106: Scenarios do not load.
[Raiden.zip](https://github.com/MegaMek/mekhq/files/2897444/Raiden.zip)

(Delete the above lines)

### Environment
What version of MekHQ does your issue apply to? 45.3
What operating system are you using?  Windows, Mac, or Linux? Windows
What java version are you using?  If unsure it will state your java version at the top of the mekhqlog.txt file. 1.8.0_201-b09

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/2897440/mekhqlog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/2897441/mekhqlog.txt)

### Description
Scenarios do not load.

### Files
[Raiden30550319.zip](https://github.com/MegaMek/mekhq/files/2897442/Raiden30550319.zip)

VicenteCartas:
Fixes #1106

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/BotForce.java
+++ b/MekHQ/src/mekhq/campaign/mission/BotForce.java
@@ -162,7 +162,9 @@ public class BotForce implements Serializable, MekHqXmlSerializable {
         int bv = 0;
         
         for(Entity entity : getEntityList()) {
-            bv += entity.calculateBattleValue(true, false);
+            if (entity != null) {
+                bv += entity.calculateBattleValue(true, false);
+            }
         }
         
         return bv;
--- a/MekHQ/src/mekhq/campaign/mission/BotForce.java
+++ b/MekHQ/src/mekhq/campaign/mission/BotForce.java
@@ -3,7 +3,7 @@ package mekhq.campaign.mission;
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.UUID;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import org.w3c.dom.Node;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/mission/BotForce.java
```diff
@@ -194,9 +195,7 @@ public void writeToXml(PrintWriter pw1, int indent) {
 
         pw1.println(MekHqXmlUtil.indentStr(indent+1) + "<entities>");
         for (Entity en : entityList) {
-            if (en != null) {
-                pw1.println(AtBScenario.writeEntityWithCrewToXmlString(en, indent + 2, entityList));
-            }
+            pw1.println(AtBScenario.writeEntityWithCrewToXmlString(en, indent + 2, entityList));
```
NickAragua: The rest of the PR seems to be about protecting vs NPEs, why remove the null check here?

MekHQ/src/mekhq/campaign/mission/BotForce.java
```diff
@@ -115,7 +116,7 @@ public void setName(String name) {
     }
 
     public void setEntityList(ArrayList<Entity> entityList) {
-        this.entityList = entityList;
+        this.entityList = new ArrayList<>(entityList.stream().filter(x -> x != null).collect(Collectors.toList()));
```
NickAragua: Could you add a comment here explaining what's being done? It looks like you're removing all null entities from the passed-in list, but it took me a minute to figure it out.

MekHQ/src/mekhq/campaign/mission/BotForce.java
VicenteCartas: There should be no reasonable way to get a null here anymore as we are filtering the collections in the constructor and in the setter. If someone adds a null item via the getter, I think that should error and then we figure out why that happened, as that seems an error worth catching.

MekHQ/src/mekhq/campaign/mission/BotForce.java
VicenteCartas: Sure!

MekHQ/src/mekhq/campaign/mission/BotForce.java
NickAragua: Perhaps then a better way would be to log the presence of a null entity and then move on. An error here would, I believe, prevent the game from being saved.

VicenteCartas: Ok, I'll add error logging here.

PR #1130, opened 2019-03-01
Ensure equipment parts don't chime in for parts not in their location #1109

sixlettervariables:
This ~~isn't strictly~~ is required after #1129, ~~but~~ this fixes the upstream problem when evaluating equipment parts on campaign load. Basically these system level parts were chiming in for any matching equipment number, no matter the location. The combination of the prior pull request and this one should fix #1109.

Issue #1109, opened 2019-02-24
Destroyed equipment not tracked properly on save/load


RexPearce:
In 0.45.3, certain destroyed equipment is not tracked properly on a save/load.

E.g., if you salvage a weapon from a mech, then save/load, the weapon that was just salvaged will be available again for salvage from the mech.

[My Campaign30250102.zip](https://github.com/MegaMek/mekhq/files/2898486/My.Campaign30250102.zip)

To reproduce:

    1. Load the attached campaign which has a mech ready to be salvaged. Note the warehouse is empty.

    3. In repair bay, select the RA location of the mech, and salvage the Large Laser.

    4. Save the campaign, then reload it. Note the warehouse correctly has a Large Laser.

    5. **Bug:** Go back to the repair bay and note the Large Laser is available for salvage again.

    6. **Bug:** Switch the mech to repair mode, note the Large Laser is marked as repairable (not replaceable), and can be repaired without consuming the Larger Laser in the warehouse.

I've noticed this seems to mainly affect the first (top) weapon installed in a location: if you repeat the above steps salvaging the RA Medium Laser instead, there is no problem.

Note: from diffing the two campaign files, it looks like the Large Laser is correctly marked as destroyed, so I'm guessing the problem is in the load routines:
```
-         <slot index="5" type="Large Laser" quirks="jettison_capable" isDestroyed="false"/>
-         <slot index="6" type="Large Laser" quirks="jettison_capable" isDestroyed="false"/>
+         <slot index="5" type="Large Laser" quirks="jettison_capable" isHit="true" isRepairable="false" isDestroyed="true"/>
+         <slot index="6" type="Large Laser" quirks="jettison_capable" isHit="true" isRepairable="false" isDestroyed="true"/>

```

MekHQ v0.45.3
Ubuntu 18.04.2
OpenJDK v1.8.0_191

HammerGS:
We need the log files as well, and any custom units.

RexPearce:
Logs attached.  (No custom units.)

[logs.zip](https://github.com/MegaMek/mekhq/files/2898525/logs.zip)

RexPearce:
@HammerGS I looked into this a bit last night and I think I found the location causing the problem:

https://github.com/MegaMek/mekhq/blob/8206566a037d86b41f576e3cc86ce0f57412f389/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java#L430-L434

[...]

PR #1132, opened 2019-03-02
Only remove duplicate parts if they are the same type on CPNX load #1109

sixlettervariables:
This is a follow-up to #1130 that adds a check for the actual part types, since it appears that MekActuators end up with the same equipment numbers as actual equipment parts (different bug?).

### DIFF
--- a/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
+++ b/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
@@ -413,7 +413,9 @@ public class CampaignXmlParser {
                         continue;
                     }
                     // Remove existing duplicate parts.
-                    if (u.getPartForEquipmentNum(((EquipmentPart) prt).getEquipmentNum(), ((EquipmentPart) prt).getLocation()) != null) {
+                    Part duplicatePart = u.getPartForEquipmentNum(((EquipmentPart) prt).getEquipmentNum(), ((EquipmentPart) prt).getLocation());
+                    if (duplicatePart instanceof EquipmentPart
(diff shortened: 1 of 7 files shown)


sixlettervariables:
I've expanded on the original commit to include some suggestions from @RexPearce. Based on the campaigns given which experience this bug, this appears to fix them and does not introduce any new odd behaviors.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -413,7 +413,9 @@ else if (xn.equalsIgnoreCase("info")) {
                         continue;
                     }
                     // Remove existing duplicate parts.
-                    if (u.getPartForEquipmentNum(((EquipmentPart) prt).getEquipmentNum(), ((EquipmentPart) prt).getLocation()) != null) {
+                    Part duplicatePart = u.getPartForEquipmentNum(((EquipmentPart) prt).getEquipmentNum(), ((EquipmentPart) prt).getLocation());
+                    if (duplicatePart instanceof EquipmentPart
```
NickAragua: I assume that duplicatePart instanceof EquipmentPart will return false rather than throwing an NPE when duplicatePart is null?

sixlettervariables: Null check would be redundant per the spec: "At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and the reference could be cast to the ReferenceType without raising a ClassCastException. Otherwise the result is false."

sixlettervariables:
This handles everything but OmniPods apparently.

PR #1133, opened 2019-03-02
Fix NPE by checking if the dialog is doing retirement or not

VicenteCartas:
Fixes a NPE in the retirement dialog preferences.


MekHQ/src/mekhq/gui/dialog/RetirementDefectionDialog.java
```diff
@@ -147,7 +147,7 @@ public RetirementDefectionDialog (CampaignGUI gui,
         }
 
         setLocationRelativeTo(gui.getFrame());
-        setUserPreferences();
+        setUserPreferences(doRetirement);
```
VicenteCartas: initComponents uses this, so I followed in setUserPreferences.

Issue #1144, opened 2019-03-11
45.3 Cannot recruit bondsmen.


Apocal43:
MekHQ 45.3
Windows 10
Java version 1.8.0_201

Cannot right-click and recruit bondsmen the same as prisoners who are willing to defect. Is this an oversight?
[Phantom Frost Supernova Trinary27.zip](https://github.com/MegaMek/mekhq/files/2951195/Phantom.Frost.Supernova.Trinary27.zip)

VicenteCartas:
I'll give this a look today.

sixlettervariables:
Fixed in #1145

PR #1145, opened 2019-03-12
Update PersonnelTableMouseAdapter.java

VicenteCartas:
Change the logic of displaying JMenu to separate the options of Ransom and Recruiting Prisonners, which allows to recruit Bondsmen but not ransoming them.


sixlettervariables:
Fixes #1144

Issue #1164, opened 2019-03-28
45.4 Can't add loot to scenario (it works in 45.3)


bsielski:
Linux versions (45.3 and 45.4).
Java version "1.8.0_201"

In 45.4 "Done" button does nothing.
In 45.3 it works.
![](https://i.imgur.com/yLnHRyg.png)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3017817/mekhqlog.txt)

sixlettervariables:
The log shows the same issue as #1161, which was fixed yesterday. I'll confirm this fixes your issue before I close this issue.

PhoenyxCullen:
I'm not having the error from #1161. I'm getting the start of this error:

Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Double
	at mekhq.gui.dialog.LootDialog.done(LootDialog.java:313)

My log is attached.
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3052335/mekhqlog.txt)

PhoenyxCullen:
Use CO Peacetime Op Costs is turned OFF.

VicenteCartas:
Not related to #1161, it's converstion to create money.

PR #1175, opened 2019-04-08
Fix for MekHQ #1164

NickAragua:
Fixes a behavior introduced in 45.4 where the user is unable to add loot to scenarios.


NickAragua:
PR #1178 is a superset of this.

PR #1190, opened 2019-04-23
Fixes accessibility issues (#1188)

### LINKED ISSUE #1188: Edit PersonnelDIalog has Poor Labels for Accessibility
The Edit personnel window is very difficult to use because none of the labels for controls are programmatically associated with their targets. Skills are difficult to edit because it's hard to be sure which skill I'm editing, among other similar problems.

VicenteCartas:
Fixes accessibility issues in CustomizePersonDialog related to the skill list.
Fixes #1188.
Cleans most Idea warnings from CustomizePersonDialog.java.


BlindGuyNW:
Okay, so I'm honestly not sure what's going on here. :) It looks like it *should* work but the edit fields still read as unlabeled, which is annoying. 

I'll try to do an audio recording at some point to show where things still seem off. It actually isn't as bad as I remember, for some reason I thought the checkboxes just didn't announce which skill they were for, but apparently that's just a bad memory, or maybe it came from my tesing of this under WINdows, I don't know.

I appreciate the work on it and am sorry I'm not more familiar with the Java accessibility API. Web stuff is more my bag.

VicenteCartas:
No probs! For the checkbox, I think adding the text to the checkbox themselves instead of as a separate label should probably make things better to announce which skill you are (I didn't want to have to change layout if possible, but if needed I'll work on that).

The audio recording will help me a lot, appreciated. I'll keep this open until we figure out what's going on.

sixlettervariables:
If you can get Java Monkey working you can "see" sort of what tools like NVDA would see. It isn't as straightforward to get it all set up however.

BlindGuyNW:
Hey,

I meant to pass this along before, but forgot. :) [this program](https://github.com/google/access-bridge-explorer) might be easier to use or at least more up-to-date than JavaMonkey.

VicenteCartas:
I'll check those programs. I was thinking on using Inspect and Windows Narrator, but they don't see anything inside the Java AWT Window :S

sixlettervariables:
Did you enable Access Bridge first? (not Google's but Java's)

VicenteCartas:
Nope, I'll check that too today at night.

Issue #1205, opened 2019-05-21
Sorting the Buy Parts Menu freezes it.


Rod995:
### Environment
`MekHQ 0.45.4`

`Windows 10 (64-bit)`

`Java version 1.8.0_191`

### Description
So I went into the menu to buy items, right?
Decided that I wanted to see the most expensive thing, so I sorted it by cost, and it just becomes unresponsive.
No, "This Application has become Unresponsive", it just stops responding to input, or at least that window does. The main campaign menu still seems responsive.

#### Steps
`Open Campaign.`
`Go to 'Purchase Parts' menu.`
`Sort by cost by clicking on the tab for cost.`
`At this time it should become unresponsive.`

### Files
`Campaign File`
[230751002.zip](https://github.com/MegaMek/mekhq/files/3204678/230751002.zip)

`Custom Units`
[2.zip](https://github.com/MegaMek/mekhq/files/3204672/2.zip)

`MekHQ Log`
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3204669/mekhqlog.txt)

sixlettervariables:
The sorters in PartsTableDialog.java are the cause:
https://github.com/MegaMek/mekhq/blob/5ea5a4fc55ea48773fc2533145f21afdd41248bb/MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java#L753-L782

sixlettervariables:
Along with repeated calls to `Campaign::getTargetForAcquisition`.

sixlettervariables:
The other cause: if you exit with an expensive column sorted, it'll launch with that sort enabled. I have a partial fix (sorting is faster), but the initial load will be slow if the last sorted column is expensive.

VicenteCartas:
If the initial sorting is problematic, we could also disable remembering the preferences for this table in concrete.

sixlettervariables:
Ok, I the partial fix has landed. In *most* cases it should be must faster sorting.

sixlettervariables:
Closed by #1207

PR #1207, opened 2019-05-24
Improve Purchase Parts performance #1205

sixlettervariables:
This improves the purchase parts dialog's performance through a few improvements:

1. Introduce proxy objects to memoize expensive properties.
2. Ensure sorting is performed on native values and not by parsing strings.
3. Directly return values to the table rather than creating strings.

This is a partial fix for #1205.

This does not address part details sorting.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
@@ -29,6 +29,7 @@ import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.util.ArrayList;
+import java.util.Objects;
 import java.util.ResourceBundle;
 
 import javax.swing.DefaultComboBoxModel;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
```diff
@@ -28,9 +28,7 @@
 import java.awt.GridBagConstraints;
```
sixlettervariables: Best reviewed with whitespace ignored.

MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
```diff
@@ -377,181 +386,381 @@ public void filterParts() {
...
+        case SG_BA_SUIT:
+            return "Battle Armor Suits";
+        case SG_OMNI_POD:
+            return "Empty OmniPods";
+        default:
+            return "?";
+        }
+    }
+
+    private Person getLogisticsPerson() {
```
NickAragua: Does this get re-initialized every time the window is re-opened? I'm perfectly happy with a cached approach, but it may run into issues if you open the parts window, close it, then your logistics person changes, then you open it back up again.

MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
```diff
@@ -377,181 +386,381 @@ public void filterParts() {
...
+            @Override
+            public int compareTo(FormattedValue<T> o) {
+                if (null == o) {
+                    return -1;
+                }
+                return getValue().compareTo(o.getValue());
+            }
+        }
+
+        public class PartProxy {
```
NickAragua: This is pretty good, could you add a comment explaining a bit about what the purpose of it is? Same with TargetProxy.

MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
sixlettervariables: Yes, these dialog instances are not cached or stored.

MekHQ/src/mekhq/gui/dialog/PartsStoreDialog.java
sixlettervariables: Yep.

Issue #1193, opened 2019-04-26
Unable to load atb campaign due to heat sinks?


MarcJC:
[loaderror.zip](https://github.com/MegaMek/mekhq/files/3122867/loaderror.zip)

Playing HQ ATB 45.4 and after finishing a mission I saved before starting the second mission.  
When opening the second mission all mech icons were missing, although bot vehicles still showed.  
Closed out MegaMek and reloaded but no change.  

Tried rebooting my laptop to see if that would fix the issue but was unable to load the most recent or even a day or two back in the campaign.  

Attached Log files saves, last autosave, and customs in zip for debugging.  

Looking at the log file on loading I see this    

mod_weapons failed for Starfire SF-1X (Prototype) - Invalid quirk!
java.lang.ClassCastException: mekhq.campaign.parts.equipment.HeatSink cannot be cast to mekhq.campaign.parts.Armor
	at mekhq.campaign.io.CampaignXmlParser.parse(CampaignXmlParser.java:590)
	at mekhq.campaign.CampaignFactory.createCampaign(CampaignFactory.java:88)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:210)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:129)
	at javax.swing.SwingWorker$1.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at javax.swing.SwingWorker.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)

I am not using a starfire at all and I have no idea why it's heat sinks would cause any type of error.  

Any help in recovering my saves would be much appreciated.

sixlettervariables:
This is related to a refit actually, the message above that is a red herring. It appears the armor allocated for a refit somehow got mixed up with a heat sink? Not sure why. I'm going to look into restoring your CPNX.

sixlettervariables:
I've added new logging and it looks like two refits are the issue:
```
21:33:17,604 ERROR [mekhq.campaign.io.CampaignXmlParser] {SwingWorker-pool-1-thread-1} 
[...]

Issue #1180, opened 2019-04-15
Unable to open campaign file


kipstafoo:
After shutting down MHQ, I am unable to reload my campaign files. The error in the log is below.

I have attached the campaign file plus all custom mechfiles for the campaign.
[bug data.zip](https://github.com/MegaMek/mekhq/files/3082227/bug.data.zip)

`
Attempted to change ammo type of non-ammo
Attempted to change ammo type of non-ammo
java.lang.ClassCastException: megamek.common.weapons.ppc.ISERPPC cannot be cast to megamek.common.AmmoType
	at mekhq.campaign.parts.equipment.LargeCraftAmmoBin.getFullShots(LargeCraftAmmoBin.java:156)
	at mekhq.campaign.parts.equipment.LargeCraftAmmoBin.updateConditionFromEntity(LargeCraftAmmoBin.java:283)
	at mekhq.campaign.unit.Unit.runDiagnostic(Unit.java:584)
	at mekhq.campaign.io.CampaignXmlParser.parse(CampaignXmlParser.java:672)
	at mekhq.campaign.CampaignFactory.createCampaign(CampaignFactory.java:88)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:210)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:129)
	at javax.swing.SwingWorker$1.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at javax.swing.SwingWorker.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
`

The offending unit is the Overlord (2762b). By removing this unit and updating anyone assigned to it, the campaign file loads just fine. I tried a smaller test-case by creating a new campaign with only this unit but the campaign file loaded and the unit entries in the two campaign files looked the same so I'm not able to currently offer more insight into why the main campaign file fails to load.

kipstafoo:
As an update, after manually removing the DS, I re-added it and re-assigned all the crew members to it (oddly, all but one had lost their skills so I had to fix that). After doing this I saved and restarted and the new campaign file works. It looks like something got corrupted along the way. I've attached the newly working campaign file in case you want to diff it.


[...]

Issue #1138, opened 2019-03-04
45.3 After refit, cant open "Repair Bay" tab in MekHQ


Arakash0:
### Environment
MeqHQ 0.45.3
Java version 1.8.0_181
Windows 7

### Description
On the day a refit of a Mech (an ON-1MA) was finished, when I pressed the repair bay tab the repair view did not appear. I believe the mech was showing as damaged in the hanger list.

After saving and reloading, the problem seemed to be fixed.

Following error is visible in the mekhqlog:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at mekhq.campaign.parts.equipment.AmmoBin.isOmniPodded(AmmoBin.java:812)
	at mekhq.campaign.parts.equipment.AmmoBin.getActualTime(AmmoBin.java:443)
	at mekhq.campaign.parts.Part.getTimeLeft(Part.java:826)
	at mekhq.campaign.parts.equipment.AmmoBin.getDesc(AmmoBin.java:497)
	at mekhq.gui.model.TaskTableModel.getValueAt(TaskTableModel.java:53)
	at javax.swing.table.TableRowSorter$TableRowSorterModelWrapper.getValueAt(Unknown Source)
	at javax.swing.DefaultRowSorter.compare(Unknown Source)
	at javax.swing.DefaultRowSorter.access$100(Unknown Source)
	at javax.swing.DefaultRowSorter$Row.compareTo(Unknown Source)
	at javax.swing.DefaultRowSorter$Row.compareTo(Unknown Source)
	at java.util.ComparableTimSort.countRunAndMakeAscending(Unknown Source)
	at java.util.ComparableTimSort.sort(Unknown Source)
	at java.util.Arrays.sort(Unknown Source)
	at javax.swing.DefaultRowSorter.sort(Unknown Source)
	at javax.swing.DefaultRowSorter.allRowsChanged(Unknown Source)
	at javax.swing.JTable.notifySorter(Unknown Source)
	at javax.swing.JTable.sortedTableChanged(Unknown Source)
	at javax.swing.JTable.tableChanged(Unknown Source)
	at javax.swing.table.AbstractTableModel.fireTableChanged(Unknown Source)
	at javax.swing.table.AbstractTableModel.fireTableDataChanged(Unknown Source)
	at mekhq.gui.model.DataTableModel.setData(DataTableModel.java:45)
	at mekhq.gui.RepairTab.refreshTaskList(RepairTab.java:798)
[...]

Issue #1123, opened 2019-02-26
45.3 ATB Campaign wont load


Arakash0:
### Environment
MeqHQ 0.45.3
Java version 1.8.0_181
Windows 7

### Description
Cant load several of the campaign saves i made in the last few hours. Ive attached the latest one that isn't working below.
I thought this might be this bug: https://github.com/MegaMek/mekhq/issues/1096
but the error message in MekHQlog seems to be different to the one listed there.

~~If there is any kind of workaround i could use to keep playing until the next release comes out i would be eternally grateful.~~
Edit: When sixlettervariables mentioned the refits might be the cause, found a workaround by deleting the refits in the save which let me load the campaign

### Files
[Campaign Save.zip](https://github.com/MegaMek/mekhq/files/2903750/Campaign.Save.zip)
[mechfiles.zip](https://github.com/MegaMek/mekhq/files/2903751/mechfiles.zip)
[mekhqlog.zip](https://github.com/MegaMek/mekhq/files/2903752/mekhqlog.zip)

sixlettervariables:
Exception from log:
```
java.lang.ClassCastException: mekhq.campaign.parts.MekLocation cannot be cast to mekhq.campaign.parts.Armor
	at mekhq.campaign.io.CampaignXmlParser.parse(CampaignXmlParser.java:595)
	at mekhq.campaign.CampaignFactory.createCampaign(CampaignFactory.java:88)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:194)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:118)
	at javax.swing.SwingWorker$1.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at javax.swing.SwingWorker.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
```

sixlettervariables:
Were you doing a refit?

https://github.com/MegaMek/mekhq/blob/83862b6d8caddfc3651f7c8d6411ae5b5147132c/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java#L595-L596

Arakash0:
Yes.
From memory several refits were ongoing in the last few hours i was playing.

[...]

PR #1216, opened 2019-06-16
Track part ID moves during campaign loads that may affect refits

sixlettervariables:
This is an experimental change that may fix the tough to track down part ID changes affecting refits at CPNX load #1193 #1180 #1138 #1123 (to name a few).

*note: best compared with whitespace ignored*

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -1554,13 +1554,36 @@ public class Campaign implements Serializable, ITechManager {
             for (Unit u : getUnits()) {
                 Refit r = u.getRefit();
                 // If there is a refit and this part matches the new armor, update the ID
-                if (null != r && r.getNewArmorSuppliesId() == p.getId()
-                    && mergedWith instanceof Armor) {
-                    MekHQ.getLogger().info(Campaign.class, "addPartWithoutId",
-                        String.format("%s (%d) was merged with %s (%d) used in a refit for %s", p.getName(),
-                            p.getId(), mergedWith.getName(), mergedWith.getId(), u.getName()));
-                    Armor mergedArmor = (Armor)mergedWith;
-                    r.setNewArmorSupplies(mergedArmor);
+                if (null != r) {
+                    if (mergedWith instanceof Armor
+                        && r.getNewArmorSuppliesId() == p.getId()) {
+                        MekHQ.getLogger().info(Campaign.class, "addPartWithoutId",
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1425,23 +1426,26 @@ public void addPart(Part p, int transitDays) {
                         ((Armor) spare).setAmount(((Armor) spare).getAmount()
                                 + ((Armor) p).getAmount());
                         MekHQ.triggerEvent(new PartChangedEvent(spare));
+                        p.setId(-1);
```
NickAragua: What does the -1 part ID signify?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1516,30 +1520,73 @@ public void addPartWithoutId(Part p) {
...
+                            }
+                        }
+                        ids = r.getNewUnitPartIds();
+                        for (int ii = 0; ii < ids.size(); ++ii) {
+                            int nid = (int)ids.get(ii);
+                            if (p.getId() == nid) {
+                                MekHQ.getLogger().info(Campaign.class, "addPartWithoutId",
+                                String.format("%s (%d) was merged with %s (%d) used in the new unit in a refit for %s", p.getName(),
+                                    p.getId(), mergedWith.getName(), mergedWith.getId(), u.getName()));
+                                ids.set(ii, mergedWith.getId());
```
NickAragua: What's the point of adding elements to the ids collection if you're not using it? Is it just a debugging thing?

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: Code which writes parts in some areas uses a full part description if the ID is less than or equal to zero. This is an attempt to force that if that code is the cause of the refit bugs.

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: Those are loaded at CPNX read and eventually get out of sync. I'm not sure if these IDs getting out of sync are the proximate cause, but it seemed worth doing if we can.

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: (The full description can't have an ID get out of sync, because it creates a brand new part)

Issue #1219, opened 2019-06-30
[Latest Snapshot] Parts in Use screen shows ammo bins instead of ammo


NickAragua:
Looks like a recent fix made some unintentional changes to the "Parts in Use" UI. Where ammo used to be listed adjacent to weapons on that screen it now lists the related ammo bin. Clicking on "buy" for the ammo bin leads an error shown in the log.

Won't be able to work on this for a while so entering it in case someone else wants to grab it.

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3342944/mekhqlog.txt)
[Spectre's Shadows c1m2.zip](https://github.com/MegaMek/mekhq/files/3342945/Spectre.s.Shadows.c1m2.zip)

sixlettervariables:
Exception from the log:
```
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at megamek.common.event.EventListener.trigger(EventListener.java:33)
	at megamek.common.event.EventBus.internalTrigger(EventBus.java:158)
	at megamek.common.event.EventBus.trigger(EventBus.java:147)
	at mekhq.MekHQ.triggerEvent(MekHQ.java:723)
	at mekhq.campaign.Campaign.addReportInternal(Campaign.java:3711)
	at mekhq.campaign.Campaign.addReport(Campaign.java:3698)
	at mekhq.campaign.Campaign.goShopping(Campaign.java:2153)
	at mekhq.campaign.market.ShoppingList.addShoppingItem(ShoppingList.java:139)
	at mekhq.gui.OverviewTab$1.actionPerformed(OverviewTab.java:306)
	at mekhq.gui.model.PartsInUseTableModel$ButtonColumn.actionPerformed(PartsInUseTableModel.java:336)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
[...]

Issue #1185, opened 2019-04-20
45.4 Unable to acquire Clan AMS Ammo Bin.


Apocal43:
MekHQ 45.4
Windows 10
Java version 1.8.0_201

Attempting to acquire Clan AMS Ammo Bin (not ammo itself) from the "Parts Acquisition" menu under the Repair Bay tab merely acquires more ammo. It can be GM acquired without issue. No other option, at least that I have found, allows for purchasing an Ammo Bin of any kind.
![image](https://user-images.githubusercontent.com/43554459/56460578-5d8ab280-63d7-11e9-870f-1612c63ab9a0.png)

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3100444/mekhqlog.txt)
[Phantom Frost Supernova Trinary59.zip](https://github.com/MegaMek/mekhq/files/3100445/Phantom.Frost.Supernova.Trinary59.zip)

PR #1220, opened 2019-06-30
Revert "Merge pull request #1203 from Krashner/ammo-bin-repair"

sixlettervariables:
PR #1203 introduced a regression in the Parts in Use dialog (#1219). This reverts commit 9e600312c6d63946e8a1d771804e7bc2e7fcbf5c, reversing
changes made to 5ea5a4fc55ea48773fc2533145f21afdd41248bb.

The ammo bin issue should be addressed, but not in a manner that regresses ammunition purchases.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/ShoppingList.java
+++ b/MekHQ/src/mekhq/campaign/market/ShoppingList.java
@@ -115,6 +115,11 @@ public class ShoppingList implements MekHqXmlSerializable {
     }
 
     public void addShoppingItem(IAcquisitionWork newWork, int quantity, Campaign campaign) {
+        //ammo bins need a little extra work here
+        if(newWork instanceof AmmoBin) {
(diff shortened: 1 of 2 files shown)


sixlettervariables:
#1185 should be reopened if this PR lands.

neoancient:
I looked into another way to fix the initial issue and I haven't figured out a better way to do it. Given that I can't reproduce the problem with the parts in use overview either, my opinion is that we should let this fix stand. If we don't want the ammo bins to appear on the parts in use tab we could filter them out, but I find that to me a minor issue compared with all the problems we've had getting ammo bins to behave in an expected way.

neoancient:
I just looked at #1219 again. I had forgotten that the issue was that it's showing `AmmoBin` instead of `AmmoStorage`, so there is an issue there that needs to be addressed. I still think a tweak of `OverviewTab` is preferable.

sixlettervariables:
Sounds good.

PR #1232, opened 2019-07-23
Campaign subset export update

NickAragua:
Fixed an issue where, if exporting a person into a campaign where their ID already exists, their kills would duplicate. 

Clear an exported unit's force ID to avoid phantom irreversible force assignments on export.

Optimization to the kills mechanism, where they are now stored in memory in a hashmap keyed by person ID, rather than re-scanning every kill in the campaign every time (this could probably lead to some hairy performance for longer campaigns).

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -199,7 +199,8 @@ public class Campaign implements Serializable, ITechManager {
     private TreeMap<Integer, Force> forceIds = new TreeMap<>();
     private TreeMap<Integer, Mission> missions = new TreeMap<>();
     private TreeMap<Integer, Scenario> scenarios = new TreeMap<>();
-    private List<Kill> kills = new ArrayList<>();
+    //private List<Kill> kills = new ArrayList<>();
+    private Map<UUID, ArrayList<Kill>> kills = new HashMap<>();
 
     private Map<String, Integer> duplicateNameHash = new HashMap<>();
 
@@ -3299,6 +3300,7 @@ public class Campaign implements Serializable, ITechManager {
         }
         removeAllPatientsFor(person);
         removeAllTechJobsFor(person);
+        removeKillsFor(person.getId());
         getRetirementDefectionTracker().removePerson(person);
 
         if (log) {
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5984,16 +5998,21 @@ public void addKill(Kill k) {
     }
 
     public List<Kill> getKills() {
-        return Collections.unmodifiableList(kills);
+        List<Kill> flattenedKills = new ArrayList<Kill>();
```
sixlettervariables:
```suggestion
        List<Kill> flattenedKills = new ArrayList<>();
```

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5984,16 +5998,21 @@ public void addKill(Kill k) {
...
     public ArrayList<Kill> getKillsFor(UUID pid) {
-        ArrayList<Kill> personalKills = new ArrayList<Kill>();
-        for (Kill k : kills) {
-            if (k.getPilotId().equals(pid)) {
-                personalKills.add(k);
-            }
+        ArrayList<Kill> personalKills = kills.get(pid);
+        
+        if(personalKills == null) {
+            return new ArrayList<Kill>();
```
sixlettervariables:
```suggestion
            return Collections.emptyList();
```

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5984,16 +5998,21 @@ public void addKill(Kill k) {
...
-        return Collections.unmodifiableList(kills);
+        List<Kill> flattenedKills = new ArrayList<Kill>();
+        for(ArrayList<Kill> personKills : kills.values()) {
+            flattenedKills.addAll(personKills);
+        }
+        
+        return Collections.unmodifiableList(flattenedKills);
     }
 
     public ArrayList<Kill> getKillsFor(UUID pid) {
```
sixlettervariables: I know this is existing code, but does `getKillsFor` need to be an ArrayList or could it just be a List like `getKills`?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -199,7 +199,7 @@
     private TreeMap<Integer, Force> forceIds = new TreeMap<>();
     private TreeMap<Integer, Mission> missions = new TreeMap<>();
     private TreeMap<Integer, Scenario> scenarios = new TreeMap<>();
-    private List<Kill> kills = new ArrayList<>();
+    private Map<UUID, ArrayList<Kill>> kills = new HashMap<>();
```
sixlettervariables: Could this be:
```suggestion
    private Map<UUID, List<Kill>> kills = new HashMap<>();
```

PR #1233, opened 2019-07-23
Update 'Mech location repair and salvage times to match StratOps #1075

### LINKED ISSUE #1075: Modifiers for salvaging a component
Modifiers for salvaging a Mech component were the same in all MekHQ versions I have checked so far, but closer inspection of StratOps on page 191, gives example in fourth paragraph, indicating that time & modifier should be based on percentage damage to component, so using Repair part of MASTER REPAIR TABLE, not Destroyed part, as it is now (there is +3 difficulty, and 240 minutes now, for all locations).

The overall description is quite lacking in limbs reclamation department, but I wonder, if by analogy to Blown Off Arms with destroyed Left/Right Torso, shouldn't all parts attached to Central Torso albo become Blown Off when CT is destroyed?

sixlettervariables:
This fixes #1075 and updates 'Mech location repair/salvage times and difficulties based on StratOps p191. It also updates our logic for breached locations to include costs for any other repairs required per StratOps p177.

Best to review this with [whitespace off](https://github.com/MegaMek/mekhq/pull/1233/files?w=1).

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MekLocation.java
+++ b/MekHQ/src/mekhq/campaign/parts/MekLocation.java
@@ -1,20 +1,20 @@
 /*
  * Location.java
- * 
+ *
  * Copyright (c) 2009 Jay Lawson <jaylawson39 at yahoo.com>. All rights reserved.
- * 
+ *
  * This file is part of MekHQ.
- * 
+ *
  * MekHQ is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


inqistor:
Excellent, however it is possible to salvage fully functioning limb, especially if you exchange it between Mechs. Line 430 will return 0 time in that instance. Page 191 says, that everything up to 25% damage takes the same amount of time. 
So, procedure works for repair, but will return wrong value for salvaging undamaged limb.

sixlettervariables:
Nice catch @inqistor, I'll merge after the checks complete.

PR #1235, opened 2019-07-28
Bot Unit SPAs

NickAragua:
Pretty simple: this lets generated units for AtB and Template scenarios have crews with SPAs, as per CamOps recommendations.


sixlettervariables:
Can you see why the LGTM build fails?

NickAragua:
I made a change to Entity.java, and forgot about it. Will push it up on the megamek stream.

Also, I'm pretty sure that the weight of an SPA is the inverse of what I need when looking to limit the SPAs added by weight. Going to look into that as well.

Issue #1238, opened 2019-08-02
BUG: Self-crewed units requiring an Engineer cannot be Activated


MaxSizeis:
Currently Mothballed dropships can't be un-mothballed in MekHQ.  Consider a GM option to un-mothball any unit.
As a current work around, the entire unit must be replaced.

Quoting SO pg 175...

The time required to place larger units (WarShips, Drop-Ships, JumpShips, Space Stations and Large Naval Vessel Support Vehicles) in mothballs is equal to one Maintenance/Repair Cycle per 500 tons. The same amount of time is required to return them to operational status.

It should be possible but it is not implemented in MHQ correctly.

sixlettervariables:
I've implemented GM mode mothballing and activation for all units. The reason Activate is not available is there is no "engineer" to perform the activation (because it is self crewed). I'll keep this open to research that bug.

sixlettervariables:
The fix in PR #1239 restores the crew to the vessel before attempting activation. This should allow self maintained units to activate themselves (if the crew is still available).

PR #1239, opened 2019-08-04
Add feature to GM Mothball/Activate units

sixlettervariables:
This adds GM Mothball and Activate to units in the hangar, contributing to #1238. This also splits out mothballing and activation to make it easier to do both as a GM action.


sixlettervariables:
Okay now this PR fixes the issue seen in #1238. Activation now will attempt a restore of the crew (and reset the engineer). This will allow vessels to be activated in most cases. However, you can see some "one turn" oddities where it says it is activating but there is no crew.

PR #1243, opened 2019-08-11
Fix logic for salvage eligibility

### LINKED ISSUE #1240: [0.45.4 : Bug] Cannot Salvage
### Environment

MekHQ 0.45.4

### Description

Mission Type: Guerrilla Warfare
Salvage Percentage: 100%

During resolution after a battle, all salvageable units were greyed out despite my contract having 100% salvage rights. Lots of error in  error log.

To replicate:
1. Load campaign
2. Select mission #052
3. Select second last scenario, hide and seek
4. Load save game
5. end the game, control battlefield

### Files

[CannotSalvageLog.txt](https://github.com/MegaMek/mekhq/files/3481031/CannotSalvageLog.txt)
[Uploading CannotSalvageCampaign.zip…]()
[Uploading CannotSalvageGame.sav.gz…]()

neoancient:
Currently when determining whether a unit is eligible to be salvaged, the current salvage percentage is calculated with the battle being resolved taken into account, and salvage is disallowed if it meets or exceeds what is allowed by the contract. This means that nothing can be salvaged if the current value (with rounding) is calculated at 100% even if the contract allows 100%. It also allows for the possibility that a unit could take the salvage percentage over the limit if the current value is under the limit before it is chosen.

This PR adjusts the logic to calculate what the salvage percentage would be after selecting the unit and compare that to the limit instead. Salvage is disallowed if it would exceed the limit.

Fixes #1240

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
@@ -407,7 +407,7 @@ public class ResolveScenarioWizardDialog extends JDialog {
             miaBtns.add(miaCheck);
             kiaCheck = new JCheckBox("");
             kiaBtns.add(kiaCheck);
-            
+
             hitSlider = new JSlider(JSlider.HORIZONTAL, 0, 5, Math.min(status.getHits(),5));
             hitSlider.setMajorTickSpacing(1);
             hitSlider.setPaintTicks(true);
@@ -440,7 +440,7 @@ public class ResolveScenarioWizardDialog extends JDialog {
             gridBagConstraints.weightx = 1.0;
             pnlPilotStatus.add(kiaCheck, gridBagConstraints);
             i++;
-            
+
             kiaCheck.addItemListener(new CheckBoxKIAListener(hitSlider, miaCheck, null));
         }
         pnlMain.add(pnlPilotStatus, PILOTPANEL);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
```diff
@@ -443,6 +443,8 @@ private void initComponents() {
             gridBagConstraints.weightx = 1.0;
             pnlPilotStatus.add(kiaCheck, gridBagConstraints);
             i++;
+
+            kiaCheck.addItemListener(new CheckBoxKIAListener(hitSlider, miaCheck, null));
```
NickAragua: Looks like a bad merge, I don't think we need to be adding two item listeners to that checkbox.

NickAragua:
Also, as a word of warning, this changes the current behavior, which *does* allow you to go over with one unit. 

Which means it'll be hard to salvage an intact grounded dropship now. :(

MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
neoancient: It's in there twice in current master. Which one should it be?

NickAragua: My bad. It should be the top one.

neoancient:
If the ability to go over is expected and intended, then maybe I should close this and just treat 100% as a special case.

NickAragua:
It's definitely expected currently, as it's been that way as long as I can remember. No idea about whether it was intended or not.

PR #1267, opened 2019-08-25
Track important dates for units

sixlettervariables:
In order to add better accounting for some things, we would need to track two dates for units:
- The date the unit was acquired.
- The date the unit was last maintained.

This adds both of those fields, and starts moving Campaign (etc) to `java.time.LocalDate` rather than `GregorianCalendar`.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -26,6 +26,7 @@ import java.io.PrintWriter;
 import java.io.Serializable;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.time.LocalDate;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -237,6 +238,7 @@ public class Campaign implements Serializable, ITechManager {
     private GregorianCalendar calendar;
     private String dateFormat;
     private String shortDateFormat;
+    private LocalDate today;
 
     private String factionCode;
     private int techFactionCode;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -429,6 +432,7 @@ public String getTitle() {
 
     public void setCalendar(GregorianCalendar c) {
         calendar = c;
+        today = LocalDate.of(c.get(Calendar.YEAR), c.get(Calendar.MONTH) + 1, c.get(Calendar.DAY_OF_MONTH));
```
NickAragua: Why maintain two different dates? We already have the "calendar", so maybe just a convenience property to convert the calendar's state to a LocalDate instead?

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4508,6 +4544,24 @@ public int getDaysSinceMaintenance() {
...
     }
 
+    /**
+     * Gets the last day the unit was maintained, or null if the unit
+     * has never had maintenance performed.
+     * @return The last day the unit was maintained, otherwise null if the
+     *         unit has never had maintenance.
+     */
+    public @Nullable LocalDate getLastMaintenanceDate() {
+        return lastMaintenanceDate;
```
NickAragua: Seems like this would supercede the "daysSinceMaintenance" property/field.

MekHQ/src/mekhq/campaign/Campaign.java
NickAragua: Ah, never mind, I read the PR description.

Issue #1078, opened 2019-02-13
Feature request: GM recharging JumpShip Drive


inqistor:
I see few similar requests already, but GM action should be easiest to implement.

It seems, game assumes that you have your own JumpShip all the time. However you should be able to hire a new one just for jump to another system, and in that system another JumpShip can be waiting with already charged Drive. Or your JumpShip could use Recharge Station.
Currently GM can only move directly to planet in a system, which takes neither time, nor money. And planet can be several weeks of travel from Jump Point, so your Drives would recharge faster, than travel takes. Just please add option for GM to Recharge Drive immediately.

pjfarland:
Perhaps a few enhancements. A random roll for available JS would be great as well. 
Also, adding functionality for Lithium-Fusion Battery powered "double" jumps.

Windchild292:
Fixed in PR #1359

PR #1280, opened 2019-08-30
System refactor update

AaronGullickson:
This is an update of PR #969, after @sixlettervariables help to get it properly updated to the master branch. From the original:

This is the beginning of a refactoring of the way we pull in and use planetary system data based on the creation of full non-canon planetary system data (including social data) that we are currently building [here](https://github.com/AaronGullickson/system_generation).

The new format will wrap all planetary information inside of a <system> node. System-level data (e.g. stellar type, recharge stations) will be kept within this node, while all planetary information will be kept in children planet nodes. The system data will identify a system position id for the primary planet and this information will be used to pull planetary information quickly to the system level.

To do this, I have created a new PlanetarySystem object that now contains some of the information formerly contained within the Planet object as well as a list of Planet objects that are contained within the system. I have also created a Systems object that will ultimately replace the Planets object for reading in the planetary data.

This is still very much a WIP but we are now at the stage with the system generation project, that we can begin to think about how to incorporate the new data back into MHQ. Ultimately, I would like to include the ability to generate a planetary system map when you double-click a system on the interstellar map. However, that will likely be a subsequent PR once this one is fully integrated and functioning. Another future RFE that would be nice would be an InterstellarMapEditor that would open separately from the main MHQ that would make future corrections and changes to the interstellar data possible through a GUI interface.

PR #1289, opened 2019-09-14
Add jumpship recharge option to map right click menu

Skoraks:
Adds the option to fully charge the jump drive on the interstellar map in GM mode. (https://github.com/MegaMek/mekhq/issues/1078).
This is my first dip into Github and open source work so please let me know if I should be doing anything differently.

### DIFF
--- a/MekHQ/src/mekhq/campaign/CurrentLocation.java
+++ b/MekHQ/src/mekhq/campaign/CurrentLocation.java
@@ -133,6 +133,13 @@ public class CurrentLocation implements Serializable {
         jumpPath = path;
     }
 
+    public void setRecharged(Campaign campaign) {
+        DateTime currentDate = Utilities.getDateTimeDay(campaign.getCalendar());
+        double neededRechargeTime = currentPlanet.getRechargeTime(currentDate);
+        rechargeTime = neededRechargeTime;
+        campaign.addReport("Jumpship drives fully charged");
+    }
+
     /**
      * Check for a jump path and if found, do whatever needs to be done to move
      * forward
--- a/MekHQ/src/mekhq/campaign/CurrentLocation.java
+++ b/MekHQ/src/mekhq/campaign/CurrentLocation.java
@@ -137,7 +137,6 @@ public class CurrentLocation implements Serializable {
         DateTime currentDate = Utilities.getDateTimeDay(campaign.getCalendar());
         double neededRechargeTime = currentPlanet.getRechargeTime(currentDate);
         rechargeTime = neededRechargeTime;
-        campaign.addReport("Jumpship drives fully charged");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/CurrentLocation.java
```diff
@@ -133,6 +133,13 @@ public void setJumpPath(JumpPath path) {
         jumpPath = path;
     }
 
+    public void setRecharged(Campaign campaign) {
+        DateTime currentDate = Utilities.getDateTimeDay(campaign.getCalendar());
+        double neededRechargeTime = currentPlanet.getRechargeTime(currentDate);
+        rechargeTime = neededRechargeTime;
+        campaign.addReport("Jumpship drives fully charged");
```
sixlettervariables: Prefer moving this reporting logic out of this and into the action listener.

Also please add something like `(GM)` to the message.

AaronGullickson:
This should wait until after PR #1280 because it will conflict with the refactoring of planets into planetary systems.

Also how are the rules applied to recharge the jump drive? Recharging is not instantaneous. These rules are covered on pg. 87 of *Strategic Operations*. Whenever possible we should follow the rules of the book.

MekHQ/src/mekhq/campaign/CurrentLocation.java
Skoraks: I think I've done what you wanted in e92de1337856ded47da3bf461191e11ddfd1c6be
Also I updated some of the existing report's charging typos

MekHQ/src/mekhq/campaign/CurrentLocation.java
```diff
@@ -133,6 +133,13 @@ public void setJumpPath(JumpPath path) {
         jumpPath = path;
     }
 
+    public void setRecharged(Campaign campaign) {
+        DateTime currentDate = Utilities.getDateTimeDay(campaign.getCalendar());
```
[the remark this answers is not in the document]
Skoraks: It does but not in the correct type, getDate() provides it as 'java.util.date', whereas getRechargeTime() is currently expecting a date time in 'org.joda.time.DateTime' - what do you recommend?

Skoraks:
@AaronGullickson - Thanks for the heads up, as for the rules I have just added a GM mode only option to charge the drives fully at the moment, it's a cheat and not using any game mechanics.

inqistor:
> This should wait until after PR #1280 because it will conflict with the refactoring of planets into planetary systems.
> 
> Also how are the rules applied to recharge the jump drive? Recharging is not instantaneous. These rules are covered on pg. 87 of _Strategic Operations_. Whenever possible we should follow the rules of the book.

It's for simulation of changing Jump Ships. Most Mercs don't have organic Jump Ships, so no reason to stick to the same one for whole travel. Technically, it should be easier to find free collar on Jump Ship jumping to one system along your path, that finding Jump Ship which will take you thorough next 6 jump points, and probably cross States Border at least once.
Anyway, IIRC charging from Recharging Stations is also not implemented. Everything is still based on Star spectrum.

AaronGullickson:
Yes, I think that makes sense. The system refactoring is now in and I don't think this will actually conflict with that, but it does it looks like there are conflicts with some of the recent changes to the InterstellarMapPanel.

sixlettervariables:
#1359 brings this up to speed.

Issue #1282, opened 2019-08-31
Unit non-cash asset value increases when transferring campaign from v0.44 stable to v0.46 stable


Bluespacer:
### Environment
MekHQ 0.46 (campaign migrated from 0.44)
Windows 10, Java version 1.8.0_221

### Description
After loading the campaign file in 0.46, the financial value of the unit's non-cash assets has increased significantly (new value between 2x-3x original amount). The amount increase is not the same with all Meks, for example. There seems to be some correlation (but not full) to unit quality rating.
This also affects contract values immediately after load.

There was some change to campaign options after load, but none that I thought would be relevant.
[Campaign.zip]

### Files
(https://github.com/MegaMek/mekhq/files/3562303/Campaign.zip)

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3562295/mekhqlog.txt)

![UnitAssetSize](https://user-images.githubusercontent.com/54739072/64062194-5f1d8000-cbec-11e9-9825-03f9b0027b76.jpg)

bczretribution:
I wonder if this is at all related to an issue I have been noticing in one of my campaigns.  I have noticed multiple units whose value seems to increment upwards almost daily, and over time they have grow to be as much as 2x or 3x the original buy value of the mech, far surpassing what their value should be even if they were 100% max quality.  I will try to get pictures and campaign files for you soon.

CaliburxZero:
![marauder value](https://user-images.githubusercontent.com/53108929/64581557-a1675f80-d358-11e9-9630-52e06a0025c0.png)
![Goshawk value](https://user-images.githubusercontent.com/53108929/64581558-a1675f80-d358-11e9-847a-6f4871a10fb8.png)
These are two examples of the value from bcz's campaign as promised.  There are many examples in this campaign of grossly-inflated value many mechs, although this is not consistent with all units.

PR #1292, opened 2019-09-20
Fix unit value explosions #1282

sixlettervariables:
You can either price parts by quantity or quality, but you're certainly going to pay more for quality.

Original line:
https://github.com/MegaMek/mekhq/blame/089d23710eff787b3d834c959dc5328bfe8155c0/MekHQ/src/mekhq/campaign/unit/Unit.java#L1118

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1141,7 +1141,7 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
 
         partsValue = partsValue
                 .plus(parts.stream()
-                    .map(x -> x.getActualValue().multipliedBy(x.getQuality()))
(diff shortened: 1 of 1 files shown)


NickAragua:
This is the second time this week I've had the opportunity to make this joke: Quantity has a quality of its own.

Bluespacer:
@sixlettervariables Thanks for the fix! Would it be possible to apply this to stable as well? (say, 0.46.2) 

I upgraded from 0.44 to 0.46 (I only play stables for the moment), but specifically this bug caused me to downgrade back to 0.44. This bug affected not just the unit value, but also contract value massively. This means it makes the strategic game too easy to play, hence why I downgraded...

neoancient:
Yes, when we do the next dev release packed with a bunch of new goodies we'll put out 0.46.1 as a bugfix-only update to 0.46.0.

PR #1300, opened 2019-09-30
Markdown for descriptions

AaronGullickson:
This PR changes all of the various descriptive elements in the views (person, unit, force, mission, scenario) to a JTextPane that parses them for the [commonmark](https://commonmark.org/) markdown syntax and then displays the resulting html. This allows players to easily create very nice looking descriptions without having to do raw html coding. There is no built-in markdown editor (yet) but markdown is so dead simple that many players will be able to pick it up. 

Here is a simple example using the mission description. 

Screen Shot 2019-09-30 at 1 56 29 PM

This is fairly basic but this has all of the features of markdown including embedded images, lists, headings, etc.

One of the nice features is that markdown also accepts raw html markup, so users *can* use html if they want. So this allows power users to really go crazy. 

The change requires the [commonmark-java library](https://github.com/atlassian/commonmark-java). I couldn't find a JAR release of it on their GH site, so I got one [here](https://jar-download.com/artifacts/com.atlassian.commonmark/commonmark). I added it to the lib directory and then to the classpath. Let me know if there is a different way to do this now.


MekHQ/src/mekhq/gui/view/ForceViewPanel.java
```diff
@@ -71,6 +76,8 @@
 	private javax.swing.JLabel lblTech1;
 	private javax.swing.JLabel lblTech2;
 	
+	private static Parser parser = Parser.builder().build();
```
AaronGullickson: I wasn't sure of the best way to do this. This the parser for the commonmark library. The example creates it dynamically, but that seemed like a waste to do every time this is called so I created it as a static variable in each View Panel. I would appreciate feedback about whether this is a smart way to do it.

AaronGullickson:
I should also note that I held off on the PlanetViewPanel because I don't want to conflict with anything from #1280 .

MekHQ/src/mekhq/gui/view/ForceViewPanel.java
AaronGullickson: Ok following @sixlettervariables advice I have moved the rendering into its own static class which calls `getRenderedHTML`. This will make it easier in the future if we want to extend the way we do rendering.

MekHQ/.classpath
```diff
@@ -20,5 +20,6 @@
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8/"/>
 	<classpathentry kind="con" path="org.eclipse.buildship.core.gradleclasspathcontainer"/>
+	<classpathentry kind="lib" path="lib/commonmark-0.13.0.jar"/>
```
sixlettervariables: @neoancient would know how to integrate this with our Gradle setup.

MekHQ/src/mekhq/gui/view/MissionViewPanel.java
```diff
@@ -147,10 +155,11 @@ private void fillStats() {
...
 		
 		txtDesc.setName("txtDesc");
-		txtDesc.setText(mission.getDescription());
 		txtDesc.setEditable(false);
-		txtDesc.setLineWrap(true);
-		txtDesc.setWrapStyleWord(true);
+		//render the markdown enabled text
+        Node document = parser.parse(mission.getDescription());
+        txtDesc.setContentType("text/html");
+        txtDesc.setText("<html>" + renderer.render(document));
```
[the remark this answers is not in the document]
sixlettervariables: Also the whitespace is off.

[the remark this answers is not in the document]
AaronGullickson: on the html ending. Apparently not, but I also discovered that it does not require the "<html>" at the beginning, so i now removed it altogether and it works fine.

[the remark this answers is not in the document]
AaronGullickson: Whitespace looks fine in eclipse. Its all tabs but thats from before.

MekHQ/.classpath
AaronGullickson: and with 3f024d8, I have changed things to do it the gradle way.

PR #1298, opened 2019-09-29
Restructuring of person view panel, double scroll issue, and other scrolling issues

AaronGullickson:
OK, this PR began simple enough as an attempt to deal with the double scroll bars on personnel and unit views in the TO&E, visible here: 

Screen Shot 2019-09-29 at 2 23 30 PM

I was able to address that issue so that there is only one scroll bar regardless of which view you have (force, person, unit). However, this led into some issues with the PersonViewPanel. Basically, as we have increased the amount of information kept there, the aesthetics of it have gone to crap. So, I re-organized things a bit and put a bunch of items into their own separate "boxes". Here is what it looks like now with pretty much all the options turned on:

Screen Shot 2019-09-29 at 10 55 26 AM

With minimal options, it still looks nice:

Screen Shot 2019-09-29 at 2 29 21 PM

However, in fixing this view, I noticed that in many cases the borders being drawn indicated that the borders were going off the edge of the view to the right. This has been a problem for a long time. The issue is that the JTextArea is not recognizing that it needs to shrink itself to allow for the vertical scroll bar. That apparently is because we have been putting JPanels in the scroll panes but JPanels do not implement the Scrollable interface. So I wrote up a ScrollablePanel extension and made all of the ViewPanels in the base GUI (but not the dialogs) use it. That fixed the problem and everything seems to be playing nicely now. 

This should be ready to go, pending review.

PR #1303, opened 2019-10-03
Various GUI fixes

AaronGullickson:
tl;dr - I fixed a bunch of small GUI annoyances mostly in the View panels. However, please scroll down to the bottom and read about two places I couldn't figure out fixes.

This PR contains a variety of small GUI fixes that have been bugging me for some time. Individually they are not much but add up to a substantial enhancement. Here are the things changed in this PR

### Asymetric Insets

Many of the views added components with an inset of 5, 5, 5, 20 where the 20 was my original failed hack for dealing with the issues around scrollbars finally addressed by PR #1298. So I have removed this inset from all views and switched them to 5,5,5,5 insets. 

### Empty white space around text descriptions

All of the various text descriptions in the View panels had an excess of white space through a variety of insets and compound borders. I have reduced that white space to something more reasonable and made it consistent across all views.

### No more JTextAreas for title: info pairs

In a variety of places in the View Panels we have a pair of components that define a title and a value (e.g. Name: Bob). The title is always defined as a JLabel but the value is sometimes defined using labels and sometimes by JTextAreas. In many cases, this looks similar but in some themes the JTextAreas look bad, as for example in Nimbus:

Screen Shot 2019-10-03 at 12 56 13 PM

In addition, a JTextArea is really overkill for the items here. So I changed all values to be represented with a JLabel which looks much nicer:

Screen Shot 2019-10-03 at 12 57 44 PM

### Word wrapping in title JLabels

Related to the point above, in some cases, the JLabel for the title was being word wrapped because it was defined as html to make it bold. I put a <nobr> tag on all JLabel text to prevent word wrapping on all of these title JLabels.

### Don't show empty items in Views

In some cases, items were shown in the views even when the user had not entered anything for them (e.g. Mission Type in MissionViewPanel). I have put checks in that prevent these from showing if they are null or empty. 

### Tabs and whitespace in Markdown

If a user enters a tab or four or more white spaces in the markdown enabled text, that will now create formatting problems, because Markdown will interpret that as a blockquote, I believe. In any case, it makes the rendering go to shit. So I put a check for these things using regular expressions into the MarkdownRenderer and replace them with two new lines so they conform to what markdown expects for paragraph breaks. 

### Keep unit scrollbar at the top in repair bay

This has been a major source of annoyance for me. Whenever you repair an item on a unit, the scrollbar for the unit goes to the bottom which can often prevent you from seeing what was just fixed. This will now keep the scrollbar at the top, where most of the fixable items will be, so that as you repair stuff you will be able to see it change on the unit without having to constantly re-scroll. 

The scrollbar starting in weird places rather than the top used to be a problem for a variety of things but should already be fixed for most other things now with previous PRs. The only remaining place where I see this happening is for the editors used in dialogs to write descriptions. However, I am working on a separate PR for markdown rich text editor that will resolve this issue. 

### Daily log improvements

In the nimbus theme, the daily log doesn't look great because of a big margin colored grey between the scrollbar and the edge of the titled panel: 

Screen Shot 2019-10-03 at 1 10 47 PM

I fixed this by removing the titled border on the panel and instead using a JLabel to do the title. I think the results look good:

Screen Shot 2019-10-03 at 1 12 53 PM

its worth noting that this is a general problem with embedding scrollpanes directly in panels with a title border because Nimbus uses this kind of cool "depression" to draw the border. 

### Markdown aware descriptions for some views

In my previous PR #1300, I added markdown aware text for the various text descriptions that show up in the Views. However, I neglected that for the contract view panel, the AtB contract view panel, and the planet view panel. So, I have added them to this PR. 

### ScrollablePanel added to AtB views

In my previous PR #1298, I change all the View JPanels to a custom ScrollablePanel that implements Scrollable and resolves some GUI issues with the Views, but I neglected to add the AtB mission and scenario views, which I have done here.

### Changed title of personnel description

The title of the description for personnel says "Description" which seems odd, so I changed it to "Biography". 

### Two things I haven't been able to fix:

There are two more things that I wanted to fix but I am stuck on. Any suggestions would be appreciated:

#### Extra line in descriptions

Screen Shot 2019-10-03 at 1 16 24 PM

As you can see in the screenshot above, there is an extra empty line at the top of the description. This is happening in all of the markdown enabled descriptions and seems to have something to do with the HTML, as I don't get the problem with plain text. The first line always starts with a `<p>` from the rendering so I am guessing that has to do with it, but I can't figure out a way to solve it.

#### Unknown grey area

Screen Shot 2019-10-03 at 1 19 36 PM

In some themes, when a view or table is too small for the component, the remainder is filled up with grey space, as can be seen above. This doesn't happen in my default Mac OS  theme, but I see it in Metal and Nimbus. I would like all of that space to be white to match the respective components. However, I cannot figure out what component that color is coming from. Its not from the tabbed pane or the scrollbars near as I can figure.


MekHQ/src/mekhq/gui/utilities/MarkdownRenderer.java
```diff
@@ -56,6 +56,18 @@ public static String getRenderedHtml(String input) {
...
+         * of them as block quotes. However, many users will probably
+         * use them to identify beginning of paragraphs. So replace all
+         * tabs with line returns. Ultimately, if we ever get up a markdown
+         * editor we should also check for tabs there and correct them in the 
+         * input
+         */
+        input = input.replaceAll("\\t+", "\n\n");
+        //This is also going to happen if indentation is used with spaces
+        input = input.replaceAll("\\s\\s\\s\\s+", "\n\n");
+
```
sixlettervariables: https://github.com/atlassian/commonmark-java/blob/bf8bea9dfcb49ee43e20372b42b43727d9a15da2/commonmark-test-util/src/main/resources/spec.txt#L340

This spec seems to say these are treated as indentation rather than blockquotes (a la `>`). I think it would be a good idea to leave this out and see what sort of "problems" show up in folks files.

sixlettervariables: Or, you could remove the `IndentedCodeBlock` from the set given to the parser builder:

https://static.javadoc.io/com.atlassian.commonmark/commonmark/0.13.0/org/commonmark/parser/Parser.Builder.html#enabledBlockTypes-java.util.Set-

AaronGullickson: As far as what happens is that it basically breaks the rendering. The text becomes much smaller and fails to wrap.

The issue is users who use indentation to begin new paragraphs without an empty line in between, which is still a common way to write. If we do nothing, the rendering will make it go haywire. If we removed the IndentedCodeBlock then I think all of their paragraphs will run together. I will test it to make sure. I have tested the current solution and it may not be elegant but it definitely works.

AaronGullickson: Ok, in b7b10d5ce, I changed the code to remove the regex substitions and use the enableBlockTypes method in the parser builder to remove potentially problematic block types.

The one down side of this approach is that users who use indentation to write paragraphs will have everything smooshed into one giant paragraph, but they will probably figure it out quickly enough. We can keep an eye on what users say.

neilfw:
Just a note... as a non developer who closely follows this project and the changes, I find Aaron's extensive update and explanation above incredibly valuable and useful.  Thank you!

AaronGullickson:
So, I figured out a solution to the grey area problem discussed above. The issue is that the viewport of the `JScrollPane` needed to have its background changed not the JScrollPane itself. I have made all of these changes in commit ff94688. 

I also realized I forgot to make all of the other changes to view panels to the JumpPathPanel view so I have updated that one in commit 84266a4. 

There is still the issue of the extra space in the descriptions, but I think that can wait for another day. So ready for this to be fully reviewed.

MekHQ/src/mekhq/gui/BriefingTab.java
```diff
@@ -307,13 +310,14 @@ public void initTab() {
         panScenarioButtons.add(btnClearAssignedUnits);
 
         scrollScenarioView = new JScrollPane();
+        scrollScenarioView.getViewport().setBackground(Color.WHITE);
         scrollScenarioView.setViewportView(null);
-        scrollScenarioView.setMinimumSize(new java.awt.Dimension(450, 600));
-        scrollScenarioView.setPreferredSize(new java.awt.Dimension(450, 600));
+        //scrollScenarioView.setPreferredSize(new java.awt.Dimension(450, 600));
```
NickAragua: Let's get rid of this commented out code, and then we're good to go.
```suggestion
```

PR #1305, opened 2019-10-04
Mac OSX key bindings on non-standard look and feel

AaronGullickson:
For mac users, when using any of the other look and feel themes besides the standard Mac OSX look and feel, the standard key bindings for cut, copy, paste, and select all with the command key do not work, and users have to instead use the control keys which is not natural for many Mac users (including me). This PR re-introduces those key bindings so Mac users can utilize the more familiar command keys when using other themes like Nimbus or Metal. It still retains the control key bindings as well. 

The code for the PR comes from [this StackOverflow thread](https://stackoverflow.com/questions/9780028/mac-keyboard-shortcuts-with-nimbus-laf).

### DIFF
--- a/MekHQ/src/mekhq/MekHQ.java
+++ b/MekHQ/src/mekhq/MekHQ.java
@@ -22,6 +22,7 @@
 package mekhq;
 
 import java.awt.*;
+import java.awt.event.KeyEvent;
 import java.beans.PropertyChangeEvent;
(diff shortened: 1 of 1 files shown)


AaronGullickson:
damn it. Branched off the wrong branch again. One moment...

AaronGullickson:
Ok, branching off the master now. Ready for review.

PR #1315, opened 2019-10-23
Rescalable images for unit view panel when fluff image exists

AaronGullickson:
On the twin premises that:

1. Battletech is first and foremost about mech porn
2. Players are so into mech images (see premise 1) that they are willing to commission [great works of art](https://bg.battletech.com/forums/index.php?topic=37902.0;topicseen) to see their personal units in action

I have therefore concluded that our UnitViewPanel could use some work to better show off fluff images when players have them. In order to do that I have moved things around a little bit when a fluff image exists and put an image which automatically adjusts to fit the space (up to its size) when it has the room. Here is an example of it in action;

Screen Shot 2019-10-23 at 9 57 39 AM

In order to do this, I created a custom ImgLabel.java object that uses the paintComponent method to paint a properly sized image. This could be used in other places as well. There is already a method similar to this in ImageUtils for rescaling images but (a) it uses getScaledImage which I have been reading is not a great way to do this, and (b) I couldn't get it to work right. 

I also increased the default of the unit view scrollbar in the hangar tab to accomodate the larger images from the get-go. 

If the player does not have a fluff image the display looks more or less as it did before except that I moved the image to the right of the text:

Screen Shot 2019-10-23 at 2 45 05 PM

### DIFF
--- a/MekHQ/src/mekhq/gui/view/UnitViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/UnitViewPanel.java
@@ -8,13 +8,15 @@ package mekhq.gui.view;
 
 import java.awt.Color;
 import java.awt.Component;
+import java.awt.Dimension;
 import java.awt.Font;
+import java.awt.Graphics;
 import java.awt.Image;
 import java.util.ResourceBundle;
 
 import javax.swing.BorderFactory;
-import javax.swing.Icon;
 import javax.swing.ImageIcon;
+import javax.swing.JLabel;
 
 import megamek.client.ui.swing.MechTileset;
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/view/UnitViewPanel.java
```diff
@@ -331,4 +347,48 @@ private Image getCamo(Unit unit) {
         }
         return camo;
     }
+    
+ /*   private class ImgLabel extends JLabel {
```
sixlettervariables: Good to delete this?

PR #1321, opened 2019-10-31
Scenario Template polish

NickAragua:
A few relatively small changes and features:
- human-readable xml format for scenario template definitions (will make a pass through the rest later)
- UI to assign forces/units explicitly from the TO&E screen to specific force templates defined in a scenario template (e.g. "Force A is going to be Main Player Force, Force B is going to be Reinforcements")
- Set deployment zone and turn for individual units as well as forces for dynamic scenarios if possible
- default "reinforcements" force template applied to all dynamic scenarios if one does not already exist
- scenario modifiers are sorted alphabetically where they are displayed


MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
```diff
@@ -504,6 +506,18 @@ public boolean isEnemyBotForce() {
                 getForceAlignment() == ForceAlignment.Third.ordinal(); 
     }
     
+    /**
+     * Convenience function that returns the displayable name of the selected unit type.
+     * @return
+     */
+    public String getAllowedUnitTypeName() {
+        if(getAllowedUnitType() > UnitType.SIZE || getAllowedUnitType() < 0) {
```
neoancient: This should be >= UnitType.SIZE

MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
```diff
@@ -504,6 +506,18 @@ public boolean isEnemyBotForce() {
...
+            return SPECIAL_UNIT_TYPES.get(getAllowedUnitType());
+        } else {
+            return UnitType.getTypeDisplayableName(getAllowedUnitType());
+        }
+    }
+    
     /**
      * Attempt to deserialize an instance of a ScenarioTemplate from the passed-in XML Node
      * @param inputFile The source file
```
neoancient: Not part of your current changes, but the documented parameter does not match the actual parameter. The same is true of the deserialize method a few lines below this, which does not appear in this changelog.

MekHQ/src/mekhq/campaign/mission/ScenarioTemplate.java
```diff
@@ -119,6 +119,7 @@ public void Serialize(File outputFile) {
             JAXBContext context = JAXBContext.newInstance(ScenarioTemplate.class);
             JAXBElement<ScenarioTemplate> templateElement = new JAXBElement<>(new QName(ROOT_XML_ELEMENT_NAME), ScenarioTemplate.class, this);
             Marshaller m = context.createMarshaller();
+            m.setProperty("jaxb.formatted.output", true);
```
neoancient: The constant you used elsewhere can be used here also.

MekHQ/src/mekhq/gui/dialog/ForceTemplateAssignmentDialog.java
```diff
@@ -0,0 +1,291 @@
...
+    private AtBDynamicScenario currentScenario;
+    private Vector<Force> currentForceVector;
+    private Vector<Unit> currentUnitVector;
+    private CampaignGUI campaignGUI;
+    
+    public ForceTemplateAssignmentDialog(CampaignGUI gui, Vector<Force> assignedForces, Vector<Unit> assignedUnits, AtBDynamicScenario scenario) {
+        currentForceVector = assignedForces;
+        currentUnitVector = assignedUnits;
+
+        //resourceMap = ResourceBundle.getBundle("mekhq.resources.ForceTemplateAssignmentDialog", new EncodeControl());
```
neoancient: Did you intend to add a string resources file?

MekHQ/src/mekhq/gui/dialog/ForceTemplateAssignmentDialog.java
```diff
@@ -0,0 +1,291 @@
...
+            Force force = (Force) value;
+            String cellValue = currentScenario.getPlayerForceTemplates().containsKey(force.getId()) ?
+                    String.format("%s (%s)", force.getName(), currentScenario.getPlayerForceTemplates().get(force.getId()).getForceName()) :
+                    force.getName();
+            ((JLabel) cmp).setText(cellValue);
+            return cmp;
+        }
+    }
+    
+    private class TemplateListCellRenderer extends DefaultListCellRenderer {
```
neoancient: This inner class does not reference an instance of the enclosing class and can be made static to take less memory and avoid a possible source of memory leaks.

NickAragua:
Addressed raised issues. UI looks like this now:

image

sixlettervariables:
> "Assign your selected **_or_** units..."

Is that or intended?

NickAragua:
> > "Assign your selected **_or_** units..."
> 
> Is that or intended?

Updated. Should be "forces or units"

PR #1325, opened 2019-11-06
Customize protomechs and advanced aero

neoancient:
When I was adding SV customization support to the MekLab tab I noticed that I had never done it for protomechs or advanced aerospace units. So here it is. I also added the fluff tab, and a preview tab for combat and support vehicles. Protomechs required some adjustments to the armor part to distinguish between standard and EDP.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/PartsStore.java
+++ b/MekHQ/src/mekhq/campaign/market/PartsStore.java
@@ -110,7 +110,7 @@ public class PartsStore implements Serializable {
 	private static final long serialVersionUID = 1686222527383868364L;
 
 	private static int EXPECTED_SIZE = 50000;
-	
+
 	private ArrayList<Part> parts;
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/campaign/market/PartsStore.java
```diff
@@ -511,7 +511,8 @@ private void stockArmor(Campaign c) {
 		/*
 		 * Protomek Armor
 		 */
-		parts.add(new ProtomekArmor(0, 100, -1, true, c));
+        parts.add(new ProtomekArmor(0, EquipmentType.T_ARMOR_STANDARD, 100, -1, true, c));
```
sixlettervariables: whitespace is a touch off

neoancient: That file had a mix of tabs and spaces. I fixed the indents.

Issue #1330, opened 2019-11-10
Pregnant people cause a NPE when selecting them in Personnel tab


MarcJC:
Playing new 47.1 HQ 
I found that when trying to select various people in the personnel tab, they would fail to load and cause a NPE error in the log.  Upon further research it seems that the link between all the personnel who are failing to load is that they are pregnant.  
[code]
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at mekhq.gui.view.PersonViewPanel.fillInfo(PersonViewPanel.java:608)
	at mekhq.gui.view.PersonViewPanel.initComponents(PersonViewPanel.java:166)
	at mekhq.gui.view.PersonViewPanel.<init>(PersonViewPanel.java:125)
	at mekhq.gui.PersonnelTab.refreshPersonnelView(PersonnelTab.java:858)
	at mekhq.gui.PersonnelTab.lambda$initTab$2(PersonnelTab.java:270)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(Unknown Source)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(Unknown Source)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(Unknown Source)
	at javax.swing.DefaultListSelectionModel.changeSelection(Unknown Source)
	at javax.swing.DefaultListSelectionModel.changeSelection(Unknown Source)
	at javax.swing.DefaultListSelectionModel.setSelectionInterval(Unknown Source)
	at javax.swing.JTable.changeSelectionModel(Unknown Source)
	at javax.swing.JTable.changeSelection(Unknown Source)
	at javax.swing.plaf.basic.BasicTableUI$Actions.actionPerformed(Unknown Source)
	at javax.swing.SwingUtilities.notifyAction(Unknown Source)
	at javax.swing.JComponent.processKeyBinding(Unknown Source)
	at javax.swing.JTable.processKeyBinding(Unknown Source)
	at javax.swing.JComponent.processKeyBindings(Unknown Source)
	at javax.swing.JComponent.processKeyEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
[...]

PR #1331, opened 2019-11-10
Fix NPE when PersonViewPanel has a pregnant individual #1330

sixlettervariables:
Fixes the NPE in PersonViewPanel by bringing all JPanel/JLabel creations local.

### DIFF
--- a/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
@@ -73,47 +73,12 @@ public class PersonViewPanel extends ScrollablePanel {
     private JPanel pnlLog;
     private JPanel pnlMissionsLog;
     private JPanel pnlInjuries;
-
-    private JLabel lblType;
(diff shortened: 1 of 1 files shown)


NickAragua:
Could you point out where the actual NPE fix is? I'm having trouble finding it for some reason.

MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -612,7 +572,7 @@ private void fillInfo() {
             gridBagConstraints.gridy = firsty;
             gridBagConstraints.fill = GridBagConstraints.NONE;
             gridBagConstraints.anchor = GridBagConstraints.NORTHWEST;
-            pnlFamily.add(lblDuedate1, gridBagConstraints);
+            pnlInfo.add(lblDuedate1, gridBagConstraints);
```
sixlettervariables: @NickAragua at this point lblDuedate1 was null (well above here).

PR #1335, opened 2019-11-11
Sort SVs into unit type and weight class when adding to force on TOE.

### LINKED ISSUE #1333: Simca Ambulance possibly causing campaign to not load (NPE with Simca Ambulance?)
### Environment
0.47.1
Windows 10

### Description
A unit I am trying to use in my campaign is likely causing an NPE when a campaign is attempting to load, after livestreaming a session with Megamek, I was purchasing some unit to create an SAR unit, including the Simca Ambulance.  I have also noticed after putting a full compliment of crew when putting the Simca Ambulance in the TO&E, The Ambulance happens to be an unspecified vehicle.  After purchasing it for the campaign (or adding it in GM mode), any time I attempted to save the campaign, then load it back up, it brings up a Campaign can't be loaded error.  

### Files
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3829720/mekhqlog.txt)

[Stormbringing Destoryers Bugged.zip](https://github.com/MegaMek/mekhq/files/3829722/Stormbringing.Destoryers.Bugged.zip)

neoancient:
When adding a unit to a force on the TO&E, support vehicles show up in the "Unsorted" catch-all category because the sorting by unit type and weight doesn't take support vehicle weight classes into account. This adds special handling for SVs so they will show up in the tank, VTOL, naval, or conventional fighter menus by their correct weight class.

Fixes #1333

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
@@ -3,12 +3,7 @@ package mekhq.gui.adapter;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.MouseEvent;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.StringTokenizer;
-import java.util.UUID;
-import java.util.Vector;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import javax.swing.JMenu;
@@ -141,7 +136,7 @@ public class TOEMouseAdapter extends MouseInputAdapter implements ActionListener
                                     if (null != u.getTech()) {
                                         u.removeTech();
                                     }
-                                    
+
                                     u.setTech(tech);
                                 } else {
                                     cantTech += tech.getName() + " cannot maintain " + u.getName() + "\n";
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -810,13 +806,25 @@ private void maybeShowPopup(MouseEvent e) {
                             weightClassForUnitType.get(weightClassMenuName).setEnabled(false);
                         }
                     }
-                    
+
+                    for (int wc = EntityWeightClass.WEIGHT_SMALL_SUPPORT; wc <= EntityWeightClass.WEIGHT_LARGE_SUPPORT; wc++) {
+                        for (int ut : svTypes) {
+                            String typeName = UnitType.getTypeName(ut);
+                            String wcName = EntityWeightClass.getClassName(wc, typeName, true);
+                            String menuName = typeName + "_" + wcName;
```
NickAragua: Does the compiler get it, or is it more efficient to use string.format for string concatenation?

neoancient: I used the same approach that's used in the surrounding code. Internally concatenation uses StringBuilder, which is significantly more efficient than String::format. String::format also introduces the possibility of an IllegalArgumentException at runtime, though a single test run will show it to work correctly in this case. I tend to use String::format when the format string itself is pulled from a resource file or in cases when it makes the code more readable and it's worth losing a nanosecond or so.

NickAragua: Gotcha, guess I have C# on my brain.

PR #1348, opened 2019-11-19
Scenario objectives - Part III

NickAragua:
Changes relating to scenario objectives, mostly relevant for template scenarios:

- Allow a scenario template modifier to add objectives to a scenario (e.g. a hostile facility may become a capture, an extraction, or whatever)
- Allow a force template to attach to other force templates for the purposes of objectives (e.g. when you add "enemy reinforcements", and there's already an objective to destroy the "primary opfor", the reinforcements become part of that objective as well)
- The rest of the scenario template files are human-readable
- Scenario modifiers that are not applicable at the current state of a template scenario are not displayed in the customize scenario UI
- Corrections to some bugs in objective completion determination and force parameter generation


NickAragua:
Just a quick update to utilize the correct calculation for arrival turn that's actually dependent on the scenario template.

PR #1355, opened 2019-11-24
General support vehicle crew

neoancient:
Adds a T_VEHICLE_CREW role for support vehicle crew other than the driver/pilot and gunner(s). This allows the unit to be fully crewed so that it can be deployed. To keep things simple there is no distinction made between different vehicle types. In order to have an experience level there needs to be a skill requirement, so I used mechanics as the most reasonable skill.

This does not enforce the officer requirement, but neither do aerospace vessels at this point. A possible future refinement is to allow crewing mobile hospital units with doctors/medics, but I leave that for another time.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -2547,6 +2552,7 @@ private void parseRoleForDesignation(int role, StringBuilder sb) {
             case T_VTOL_PILOT:
             case T_VEE_GUNNER:
             case T_CONV_PILOT:
+            case T_VEHICLE_CREW:
                 sb.append("Lambda");
```
NickAragua: I gotta ask, any idea what the deal is with this?

sixlettervariables: ComStar designations I thought.

Issue #1367, opened 2019-12-18
Sol's Spectral Classification is Incorrect


Windchild292:
MekHQ 0.47.2 and 0.46.0 have Sol's spectral classification as a G9V, while Sol is a G2V star.

PR #1368, opened 2019-12-18
1367: Fixed Sol's Spectral Class

Windchild292:
Fixing #1367, which changed Sol's Spectral Class to G2V from its erroneous G9V.


NickAragua:
Thanks, we appreciate the help.

PR #1378, opened 2019-12-31
Unit Rating: Varied Minor Fixes

### LINKED ISSUE #1100: Overview -> Unit Rating does not take into account small craft Infantry Bays
### Environment
0.44.0
Windows

### Description
When using small craft to hold your infantry platoons, while docked in your dropship, the Overview -> Unit Rating, does not take into account these bays and gives the unit a penalty.

Overview -> Transport Breakdown
Infantry Bays (Occupied):                   6 (   6)
Small Craft Bays (Occupied):                2 (   2)

Overview -> Unit Rating
Small Craft Bays:            2 needed /   2 available
Infantry Bays:               7 needed /   0 available

The unit rating also seems to think I need one more bay, than I actually have platoons for some reason.

### Files
Attached the campaign and custom unit files.

[Morkath's Marauders_FullCustom_Company.zip](https://github.com/MegaMek/mekhq/files/2890291/Morkath.s.Marauders_FullCustom_Company.zip)
[CustomUnits.zip](https://github.com/MegaMek/mekhq/files/2890288/CustomUnits.zip)

Windchild292:
This fixes #1100.

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
+++ b/MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
@@ -27,20 +27,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import megamek.common.ASFBay;
-import megamek.common.BattleArmor;
-import megamek.common.BattleArmorBay;
-import megamek.common.Bay;
-import megamek.common.Dropship;
-import megamek.common.Entity;
-import megamek.common.HeavyVehicleBay;
-import megamek.common.Infantry;
-import megamek.common.InfantryBay;
-import megamek.common.Jumpship;
-import megamek.common.LightVehicleBay;
-import megamek.common.MechBay;
-import megamek.common.SmallCraftBay;
-import megamek.common.UnitType;
+import megamek.common.*;
 import megamek.common.logging.LogLevel;
 import megamek.common.logging.MMLogger;
 import mekhq.MekHQ;
@@ -49,6 +36,7 @@ import mekhq.campaign.mission.Mission;
 import mekhq.campaign.personnel.Person;
 import mekhq.campaign.personnel.Skill;
 import mekhq.campaign.unit.Unit;
+import sun.jvm.hotspot.gc.shared.Space;
 
 /**
  * @author Deric Page (deric (dot) page (at) usa.net)
@@ -95,6 +83,7 @@ public abstract class AbstractUnitRating implements IUnitRating {
     private int smallCraftBayCount = 0;
     private int lightVeeBayCount = 0;
     private int heavyVeeBayCount = 0;
+    private int superHeavyVeeBayCount = 0;
     private int baBayCount = 0;
     private int infantryBayCount = 0;
     private int dockingCollarCount = 0;
@@ -402,14 +391,14 @@ public abstract class AbstractUnitRating implements IUnitRating {
      */
     BigDecimal getUnitValue(Unit u) {
         BigDecimal value = BigDecimal.ONE;
-        if (isConventionalInfanry(u) &&
+        if (isConventionalInfantry(u) &&
             (((Infantry) u.getEntity()).getSquadN() == 1)) {
             value = new BigDecimal("0.25");
         }
         return value;
     }
 
-    boolean isConventionalInfanry(Unit u) {
+    boolean isConventionalInfantry(Unit u) {
         return (u.getEntity() instanceof Infantry) &&
                !(u.getEntity() instanceof BattleArmor);
     }
@@ -480,6 +469,7 @@ public abstract class AbstractUnitRating implements IUnitRating {
         setBaBayCount(0);
         setLightVeeBayCount(0);
         setHeavyVeeBayCount(0);
+        setSuperHeavyVeeBayCount(0);
         setInfantryBayCount(0);
         setDockingCollarCount(0);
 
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
```diff
@@ -49,6 +36,7 @@
 import mekhq.campaign.personnel.Person;
 import mekhq.campaign.personnel.Skill;
 import mekhq.campaign.unit.Unit;
+import sun.jvm.hotspot.gc.shared.Space;
```
sixlettervariables: Is this required?

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
```diff
@@ -498,28 +488,48 @@ protected void initValues() {
...
+                } else if (bay instanceof ASFBay) {
+                    setFighterBayCount(getFighterBayCount() + (int) bay.getCapacity());
+                } else if (bay instanceof SmallCraftBay) {
+                    setSmallCraftBayCount(getSmallCraftBayCount() + (int) bay.getCapacity());
+                }
+            }
+        }
+    }
+
+/*    private void updateBayCount(Dropship ds) {
```
sixlettervariables: Do we need to keep the commented out code?

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
Windchild292: I have no idea how that got added... interesting...

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
Windchild292: Nope, just forgot to remove it

Windchild292:
It appears my IDE has decided to make some change to the code automatically, without even notifying me.

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
Windchild292: I wonder if that caused the build failures?

Windchild292: I've built it without it, so nope. I've reverted the automatic changes from my IDE.

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
```diff
@@ -498,48 +502,26 @@ protected void initValues() {
...
-
-    private void updateBayCount(Jumpship jumpship) {
-        for (Bay bay : jumpship.getTransportBays()) {
-            if (bay instanceof ASFBay) {
-                setFighterBayCount(getFighterBayCount() +
-                                   (int) bay.getCapacity());
-            } else if (bay instanceof SmallCraftBay) {
-                setSmallCraftBayCount(getSmallCraftBayCount() +
-                                      (int) bay.getCapacity());
+        if ((e instanceof Jumpship) || (e instanceof SmallCraft)) {
```
sixlettervariables: `Entity::getTransportBays` always returns a non-null Vector. I wonder if this `if` test can be dropped to simplify the code (and perhaps future proof it). Thoughts @neoancient @mkerensky @NickAragua ?

neoancient: Unless I missed something, this count is to determine the unit capacity for interstellar transport, so it should only be checking Jumpships, Warships, and Dropships. While small craft mech/aero/vehicle bays are impractical, they can have infantry bays that should not count toward transport, since the infantry are not going to be transported in an infantry bay of a small craft inside a small craft bay. In the near future when large naval and airship support vehicles are implemented, it would also be possible for them to have a variety of bays that also should not count toward transport capacity. `if (e.isLargeCraft())` would include dropships, jumpships, and warships. But it would also include space stations, which also shouldn't count toward transport capacity.

Windchild292:
Will fix those probably tomorrow @neoancient.

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
Windchild292: You could easily have people living in small craft bays, although I wouldn't recommend it. Is there be a way to put a morale penalty on the unit inside if they live there long-term?
Also, no matter how this is programmed Space Stations will be included, and there is no easily way that I've found to remove them from the count.

Windchild292: And also, removing space stations from transport capacity opens up a whole bag of worms. First of all, why does a mercenary company have a space station? Is it because it is a merc company base? In that case, we definitely need to count it, for base requirements (given how you can easily simulate having a base with the splitting force mechanics) should show any bays the company has.

neoancient: I was simply responding to @sixlettervariables question about whether I thought the if statement was necessary, and I believe it is. I don't intend to get pulled into a debate about how it should work. But in response to your objections:
1. You can suggest all sorts of ways to transport infantry in a make-shift way. This should be kept simple. If you want to claim that infantry can live in a bay in a docked small craft you would then need to make sure the unit has transport capacity for that small craft. That's an unnecessary complication.
2. Space stations can easily be excluded from the bay count in the if statement in question.
3. MekHQ works for any force, not just mercenaries. FMM is designed for merc units but the CO rating is for any unit.
4. Transport capacity is about the ability to move forces between systems, not the ability to house them. Space stations do not contribute to transport ability.

Windchild292: My apologies, was spitballing ideas and forgot that you aren't used to how I tend to operate.
For 1, this isn't a makeshift bay. This is a bay that is the same size and with the same capacity as that on a dropship. I was thinking a morale penalty might be a nice way to show that the extra space is a bit cramped, but... RAW they are the same size and ability as the dropship ones.
For the rest, was just trying to throw out ideas, things to think about and consider.

Windchild292:
One of the unit tests failed because of an actual bug in calculations, which will need to be fixed.
The others failed because of reporting changes, and those have been rectified.

Windchild292:
I've added code to handle super heavy vehicles and their bays, but have left it disabled as I've not tested it.

Still trying to fix the last annoying unit test, I've not yet figured out why it is occurring. But, will hopefully have that one done by end of day today.

Windchild292:
The final failure is a pre-exisiting bug, I've managed to recreate it by reordering the statements in an older code base. It occurs based on a difference in ordering, which should not have that effect.

Windchild292:
The bug was that it was returning True for ExcessCapacity, because of a logical error in calculation. This has now been fixed. Ready for review

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
sixlettervariables: Thank you for the clarification @neoancient

MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
```diff
@@ -503,41 +503,70 @@ public int getTransportValue() {
...
+        //int excessSuperHeavyVeeBays = Math.max(getSuperHeavyVeeBayCount() - getSuperHeavyVeeCount(), 0);
+        //int excessHeavyVeeBays = Math.max(getHeavyVeeBayCount() + excessSuperHeavyVeeBays - getHeavyVeeCount(), 0);
         int excessHeavyVeeBays = Math.max(getHeavyVeeBayCount() - getHeavyVeeCount(), 0);
         int excessSmallCraftBays = Math.max(getSmallCraftBayCount() - getSmallCraftCount(), 0);
 
         // We need to subtract any filled bays from the count. This follows the following logic:
         // Assume you have 2 heavy vehicle bays, and 4 light vehicle bays, and are trying to store 1 heavy and 5 light vehicles
         // You have 1 more light vehicle than light vehicle bays to store them in, so you check how many free heavy vehicle bays
         // there are. Finding 1, you can store the light vehicle there, and it doesn't count as having excess
+        /*int superHeavyVeeBaysFilledByLighterVees, heavyVeeBaysFilledByLights, smallCraftBaysFilledByFighters;
```
sixlettervariables: Let's leave a TODO here explaining why this stuff is commented out.

MekHQ/src/mekhq/campaign/rating/FieldManualMercRevDragoonsRating.java
```diff
@@ -780,17 +780,22 @@ String getTransportationDetails() {
 
         out.append(String.format("%-" + HEADER_LENGTH + "s %3d", "Transportation", getTransportValue())).append("\n");
 
-        int excessSmallCraftBays = Math.max(0, getSmallCraftBayCount() - getSmallCraftCount());
-        int excessHeavyVeeBays = Math.max(0, getHeavyVeeBayCount() - getHeavyVeeCount());
+        //int excessSuperHeavyVeeBays = Math.max(getSuperHeavyVeeBayCount() - getSuperHeavyVeeCount(), 0);
```
sixlettervariables: likewise another TODO comment here

MekHQ/src/mekhq/campaign/rating/CampaignOpsReputation.java
Windchild292: Done

MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
neoancient: This still includes small craft and space station bays. FMM(r), p. 154 and Campaign Operations, p. 35 both state explicitly that transport capacity is based on dropships, jumpships, and warships.

Windchild292: I've made those changes. That means that #1100, which started this, is not a bug at all. However, I've made some internal changes to allow for a new unit rating that decides its bay count separately from the main two, and there are some other fixes in this PR. So, I think this PR should go in, but renamed (probably to something like Unit Rating: Minor bug fixes).

Also, I am considering making a new type of rating to handle stationary forces or single system forces, but that's future ideas/plans.

Morkath: > 
> 
> This still includes small craft and space station bays. FMM(r), p. 154 and Campaign Operations, p. 35 both state explicitly that transport capacity is based on dropships, jumpships, and warships.

And neither actually excludes sub-bays on the larger vessels.  You are the one making an arbitrary decision on what equipment is and is not valid.

PR #1379, opened 2020-01-01
Feature: Adds hyperlink from origin system to the map panel

sixlettervariables:
[*Best viewed with whitespace differences ignored.*](https://github.com/MegaMek/mekhq/pull/1379/files?w=1)

This adds a hyperlink from the person view panel to:
- Jumps to the system in the big view if their origin planet is the primary planet for the system.
- Jumps into the system if their origin planet is within a system itself.

image

image

image

image


VicenteCartas:
Looks really good to me!

Issue #1262, opened 2019-08-24
RFE: In Interstellar Map, indicate which systems are capitals or hiring halls


RexPearce:
When playing AtB, capitals and hiring halls will generate additional contracts each month. This can be important when starting out and Dragoon rating is low. However, I haven't figured out any way to identify these systems, other than looking in the XML for which system is the capital for faction for a given era.

I'd be nice if the map showed capital systems and hiring halls with a different icon. (maybe a star instead of a dot for capitals?) If that can't easily be done, even just a note in the text description that's updated as capitals/halls move around by era would be helpful.

NickAragua:
Implemented with #1396

PR #1396, opened 2020-01-17
#1262 identify on the interstellar view planets that are eligible for additional contracts

swang300:


MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -738,6 +738,16 @@ protected void paintComponent(Graphics g) {
 	                        if(null != factions && !isSystemEmpty(system)) {
 	                            int i = 0;
 	                            for(Faction faction : factions) {
+	                            	if (faction.getStartingPlanet(campaign.getGameYear()).equals(system.getId())) {
```
sixlettervariables: Hoist `faction.getStartingPlanet(campaign.getGameYear())` into a local variable outside the for loop.

swang300:
> This shows for all campaigns but is only relevant to AtB campaigns.

Only has a game effect in AtB campaigns, but Hiring Halls and Capitals are still relevant in non-AtB games as well?

neoancient:
The stated purpose of this PR is identify the systems where extra contracts are available is only relevant for AtB games. Identifying capitals for information purposes is generally relevant to non-AtB games, but the hiring halls are much more AtB-specific.

Issue #1302, opened 2019-10-01
Ratdata/Rat Issue: ConventionalFighter Rat


AlvarOldebrecht:
### Environment
MekHQ 45,5 and MekHQ46.0
Windows 10
Java version 1.8.0_221 or Java version 12.0.2

### Description
At the moment i'm creating a "Age of War"-RAT and found an error. If i put in the ratdata xml-file the ConventionalFighter-RATs above Mek-RATs, MekHQ will always roll for Meks on the ConventionalFighter-RAT. 

![ConventionalFigher Rat above Mek](https://user-images.githubusercontent.com/56050031/66003402-53043700-e4a6-11e9-941b-5b35f6c23871.png)

The Error is fixed, when the Mek-RATs are written above the ConventionalFighter-RATs.

![Mek Rat above ConventionalFigher Rat](https://user-images.githubusercontent.com/56050031/66003425-60212600-e4a6-11e9-907f-b89d11093824.png)

But as far as i know, this shouldn't happen.

### Files

[My Campaign25300320.zip](https://github.com/MegaMek/mekhq/files/3678630/My.Campaign25300320.zip)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/3678608/mekhqlog.txt)
[MekHQ46.zip](https://github.com/MegaMek/mekhq/files/3678615/MekHQ46.zip)

Windchild292:
I've looked into this, and, while it has resulted in some code cleanup, I've not found the cause of this issue. I will be opening a PR for this code cleanup, however.

I've also found this issue occurs when you try to select a weight class of mech that is not part of the RAT. In the aforementioned case, it appears that it is defaulting to a ConventionalFighter roll or selecting the top ConventionalFighter instead of saying nothing can be found.

Based on my read of the code, I believe this may be an issue in MegaMek, not in MekHQ.

Windchild292:
Took a new look. This is a user error. "Conventional Fighter" is the proper unit type, not "ConventionalFighter". The difference causes parsing to fail, and the internal fallbacks then work properly.

PR #1399, opened 2020-01-18
Cleaning up code in GMTools and RAT Management files

Windchild292:
This originated in me looking for the issue for #1302, but is only a code cleanup. It does not fix the aforementioned issue, which is likely in MegeMek instead of MekHQ

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
@@ -349,21 +349,23 @@ public class GMToolsDialog extends JDialog implements ActionListener {
                 lastRolledUnit = performRollRat();
             }
 
-            Entity e;
-            try {
-                e = new MechFileParser(lastRolledUnit.getSourceFile(), lastRolledUnit.getEntryName()).getEntity();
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
```diff
@@ -331,72 +331,60 @@ private void setUserPreferences() {
...
         }
-        
+
         if(event.getActionCommand().equals(RAT_ADDER_TOOL)) {
-            if (null == lastRolledUnit) {
+            if (lastRolledUnit == null) {
                 lastRolledUnit = performRollRat();
             }
 
-            if (null != lastRolledUnit) {
```
neoancient: The second null check is here because performRollRat() can return null.

Windchild292: You are correct, this is a dumb mistake on my side. It has been fixed.

neoancient: We all miss simple stuff. That's why we double-check each others' work.

PR #1401, opened 2020-01-19
1263: Fixing Marriages Without Surnames

Windchild292:


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -2632,15 +2679,16 @@ private JMenuItem newMenuItem(String text, String command, boolean enabled) {
...
         result.setSelected(selected);
         result.setActionCommand(command);
         result.addActionListener(this);
         result.setEnabled(true);
         return result;
     }
+
+    private boolean IsNullOrEmpty(String str){
+        return str.trim().length() == 0;
+//        return ((str != null) && (!str.trim().isEmpty()));
```
NickAragua: We have a method like this in megamek.common.util, let's use that instead of reinventing the wheel.

Windchild292: Ah, I forgot to hit enter to ask if there was pre-existing way to do this in MekHQ. And otherwise, I would have suggested making one, cause the fact it isn't standard in Java is bloody annoying.

Windchild292:
Commit name is a typo, my bad on that

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -528,34 +529,81 @@ public void actionPerformed(ActionEvent action) {
...
-                String spouseSurname = spouse.getName().split(SPACE, 2)[1];
-                String selectedGivenname = selectedPerson.getName().split(SPACE, 2)[0];
-                String spouseGivenname = spouse.getName().split(SPACE, 2)[0];
+                String[] selectedName = selectedPerson.getName().split(SPACE, 2);
+                String[] spouseName = spouse.getName().split(SPACE,2);
+                String selectedGivenName = "", selectedSurname = "", spouseGivenName = "", spouseSurname = "";
+
+                if (selectedName.length >= 2) {
+                    selectedGivenName = selectedName[0];
+                    selectedSurname = selectedName[1];
```
sixlettervariables: If >= 2 wouldn't the surname be selectedName[selectedName.length - 1] ?

Windchild292: You are correct, and the GivenName would be the rest of the name with the removal of the Surname. I'm going to assume a single last name for this PR, and create a RFE to permit multiple last names

Issue #1366, opened 2019-12-18
Empty Procurement List on Reload


Windchild292:
Description:
This bug is that items do not show up on the Procurement List after reloading MekHQ, which can cause items to get stuck in trying to purchase (if, for example, you accidentally order something you can't afford, you will then get daily messages about it).

Reproduction:
Try to order something you can't afford. It will then get added to the Procurement List. Save and close MekHQ. After reloading, the item will not be in Procurement List.

Environment
I've encountered this on MekHQ 0.46.0 and 0.47.2.
I am running Windows 10, Java 13.0.1 (with JDK installed).

Issue #1345, opened 2019-11-16
Part acquisition improper admin


bczretribution:
mekhq 47.2 (was also occurring in 45.2 before I updated)

Supply and acquisition are set to 2 acquisitions per day per admin.  As can be seen here the same admin is making all acquisition rolls, every day. Ignoring the 2 per person limit.
![image](https://user-images.githubusercontent.com/34722287/68997413-28aee300-0874-11ea-95bf-438dd603d2cf.png)

campaign file here
[error.zip](https://github.com/MegaMek/mekhq/files/3854547/error.zip)

CrimsonThunderSB:
is this being acquired in Bulk or Individually?

bczretribution:
These particular ones were acquired in bulk from the "parts in use" tab.  Though I believe I placed orders through the purchase parts menu prior and had a similar issue.

sixlettervariables:
Fixed by #1407

PR #1407, opened 2020-01-19
Bug: Use current shopping list in ProcurementTableModel and restrict acquisitions #1366 #1345

sixlettervariables:
The ProcurementTableModel defaulted to loading without the campaign's current parts list, causing CPNX reloads to display an empty procurement table in the warehouse tab #1366.


sixlettervariables:
I've also fixed the max acquisition bug when shopping. The standard shopping routine will now rotate through each capable logistics persons to acquire goods until all of them have exhausted their acquisitions. I'm looking into reworking the planetary acquisition method to rotate through logistics personnel (currently it just exhausts one of them).

Issue #1406, opened 2020-01-19
Personnel: Add Hyperlinks in Family Information


Windchild292:
Add hyperlinks between people in the family information section of the Personnel table to make it easier to move between them.

Windchild292:
I'm picking this one up, using @sixlettervariables's system link as the basis for the code.

sixlettervariables:
Closed by #1408 (and #1409)

PR #1409, opened 2020-01-19
1405: Display Mother and Father for Children in the Force

Windchild292:
This PR adds lines for Mothers and Fathers, with links to them.
This PR includes all of #1406, which has not gone into master at the time of writing.


MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -748,27 +759,99 @@ private JPanel fillFamily() {
...
+                gridBagConstraints = new GridBagConstraints();
+                gridBagConstraints.gridx = 1;
+                gridBagConstraints.weightx = 1.0;
+                gridBagConstraints.insets = new Insets(0, 10, 0, 0);
+                gridBagConstraints.fill = GridBagConstraints.NONE;
+                gridBagConstraints.anchor = GridBagConstraints.NORTHWEST;
+
+                for (Person child : person.getChildList()) {
+                    gridBagConstraints.gridy = firsty;
+                    lblChildren2 = new JLabel();
```
sixlettervariables: This should be a loop local variable, otherwise the field becomes the last child.

Windchild292: By re-declaring it like this, it works correctly. The previous JLabel is already part of the pnlFamily, so it works well. I've tested it on a family with 2 children, and they link correctly. Looking at it did find a wasteful declaration, so I've cleaned that up.

PR #1419, opened 2020-01-22
Improve performance of purchasing parts and fix refit bug

sixlettervariables:
While investigating an issue where ammo could not be bought, I noticed a number of huge performance slowdowns in the Purchase Parts screen. I also found a bug w.r.t. refit ID's.

1. Cache CurrencyManager::getDefaultCurrency. This was extremely expensive and called constantly 25bed25.
2. Remove text "sanitizing" from PartsInUse 21d2f0f.
3. Add Campaign::streamSpareParts and use it where allocating a huge ArrayList is unnecessary 51ffcd3.
4. Add Campaign::findSparePart to take a predicate to quickly find a matching part e8b0652.
5. Speed up AmmoStorage 43da701, Armor 72ce2f3, BaArmor b1c5751, ProtomekArmor c9fc5c5, and SVArmor a0ab221 bc100ad with findSparePart.
6. Speed up AmmoBin with findSparePart and streamSpareParts 18a608a. **This one could use some eyes**
7. Fix incorrect usage of reference equality of Refit ID's when value equality was intended a90d59b.

The time it takes to purchase 1000 units of AC/20 ammo by clicking 'Buy' is cut by ~70% on my desktop.

These commits are best viewed one by one with whitespace ignored (although I tried to minimize whitespace changes).

### DIFF
--- a/MekHQ/src/mekhq/campaign/finances/CurrencyManager.java
+++ b/MekHQ/src/mekhq/campaign/finances/CurrencyManager.java
@@ -43,6 +43,8 @@ import javax.xml.parsers.DocumentBuilder;
 import java.io.FileInputStream;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -57,11 +59,13 @@ import java.util.Set;
 public class CurrencyManager extends CurrencyUnitDataProvider {
     private static final CurrencyManager instance = new CurrencyManager();
 
+    private DateTime lastChecked;
     private Campaign campaign;
-    private ArrayList<Currency> currencies;
-    private HashMap<String, String> currencyCodeToNameMap;
-    private HashMap<String, String> currencyCodeToSymbolMap;
+    private List<Currency> currencies;
+    private Map<String, String> currencyCodeToNameMap;
+    private Map<String, String> currencyCodeToSymbolMap;
     private Currency backupCurrency;
+    private Currency defaultCurrency;
 
     private MoneyFormatter xmlMoneyFormatter;
     private MoneyFormatter uiAmountPrinter;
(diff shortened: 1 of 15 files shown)


MekHQ/src/mekhq/campaign/finances/CurrencyManager.java
```diff
@@ -116,17 +120,26 @@ MoneyFormatter getUiAmountAndNamePrinter() {
         return this.uiAmountAndNamePrinter;
     }
 
-    Currency getDefaultCurrency() {
-        if (campaign != null) {
-            HashMap<String, Currency> possibleCurrencies = new HashMap<>();
+    synchronized Currency getDefaultCurrency() {
```
VicenteCartas: Nice idea saving the currency for the day!

NickAragua: What happens if the user moves around between systems (e.g. using gm move?) We probably want to invalidate the cached value in that case as well.

MekHQ/src/mekhq/campaign/parts/PartInUse.java
```diff
@@ -1,17 +1,13 @@
...
 import java.util.Objects;
-import java.util.regex.Pattern;
 
 import megamek.common.AmmoType;
 import mekhq.campaign.finances.Money;
 import mekhq.campaign.unit.Unit;
 import mekhq.campaign.work.IAcquisitionWork;
 
 public class PartInUse {
-    private static final Pattern cleanUp1 = Pattern.compile("\\d+\\shit\\(s\\),\\s"); //$NON-NLS-1$
```
NickAragua: lol

neoancient: The purpose of this is to keep the parts overview from being cluttered by showing number of hits the part has taken.
Screenshot from 2020-01-23 10-36-19

sixlettervariables: Woops, late night editing missed the Java regex difference. .NET regex that'd be an escaped `\` followed by `shit`. I'll find a better way than regex to remove those.

MekHQ/src/mekhq/campaign/finances/CurrencyManager.java
sixlettervariables: I'll invalidate the cache for that too, good catch.

MekHQ/src/mekhq/campaign/parts/PartInUse.java
neoancient: I'm not going to pretend I've never snickered at it. My initial thought is a boolean parameter to getDetails() that would determine whether to include the hit data.

sixlettervariables: I've added an overload to handle that situation. One interesting find is that PodSpace seems to do something a bit different; but ultimately it isn't affected by this change.

PR #1425, opened 2020-01-24
Upgrade to Gradle 5.6

sixlettervariables:
This PR would update Gradle to version 5.6. I'm not savvy with our build system, so this is up for people to evaluate. If this upgrade is reasonable, I can complete companion PR's for MM and MML.

(aside: I needed Gradle 5.2+ to run SpotBugs as a plugin)


settings.gradle
```diff
@@ -2,8 +2,6 @@ rootProject.name = 'MekHQRoot'
 
 include 'MekHQ'
 
-enableFeaturePreview('STABLE_PUBLISHING')
```
neoancient: If memory serves, there was going to be a change in the way publishing tasks are handled in 5.0 (which wasn't released yet), and adding this allowed using it with 4.9 so it wouldn't have to be rewritten if we switch to a later version. So it served its purpose.

neoancient:
The build failed during release packaging when trying to execute tasks in the cloned megamek repo, which is to be expected since it includes the STABLE_PUBLISHING preview, which won't run with 5.0+.

neoancient:
Closing as obsolete

PR #1435, opened 2020-01-29
Fix and log invalid lc ammo bin

mkerensky:
Corrects an issue where campaigns fail to load due to an invalid class cast to largecraftammobin

This happens infrequently with dropship refits.  Fix allows the campaign to load anyway and logs which ship and equipment number are bugged so the unit can be repaired.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
@@ -270,9 +270,14 @@ public class LargeCraftAmmoBin extends AmmoBin {
 
     @Override
     public void updateConditionFromEntity(boolean checkForDestruction) {
+        final String METHOD_NAME = "updateConditionFromEntity()"; //$NON-NLS-1$
         if(null != unit) {
             Mounted mounted = unit.getEntity().getEquipment(equipmentNum);
-            if(null != mounted) {
+            if (!(mounted.getType() instanceof AmmoType)) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
```diff
@@ -270,9 +270,14 @@ public MissingPart getMissingPart() {
...
     public void updateConditionFromEntity(boolean checkForDestruction) {
+        final String METHOD_NAME = "updateConditionFromEntity()"; //$NON-NLS-1$
         if(null != unit) {
             Mounted mounted = unit.getEntity().getEquipment(equipmentNum);
-            if(null != mounted) {
+            if (!(mounted.getType() instanceof AmmoType)) {
+                MekHQ.getLogger().log(LargeCraftAmmoBin.class, METHOD_NAME, LogLevel.WARNING,
+                        unit.getName() + " has invalid ammo bin at equipment number " + equipmentNum);
+            }
+            if(null != mounted && mounted.getType() instanceof AmmoType) {
```
apokrovski: You definitely need a null check on mounted before calling mounted.getType(). I'd suggest moving the logic from the if statement on line 280 into the if statement on line 276, then making 280 an "else".

neoancient: Mounted can't be null at this point because if it were it would have thrown a NullPointerException on line 276. If it's possible that this is null, then that needs to be part of the check (which should also be logged, since that's also a problem).

mkerensky: Fixed.

Issue #1413, opened 2020-01-21
Overview Hangar Report: Add Mothballed to Mothballed Units in Hangar


Windchild292:

Windchild292:
You can assign this one to me, I'll be doing it now as a break from working on compilers

PR #1453, opened 2020-02-05
1413/1415: Adding (Mothballed) to units in the Hangar Report tab and Order Alphabetically

Windchild292:
This implements #1413


Windchild292:
Yeah, sorry about that. Didn't realize I had included that file until after I opened up the PR.

PR #1472, opened 2020-02-11
Upgrade gradle version to 5.6

neoancient:
See MegaMek/megamek#1663


MekHQ/build.gradle
```diff
@@ -125,7 +127,7 @@ task mmJarLibDir (type: Jar, dependsOn: compileMM) {
             .findAll { it.name.endsWith(".jar") && !it.name.toLowerCase().startsWith("megamek") }
             .collect { it.name }.join(' ')
     }
-    ext.jarLocation = "${buildDir}/libs/${archiveName}"
+    ext.jarLocation = "${buildDir}/libs/${archiveFileName}"
     inputs.dir "${mmDir}/build/classes/java/main"
```
sixlettervariables: Does these paths need `megamek` added too?

MekHQ/build.gradle
```diff
@@ -140,26 +142,26 @@ task compileMML (type: GradleBuild) {
 }
 
 task mmlJar (type: Jar, dependsOn: compileMML) {
-    archiveName = "MegaMekLab.jar"
+    archiveFileName = "MegaMekLab.jar"
     from "${mmlDir}/build/classes/java/main"
```
sixlettervariables: Do these need `megameklab`?

MekHQ/build.gradle
```diff
@@ -140,26 +142,26 @@ task compileMML (type: GradleBuild) {
...
     manifest {
         attributes "Main-Class": 'megameklab.com.MegaMekLab'
-        attributes 'Class-Path' : "${mmJar.archiveName} " + (project.sourceSets.main.runtimeClasspath.files
+        attributes 'Class-Path' : "${mmJar.archiveFileName} " + (project.sourceSets.main.runtimeClasspath.files
             .findAll { it.name.endsWith(".jar") && !it.name.toLowerCase().startsWith("megamek")}
             .collect { "${lib}/${it.name}" }.join(' '))
     }
-    ext.jarLocation = "${buildDir}/libs/${archiveName}"
+    ext.jarLocation = "${buildDir}/libs/${archiveFileName}"
     inputs.dir "${mmlDir}/build/classes/java/main"
```
sixlettervariables: And here.

MekHQ/build.gradle
neoancient: No. Those that were added to the compileMM task were to find the class and resource files to place into the jar in the build directory of the megamek subdirectory rather than the top level. But once built, the jar is placed in the libs subdirectory of the current build directory, which is where it looks to find it in this task.

MekHQ/build.gradle
neoancient: All the mml source is in the top-level directory, unlike MM and MekHQ, which put everything in a subdirectory.

MekHQ/build.gradle
neoancient: Same here. The MML build directory is in the top level.

PR #1490, opened 2020-02-16
Ship Autoload ui fixes 2

mkerensky:
Improvements to the UI of assigning units to ships from the TOE tab
Fixes some bugs that causes the Assign to Ship menu item to display when it shouldn't, as well as not display when it should.


MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1168,15 +1172,30 @@ private void maybeShowPopup(MouseEvent e) {
                 JMenu i_trn = new JMenu(TOEMouseAdapter.INFANTRY_CARRIERS);
                 JMenu a_trn = new JMenu(TOEMouseAdapter.ASF_CARRIERS);
                 JMenu sc_trn = new JMenu(TOEMouseAdapter.SC_CARRIERS);
+                JMenu singleUnitMenu = new JMenu("");
```
Windchild292: Why are you initializing this with a "" instead of just not including text?

mkerensky: No reason... <whistles>    Fixed.

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
```diff
@@ -405,6 +405,42 @@ public static boolean doAllUnitsHaveC3Master(Vector<Unit> units) {
...
+    
+    /**
+     * Used to test a selection of Units provided by the player and determine whether they all share a designated unitType. 
+     * @param units Vector of units that the player has selected
+     * @return false if any unit in the passed-in Vector does not have the specified unit type
+     */
+    public static boolean areAllUnitsSameType(Vector<Unit> units, int unitType) {
+        // For our purposes, a selection of tanks isn't the same unless they're all the same weight class
+        // Set this based on the first unit in the selection
+        double firstUnitWeight = units.get(0).getEntity() != null ? units.get(0).getEntity().getWeight() : 0;
```
sixlettervariables: This should exit early if `units` is empty.

NickAragua:
Sorry this was left hanging for so long, could you do the pending merge? Then we'll get it merged in.

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
```diff
@@ -405,6 +405,42 @@ public static boolean doAllUnitsHaveC3Master(Vector<Unit> units) {
...
+        boolean light = firstUnitWeight > 0 && firstUnitWeight <= 50;
```
sixlettervariables: Is the > 0 necessary? Won't they all have at least some weight (if you have some unit).

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
```diff
@@ -405,6 +405,42 @@ public static boolean doAllUnitsHaveC3Master(Vector<Unit> units) {
...
+        boolean heavy = firstUnitWeight > 50 && firstUnitWeight <= 100;
+        boolean superheavy = firstUnitWeight > 100 && firstUnitWeight <= 200;
```
sixlettervariables: Is there something > 200 for units?

mkerensky: There is not. I know you can construct such units, but the SH Vehicle *bay* only holds up to 200 tons - anything heavier is Cargo - so I think I'm ok here.

Windchild292: There are a few examples, like the Koi and Ryu WiGE.

neoancient: But mkerensky is correct that vehicles over 200 tons cannot be transported in a bay.

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
mkerensky: Fixed.  I thought I only called this in the first place after checking that in TOEMouseAdapter, but that is only done in one of two locations.

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
mkerensky: Seemed like a good idea, in case we managed to get a botched unit with negative or 0 weight

mkerensky:
I think I've got everything fixed here. I also found a couple of cases where I was checking for a null ArrayList instead of an empty one and fixed those.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -280,22 +281,27 @@ public void actionPerformed(ActionEvent action) {
...
             }
         }
         if (command.contains(UNASSIGN_FROM_SHIP)) {
-            for (Unit u : units) {
-                for (UUID oldShipId : u.getTransportShipId().keySet()) {
-                    Unit oldShip = gui.getCampaign().getUnit(oldShipId);
-                    if (oldShip != null) {
-                        oldShip.unloadFromTransportShip(u);
-                        MekHQ.triggerEvent(new UnitChangedEvent(oldShip));
+            if (!units.isEmpty()) {
```
sixlettervariables: This is superfluous with the for loop below. It'll do nothing if units is empty.

PR #1509, opened 2020-02-21
Adding Monthly and Yearly Autosave Options

Windchild292:
This adds monthly and yearly autosave options to MekHQ, in addition to some protection against bad directories and file names.

### DIFF
--- a/MekHQ/src/mekhq/service/AutosaveService.java
+++ b/MekHQ/src/mekhq/service/AutosaveService.java
@@ -128,7 +128,8 @@ public class AutosaveService implements IAutosaveService {
                     .collect(Collectors.toList());
 
             // Delete older autosave files if needed
-            int maxNumberAutosaves = this.userPreferences.getInt(MekHqConstants.MAXIMUM_NUMBER_SAVES_KEY, MekHqConstants.DEFAULT_NUMBER_SAVES);
+            int maxNumberAutosaves = this.userPreferences.getInt(MekHqConstants.MAXIMUM_NUMBER_SAVES_KEY,
+                    MekHqConstants.DEFAULT_NUMBER_SAVES);
             while (autosaveFiles.size() >= maxNumberAutosaves && autosaveFiles.size() > 0) {
                 autosaveFiles.get(0).delete();
                 autosaveFiles.remove(0);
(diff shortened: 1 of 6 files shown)


MekHQ/src/mekhq/service/AutosaveService.java
```diff
@@ -73,20 +82,33 @@ private boolean isWeeklyAutosaveEnabled() {
...
             String fileName = this.getAutosaveFilename(campaign);
-
-            try (FileOutputStream fos = new FileOutputStream(fileName);
-                 GZIPOutputStream output = new GZIPOutputStream(fos)) {
-                PrintWriter writer = new PrintWriter(new OutputStreamWriter(output, "UTF-8"));
-                campaign.writeToXml(writer);
-                writer.flush();
-                writer.close();
+            if (!StringUtil.isNullOrEmpty(fileName)) {
+
```
VicenteCartas: nit: whiteline

Windchild292: Whoops, fixed

MekHQ/src/mekhq/service/AutosaveService.java
```diff
@@ -98,39 +119,51 @@ private String getAutosaveFilename(Campaign campaign) {
...
+            index = 1;
+            while (repeatedName) {
+                fileName = String.format(
+                        "Autosave-%d-%s-%s.cpnx.gz",
+                        index++,
+                        campaign.getName(),
+                        campaign.getShortDateAsString());
+
+                repeatedName = false;
+                for (File file : autosaveFiles) {
```
NickAragua: Seems like if you turned autosaveFiles into a map, you could get an O(1) run time here, rather than O(#files)

Windchild292: Given an expected n of 5, I would expect it would actually be a performance drop to switch it out

Windchild292: @NickAragua 
Thoughts on the above?

NickAragua: I suppose it's unlikely that you'll encounter too many name overlaps like this. May as well leave it.

PR #1514, opened 2020-02-22
1429: Adding Bulk Add to the Parts Marketplace

### LINKED ISSUE #1429: Enhancement to Purchase Parts menu option
### Environment
Using Snapshot 47.3 

### Description
Under the Overview tab, Parts in use, there's an option while GM mode is engaged to add parts in bulk.  Under the Marketplace, Purchase Parts, while in GM mode you can only add one part at a time.  Would it be possible to *add* add parts in bulk to that section?

Windchild292:
This fixes #1429. It also reorders the buttons to be consistent with them elsewhere in the program and hides GM buttons when not in GM mode.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/PopupValueChoiceDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/PopupValueChoiceDialog.java
@@ -1,7 +1,20 @@
 /*
- * AlertPopup.java
+ * Copyright (c) 2013 The MegaMek Team. All rights reserved.
  *
- * Created on Jan 6, 2010, 10:46:02 PM
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/gui/dialog/PopupValueChoiceDialog.java
```diff
@@ -1,28 +1,31 @@
 /*
- * AlertPopup.java
+ * Copyright (c) 2013 The MegaMek Team. All rights reserved.
```
sixlettervariables:
```suggestion
 * Copyright (c) 2010, 2013 The MegaMek Team. All rights reserved.
```
Per:
> `* Created on Jan 6, 2010, 10:46:02 PM`

Windchild292: The file is actually 2013, the header was copied and never updated with the date

sixlettervariables: As in the header was copied from another file?

Windchild292: Yep

PR #1515, opened 2020-02-22
Scale Unit sell values by their Entity's price multiplier

### LINKED ISSUE #1497: Unit sell value does not account for price multipliers
### Environment
- MekHQ 0.47.4
- Linux Java 8

### Description
Dropship sell values are not being computed correctly, they're low by several orders of magnitude. For example: a basic Mule dropship has a MegaMekLab / buy value of ~173 million, but the sell value is only ~3 million.

This throws off all campaign finance calculations.

Looking at Unit.java, I suspect that the dropship's parts (in particular but not limited to the engines) are not being added correctly. There are several places where parts are added for SmallCraft and Jumpships, but not Dropships.

### Files
 
Mule_Correct_Price

Mule_Hanger_Price

Mule_Sell_Price

RAldrich:
Fix #1497

When computing construction costs, the final cost of many units is scaled by a multiplier.
Generally this is related to the unit's weight class or type (e.g. aerodyne dropships cost 36 times
the sum of their components costs).

This change causes getSellValue() to take this into consideration, generally resulting in higher
sell prices. Ideally the sell value of a perfect condition, undamaged Unit would match it's buy cost.
This change doesn't get all the way there (things like SI are still not accounted for), but this is
a step towards that goal that doesn't require a major redesign of the Entity cost calculation methods.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1158,6 +1158,9 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
             partsValue = partsValue.plus(2000.0 * sinks);
         }
 
+        // Scale the final value by the entity's price multiplier
+        partsValue = partsValue.multipliedBy(entity.getPriceMultiplier());
+
         return partsValue;
     }
 
@@ -1572,7 +1575,7 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
             updateBayCapacity(unitType, unitWeight, false, bayNumber);
         }
     }
-    
+
     /**
      * Calculates transport bay space required by an infantry platoon,
      * which is not the same as the flat weight of that platoon
@@ -1848,7 +1851,7 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
         } else {
             retVal.id = UUID.fromString(idNode.getTextContent());
         }
-        
+
         //Temp storage for used bay capacities
         boolean needsBayInitialization = true;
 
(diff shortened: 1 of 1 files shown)


RAldrich:
This PR is dependent on https://github.com/MegaMek/megamek/pull/1711 and cannot be merged before it.

RAldrich:
This was manually tested for a variety of Unit types. The actual impact this has on a particular type depends on how much of that type's construction cost is models as Parts - anything that's not modeled still isn't getting accounted for. In particular, Infantry are still way off.

That said this gets us much closer to the ideal. The biggest impact is on Dropships, as they have a very large flat multiplier (x28-x36). Using the example Mule from issue #1497, the sell value for a condition "D" unit goes from ~4 million to ~80 million. With no quality modifier it goes up to ~150 million - still short of the correct ~170 million, but much less noticeable.

sixlettervariables:
I figure if I buy a car today for $25k, I certainly can't sell it for that once it rolls off the lot!

Windchild292:
Very true, but I'm almost certain that it shouldn't be less than a salvage mech @sixlettervariables

RAldrich:
> Looks good, pending [MegaMek/megamek#1711](https://github.com/MegaMek/megamek/pull/1711)

Thanks!

> I figure if I buy a car today for $25k, I certainly can't sell it for that once it rolls off the lot!

Very true! But that's accounted for by the used parts and part quality multipliers in the campaign options. I only expect the sell price to match the buy price if those are disabled.

sixlettervariables:
I'm rerunning CI, but I don't think it'll pick up the changes. In any case once CI finishes this I'll merge it.

PR #1521, opened 2020-02-24
1482: Hire Full Complement Now Records Time in Service

### LINKED ISSUE #1482: Hire Full Complement does no record recruitment for tracking time in service
### Environment
MMHQ 0.46.1 & 0.47.4
Windows 10
JAVA 1.8.0_241

### Description
When in the hanger when you right click on a unit and select "hire full complement" the new personnel do not get a <recruitment> tag in the campaign XML file with the track time in service selected. Their recruit date does shows in the personal log. 

Hiring personnel in all other ways appears to work normally.

Windchild292:
This fixes #1482, includes some refactoring, and implements some changes used in further personnel changes.

Anything deprecated here will be removed in future personnel changes that are currently undergoing testing.

### DIFF
--- a/MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
+++ b/MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
@@ -166,7 +166,7 @@ public class ResolveScenarioTracker {
         } else {
             nu.setId(UUID.fromString(e.getExternalIdAsString()));
         }
-        
+
         for (Part part : nu.getParts()) {
             part.setBrandNew(false);
         }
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -1160,18 +1153,18 @@ public void resolveScenario(int resolution, String report) {
...
                 continue;
             }
             MekHQ.triggerEvent(new PersonBattleFinishedEvent(person, status));
-            if(status.isDead()) {
+            if (status.isDead()) {
                 continue;
             }
             if (status.isCaptured()) {
-                getCampaign().recruitPerson(person, true, true);
+                getCampaign().recruitPerson(person, true, false, false, true);
```
sixlettervariables: I think this PR is fine, but it is a shame Java doesn't have named parameters to make booleans like this easier to read.

The alternative would be to use Enumerations.

Windchild292: Yeah, thankfully Idea shows what each mean when you use true/false variables

PR #1532, opened 2020-02-28
relax bomb selection restrictions

NickAragua:
This allows users in MekHQ to load whatever bombs they want onto their aerospace fighters, limited only by availability. Previously, there was a restriction on that based on ... uh... a setting that's no longer present in MegaMek. ::shrug::


neoancient:
Pending MegaMek/megamek#1723

Issue #1542, opened 2020-03-04
Suggested Enhancement(s)


MageOhki:
Current versions of MekHQ

1: Separate out ground support vehicles from ground combat vehicles in hanger.  Currently all are under tanks.
2: I _don't_ know if this is possible (or practical), but the following changes in personnel areas for the tab may work better:
A: Separate out VTOL personnel from Vehicle crews.
B: once assigned to a combat vee or a support vee, have them show up on seperate blocks?
C: Same as above for Small Craft, Dropships and jumpships?

3: When showing techs for refits, would it be possible to sort them by the skill roll (ie, 4+, 5+), best roll to worst roll, then a secondary sort by time open?

4: Mass Training
A: Add special abilities 
B: set it so you can select those with the skill you're targeting, or without it at all.

5: Training/Spending XP:
special abilities: While Clan tech knowledge is buyable, Tech Speicalists (Internal, Armor, Weapons) are not, can that be added?

6: Weigh Secondary skills more suited for the role the hiree has.
(Example: I get a Mechwarrior with Tech: Aero, or a Aerospace pilot with Tech: Battle Armor, *really*?)

7: GM Tools multi select on mothball and activate items off right click menu.  At this time, you have to select one at a time...

Windchild292:
Each one of these should be their own issue, and I'm pretty sure you've already opened an issue for 7. 6 is a can of worms that I've started initial planning for (zero promises), and I think a four tiered search for 3 is best (Tech type, skill, rank, time open)

Windchild292:
I've actually run into issues relating to number 7 for my testing, so I'll fix that one.

Windchild292:
Number 3 will be implemented in #1544, just realised it now when reviewing the two big groupings of RFEs.

Windchild292:
1: #1571
2: #1572
3: Implemented in #1544
4: #1573 and #1574
5: Clan Tech Knowledge is purchasable? Also, do we have ideas on how much they should cost? 
6: #1576
7: Implemented in #1544 

Need replies for the fifth suggestion, but once that is done this story can be closed.

PR #1544, opened 2020-03-04
Hangar Tab Popup List: Bulk Selection and Rank Fixes

### LINKED ISSUE #1417: GM Tools, Hanger improvement
### Environment
NA
### Description
In the Hanger, in GM tools, the option to mothball/activate instantly has been put in.   It however only applies to *single* units, you cannot choose several units to demothball or mothball.  Would it be possible to fix this?

### Files
NA

Windchild292:
This is a bunch of changes to improve and modify the hangar tab popup list. I've modified most of the commands so that they work with a single unit or multiple, removes items from the list based on whether or not they are applicable, and added rank to the Hangar Mothball/Activate inputs.
Additionally, I added rank to the Warehouse and Repair bay tabs to make it easier to figure out the rank of our active techs.
As part of this, number 7 of #1542 will be implemented, and it fixes #1417.


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1887,14 +1898,22 @@ public Person findBestInRole(int role, String skill) {
...
             }
         }
 
         // Return the tech collection sorted worst to best
         // Reverse the sort if we've been asked for best to worst
         if (sorted) {
+            // First order by the amount of time the person has remaining, with the highest amount
+            // of time being at the top of the list
+            if (eliteFirst) {
+                techs.sort(Comparator.comparingInt(Person::getMinutesLeft));
```
NickAragua: How is this "elite first"?

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
```diff
@@ -4,7 +4,9 @@
 import java.util.UUID;
 import java.util.Vector;
 
+import megamek.common.BattleArmor;
```
NickAragua: Why add these imports?

MekHQ/src/mekhq/gui/dialog/MassMothballDialog.java
```diff
@@ -106,14 +102,16 @@ public MassMothballDialog(Frame parent, Unit[] units, Campaign campaign, boolean
         gbc.weightx = 3;
         JLabel instructionLabel = new JLabel();
         instructionLabel.setBorder(new LineBorder(Color.BLUE));
-        instructionLabel.setText("<html>Choose the techs to carry out mothball/reactivation operations on the displayed units. <br/>A * indicates that the tech is currently maintaining units.</html>");
```
NickAragua: While you're in this file, there's a bug in this dialog that shows activation time cost to be the same as mothball time cost, where it should be 1/2 the time. Do you think you can track it down?

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: It works in practice, but I can't for the life of me understand why.

MekHQ/src/mekhq/gui/utilities/StaticChecks.java
Windchild292: Weird, the changes looked like they had been reverted when i first opened this PR. They were part of one stage of this branch, but I ended up reverting them

Windchild292: Reverted

MekHQ/src/mekhq/gui/dialog/MassMothballDialog.java
Windchild292: I'll take a look, and I've actually got an idea of where it is.

Windchild292: Fixed, and found another bug that is somewhat related that I fixed at the same time.

MekHQ/src/mekhq/campaign/Campaign.java
sixlettervariables: This doesn't sort elites first. The code which does is on line 1934:
```
                // Return, swapping the value if we're looking to have Elites ordered first
                return eliteFirst ? -retVal : retVal;
```
I think the logic should be revised for this block, as it stands it doesn't make much sense (even if it produces the desired results).

MekHQ/src/mekhq/gui/RepairTab.java
```diff
@@ -856,7 +857,7 @@ public void refreshTaskList() {
 
     public void refreshTechsList() {
         int selected = techTable.getSelectedRow();
-        ArrayList<Person> techs = getCampaign().getTechs(true, null, false, false);
+        List<Person> techs = getCampaign().getTechs(true, null, true, false);
```
sixlettervariables: A comment explaining the unnamed arguments would be nice. Hard to say what this does outside of an IDE (and even within you have to hover).

MekHQ/src/mekhq/gui/WarehouseTab.java
```diff
@@ -670,8 +668,7 @@ protected void updateTechTarget() {
     }
 
     public void refreshTechsList() {
-        ArrayList<Person> techs = getCampaign().getTechs(true, null, false, false);
-        techsModel.setData(techs);
+        techsModel.setData(getCampaign().getTechs(true, null, true, false));
```
sixlettervariables: A comment here as well would be nice. Also, if `getTechs(true, null, true, false)` is the new default, how about a descriptively named method on Campaign?

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -244,60 +311,43 @@ public void actionPerformed(ActionEvent action) {
...
-                                            "A file already exists for this unit, cannot tag as custom. (Unit name and model)",
-                                            "File Already Exists",
-                                            JOptionPane.ERROR_MESSAGE);
-                            return;
-                        }
-                        FileOutputStream out = new FileOutputStream(
-                                fileNameCampaign);
-                        PrintStream p = new PrintStream(out);
+                String fileName = unit.getEntity().getChassis() + " " + unit.getEntity().getModel();
+                if (unit.getEntity() instanceof Mech) {
```
sixlettervariables: Could this be moved into a method? This massive if...elseif block is unruly as it stands.

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -605,454 +629,508 @@ public void mouseReleased(MouseEvent e) {
...
+                    popup.add(menuItem);
+                }
+
+                // sell unit
+                if (gui.getCampaign().getCampaignOptions().canSellUnits()) {
+                    popup.addSeparator();
+                    menuItem = new JMenuItem("Sell Unit");
+                    menuItem.setActionCommand(COMMAND_SELL);
+                    menuItem.addActionListener(this);
+                    menuItem.setEnabled(!unit.isDeployed());
```
sixlettervariables: Should this be protected by `oneSelected`? It currently checks if the unit is deployed.

Windchild292: This should be hidden behind a check that ensures that one is not deployed, instead of enabling/disabling if the first unit is deployed. I've fixed that

MekHQ/src/mekhq/gui/RepairTab.java
Windchild292: Ah, within Idea it shows all the variable names inline if you use values like true, null, "", or false.
I'll add a comment.

Windchild292: image

MekHQ/src/mekhq/gui/WarehouseTab.java
Windchild292: It isn't the default, just used in these two places. Added the comment though

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: I've added comments which should make it easier to understand.

sixlettervariables: I guess I don't follow. If you *don't* want elites first, why wouldn't you still want the person with the most time left?

Windchild292: Yes, this is used for Mothballing, which you always want a person with the most time left sorted to the top while the best personnel are the worst skilled. Annoying logic, but it works and looks good in practice.

Issue #1250, opened 2019-08-18
Warehouse sorting by total value is wrong.


plutonick:
When in the Warehouse tab, when sorting by Total Value, the sorting messed up.

![image](https://user-images.githubusercontent.com/8420602/63218814-9c096180-c16c-11e9-8289-74c919a10125.png)

I am using version 0.45.4

sixlettervariables:
Classic case of sorting.by string and not the value itself.

sixlettervariables:
Closed by #1550

PR #1550, opened 2020-03-08
Added FormattedNumberSorter to Warehouse total cost column (Issue #1250)

Skoraks:
Hi,
I added the existing 'FormattedNumberSorter' to the Total Cost column in the Warehouse tab to resolve the ordering issue where it is treated as a string.


sixlettervariables:
Thank you!

PR #1554, opened 2020-03-12
Finishing Switch to MekHQ.getLogger() instead of System.out/err/printStackTrace

Windchild292:
This should fix the logger not working in current nightly builds, and completely moves MekHQ to use DefaultMMLogger instead of System.out, System.err, and e.printStackTrace.

To review:
Hide whitespace differences should be on (a bunch of these files have not been opened recently, and it was an automatic thing by my IDE)
View the first three commits together, as they are the same thing (switching to using the getLogger()). Commit four consists of whitespace changes (mostly relating to tabs) that I discovered when reviewing this story.
Only logical change is commented, as are the main questions.

### DIFF
--- a/MekHQ/src/mekhq/MekHQ.java
+++ b/MekHQ/src/mekhq/MekHQ.java
@@ -89,9 +89,9 @@ import mekhq.service.IAutosaveService;
  * The main class of the application.
  */
 public class MekHQ implements GameListener {
-	//TODO: This is intended as a debug/production type thing.
-	// So it should be backed down to 1 for releases...
-	// It's intended for 1 to be critical, 3 to be typical, and 5 to be debug/informational.
+	// TODO : This is intended as a debug/production type thing.
+	// TODO : So it should be backed down to 1 for releases...
+	// TODO : It's intended for 1 to be critical, 3 to be typical, and 5 to be debug/informational.
 	public static int VERBOSITY_LEVEL = 5;
 	public static String CAMPAIGN_DIRECTORY = "./campaigns/";
 	public static String PREFERENCES_FILE = "mmconf/mekhq.preferences";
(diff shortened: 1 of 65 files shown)


MekHQ/src/mekhq/MekHQ.java
```diff
@@ -366,8 +366,12 @@ public static void main(String[] args) {
         System.setProperty("com.apple.mrj.application.apple.menu.about.name","MekHQ");
 
         SwingUtilities.invokeLater(() -> {
-            //redirect output to log file
-            redirectOutput();
+            // TODO : logFileName should probably not be inline
+            String logFileName = "logs" + File.separator + "mekhqlog.txt";
+            // Reset the log file - TODO : Maybe move this to being a call to the logger?
+            MegaMek.resetLogFile(logFileName);
```
Windchild292: Should this be added to the logger?

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -860,15 +858,11 @@ private void loadUnitsAndPilots(File unitFile) throws IOException {
...
-            // Read a Vector from the file.
-            try {
+            try (InputStream listStream = new FileInputStream(unitFile)) {
+                // Read a Vector from the file.
                 parser.parse(listStream);
-                listStream.close();
-            } catch (Exception excep) {
-                excep.printStackTrace(System.err);
-                // throw new IOException("Unable to read from: " +
-                // unitFile.getName());
```
Windchild292: This is the only logical change, I've changed listStream into a try (with resources)

Windchild292:
That resource leak is the logger, and is expected.

MekHQ/src/mekhq/MekHQ.java
sixlettervariables: I don't know but I'm confused as to why the log file bits have to be invoked on the UI thread.

Windchild292:
There are places we could really improve for logging, but it could be worse (and I think this is a step in the correct direction)

MekHQ/src/mekhq/MekHQ.java
Windchild292: Good point, I'll test and see what happens outside the UI thread.

Windchild292: I've got no idea why it was in the UI thread, but I've moved it to its over thread and added a reset log file method to the default logger.

Windchild292:
One change, and this now requires https://github.com/MegaMek/mekhq/pull/1554

PR #1556, opened 2020-03-13
Updating Unit Rating Wording

Windchild292:
We have the space to display using the full terms and proper spelling, so I switched it to use them instead of any shortforms. I also added ranks to the names used for commanders and fixed some capitalization issues along the way.


Windchild292:
Merge conflicts have been fixed.

PR #1559, opened 2020-03-15
Quick fix for logger

Windchild292:
The only real change here is that we no longer remove the close() calls for the logger (the other changes are just moving calls outside of the try{} catch), which seems to have broken the logger in MekHQ. As it works on my local branch, I think we should merge this ASAP and then test on the nightly build to ensure this fixes the issue. If it does not, I will take another look into the problem.
This will cause an LGTM issue, as I'm removing a potential resource link.


Windchild292:
Closing as the primary logger branch was merged instead.

PR #1533, opened 2020-02-29
Windchild's Personnel Improvements

Windchild292:
This is my suite of personnel changes. The major improvements are Random Marriages, Given Names and Surnames, expanded Family Display, Former Spouse Tracking, and a ton of new options with which you can customize the family-side of MekHQ, in addition to a suite of smaller changes and improvements.

MekHQ PRs #1438, #1462, #1463, #1468, #1469, #1499, #1517, #1518, #1519, #1520, #1521, and #1523 were originally part of this PR and have been split off to reduce the number of changes or to include changes in 0.47.4.
Similarly, MegaMek PRs https://github.com/MegaMek/megamek/pull/1649, https://github.com/MegaMek/megamek/pull/1681, https://github.com/MegaMek/megamek/pull/1712, https://github.com/MegaMek/megamek/pull/1716, and https://github.com/MegaMek/megamek/pull/1717 have been split off from the MegaMek-side PR.
However, it is still a pretty large suite of changes, and because of the splits a few files have ended up being pretty much just whitespace changes.

The changes I considered important enough to track are listed below:
> Part 1: Display Grandparents, Aunts/Uncles, and cousins in the personnel view
> Part 2: Option for both members of the marriage to use the same surname
> Part 3: Enable the ability to increase odds of random procreation
> Part 4: Former Spouse tracking
> Part 5: Option for baby to use either the mothers or fathers surname
> Part 6: #876: Marriage for either male or female
> Part 7: #1404: Given Name(s) Surname(s) split
> Part 8: Add honorific support
> Part 9: Add conceived tracking to fathers
> Part 10: Expected Due Date versus True Due Date for pregnant personnel
> Part 11: Add given name, surname and honorific to personnel tables
> Part 12: Add Bloodname logic for Clanner factions to display Bloodnames instead of surnames in the personnel tables
> Part 13: Setting to decide if you want to keep the same name when removing a widowed spouse or revert to maiden name
> Part 14: Option to remove spouse and keep married name for both or either member of the marriage
> Part 15: Removed forced prisoner marriage to members of the merc company
> Part 16: Add gender tracking to MegaMek Crew
> Part 17: Add extraData fields to MegaMek Crew
> Part 18: #1482: Hire Full Complement does not record recruitment for tracking time in service
> Part 19: Standardize personnel hiring in MekHQ
> Part 20: #765: Random marriages
> Part 21: #1486: Export Personnel to CSV does not work 0.47.4
[...]

PR #1575, opened 2020-03-18
Batch XP: Fixing Dialog Title

Windchild292:


Windchild292:
Forgot that most of these were implemented in #1533

PR #1583, opened 2020-03-20
Adding Menu Bar Mnemonics (Shortcut Keys)

Windchild292:
I'd recommend reviewing the March 19th Commits together, then the refactoring commit, then the remaining commits can be reviewed in any order.


MekHQ/resources/mekhq/resources/CampaignGUI.properties
```diff
@@ -3,26 +3,27 @@
 # Menu Resources
 # File Menu
 fileMenu.text = File
-menuLoad.text=Load Campaign...
-menuSave.text=Save Campaign...
+menuLoad.text=Load Campaign
```
sixlettervariables: Why remove the ellipses? That's standard menu speak for "I open a dialog". (I understand for Load and Save in some respects, but the rest?)

Windchild292: Late night mistake, I'll revert it once I finish tracking down the medical rating issue.

Windchild292: Reverted

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -909,31 +987,36 @@ private void initMenu() {
...
-        miGMToolsDialog.setEnabled(true);
+        miGMToolsDialog.setMnemonic(KeyEvent.VK_G);
         miGMToolsDialog.addActionListener(evt -> showGMToolsDialog());
         menuManage.add(miGMToolsDialog);
 
-        JMenuItem miAdvanceMultipleDays = new JMenuItem(resourceMap.getString("miAdvanceMultipleDays.text"));
-        miAdvanceMultipleDays.setEnabled(getCampaignController().isHost());
-        miAdvanceMultipleDays.addActionListener(evt -> showAdvanceDaysDialog());
-        menuManage.add(miAdvanceMultipleDays);
+        if (getCampaignController().isHost()) {
```
NickAragua: ?

Windchild292: I'm hiding the menu item if it can't be shown, unlike the current variant of disabling it

Windchild292: As this is never changed after the menu is enabled, this is the better way to implement it

Windchild292: Actually, thinking more about it this should be visible based on the current host, and enabled when that changes. I'll set that up now.

Windchild292: Switched to enabling the visual based on whether the campaign is the host or not, and added a method to switch it out. However, there is no obvious way to swap based on the current CampaignController code

MekHQ/src/mekhq/campaign/CampaignController.java
```diff
@@ -57,11 +57,15 @@ public UUID getHost() {
...
-     * @param id The unique identifier of the host compaign.
+     * @param id The unique identifier of the host campaign.
      */
     public void setHost(UUID id) {
         host = id;
         isHost = getLocalCampaign().getId().equals(id);
+
+        // We need to change these based on which campaign is currently the host
+        // TODO: Add the following:
+        // something that returns CampaignGUI.showAdvanceMultipleDays(isHost);
```
sixlettervariables: I don't understand this comment.

Windchild292: This was me adding a TODO for however you decided to handle campaigns switching between being host or not, as this should be enabled on that switch.

Otherwise, we might want a getCampaignController Host current date or such.

Windchild292: Removed based on that presumption.

PR #1590, opened 2020-03-24
Campaign Option: Dependents don't leave the force

Windchild292:
This adds a new option that disables dependents leaving the force, which is useful for certain forms of force.


NickAragua:
Could you elaborate on this one a little? I feel like I'm missing the point here.

pheonixstorm:
In AtB dependents can get removed just like other personnel. If I read it right this adds an option to the atb options tab to stop this from happening.

Windchild292:
That is correct @pheonixstorm. This is pretty much the dependent equivalent of disabling retirement

PR #1609, opened 2020-04-01
Person LocalDate

Windchild292:
This PR switches the Person class to use java.time.LocalDate instead of the mixture of Joda, Gregorian Calendar, and Date.

LocalDate for campaign is currently just a conversion call, as we don't need to track the same thing three times.

This was stirred on to make it easier to implement parts of the RandomDeath code.

### DIFF
--- a/MekHQ/src/mekhq/campaign/MercRosterAccess.java
+++ b/MekHQ/src/mekhq/campaign/MercRosterAccess.java
@@ -473,10 +473,10 @@ public class MercRosterAccess extends SwingWorker<Void, Void> {
                 preparedStatement.setInt(6, forceId);
                 preparedStatement.setInt(7, 1);
                 //TODO: get joining date right
-                preparedStatement.setDate(8, new java.sql.Date(p.getBirthday().getTimeInMillis()));
+                preparedStatement.setDate(8, java.sql.Date.valueOf(p.getBirthday()));
                 //TODO: combine personnel log with biography
                 preparedStatement.setString(9, p.getBiography());
-                preparedStatement.setDate(10, new java.sql.Date(p.getBirthday().getTimeInMillis()));
+                preparedStatement.setDate(10, java.sql.Date.valueOf(p.getBirthday()));
                 preparedStatement.setString(11, p.getId().toString());
                 if(preparedStatement.executeUpdate() < 1) {
                     //no prior record so insert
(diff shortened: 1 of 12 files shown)


MekHQ/src/mekhq/campaign/MercRosterAccess.java
```diff
@@ -473,10 +473,10 @@ private void writePersonnelData() {
                 preparedStatement.setInt(6, forceId);
                 preparedStatement.setInt(7, 1);
                 //TODO: get joining date right
-                preparedStatement.setDate(8, new java.sql.Date(p.getBirthday().getTimeInMillis()));
+                preparedStatement.setDate(8, java.sql.Date.valueOf(p.getBirthday()));
```
NickAragua, line 476: I hate to ask, but is this going to impact this type of export? Not that I've ever even heard of a MercRoster database, but...

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -2228,13 +2209,27 @@ public static Person generateInstanceFromXML(Node wn, Campaign c, Version versio
...
+                        }
+                    } else {
+                        retVal.dueDate = LocalDate.parse(wn2.getTextContent().trim(),
+                                DateTimeFormatter.ofPattern(DATE_SAVE_FORMAT));
+                    }
                 } else if (wn2.getNodeName().equalsIgnoreCase("expectedDueDate")) {
-                    SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss");
-                    retVal.expectedDueDate = (GregorianCalendar) GregorianCalendar.getInstance();
-                    retVal.expectedDueDate.setTime(df.parse(wn2.getTextContent().trim()));
+                    if (version.isLowerThan("0.47.6")) {
```
NickAragua, line 2223: Is there a way to just parse a date without specifying the exact format? Seems like a built-in library should be smart enough to contain that logic already.

Windchild292: I didn't find one while researching, annoyingly.

MekHQ/src/mekhq/campaign/MercRosterAccess.java
HammerGS: MercRoster that is a name I haven't heard in a long time. It was one of Taharqa projects.

https://sourceforge.net/projects/mercroster/

MekHQ/src/mekhq/campaign/personnel/Person.java
sixlettervariables: You can but it fails on machines with different cultures from what wrote the date, can't pass around CPNX's.

MekHQ/src/mekhq/campaign/MercRosterAccess.java
Windchild292: This method was added for reverse compatibility, so it shouldn't?
https://www.javatpoint.com/java-sql-date-valueof-method

Windchild292: As for MercRoster, see Hammer's reply... although I've seen it surprisingly often when looking at personnel (probably because of how tied in the two are)

MekHQ/src/mekhq/campaign/personnel/Person.java
NickAragua: All right, in that case, let's consolidate this logic into a separate function.

Windchild292: I plan on doing so for all of the personnel migration cases, but that will be in a different PR.

PR #1617, opened 2020-04-03
Add basic GitHub Actions CI

sixlettervariables:
Mirrors the work for MegaMek and MegaMekLab.


sixlettervariables:
I've cleaned this up and squashed it all down to the correct bits. Example run here:

https://github.com/sixlettervariables/mekhq/pull/213/checks?check_run_id=560832198

NickAragua:
I'd approve it, but I have no idea what any of these things do, so better for @neoancient to review it.

Windchild292:
I'm in the same boat as the above

sixlettervariables:
I've added comments explaining the bits and pieces. It wasn't immediately obvious to me either, but git rebase hides my shame.

sixlettervariables:
Okay this update to the action includes the following:
1. If there is a "sister" branch on the submitter's MM and/or MML repo, that branch is used, otherwise it uses MM/master and MML/master.
2. Composite builds have been corrected.

Windchild292:
The changes look and sound great based on the comments in the file, however I also don't know the language, so I don't feel comfortable approving this PR.

PR #1621, opened 2020-04-03
1613: Automatically Output Finances Table to File on New Year

### LINKED ISSUE #1613: RFE: Automatically Output Finances Table to File on New Year
This is an RFE for the option to automatically output the finances to csv for the year before the table is reset.

Windchild292:
This fixes #1613.

To review: view the commits one by one, ignoring the last (unless you want to read through a reorder and grouping of the finances methods and groupings). (this is actually a small group of changes, the reorder just makes it look significantly larger than it is)

Because of the reorder in CampaignOptions, I discovered that we were not writing infantryDontCount to file. That has been fixed.


MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -239,6 +235,12 @@ public void newDay(Campaign campaign) {
 
         // check for a new year
         if (calendar.get(Calendar.MONTH) == Calendar.JANUARY && calendar.get(Calendar.DAY_OF_MONTH) == 1) {
+            if (campaignOptions.getYearlyFinancesToCSVExport()) {
```
sixlettervariables, line 238: I think this would make more sense moving it into the `newFiscalYear` code at the beginning.

Windchild292: 🤦‍♂ 
Fixed

PR #1622, opened 2020-04-03
1134: Option to Disable Dependent Automatic Addition in AtB

### LINKED ISSUE #1134: Feature Request: In ATB Campaign, option to disable automatically added Dependants
MekHQ 45.3 ATB Campaign
I would like an option that lets us disable the automatic adding of Dependants/Astechs.
Ideally this would stop Dependants being added
- On the first day of each Year
- Due to the monthly special event in a contract

Having them added is fine for normal play, but once your campaign is 10-20 years old you get far more than you need. As an example, in my current Campaign 3020-3054, i have 1000 dependents, proving 3x the astech hours that I need for my Regiment sized unit.

Windchild292:
This fixes #1134.


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -673,10 +673,12 @@ public void doBonusRoll(Campaign c) {
         switch (roll) {
         case 1: /* 1d6 dependents */
             number = Compute.d6();
-            c.addReport("Bonus: " + number + " dependent" + ((number>1)?"s":""));
-            for (int i = 0; i < number; i++) {
-                Person p = c.newDependent(Person.T_ASTECH, false);
-                c.recruitPerson(p, false, true, false, true);
+            c.addReport("Bonus: " + number + " dependent" + ((number > 1) ? "s" : ""));
```
sixlettervariables, line 676: This should only be reported if you're actually adding them.

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -2350,6 +2352,14 @@ public void setFoundersNeverRetire(boolean mods) {
         foundersNeverRetire = mods;
 	}
 
+	public boolean getAddDependents() {
```
sixlettervariables, line 2355: Better: `canAtBAddDependents` or just `canAddDependents`

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -2350,6 +2352,14 @@ public void setFoundersNeverRetire(boolean mods) {
...
+        return addDependents;
+    }
+
+    public void setAddDependents(boolean b) {
```
sixlettervariables, line 2359: `setCanAtBAddDependents`

MekHQ/src/mekhq/campaign/mission/AtBContract.java
Windchild292: That was outside to verify for testing, moved back inside so it is only reported in that case. Fixes for review changes have been pushed.

Issue #1625, opened 2020-04-03
RFE: Time in Service (TIS) and Time in Rank (TIR) should also have an option to display in months, years and combo.


PhoenyxCullen:
Time in Service (TIS) and Time in Rank (TIR) should also have an option to display in months, years and combo.

Windchild292:
I'll handle this once #1628 is merged.

PR #1628, opened 2020-04-03
1626: Adding Time in Rank to MekHQ

### LINKED ISSUE #1626: RFE: TIR should be added, as others are using this method to promote personnel.
TIR should be added, as others are using this method to promote personnel.

Windchild292:
This fixes #1626 , and implements it with a standard display in months before the addition of #1625

### DIFF
--- a/MekHQ/resources/mekhq/resources/PersonViewPanel.properties
+++ b/MekHQ/resources/mekhq/resources/PersonViewPanel.properties
@@ -1,3 +1,9 @@
+# General Properties
+Time.YearYears=year(s)
+Time.MonthMonths=month(s)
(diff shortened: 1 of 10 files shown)


MekHQ/resources/mekhq/resources/PersonViewPanel.properties
```diff
@@ -1,3 +1,9 @@
+# General Properties
+Time.YearYears=year(s)
```
sixlettervariables, line 2: YearYears?

Windchild292: Couldn't think of a better way to say year(s)

Issue #1612, opened 2020-04-02
Cannot scroll through spouses


FastJack-MD:
Love that you set up the non-cis family options, but when trying to choose a spouse for a character, if you have an amount of PCs that fills out the menu, it doesn't allow you to scroll through the list.

Windchild292:
This is on my list for the near future, just waiting on a related PR before adding it.

FastJack-MD:
Is there a way to work around it before the next release?

Windchild292:
🤔 
Marrying most of your force will work, and the list is smaller than it was before (it formerly included prisoners, and personnel that were dead, missing, or retired). As the list is by age then alphabetically by name, you should be able to marry anyone to the youngest grouping of personnel, then continue that forward. However, I do know that is not the easiest thing to do.

Annoyingly, I added the scrollbar check to the wrong menu, and I apologize about the inconvenience that is going to cause.

Also, this comment led me to discover a very well hidden bug in the potential spouse logic, which I have now fixed.

PR #1630, opened 2020-04-04
Prisoners Marriage Logic Fix

Windchild292:
This is a bug I found when asked about a workaround as part of #1612 . Prisoners were supposed to be able to marry whomever (specifically that a prisoner can marry anyone, but nobody else can marry a prisoner), however they are currently only able to marry other prisoners.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -1285,7 +1285,7 @@ public class Person implements Serializable, MekHqXmlSerializable {
                 !this.equals(p)
                 && !p.hasSpouse()
                 && p.oldEnoughToMarry()
-                && (!p.isPrisoner() || (p.isPrisoner() && isPrisoner()))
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1285,7 +1285,7 @@ public boolean safeSpouse(Person p) {
                 !this.equals(p)
                 && !p.hasSpouse()
                 && p.oldEnoughToMarry()
-                && (!p.isPrisoner() || (p.isPrisoner() && isPrisoner()))
+                && (!p.isPrisoner() || isPrisoner())
```
sixlettervariables, line 1288: So this logic is we can marry the person if:
A. They are not a prisoner, or...
B.  we're a prisoner.

?

Windchild292: That is correct. That enables people to go to the prisoner list and marry prisoners to personnel if they want to, but not the other way around.

PR #1639, opened 2020-04-07
1218: Personnel Total Earnings

### LINKED ISSUE #1218: Feature request: Mechwarrior total earnings
Hi,

As always, fab work!

I was wondering if you would implement somewhere where we can see a total earnings that a mechwarrior has earned since joining the company, including money from shares. Perhaps under 'Time in service' on the personnel tab?

I thought it would add a nice bit of immersion/fluff to see how much each person has earned since they joined the company, taking into account their shares (I usually set my share % to 50). I had this thought whilst deciding if any of my MWs are rich enough to set up their own spin off Merc company when one of my games was getting a little stale.

Thanks!

Windchild292:
This fixes #1218 by implementing tracking for the total earnings of personnel.

To review: whitespace differences off, and there is a comment on what has changed in the functionality I moved to a private method.


MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -407,6 +391,43 @@ public void newDay(Campaign campaign) {
...
+                    boolean sharesForAll = campaign.getCampaignOptions().getSharesForAll();
+                    for (Person person : campaign.getActivePersonnel()) {
+                        numberOfShares += person.getNumShares(sharesForAll);
+                    }
+
+                    Money singleShare = shares.dividedBy(numberOfShares);
+                    for (Person person : campaign.getActivePersonnel()) {
+                        person.payPersonShares(singleShare, sharesForAll);
+                    }
+                }
```
Windchild292, line 417: This is the new addition to track shares.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3755,24 +3755,18 @@ public Money getPayRoll() {
     }
 
     public Money getPayRoll(boolean noInfantry) {
-        if(!campaignOptions.payForSalaries()) {
+        if (getCampaignOptions().payForSalaries()) {
```
NickAragua, line 3758: This one seems like an "if it ain't broke don't fix it" scenario.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3806,11 +3800,11 @@ public Money getWeeklyMaintenanceCosts() {
     }
 
     public Money getOverheadExpenses() {
-        if(!campaignOptions.payForOverhead()) {
+        if (getCampaignOptions().payForOverhead()) {
```
NickAragua, line 3803: Same here, not really sure of the point of rearranging the if block.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -4309,37 +4362,33 @@ public void setOriginalUnit(Unit unit) {
     }
 
     public int getNumShares(boolean sharesForAll) {
```
NickAragua, line 4364: I think this method is going to require an explanation of the sharesForAll parameter. Also, I'm wondering if it'll be a little more efficient if we cache the results of all this logic. No need to store it permanently, just invalidate it when something happens to change the number of shares.

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: This is to prevent potential issues from the statement being false first and true second (I don't really care, but I've seen this result in bugs from lazy devs who didnt check for a false statement first more often than you'd expect)

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: See the next one for my reasoning on the change.

MekHQ/src/mekhq/campaign/personnel/Person.java
Windchild292: Oh, it will be... but that's for a future PR (trying to keep features separate, and that is definitely a new internal feature). As for the comment, that has been added.

Issue #1636, opened 2020-04-07
Failure to save game


KiloWhiskey1:
### Environment
What version of MekHQ does your issue apply to? 47.5
What operating system are you using?  Windows, Mac, or Linux? Windows 10
Java version 1.8.0_231

### Description
Game will not save megamelog is tossing out a null pointer exception.

18:57:41,266 INFO [mekhq.gui.CampaignGUI] {AWT-EventQueue-0} 
Saving campaign...

18:57:45,973 INFO [mekhq.Utilities] {AWT-EventQueue-0} 
File copied.

18:57:46,187 ERROR [mekhq.gui.CampaignGUI] {AWT-EventQueue-0} 

java.lang.NullPointerException
	at mekhq.campaign.ExtraData.isEmpty(ExtraData.java:182)
	at mekhq.campaign.personnel.Person.writeToXml(Person.java:2124)
	at mekhq.campaign.Campaign.writeMapToXml(Campaign.java:4846)
	at mekhq.campaign.Campaign.writeToXml(Campaign.java:4645)
	at mekhq.gui.CampaignGUI.saveCampaign(CampaignGUI.java:1418)
	at mekhq.gui.CampaignGUI.menuSaveXmlActionPerformed(CampaignGUI.java:1384)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.AbstractButton.doClick(Unknown Source)
	at javax.swing.plaf.basic.BasicMenuItemUI.doClick(Unknown Source)
	at javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
[...]

PR #1642, opened 2020-04-07
Fixing Popup Display Text Spacing Issues in DataLoadingDialog

Windchild292:
This fixes a few issues I noticed when working on #1636, notably that the display did not show the spaces as expected which forced words together.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
@@ -1,5 +1,5 @@
 /*
- * MegaMek - Copyright (C) 2000-2002 Ben Mazur (bmazur@sev.org)
+ *  MegaMek - Copyright (C) 2000-2002 Ben Mazur (bmazur@sev.org)
  *
  *  This program is free software; you can redistribute it and/or modify it
  *  under the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
```diff
@@ -212,13 +202,13 @@ public Void doInBackground() {
...
-                                    + "before starting a new version of MekHQ.\nIf you believe the units"
-                                    + "listed are not customs, then try deleting the file data/mechfiles/units.cache"
-                                    + "and restarting MekHQ.\nIt is also possible that unit chassi"
-                                    + "and model names have changed across versions of MegaMek."
-                                    + "You can check this by\nopening up MegaMek and searching for the units."
+                            "The following units could not be loaded by the campaign:\n "
+                                    + e.getError() + "\n\nPlease be sure to copy over any custom units "
+                                    + "before starting a new version of MekHQ.\nIf you believe the units "
+                                    + "listed are not customs, then try deleting the file data/mechfiles/units.cache "
+                                    + "and restarting MekHQ.\nIt is also possible that unit chassi "
```
NickAragua, line 209:
```suggestion
                                    + "and restarting MekHQ.\nIt is also possible that unit chassis "
```

Windchild292: Actually, chassi is proper English in this case. This is because Chassi is the plural of chassis, which is what makes the most sense for the dialog.

Windchild292:
A rare PR for me because of that @NickAragua 😆

PR #1645, opened 2020-04-09
Fix undesirable deployment zone for reinforcements

NickAragua:
This small change fixes a behavior where any forces assigned to most AtB scenarios (especially ones generated from templates) as reinforcements would default to ANY as a deployment zone, which makes very little sense.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
@@ -598,6 +598,8 @@ public class ScenarioForceTemplate implements Comparable<ScenarioForceTemplate> 
         rft.setForceName(REINFORCEMENT_TEMPLATE_ID);
         rft.setGenerationMethod(ForceGenerationMethod.PlayerSupplied.ordinal());
         rft.setGenerationOrder(1);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
```diff
@@ -598,6 +598,8 @@ public static ScenarioForceTemplate getDefaultReinforcementsTemplate() {
         rft.setForceName(REINFORCEMENT_TEMPLATE_ID);
         rft.setGenerationMethod(ForceGenerationMethod.PlayerSupplied.ordinal());
         rft.setGenerationOrder(1);
+        rft.setActualDeploymentZone(Board.START_NONE);
+        rft.setSyncDeploymentType(SynchronizedDeploymentType.None);
```
Windchild292, line 602: Wouldn't this now generate a random edge to enter from? Do we want it to instead be based on the player's edge as the base?

NickAragua: That's the idea, yes. This is the default template. If someone wants to define a scenario template where the reinforcements come from a fixed position or synchronized with the primary force, they can do so by creating a "Reinforcements" force template within the scenario template, which will override this default template.

PR #1647, opened 2020-04-09
Switching Person to use writeSimpleXmlTag

Windchild292:
This makes it easier for me to add and ensure continuity in the writeToXml functions, and adds a few small methods to the util to do so.


Windchild292:
I actually prefer it hand rolled, makes it easier to fix any issues that come up.

I've fixed far too many bugs that have originated in the marshaller/unmarshaller process.

PR #1648, opened 2020-04-09
Fixing Personnel Import from .psnx File

Windchild292:
This fixes a few issues (the big ones being a pair of NPEs) raised on Discord relating to importing this type of file, and some code cleanup relating to tracking down those issues (adding missing JavaDoc comments and removing useless null checks).


Windchild292:
Misclicked when opening this, still has a bug relating to divorces that I'm going to need to iron out first.

PR #1656, opened 2020-04-10
Avoid NPE in Person::isMothballing

sixlettervariables:
Sometimes a unit would be marked as mothballing but the tech would be cleared, leading to an NPE when checking if the IDs matched. This switches the logic to only search the units the person is tech'ing.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -3635,8 +3635,9 @@ public class Person implements Serializable, MekHqXmlSerializable {
         if (!isTech()) {
             return false;
         }
-        for (Unit u : campaign.getUnits()) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -3635,8 +3635,9 @@ public boolean isMothballing() {
         if (!isTech()) {
             return false;
         }
-        for (Unit u : campaign.getUnits()) {
-            if (u.isMothballing() && u.getTech().getId().equals(id)) {
+        for (UUID unitId : techUnitIds) {
+            Unit u = campaign.getUnit(unitId);
+            if (u != null && u.isMothballing()) {
```
Windchild292, line 3640:
```suggestion
            if ((u != null) && u.isMothballing()) {
```

PR #1659, opened 2020-04-11
Log Maintenance

Windchild292:
This actually implements the option Log Maintenance, which I learned wasn't enabled by Discord.


MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -257,7 +257,8 @@
     private int adminXPPeriod;
     private int edgeCost;
 
-    //repair related
+    //region Repair and Maintenance Tab
```
sixlettervariables, line 260: Can we roll back all but the actual addition of `logMaintenance` ? Order of XML elements is significant, while perhaps we don't define it as such, it impacts XML diff tools' semantic analysis.

Windchild292: I'll roll back that part, but we should probably think about developing a final syntax order instead of the jumble we currently have.

Windchild292: Rolled back the XML Write changes

PR #1679, opened 2020-04-18
ROM Designation Enum

Windchild292:
This moves the ROM Designations integer list to be an Enum instead.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1937,9 +1917,17 @@ public static Person generateInstanceFromXML(Node wn, Campaign c, Version versio
                 } else if (wn2.getNodeName().equalsIgnoreCase("acquisitions")) {
                     retVal.acquisitions = Integer.parseInt(wn2.getTextContent());
                 } else if (wn2.getNodeName().equalsIgnoreCase("primaryDesignator")) {
-                    retVal.primaryDesignator = Integer.parseInt(wn2.getTextContent());
+                    if (version.isLowerThan("0.47.6")) {
```
sixlettervariables, line 1920: Move this to the actual Enum as a parse method (version independent). Have it support text and ordinal values, preferring text and falling back on ordinal.

sixlettervariables: Or store an integer value in the Enum as well if you'd like the order to be modifiable in the future. (Needs a lookup map built in that case)

Also a call to values() creates a new array each time.

Windchild292: Done, and good to know on the latter. I'm thinking that the enum list here won't likely change, so we are good on the lookup map.

PR #1681, opened 2020-04-18
Change refit model weight calculation to work with BA

mkerensky:
Fixes an undocumented bug where the available refit dialog won't appear due to an index out of bounds exception. This happens when refitting squads that have models with different numbers of troopers, such as IS/WoB versions of the Phalanx.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -369,17 +369,29 @@ public class Utilities {
         CampaignOptions options = campaign.getCampaignOptions();
         ArrayList<String> variants = new ArrayList<>();
         for(MechSummary summary : MechSummaryCache.getInstance().getAllMechs()) {
-            // If this isn't the same chassis, is our current unit, or is a different weight we continue
-            if(!en.getChassis().equalsIgnoreCase(summary.getChassis())
+            // If this isn't the same chassis, is our current unit, we continue
+            if (!en.getChassis().equalsIgnoreCase(summary.getChassis())
                     || en.getModel().equalsIgnoreCase(summary.getModel())
-                    || summary.getTons() != en.getWeight()
                     || !summary.getUnitType().equals(UnitType.determineUnitType(en))) {
                 continue;
             }
             // If we only allow canon units and this isn't canon we continue
-            if(!summary.isCanon() && options.allowCanonRefitOnly()) {
+            if (!summary.isCanon() && options.allowCanonRefitOnly()) {
                 continue;
             }
+            
+            // Weight of the two units must match or we continue, but BA weight gets
+            // checked differently
+            if (en instanceof BattleArmor) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/Utilities.java
```diff
@@ -369,17 +369,29 @@ public static File lastFileModified(String dir, FilenameFilter filter) {
...
             // If we only allow canon units and this isn't canon we continue
-            if(!summary.isCanon() && options.allowCanonRefitOnly()) {
+            if (!summary.isCanon() && options.allowCanonRefitOnly()) {
                 continue;
             }
+            
+            // Weight of the two units must match or we continue, but BA weight gets
+            // checked differently
+            if (en instanceof BattleArmor) {
+                if (((BattleArmor)en).getTroopers() != (int) summary.getTWweight()) {
```
Windchild292, line 382:
```suggestion
                if (((BattleArmor) en).getTroopers() != (int) summary.getTWweight()) {
```

MekHQ/src/mekhq/Utilities.java
```diff
@@ -369,17 +369,30 @@ public static File lastFileModified(String dir, FilenameFilter filter) {
         CampaignOptions options = campaign.getCampaignOptions();
         ArrayList<String> variants = new ArrayList<>();
         for(MechSummary summary : MechSummaryCache.getInstance().getAllMechs()) {
-            // If this isn't the same chassis, is our current unit, or is a different weight we continue
-            if(!en.getChassis().equalsIgnoreCase(summary.getChassis())
+            // If this isn't the same chassis, is our current unit, we continue
+            if (!en.getChassis().equalsIgnoreCase(summary.getChassis())
                     || en.getModel().equalsIgnoreCase(summary.getModel())
-                    || summary.getTons() != en.getWeight()
                     || !summary.getUnitType().equals(UnitType.determineUnitType(en))) {
```
Windchild292, line 375: Can we swap this to use UnitType.getTypeName(e.getUnitType()), as the current method is deprecated.

MekHQ/src/mekhq/Utilities.java
Windchild292: Also, should this be based on the ADVANCED_TACOPS_BA_WEIGHT option?

MekHQ/src/mekhq/Utilities.java
mkerensky: Done.

MekHQ/src/mekhq/Utilities.java
mkerensky: No, this should be TWweight to work as intended when compared to getTroopers.  I'm pretty sure MechSummary was written before BA/Infantry was fully implemented, as it doesn't have any functionality for tracking the number of troopers.  Infantry/BA refits bear a more comprehensive look at some point, but that's not a rabbithole I want to go down at present.

PR #1685, opened 2020-04-19
Adding All Role and No Role options for Portraits

Windchild292:
This fills a request from the MegaMek forums, that I thought would be a nice break to implement.


MekHQ/src/mekhq/gui/dialog/CampaignOptionsDialog.java
```diff
@@ -3327,9 +3327,37 @@ public void actionPerformed(ActionEvent e) {
         panRandomPortrait.setName("panRandomPortrait"); // NOI18N
         panRandomPortrait.setLayout(new BorderLayout());
 
-        JPanel panUsePortrait = new JPanel(new GridLayout(nRow, 4));
-
+        JPanel panUsePortrait = new JPanel(new GridLayout((int) Math.ceil((Person.T_NUM + 2) / 4.0), 4));
```
NickAragua, line 3330: This is begging for a comment to explain what all that math is for.

Windchild292: Added

PR #1697, opened 2020-04-27
Fixing Medical Age Display Calculation

Windchild292:


Windchild292:
It was one day off of what it should have been, and calculating using a less efficient method.

PR #1700, opened 2020-04-28
fix phantom scenario assignment npe

NickAragua:
This PR fixes an issue which prevents the personnel tab from displaying when one of the units involved has been assigned to a scenario but has somehow not been properly unassigned.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -4235,7 +4235,7 @@ public class Campaign implements Serializable, ITechManager {
         }
         
         // clean up units that are assigned to non-existing scenarios
-        for(Unit unit : this.getUnits()) {
+        for (Unit unit : this.getUnits()) {
             if(this.getScenario(unit.getScenarioId()) == null) {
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4233,6 +4233,13 @@ public void cleanUp(){
                 force.removeUnit(unitID);
             }
         }
+        
+        // clean up units that are assigned to non-existing scenarios
+        for(Unit unit : this.getUnits()) {
```
Windchild292, line 4238:
```suggestion
        for (Unit unit : this.getUnits()) {
```

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4233,6 +4233,13 @@ public void cleanUp(){
...
+            if(this.getScenario(unit.getScenarioId()) == null) {
```
Windchild292, line 4239:
```suggestion
            if (this.getScenario(unit.getScenarioId()) == null) {
```

PR #1701, opened 2020-04-28
1698: Determine Father Based on Spouse at Time of Birth

### LINKED ISSUE #1698: Determine Father Based on Spouse at Time of Birth
In 47.5 I had two married people; Christopher Bachleda married Ruth Hale and then she got pregnant. 

When Ruth had a baby boy she is listed as Jaicus Hale parent. 

It also lists Christopher as her spouse.  

However Christopher is only listed as having Ruth Hale as a spouse and has no link in his family information to his son.

Jaicus Hale is listed as having Ruth as parent but has no link to Christopher.  

Also when attempting to click Jaicus Hale on the daily log nothing happens.  All other names in the daily log properly link to the personnel records  

Campaign zip with log and attached customs  attached  
[Ulric's Behemoths30581102.zip](https://github.com/MegaMek/mekhq/files/4541827/Ulric.s.Behemoths30581102.zip)

A way to right click on people and add family links would be very helpful

Windchild292:
This fixes #1698 by adding an option to determine the father of a child at the time of birth rather than the time of conception. I also renamed useParentage to displayParentage, as it was only used for display at this point, and moved the baby recruitment around to fix an issue with the baby linking from the daily log.


purpledragon2:
Just wondering why it is done at time of birth instead of time of conception?

PR #1704, opened 2020-04-28
1691: Personnel are now filtered by primary and secondary roles in the Personnel Tab

### LINKED ISSUE #1691: Filter personal by secondary role as well as primary role
Would it be possible to have personal be filtered by secondary roles as well as primary roles in the personal filters?  By this I mean that if my commander is both a MechWarrior and a Admin/Command that he would show up in both the MechWarrior and the Administrators filters.

Windchild292:
This fixes #1691, and implements fixes for a few missing roles in the sorting for the Personnel Tab.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -792,7 +792,8 @@ public class Person implements Serializable, MekHqXmlSerializable {
         this.portraitFileOverride = s;
     }
 
-
+    //region Personnel Roles
+    // TODO : Move me into an enum with checks for hasAdministratorRole, hasTechRole, hasMedicalRole, etc.
     public int getPrimaryRole() {
         return primaryRole;
     }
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -831,6 +832,104 @@ public boolean hasRole(int role) {
...
     }
 
+    /**
+     * This is used to determine whether the person has either a primary or secondary role within the
+     * two bounds. This is inclusive of the two bounds
+     * @param minimumRole the minimum role bound (inclusive)
+     * @param maximumRole the maximum role bound (inclusive)
+     * @return true if they have a role within the bounds (inclusive), otherwise false.
+     */
+    public boolean hasRoleWithin(int minimumRole, int maximumRole) {
```
NickAragua, line 842: I'd almost suggest abstracting these, but it looks like you've got further plans for improvement in this area, so we can leave it as is for now.

Windchild292: I do, this is just a step towards the final plans (likely using an enum, but abstraction is under consideration)

PR #1705, opened 2020-04-28
LGTM and String Comparison Fixes

Windchild292:
This fixes most of the remaining LGTM issues, all remaining string comparisons using ==, adds some null and/or failure protections, and resolves a few other bugs with object comparisons (especially for Person).

To review: Whitespace differences off, and all at once (I overwrote a few of my initial changes later in the PR)

Changes (larger list):
Added logging to IconPackage to help with debugging (this has actually came up recently for me)
Implemented Person::equals(Object), which determines if the two are equal based on their UUID id
Fixed a bunch of String comparisons using == and !=
Fixed some whitespacing issues on for and if statements
Fixed a bunch of NPEs
Removed some redundant comparisons
Added some additional logging for failure cases
Fixed some non-final enum fields (IDE spot-bugs recommendation)


MekHQ/src/mekhq/MekHqXmlUtil.java
```diff
@@ -328,14 +328,14 @@ public static String writeEntityToXmlString(Entity tgtEnt, int indentLvl, ArrayL
         }
 
          if ((null != tgtEnt.getCamoCategory())
-                 && (tgtEnt.getCamoCategory() != IPlayer.NO_CAMO)
+                 && (!IPlayer.NO_CAMO.equals(tgtEnt.getCamoCategory()))
```
NickAragua, line 331: I guess we don't need the enclosing parens if there's no !=

MekHQ/src/mekhq/MekHqXmlUtil.java
```diff
@@ -328,14 +328,14 @@ public static String writeEntityToXmlString(Entity tgtEnt, int indentLvl, ArrayL
...
                  && !tgtEnt.getCamoCategory().isEmpty()) {
              retVal += "\" camoCategory=\"";
              retVal += String.valueOf(escape(tgtEnt.getCamoCategory()));
          }
 
          if ((null != tgtEnt.getCamoFileName())
-                 && (tgtEnt.getCamoFileName() != IPlayer.NO_CAMO)
+                 && (!IPlayer.NO_CAMO.equals(tgtEnt.getCamoFileName()))
```
NickAragua, line 338: Unnecessary enclosing parens

MekHQ/src/mekhq/campaign/mission/AtBScenario.java
```diff
@@ -1183,8 +1182,8 @@ private void addStar(ArrayList<Entity> list, String faction, int skill, int qual
                 list.add(en);
             }
         }
-        if (forceType == FORCE_NOVA || forceType == FORCE_VEENOVA) {
-            unitType = forceType == FORCE_VEENOVA ? UnitType.INFANTRY : UnitType.BATTLE_ARMOR;
+        if (forceType == FORCE_NOVA) {
```
NickAragua, line 1185: I don't see a reason to eliminate the VEENOVA logic.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1886,12 +1886,9 @@ public void writeToXml(PrintWriter pw1, int indent) {
     public static Person generateInstanceFromXML(Node wn, Campaign c, Version version) {
         final String METHOD_NAME = "generateInstanceFromXML(Node,Campaign,Version)"; //$NON-NLS-1$
 
-        Person retVal = null;
+        Person retVal = new Person(c);
```
NickAragua, line 1889: Do we really want a non-null person in case we fail to read a person in?

MekHQ/src/mekhq/campaign/mission/AtBScenario.java
Windchild292: Unused and has never been used, years after it was originally implemented.

MekHQ/src/mekhq/campaign/personnel/Person.java
Windchild292: We already will have a non-null person, as the only way it will return null currently is if it fails to create the person object initially.

However, by adding retVal = null on the catch would do that, which seems to have been the original intent in the code. Pushed a fix to ensure the original intent works.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -923,7 +922,7 @@ private void processPrisonerCapture(List<TestUnit> unitsToProcess) {
                 }
                 if (en.hasWorkingMisc(MiscType.F_COMMAND_CONSOLE)) {
                     for (Person p : crew) {
-                        if (p != commander && p != driver) {
+                        if (!p.equals(commander) && !p.equals(driver)) {
```
sixlettervariables, line 925: I think reference equality is okay here actually. We wouldn't want two instances with the same UUID. In any case this is safe for now.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -2654,6 +2652,25 @@ public String toString() {
...
+     * Two people are determined to be equal if they have the same id
+     * @param object the object to check if it is equal to the person or not
+     * @return true if they have the same id, otherwise false
+     */
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof Person)) {
+            return false;
+        } else {
+            return getId().equals(((Person) object).getId());
```
sixlettervariables, line 2665: Should have a reference equality check first.

Windchild292: Is that for performance reasons?

sixlettervariables: Part of the recommendations for implementing Object::equals.

Windchild292: Implemented the fix

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
Windchild292: It is okay for now, but I'd rather future proof and implement it properly with equals comparisons.

PR #1713, opened 2020-05-01
Marriage Surname Enum

Windchild292:
This moves marriage surnames to an enum and adds spaced formats.

It also makes the surname type make a lot more sense to anyone who isn't me (this originates with a complaint I got, which makes a bunch of sense), with styles renamed to be in a single, consistent system (i.e. anything yours changes your spouse's surname, anything spouse changes your surname).

The spaced formats were missed when I finished working on the name changes, and as such are implemented here.

The migration code is not great nor nice, but it does work well. Alternatively, this could be ignored during migration as it has only been there since 0.47.5.


Windchild292:
I'm moving this to draft as I will be moving all the marriage code into the enum.

Windchild292:
Opening it up as I've finished the rename and moving the code into the enum instead.

PR #1714, opened 2020-05-01
Random Names Rework Part 2 - MekHQ Changes

Windchild292:
This is part two of my rework of the random name system.

The MekHQ change is that the origin faction is now used to generate instead of the player's faction, as that provides a better format for generating names.

This looks far worse a PR than it is because of the data file changes. Most of the files that are changed are just a few lines of changes, but gender and the generators do touch on a bunch of files. The data files MUST be updated immediately or MekHQ will not be able to generate names.


MekHQ/src/mekhq/campaign/mission/AtBScenario.java
```diff
@@ -1660,7 +1660,7 @@ public static String writeEntityWithCrewToXmlString(Entity tgtEnt, int indentLvl
             crew.append(MekHqXmlUtil.indentStr(indentLvl + 2)).append("<crewMember slot=\"")
                     .append(pos).append("\" name=\"").append(MekHqXmlUtil.escape(tgtEnt.getCrew().getName(pos)))
                     .append("\" nick=\"").append(MekHqXmlUtil.escape(tgtEnt.getCrew().getNickname(pos)))
-                    .append("\" gender=\"").append(tgtEnt.getCrew().getGender(pos))
+                    .append("\" gender=\"").append(tgtEnt.getCrew().getGender(pos).name())
```
NickAragua, line 1663: Is this the code that writes out a MUL entry? Can the MUL Parser handle this change?

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -2103,7 +2097,7 @@ public static Person generateInstanceFromXML(Node wn, Campaign c, Version versio
                 } else if (wn2.getNodeName().equalsIgnoreCase("hits")) {
                     retVal.hits = Integer.parseInt(wn2.getTextContent());
                 } else if (wn2.getNodeName().equalsIgnoreCase("gender")) {
-                    retVal.gender = Integer.parseInt(wn2.getTextContent());
+                    retVal.gender = Gender.parseFromString(wn2.getTextContent().trim());
```
NickAragua, line 2100: Does parseFromString handle gender entries from older campaigns?

Windchild292: It does. It defaults to valueOf but does enable integer parsing.

MekHQ/src/mekhq/campaign/mission/AtBScenario.java
Windchild292: It can, using the same Gender::parseFromString method as Person does to load.

PR #1715, opened 2020-05-01
Prisoner Status Enum

Windchild292:
This PR moves Prisoner Status to an Enum, and changes the logic to occur in a single method instead of across multiple files (which fixes a few minor bugs originating from the spread out code)

This was a much larger PR than I expected (I am sorry about this, but it does fix a few minor bugs because of a lack of standardization), and should be reviewed with whitespace differences off.


MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -951,24 +952,21 @@ private void processPrisonerCapture(List<TestUnit> unitsToProcess) {
...
-                // Give them a UUID. We won't actually use this for the campaign, but to
-                //identify them in the prisonerStatus hash
-                UUID id = p.getId();
-                if (null == id) {
-                    id = UUID.randomUUID();
-                    p.setId(id);
-                }
-                PrisonerStatus status = new PrisonerStatus(p.getFullName(), u.getEntity().getDisplayName(), p);
+            for (Person p : crew) {
+                // We need to assign them a UUID now to place them into the hash
```
NickAragua, line 959: You probably still need to say which hash.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -516,44 +507,75 @@ public void setDependent(boolean tf) {
...
 
-    public int getPrisonerStatus() {
-        return prisonerStatus;
-    }
+        // Now, we need to fix values and ranks based on the Person's status
+        switch (prisonerStatus) {
+            case PRISONER:
+            case PRISONER_DEFECTOR:
+                // They don't get to have a rank. Their rank is Bondsman.
+                getCampaign().changeRank(this, Ranks.RANK_BONDSMAN, true);
```
NickAragua, line 527: Seems almost identical to the Bondsman functionality. Is this a) intended, and b) if it is, can it be consolidated?

Windchild292: It is intended, albeit with a typo on the line in question (it should say RANK_PRISONER). However, I consolidated it. I'm also going to create a story to remove the RANK_BONDSMAN and RANK_PRISONER ranks and replacement with leading displays for ranks, as that is significantly more realistic.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
Windchild292: Added clarification to the comment in question.

PR #1756, opened 2020-05-19
fix sandblaster issues

NickAragua:
A recent MegaMek PR changed the Sandblaster SPA so that it points to a particular weapon, but caused a bunch of issues in MekHQ with personnel generation, in multiple areas. This PR fixes those errors, and also limits randomly generated personnel with Sandblaster to just weapons that can potentially use cluster tables.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
+++ b/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
@@ -114,6 +114,10 @@ public class CrewSkillUpgrader {
         // that cost less XP than the remaining cap
         for(int x = 0; (x < spaCap) && (xpCap > minAbilityCost); x++) {
             int spaCost = addSingleSPA(entity, xpCap);
+            // if we didn't assign an SPA, let's reroll it.
+            if (spaCost == 0) {
+                x--;
+            }
             xpCap -= spaCost;
         }
     }
@@ -176,6 +180,11 @@ public class CrewSkillUpgrader {
         default:
             entity.getCrew().getOptions().getOption(spa.getName()).setValue(true);
             return spa.getCost();
+        }
+        
+        // if we fail to pick a random weapon/specialization for whatever reason, don't assign the SPA
+        if (spaValue.equals(Crew.SPECIAL_NONE)) {
+            return 0;
         }
         
         entity.getCrew().getOptions().getOption(spa.getName()).setValue(spaValue);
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
```diff
@@ -216,7 +221,16 @@ private boolean extraEligibilityCheck(SpecialAbility spa, Entity entity) {
...
+        List<Mounted> eligibleWeapons = new ArrayList<>();
+        
+        for(Mounted weapon : weapons) {
+            if(SpecialAbility.isWeaponEligibleForSPA(weapon.getType(), Person.T_NONE, clusterOnly)) {
+                eligibleWeapons.add(weapon);
+            }
+        }
+        
         int weaponIndex = Compute.randomInt(entity.getIndividualWeaponList().size());
         return entity.getIndividualWeaponList().get(weaponIndex).getName();
```
neoancient, line 235: After building a list of all eligible weapons, you're still picking randomly from all weapons.

MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
```diff
@@ -216,9 +221,18 @@ private boolean extraEligibilityCheck(SpecialAbility spa, Entity entity) {
...
+        List<Mounted> eligibleWeapons = new ArrayList<>();
+        
+        for(Mounted weapon : weapons) {
+            if(SpecialAbility.isWeaponEligibleForSPA(weapon.getType(), Person.T_NONE, clusterOnly)) {
+                eligibleWeapons.add(weapon);
+            }
+        }
+        
+        int weaponIndex = Compute.randomInt(eligibleWeapons.size());
+        return eligibleWeapons.get(weaponIndex).getName();
```
neoancient, line 235: What happens if the entity doesn't have any eligible weapons?

NickAragua: NPE, probably. I'll add some more defensive coding when I get back to it tonight.

NickAragua: Fixed. I suppose I could have it re-roll in that case, but it's a lot more work and I don't think anybody's going to complain about bot unit crews having a tiny amount less SPAs.

neoancient: Does this have the effect of giving the crew the SPA, but making it ineffective because it doesn't match any weapon?

NickAragua: Basically. I suppose a better solution would be to have the calling function check whether this method returned anything and avoid adding the SPA in that case.

neoancient: It's functional as is, but probably clearer without the non-functional SPA.

Issue #1447, opened 2020-02-02
Planetary Acquisition can infinite loop under certain conditions


sixlettervariables:
Updated information: It appears an issue with Planetary acquisition in Campaign Options is the root of this bug.  While solved on a short term basis, leaving this for future attempts to resolve the underlying problem for others.  (Turning Planetary acquisition off solved all the problems, my suspicion is it's related to purchasing in general, but I'm not a coder anymore, and I have no clue.)

Edit note: Other issue about internal to MHQ MML in unoffical tech level *not* showing battlearmor bays in Transport bays remains.

_Originally posted by @MageOhki in https://github.com/MegaMek/mekhq/issues/1444#issuecomment-581164000_

servo01:
ANY part acquisition roll is causing my fresh install of 47.3 to lock completely. - Even with planetary acquisition turned off.

sixlettervariables:
Fixed that as well @servo01, thanks for the heads up.

sixlettervariables:
Fixed in #1448.

PR #1758, opened 2020-05-20
Planet acquisition fixes

AaronGullickson:
I noticed a few issues with planetary acquisitions that I wanted to clean up for the stable if possible, especially now that planetary acquisitions should work nicely with the new planetary system data. 

I also see that there was an issue with infinite loops in #1447  that was apparently fixed with #1451. Is that fully resolved or are there still issues with planetary acquisitions?

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2729,7 +2729,7 @@ public class Campaign implements Serializable, ITechManager {
      * @return true if your target roll succeeded.
      */
     public boolean findContactForAcquisition(IAcquisitionWork acquisition, Person person, PlanetarySystem system) {
-        if (person.getAcquisitions() >= getCampaignOptions().getMaxAcquisitions()) {
+        if (null == person || !canAcquireParts(person)) {
             return false;
         }
         DateTime currentDate = Utilities.getDateTimeDay(getCalendar());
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -2729,7 +2729,7 @@ public boolean canPayFor(IAcquisitionWork acquisition) {
      * @return true if your target roll succeeded.
      */
     public boolean findContactForAcquisition(IAcquisitionWork acquisition, Person person, PlanetarySystem system) {
-        if (person.getAcquisitions() >= getCampaignOptions().getMaxAcquisitions()) {
+        if (null == person || !canAcquireParts(person)) {
```
AaronGullickson, line 2732: This was causing the method to return false when the user left max acquisitions at zero which is supposed to be unlimited. Using the `canAcquireParts` method instead fixes this because it already checks that properly.

MekHQ/src/mekhq/campaign/parts/AmmoStorage.java
```diff
@@ -283,7 +283,6 @@ public Object getNewEquipment() {
 
     @Override
     public String failToFind() {
-        resetDaysToWait();
```
AaronGullickson, line 286: The AmmoStorage, Armor, and Refit parts objects all add `resetDaystoWait()` to the `failToFind` method. However, this is redundant, because each of the `Campaign.goShopping` methods already does this reset at the proper place for all IAcquisitionWork. In this case, it was causing some issues with planetary acquisition because we don't want to reset that until we have failed to find stuff on all the planets on our list.

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: This is already checked in goShoppingByPlanet(ShoppingList), which is the only place this method is called. A null person is also not possible from the current path.

AaronGullickson: Good point. I have removed this entirely now as its redundant.

Issue #347, opened 2017-03-14
Feature Request:TOE Layered Icon Changes


trent1098s:
Three simple requests in one;

#1 In the XML file for the campaign the items selected on each layer are stored in the XML file. Can a change be made to the Change Force Icon context menu code so that when the dialog opens, the _currently selected layers_ are pre-populated? Currently if you build a layered icon then edit the layered icon you have to start from scratch again. (This is aggravating if you have a 4 or 5 layer icon and just needed to tweak the Formation, or whatever, you have to go from tab to tab rebuilding the icon from scratch)

#2 A "none" selection in each tab for building the layered icon. Currently if you accidentally click/tap a section (such as select a background image) there is no way of clearing that out other than starting all over again.

#3 Add "Copy Icon" to the context menu and "Paste Icon" to the context menu on the TOE page. This would allow you to more quickly build a TOE by copying an icon from one force to another. 

These three things would rapidly accelerate the TOE building process for larger forces. I'm running a campaign where there are four regimental sized units on the planet plus air wing and associated other units, and it took 8 hours just to build the icons for the TOE on the units. :)

Dylan-M:
1. This is a known limitation to an unfinished feature. It isn't trivial to do.

2. You can clear it using ctrl-click for most, if not all, selections. A "None" is planned.

3. This isn't trivial either.

Regardless, these should be three separate tickets. Please see:
https://github.com/MegaMek/megamek/wiki/Making-a-Bug-Report

Specifically the section regarding One Issue Per Ticket

trent1098s:
Thanks, in the future I'll split up requests to multiple tickets. 

It doesn't sound like there's interest in doing these changes; I can probably just write something in Visual Studio that lets me manipulate these campaign files outside of mekHQ, so I can set up the icons in bulk. It's a real time consuming nightmare trying to do a 'prettied up' regiment size unit in the current interface.

For completeness this is the data relating to the above in the campaign file.

						<force id="81" type="mekhq.campaign.force.Force">
							<name>1st Squadron</name>
							<desc></desc>
							<iconCategory>Layered</iconCategory>
							<iconHashMap>
								<iconentry key="Pieces/Frames/">
									<value name="Frame.png"/>
								</iconentry>
								<iconentry key="Pieces/Type/">
									<value name="Vehicle Tracked.png"/>
								</iconentry>
								<iconentry key="Pieces/Formations/">
									<value name="08 Battalion.png"/>
								</iconentry>
[...]

PR #1761, opened 2020-05-22
347/1662: Remember Layered Force Icon Setup

### LINKED ISSUE #1662: Enhancement Request: Remember Layered Force Icon Settings
So I use a lot of layered icon in my game, and I have a tendency to change them pretty often to denote new situations (mainly how damaged a lance, etc is). It's pretty time consuming to redo each individual part of the layered icon each time I make a change, and I was wondering if it would be a lot of work for the system to remember the current settings for a force icon when you go in to modify it?

Untitled

Windchild292:
This fixes #1662 and two of the three requests in #347 (copy/paste icons being left). The last one is possible, but would require significant refactoring outside of the scope of this.

There's a bunch of rework and refactoring to do this, and I'd recommend reviewing with whitespace differences off.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
@@ -80,7 +80,6 @@ public class ImageChoiceDialog extends JDialog {
     private boolean changed = false;
 
     //region Layered Images Support
-    private ImageIcon imageIcon = null;
     private JLabel preview = new JLabel();
     private JTabbedPane tabbedPane = new JTabbedPane();
     private JTabbedPane layerTabs = new JTabbedPane();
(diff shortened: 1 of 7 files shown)


trent1098s:
Wow, this was an old request. Thanks for putting in the work @Windchild292!!

rroby1:
Absolutely thank you. I’ve been kicking around writing a Python script to do it for myself. Mainly to help learn Python, but also because it is a pain to work with the icons. Thank you @Windchild292 for this. It’s a great start.

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
```diff
@@ -74,125 +76,100 @@
...
+        String match = null;
+        categoryModel.addElement(Crew.ROOT_PORTRAIT);
+        Iterator<String> names = (imageItems != null)
+                ? imageItems.getCategoryNames() : Collections.emptyIterator();
+        while (names.hasNext()) {
+            String name = names.next();
+            if (!"".equals(name)) {
+                categoryModel.addElement(name);
+                if (category.equals(name)) {
+                    match = name;
```
sixlettervariables, line 148: Should you break here?

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
```diff
@@ -244,132 +200,67 @@ private void initComponents() {
...
+
+                // Initialize Initial Values, provided the Icon Map is not empty
+                if (!iconMap.isEmpty()) {
+                    if (iconMap.containsKey(layeredForceIcons[i].getLayerPath())) {
+                        if (layeredForceIcons[i].getListSelectionModel() == ListSelectionModel.SINGLE_SELECTION) {
+                            // Determine the current selected value
+                            String selected = iconMap.get(layeredForceIcons[i].getLayerPath()).get(0);
+                            for (int k = 0; k < tableModel.getRowCount(); k++) {
+                                if (tableModel.getValueAt(k, 0).equals(selected)) {
+                                    layeredTables[i].setRowSelectionInterval(k, k);
```
sixlettervariables, line 241: A comment as to why this is `, k` would help. The iteration makes it seem like it would be `, 0` or `, 1`.

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
```diff
@@ -478,99 +382,39 @@ public String getFileName() {
...
-            iconMap.remove(IconPackage.FORCE_LOGOS, iconMap.get(IconPackage.FORCE_LOGOS));
-        } else {
-            tmp = new Vector<>();
-            for (int index : tableLogos.getSelectedRows()) {
-                tmp.add((String) tableLogos.getValueAt(index, 0));
-                iconMap.put(IconPackage.FORCE_LOGOS, tmp);
+        Vector<String> temp;
+        LayeredForceIcon[] layeredForceIcons = LayeredForceIcon.values();
+        for (int i = 0; i < layeredTables.length; i++) {
+            if (layeredTables[i].getSelectedRow() <= 0) {
```
sixlettervariables, line 397: A comment as to why this is `<=` (no selection or the first) would be good.

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
Windchild292: It can only match once, so it definitely should.  Fixed.

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
Windchild292: Added.

MekHQ/src/mekhq/gui/dialog/ImageChoiceDialog.java
Windchild292: This is just because of how the backend is done, unfortunately. It actually sets it as having selected from k to k, inclusively, although the single selection only actually cares about the second value... Not the easiest logic to read, and I've added a comment for both places this occurs.

Windchild292:
Welcome @trent1098s and @rroby1 :)

PR #1765, opened 2020-05-24
1569/1570 Phase 1: Genealogy

Windchild292:
This is the Genealogy rework, which removes Ancestors and replaces it with proper Genealogical tracking for family units (parents, siblings, children, spouse, former spouses).
This decreasing the size of Person by roughly 500 lines of code while making it easier to modify genealogical connections for future development.

Please note that as part of Phase 1, this adds some code that is not used currently (mostly relating to family relationship types and connections). This code will be used or removed before the two linked stories are merged.

Here are the two campaign files I primarily used while testing, which have significant familial groups:
[TestCampaigns.zip](https://github.com/MegaMek/mekhq/files/4673796/TestCampaigns.zip)


MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
```diff
@@ -0,0 +1,504 @@
...
+
+    /**
+     * @return true if the Person has any parents, otherwise false
+     */
+    public boolean hasParents() {
+        return (getFamily().get(FamilialRelationshipType.PARENT) != null);
+    }
+
+    // Recursive Search Format to determine mutual ancestry
+    public boolean checkMutualAncestors(Person person, Campaign campaign) {
```
sixlettervariables, line 178: I think this is ripe for unit tests, along with removing relations to make sure the tree is pruned.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1901,4 +1861,72 @@ private static void updatePlanetaryEventsFromXML(Node wn) {
             }
         }
     }
+
+
+    //region Migration Methods
+    //region Ancestry Migration
+    private static Map<UUID, List<Person>> ancestryMigrationMap = new HashMap<>();
+
+    public static void addToAncestryMigrationMap(UUID ancestorsId, Person person) {
```
NickAragua, line 1870: Let's put a comment on this public method.

MekHQ/src/mekhq/campaign/personnel/enums/FamilialRelationshipType.java
```diff
@@ -0,0 +1,172 @@
...
+package mekhq.campaign.personnel.enums;
+
+import megamek.common.Crew;
+import megamek.common.util.EncodeControl;
+import mekhq.MekHQ;
+import mekhq.campaign.personnel.Person;
+
+import java.util.ResourceBundle;
+
+public enum FamilialRelationshipType {
```
NickAragua, line 28: Seems like some comments would be helpful for this enum in general.

MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
NickAragua: And while you're at it, let's put some comments on these public methods, as I don't really have any idea what they do just looking at the name.

Windchild292:
Added the missing JavaDoc comments, but discovered two issues that will require some deeper thought (both relating to sibling tracking).

Issue #1768, opened 2020-05-27
Scenario view in Briefing room cuts off TO&E


AaronGullickson:
### Environment
MekHQ v0.47.6
Mac OSX 
Java 8

### Description
The view of the TO&E in the scenario view on the Briefing Room tab is cut off. You can use the arrow key to move up and down it but you can't see more than the first few units listed. I have experimented with multiple themes and the problem persists. The two text areas also take up a bunch of unnecessary empty space. 

<img width="588" alt="Screen Shot 2020-05-27 at 10 07 23 AM" src="https://user-images.githubusercontent.com/16107446/83051093-5a162180-a002-11ea-8b3b-da2ba2c65010.png">

I suspect the issue has to do with some AtB GUI elements that are not present in this campaign and the way free space is being picked up.

Windchild292:
I wonder if this is the same as #1506

AaronGullickson:
Yes, it is. I looked for other reports, but my search-fu failed. I will add it to the linked issues.

PR #1769, opened 2020-05-27
Briefing room fixes

AaronGullickson:
This PR addresses several graphical issues in the Briefing Room tab. The main fix is for #1768 where the TO&E reports for a scenario were being cut off. However, I was also annoyed about how the various text areas were leaving a bunch of white space, so I made some changes to the grid bag layout to fix that.

Here is a shot of what the it looks like before:
Screen Shot 2020-05-27 at 2 06 21 PM

Here is a shot of what it looks like after:
Screen Shot 2020-05-27 at 2 05 45 PM

I did have to unsnarl a bit of code to get it all working. Most importantly I passed the scenarioTable object into the MissionViewPanel where it is displayed so that there is only one scroll bar on the left. In order to do that without duplicating a bunch of code, I removed the separate ContractViewPanel and AtBContractViewPanel classes and consolidated them with MissionViewPanel. The only area where they differed was the `fillStats` function so this was relatively straightforward.


MekHQ/src/mekhq/gui/view/MissionViewPanel.java
```diff
@@ -9,15 +9,17 @@
 import java.awt.*;
```
sixlettervariables, line 9: Ended up mixed tabs and spaces.

AaronGullickson: Just switched over to intelliJ and probably need to futz with some settings

sixlettervariables: I need to get us an editorconfig so they'll all just work.

AaronGullickson: Tabs should all be gone now.

PR #1779, opened 2020-06-01
1766: Adding an option for attached units to use the faction camo

### LINKED ISSUE #1766: Attached units using faction camo
Is it possible to have allied 'Mechs that are player controlled use that allied faction camo instead of the player camo?

Windchild292:
This fixes #1766


NickAragua:
Does this do the trick for standard AtB scenarios as well?

Windchild292:
It does, both generation types use that method for moving attached units to the player's control.

PR #1794, opened 2020-06-03
1781: Adding Sorting to the Scenario Table

### LINKED ISSUE #1781: RFE: Ascending/Descending Date sort for scenarios in MHQ Briefing room Tab
I propose changing the date order of the scenarios section of the briefing room tab so that the most recent scenario is displayed on top, instead of the current sort, which shows the oldest scenario on top. This will improve the quality of life for long running contracts, especially in the case of garrison contracts.

Windchild292:
This fixes #1781 by adding sorting to the scenario table, fixing a few NPEs, adding a missing MissionRemovedEvent, fixing the ScenarioRemovedEvent, and fixing the GM button refresh.

The NPEs were in ScenarioTableModel.java according to the logs, but I was unable to originally figure out why they were happening (did not expect it to be the wrong event firing when a scenario was removed). After trying a bunch of things out, I found a comparable model (the UnitMarketTableModel), and refactored the two to be nigh-on identical while actually using the inherited methods of their parent properly. That is why those two files are both included with heavy refactors, even though they would otherwise be out of scope.


MekHQ/src/mekhq/gui/sorter/DateStringComparator.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public int compare(String o1, String o2) {
+        LocalDate dateA;
+        LocalDate dateB;
+        try {
+            dateA = LocalDate.parse(o1, DateTimeFormatter.ofPattern(
+                    campaign.getCampaignOptions().getDisplayDateFormat()));
+        } catch (Exception ignored) {
+            if ("-".equals(o2) && "-".equals(o1)) {
```
sixlettervariables, line 45: Why not try these first?

MekHQ/src/mekhq/gui/sorter/ScenarioStatusComparator.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public int compare(String o1, String o2) {
+        int a = -1;
+        int b = -1;
+
+        // First we need to determine the numbers based on the name
+        for (int i = 0; i < Scenario.S_NUM; i++) {
+            if (Scenario.getStatusName(i).equals(o1)) {
+                a = i;
+            } else if (Scenario.getStatusName(i).equals(o2)) {
```
sixlettervariables, line 38: Could `o1` and `o2` be equal?

MekHQ/src/mekhq/gui/sorter/ScenarioStatusComparator.java
```diff
@@ -0,0 +1,65 @@
...
+                b = i;
+            }
+        }
+
+        // Now we need to fix the references to Defeat and Draw so they sort nicely
+        switch (a) {
+            case 3:
```
sixlettervariables, line 45: Comments for which is defeat and which is draw would be nice.

MekHQ/src/mekhq/gui/sorter/ScenarioStatusComparator.java
```diff
@@ -0,0 +1,65 @@
...
+public class ScenarioStatusComparator implements Comparator<String>, Serializable {
+    private static final long serialVersionUID = -6287998488809978029L;
+
+    @Override
+    public int compare(String o1, String o2) {
+        int a = -1;
+        int b = -1;
+
+        // First we need to determine the numbers based on the name
+        for (int i = 0; i < Scenario.S_NUM; i++) {
```
sixlettervariables, line 35: I think it would be more natural to have a helper method that converted the string to its sort order integer, then just compare the two integers.

MekHQ/src/mekhq/gui/sorter/DateStringComparator.java
Windchild292: Fixed, airhead mistake

Windchild292:
Applied changes from review and fixed a mistake whereby the sorting for incorrect values was done improperly.

MekHQ/src/mekhq/gui/sorter/DateStringComparator.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private Campaign campaign;
+
+    public DateStringComparator(Campaign campaign) {
+        this.campaign = campaign;
+    }
+
+    @Override
+    public int compare(String o1, String o2) {
+        if (o1.equals(o2)) {
```
sixlettervariables, line 39: This will fail on a null string, prefer `Objects.equals` here to make that case moot.

MekHQ/src/mekhq/gui/sorter/ScenarioStatusComparator.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * This class compares two {@link Scenario} status integers
+ */
+public class ScenarioStatusComparator implements Comparator<String>, Serializable {
+    private static final long serialVersionUID = -6287998488809978029L;
+
+    @Override
+    public int compare(String o1, String o2) {
+        if (o1.equals(o2)) {
```
sixlettervariables, line 35: Null check issues again.

Windchild292:
Fixed null handling issues.

PR #1802, opened 2020-06-12
Updates to special abilities in all presets

AaronGullickson:
This PR updates all of the presets that ship with MHQ to include some SPAs that have not been included. This varied from preset to preset, but in all cases it included Zweihander and all of the options for tech special abilities except clan tech knowledge. I have also set default values for all of these special abilities in the defaultspa.xml file. I have tried to be consistent with the existing point ranges and weights for all presets.

### DIFF
--- a/MekHQ/data/universe/defaultspa.xml
+++ b/MekHQ/data/universe/defaultspa.xml
@@ -79,6 +79,7 @@
 		<xpCost>15</xpCost>
 		<weight>6</weight>
         <prereqAbilities>melee_specialist</prereqAbilities>
+        <invalidAbilities>zweihander</invalidAbilities>
         <skillPrereq>
 	    	<skill>Piloting/Mech::Regular</skill>
 	        <skill>Gunnery/Protomech::Regular</skill>
@@ -89,6 +90,7 @@
         <xpCost>10</xpCost>
         <weight>6</weight>
         <prereqAbilities>melee_specialist</prereqAbilities>
+        <invalidAbilities>melee_master</invalidAbilities>
         <skillPrereq>
             <skill>Piloting/Mech::Regular</skill>
         </skillPrereq>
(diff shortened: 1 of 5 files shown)


MekHQ/data/universe/defaultspa.xml
```diff
@@ -78,11 +78,23 @@
...
 	    	<skill>Piloting/Mech::Regular</skill>
 	        <skill>Gunnery/Protomech::Regular</skill>
 	    </skillPrereq>
 	</ability>
+    <ability>
+        <lookupName>zweihander</lookupName>
+        <xpCost>10</xpCost>
+        <weight>6</weight>
+        <prereqAbilities>melee_specialist</prereqAbilities>
+        <invalidAbilities>melee_master</invalidAbilities>
```
AaronGullickson, line 93: Its not exactly clear from Camo Ops, but it seems like the intent is that Zweihander and melee master are intended to be mutually exclusive. Users can change this if they wish.

MekHQ/data/universe/defaultspa.xml
```diff
@@ -212,7 +224,8 @@
 	        <skill>Gunnery/Vehicle::Regular</skill>
 	        <skill>Gunnery/Battlesuit::Regular</skill>
 	        <skill>Gunnery/Protomech::Regular</skill>
-	    </skillPrereq>
+            <skill>Artillery::Regular</skill>
```
AaronGullickson, line 227: I added artillery as a skill prereq for oblique artileryman

AaronGullickson:
Note that a lot of the changes in specific presets are due to new options that have been added since we last updated the presets that are set to their default value.

AaronGullickson:
Yes, they are all working. I built them through MekHQ using the save as preset option. I agree that the descriptions are probably redundant. But its probably because the presets are just saved from an existing set of campaign options.

PR #1830, opened 2020-06-30
order refit techs by role, then ascending skill

NickAragua:
When picking techs to carry out refits, it's very annoying to have to search through a large list looking for the right type of tech. Now, it's ordered by tech type (mech/aero/ba/etc), then skill in descending order.


NickAragua:
The existing code constructed a hash look up so that when the showdialog call returns, the result can be used to look up the tech ID. I left that logic alone for now, although it can definitely be done better.

PR #1835, opened 2020-07-03
DateTime to LocalDate Switchover

### LINKED ISSUE #1797: News.xml not being displayed in the daily feed
Mekhq47.6
Mac

47.6 does not appear to be reading the news.xml into the daily feed. I started a fresh campaign, set the date to the day before a news item (10/21/3013) and advanced the day and saw nothing.

I *think* I saw it in previous versions, around Victor SD's birthday, but I couldn't replicate it this morning.

### LINKED ISSUE #1820: 47.7 Enabling - Faction Specific - for System of Origin has no effect
47.7
Linux
Java 11.0.5

Enabling Faction Specific from the edit personnel screen does not limit the range of systems/planets available. 

To recreate:
Main Screen/Personnel/GM Mode/Edit/Faction Specific

Windchild292:
This removes Joda DateTime completely from MekHQ, and replaces it with LocalDate. This needed to be done in a single large PR because of the integration between the different places that use DateTime.

Part of this fixes #1797, fixes #1820, because the time zone difference no longer applies.

To review: Whitespace differences off, and I apologize for the difficulty in this review.

### DIFF
--- a/MekHQ/src/mekhq/campaign/universe/Planet.java
+++ b/MekHQ/src/mekhq/campaign/universe/Planet.java
@@ -235,13 +235,10 @@ public class Planet implements Serializable {
                 nameChangeYear = Integer.parseInt(yearString);
             }
 
-            // this is a dirty hack: in order to avoid colliding with faction changes, we
-            // set name changes to be a second into the new year
-            LocalDate nameChangeYearDate = new DateTime(nameChangeYear, 1, 1, 0, 0, 1, 0);
+            LocalDate nameChangeYearDate = LocalDate.ofYearDay(nameChangeYear, 1);
 
             String altName;
             String primaryName = nameString;
-            PlanetaryEvent nameChangeEvent = null;
             int parenIndex = nameString.indexOf('(');
             int closingParenIndex = nameString.indexOf(')');
             // this indicates that there's an (Alternate Name) sequence of some kind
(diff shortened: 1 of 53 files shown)


sixlettervariables:
I'll handle the review on this as I've actually done this before (about a year ago) and just never got around to pushing my changes up for review.

MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
```diff
@@ -226,7 +226,9 @@ public void mouseClicked(MouseEvent e) {
             mainPanel.add(lblDistance, gridBagConstraintsLabels);
 
             JumpPath path = campaign.calculateJumpPath(campaign.getCurrentSystem(), contract.getSystem());
-            int days = (int) Math.ceil(path.getTotalTime(Utilities.getDateTimeDay(contract.getStartDate()),
+
+            int days = (int) Math.ceil(path.getTotalTime(
+                    contract.getStartDate().toInstant().atZone(ZoneId.systemDefault()).toLocalDate(),
```
NickAragua, line 231: This seems more awkward than the previous syntax, is there any way to make it less, uh...

MekHQ/src/mekhq/campaign/universe/Planet.java
```diff
@@ -314,18 +297,22 @@ public Planet(String tsvData, List<DateTime> years) throws Exception {
...
+                    LocalDate eventDate = years.get(x - 3);
+
+                    PlanetaryEvent pe = events.getOrDefault(eventDate, new PlanetaryEvent());
+                    pe.faction = new ArrayList<>();
                     pe.faction.add(newFaction);
-                    pe.date = eventDate;
 
-                    this.eventList.add(pe);
-                    this.events.put(eventDate, pe);
+                    if (!events.containsKey(eventDate)) {
```
NickAragua, line 307: This seems like it's a behavior change from existing behavior, can you go into a little detail on why it was done?

MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
Windchild292: This is temporary, until I switch contracts to LocalDate

NickAragua: Sounds good.

MekHQ/src/mekhq/campaign/universe/Planet.java
Windchild292: This was done to improve the code in question (as a single event can change anything on a given day) by removing the same day event hack of placing an event slightly later in the day, which is no longer possible (LocalDate only stores year, month, day). It was also significantly easier to do this than replace it with LocalDateTime, which would require conversions.

NickAragua: Sounds good in theory, although in practice, we'll need to take a look at the planetary events data and see if we have any events like that which will need to be merged (in order to avoid throwing away any events past the first).

Windchild292: That is already handled in the code, this is just a place that was missed.

Windchild292: Although the changes should use that method, so I'll fit that bit.

Windchild292: And the changes are fixed, it properly adds it to the eventList now without causing issues from duplicate events (which currently will occur).

PR #1842, opened 2020-07-07
Fixing Duplicate Transient Campaign Run Failure

Windchild292:
This fixes MekHQ failing to load a campaign because of the transient property.


NickAragua:
Is this going to wind up bloating the size of the .cpnx files by putting the default events (defined in the data/universe/planetary_systems/system_events.xml) into the save file? I feel like we want to avoid that.

Windchild292:
No, just a mistaken addition on my end (the XML is marked as transient above, not inline)

NickAragua:
Gotcha.

PR #1850, opened 2020-07-09
Phenotype Enum Standardization and Fully Implementing Vehicle, ProtoMech, and Naval Phenotypes

Windchild292:
This moves Phenotypes into a single enum instead of two divergent sets of Magic Numbers, and finishes the implementation of Naval, Vehicle, and ProtoMech phenotype tracking.

To review: Whitespace differences off, and I'd go through the files with few changes first then the files with more changes.


MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -1301,66 +1318,6 @@ public static Entity createEntityWithCrew(String faction, int skill, Campaign ca
...
-     * Generates a new Entity without using a RAT. Useful for "persistent" or fixed units.
-     *
-     * @param name            Full name (chassis + model) of the entity to generate.
-     * @param fName            Faction code to use for crew name generation
-     * @param skill            RandomSkillsGenerator.L_* constant for the average force skill level.
-     * @param campaign
-     * @return                A new Entity
-     */
-    @Deprecated //Deprecated 15-Feb-2020 as unused, needs the current name system to be implemented if activated
-    private static Entity getEntityByName(String name, String fName, int skill, Campaign campaign) {
```
NickAragua, line 1314: Let's leave this here. It is extremely useful for debugging, or if you want to generate some specific units.

MekHQ/src/mekhq/campaign/personnel/Clan.java
```diff
@@ -0,0 +1,293 @@
...
+
+    public static Collection<Clan> getClans() {
+        return allClans.values();
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public String getGenerationCode() {
```
NickAragua, line 91: Let's go through this file and put some documentation on the methods.

NickAragua: At the very least the public ones.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -2156,6 +2109,36 @@ public static Person generateInstanceFromXML(Node wn, Campaign c, Version versio
                 retVal.setExpectedDueDate(retVal.getDueDate());
             }
 
+            //versions before 0.3.4 did not have proper clan phenotypes
+            if (version.isLowerThan("0.3.4") && c.getFaction().isClan()) {
```
NickAragua, line 2113: I wonder if anyone actually has saves from before 3.4. I would suggest moving this out into a separate method.

Windchild292: Next part of my person changes following this and genealogy is going to be reworking the xml read/write into a simpler format. This will be part of doing so.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
Windchild292: This will need to be modified then, as it doesn't work properly as it is currently because of personnel changes, and has not since February. If you want me to I can restore it and fix it?

MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
```diff
@@ -1315,36 +1295,39 @@ public int getAge() {
...
+                    case ELEMENTAL:
+                        skillBonus.get(SkillType.S_GUN_BA).setValue(1);
+                        break;
+                    case VEHICLE:
+                        skillBonus.get(SkillType.S_GUN_VEE).setValue(1);
+                        skillBonus.get(SkillType.S_PILOT_GVEE).setValue(1);
+                        skillBonus.get(SkillType.S_PILOT_NVEE).setValue(1);
+                        skillBonus.get(SkillType.S_PILOT_VTOL).setValue(1);
+                        break;
+                }
```
Windchild292, line 1325: Forgot to add to this: Should there be bonuses here for the other phenotypes?

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
Windchild292: Refactored it down so that it just uses the createEntityWithCrew but first grabs the MechSummary. This keeps the current behaviour while being less code.

MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
NickAragua: I'm not familiar with the rules for it, so I can't say one way or the other. Shouldn't be too hard to implement if that's the case, though.

PR #1852, opened 2020-07-09
Moving Retirement to LocalDate

Windchild292:
This moves AtB retirement to LocalDate


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8395,8 +8395,8 @@ public boolean checkRetirementDefections() {
 
     public boolean checkYearlyRetirements() {
         if (getCampaignOptions().getUseAtB()
-                && Utilities.getDaysBetween(getRetirementDefectionTracker()
-                        .getLastRetirementRoll().getTime(), getDate()) == 365) {
+                && (Math.toIntExact(ChronoUnit.YEARS.between(getRetirementDefectionTracker()
+                .getLastRetirementRoll(), getLocalDate())) == 1)) {
```
sixlettervariables, line 8399: `ChronoUnit.YEARS.between` returns the whole number of years between two dates. This isn't the same logic as before (but perhaps *roughly* the same logic).

MekHQ/src/mekhq/gui/dialog/RetirementDefectionDialog.java
```diff
@@ -173,8 +184,8 @@ private void initComponents(boolean doRetirement) {
             }
             if (null == contract) {
                 instructions += "\n\nDays since last retirement roll: "
-                        + Utilities.countDaysBetween(rdTracker.getLastRetirementRoll().getTime(),
-                                hqView.getCampaign().getDate());
+                        + Math.toIntExact(ChronoUnit.DAYS.between(rdTracker.getLastRetirementRoll(),
```
sixlettervariables, line 187: Why `toInt` here and above, Is `long` not going to work?

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: That is correct, but the rules do say yearly so this is proper

MekHQ/src/mekhq/gui/dialog/RetirementDefectionDialog.java
Windchild292: Force of habit to move from long to int, swapped them both to long because they do work as intended as longs.

Issue #1843, opened 2020-07-07
Order the personnel by date of enter service/captured?


binaryspica:
### Environment
Ver 0.47.7

### Description
That's mainly the problem on prisoners. If you have many prisoners, it is merged to only one prison and I have no idea who is the prisoner captured on what battle at all. You can know individual prisoners by select the prisoner, but it only shows one person's information. But I want to arrange them by captured date.

It would be better if I can grouping them, and perhaps order by original faction seems works too on some situation, but the date of enter service/captured seems a easy solution for this.

Windchild292:
Subsumed by 2207

PR #1857, opened 2020-07-10
Personnel biographical view and last name simplification 

AaronGullickson:
Primarily the purpose of this pull request was to add information on recruitment date and death date as a sortable field in the personnel table. Since it would get a little scrunched to add that to fluff, I created a new view called "Biographical Information" that includes these dates and I moved age, origin faction, and origin planet over from fluff information since that table was a bit cluttered. Here is a screenshot of the new view:

Screen Shot 2020-07-10 at 1 19 25 PM

We could add other dates here as well. I wanted to add retirement date and PR #1843 asks for the date when people were taken as prisoners. However, we currently don't track those dates on the person, just as one of many possible events in their personnel log. 

Since I was messing about on the personnel table, I decide to fix up another issue as well. I don't like the way surname and bloodname were being distinguished based on the type of faction for the overall campaign. Players may want to run campaigns that involve a mix of non-clan and clan personnel and the dark age sort of assumes this sort of structure for many factions, with partially assimilated clanners. They may also want to just give all of their non-bloodnamed warriors the surname "Wolf" which there is also precedent for in canon. This seemed like an unnecessary restriction on how players might want to construct their campaigns. 

So I got rid of all of the checks for clan or non-clan faction on surname. Instead all fields except fluff now contain a "Last Name" column and last name here is either surname or bloodname plus any honorific. If both are given, then bloodname takes precedent. This logic is now built into `Person.getFullName` so it should show up correctly anywhere that method is used.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -644,27 +644,36 @@ public void setMaidenName(String n) {
...
+     * A bloodname will overrule a surname but we do not disallow surnames for clanners, if the
+     * player wants to input them
+     * @return a String of the person's last name
+     */
+    public String getLastName() {
+        String lastName = "";
+        if (!StringUtil.isNullOrEmpty(bloodname)) {
+            lastName = bloodname;
+        }
+        else if (!StringUtil.isNullOrEmpty(surname)) {
```
Windchild292, line 658:
```suggestion
        } else if (!StringUtil.isNullOrEmpty(surname)) {
```

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -644,27 +644,36 @@ public void setMaidenName(String n) {
...
+            lastName = surname;
+        }
+        if (!StringUtil.isNullOrEmpty(honorific)) {
```
Windchild292, line 661:
```suggestion
        }

        if (!StringUtil.isNullOrEmpty(honorific)) {
```

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1146,6 +1155,15 @@ public LocalDate getDateOfDeath() {
         return dateOfDeath;
     }
 
+    public String getDeathDateAsString(Campaign campaign) {
+        if(getDateOfDeath() == null) {
```
Windchild292, line 1159:
```suggestion
        if (getDateOfDeath() == null) {
```

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1100,6 +1109,15 @@ public LocalDate getDateOfDeath() {
         return dateOfDeath;
     }
 
+    public String getDeathDateAsString(Campaign campaign) {
+        if (getDateOfDeath() == null) {
+            return "";
+        } else {
+            return getDateOfDeath().format(DateTimeFormatter.ofPattern(
+                    campaign.getCampaignOptions().getDisplayDateFormat()));
```
Windchild292, line 1117:
```suggestion
            return campaign.getCampaignOptions().getDisplayFormattedDate(getDateOfDeath());
```

Windchild292: This was a LocalDate change I made to make it easier to read.

PR #1862, opened 2020-07-13
1861: tryingToMarry and tryingToConceive flags are now saved properly

### LINKED ISSUE #1861: The flag Trying to Marry and Trying to Conceive are not saved.
### Environment
Ver 0.47.6~7

### Description
Whatever you have changed it, you cannot save the state. 

The flag was initiated on 0.47.6, and I found that both versions are not able to save the flag's state at all.

Windchild292:
This fixes #1861


Windchild292:
The default is true so only modifications are serialized

PR #1863, opened 2020-07-13
Retirement Date Tracking

Windchild292:
This implements Retirement Date tracking for Personnel in MekHQ and cleans up a few more minor things with Time in Service and Time in Rank (whose implementations were largely followed to implement this).


MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java
```diff
@@ -581,6 +588,28 @@ public Component getListCellRendererComponent(final JList<?> list,
             y++;
         }
 
+        if (campaign.getCampaignOptions().useRetirementDateTracking() && (retirement != null)) {
```
NickAragua, line 591: Seems like it'd be possible for retirement to be null, but the user wants to explicitly set a retirement date. This logic would disallow that workflow.

Windchild292: Same logic is used for time in service and time in rank.

PR #1876, opened 2020-07-21
198: Adding Ransom Mech and Ransom Prisoner Options for Scenario Resolution

### LINKED ISSUE #198: Add Ransom Mech to Salvage Options in Mission Resolution
Can we get an option to ransom a mech instead of salvaging it in the mission resolution screen. This would be some % of the mechs worth, probably set in the campaign options.

Similarly, I'd like to see an option to ransom prisoners instead of just setting them free added to the prisoner options.

Windchild292:
This fixes #198 and further cleans up and cleans up the Salvage panel to resemble the other panels in the GUI.


MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -1299,25 +1301,38 @@ private void initUnitsAndPilotsWithoutBattle() {
...
-        TestUnit salvageUnit = potentialSalvage.get(i);
-        actualSalvage.add(salvageUnit);
+        TestUnit salvageUnit = getPotentialSalvage().get(i);
+        getActualSalvage().add(salvageUnit);
     }
 
-    public void dontSalvageUnit(int i) {
-        leftoverSalvage.add(potentialSalvage.get(i));
+    public void ransomUnit(int i) {
+        TestUnit ransomUnit = potentialSalvage.get(i);
```
sixlettervariables, line 1330: These are public and should probably include a check on `i`.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -1299,25 +1301,38 @@ private void initUnitsAndPilotsWithoutBattle() {
...
+        getRansomedSalvage().add(ransomUnit);
+    }
+
+    public void doNotSalvageUnit(int i) {
+        getLeftoverSalvage().add(getPotentialSalvage().get(i));
```
sixlettervariables, line 1335: These are public and should probably include a check on `i`.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -1299,25 +1301,38 @@ private void initUnitsAndPilotsWithoutBattle() {
...
+    }
+
+    public List<TestUnit> getLeftoverSalvage() {
+        return leftoverSalvage;
+    }
+
     public void salvageUnit(int i) {
-        TestUnit salvageUnit = potentialSalvage.get(i);
-        actualSalvage.add(salvageUnit);
+        TestUnit salvageUnit = getPotentialSalvage().get(i);
```
sixlettervariables, line 1325: These are public and should probably include a check on `i`.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
Windchild292: They are used by the ResolveScenarioWizardDialog, so they need to be. I've added the check on the i value to be safe, but it is not necessary based on how they are used at this time.

PR #1879, opened 2020-07-22
Personnel XP Standardization

Windchild292:
This PR standardizes how personnel experience is done, removing the need (if there even was one, cause I can't see a reason for it after going over the code) for a separate count for engineering xp. I also used this to make some more formatting changes to improve readability.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -3961,13 +3961,11 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
         }
         int minutesLeft = TECH_WORK_DAY;
         int overtimeLeft = TECH_WORK_DAY / 2;
-        int edgeLeft = 0;
         boolean breakpartreroll = true;
         boolean failrefitreroll = true;
         if (null != engineer) {
             minutesLeft = engineer.getMinutesLeft();
             overtimeLeft = engineer.getOvertimeLeft();
-            edgeLeft = engineer.getEdge();
         } else {
             //then get the number based on the least amount available to crew members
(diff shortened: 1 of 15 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3956,35 +3956,32 @@ private void assignToCrewSlot(Person p, int slot, String gunType, String driveTy
...
+                if (p.getMinutesLeft() < minutesLeft) {
                     minutesLeft = p.getMinutesLeft();
                 }
-                if(p.getOvertimeLeft() < overtimeLeft) {
+
+                if (p.getOvertimeLeft() < overtimeLeft) {
                     overtimeLeft = p.getOvertimeLeft();
                 }
-                if(p.getEdge() < edgeLeft) {
-                    edgeLeft = p.getEdge();
```
NickAragua, line 3982: What was this for?

MekHQ/src/mekhq/service/MassRepairService.java
```diff
@@ -1059,7 +1062,7 @@ private static WorkTimeCalculation calculateNewMassRepairWorktime(IPartWork part
         }
 
         WorkTime newWorkTime = partWork.getMode();
-        WorkTime previousNewWorkTime = newWorkTime;
+        WorkTime previousNewWorkTime;
```
NickAragua, line 1065: Does this need a default value still?

Windchild292: Nope, it is always overwritten

MekHQ/src/mekhq/campaign/unit/Unit.java
Windchild292: Completely unused, so I removed the logic.

PR #1886, opened 2020-07-31
AbstractPerson

Windchild292:
This PR creates an AbstractPerson to reduce and simplify the Person class. However, code reduction and simplification is only stage one. The end plan is to separate Dependants from standard Personnel, while adding a third type of personnel, the far lighter Ancestor.

I've selected the following as the MVP for a Person. Anything here will form the most basic Person object, with a slight increase of information.
1) id
2) The Person's name (this does not include Bloodname, as that is a bit more complex than a MVP)
3) Maiden Name (this is for marriage tracking, but the inclusion is questionable. It is here because I think it would be needed for the Ancestor type)
4) Callsign (this one is questionable, but I think it should be included)
5) Gender
6) PersonnelStatus (this one is added because of future plans, which make it a necessity to have)
7) Portrait
8) Birthday
9) Death Date
10) Genealogy
11) Biography


binaryspica:
Maiden name should be included. This is the person's default birth name, and can be changed depend on the marriage. Technically it is the default surname for the mechanism, so it must be one of the the person's basic data.

Else, it is also possible to keep the basic surname and make the 'current' surname instead. For example, set the name data to 'given name', 'middle name', 'surname', 'marriage name', then, only change the marriage name on the marriage process, and left given name and surname unchanged. Then regardless how much the person marriages, the person's basic surname is left unchanged and is always remembered.


Are faction, rank and phenotype should be included as well? Are these already included on 6?


Your plan on 'ancestor' type means the person only there for describe the heritage of current person? I appreciated for the idea, for I want exactly same thing when I try to make the family tree manually. For now all I can is just make the person retire.

sixlettervariables:
How would Dependents being separated work? When would they transition to a Person? (Adding a role?)

Likewise, aren't all Ancestors tracked within MHQ personnel on the roster? Or is this for adding people who will never appear in a role?

binaryspica:
I think that what Windchild want to say is, 'ancestors' is only there for the blood relationship of your company's current member, and they are not expected to be part of the company.

Windchild292:
Ancestors allow for an expanded ancestry for personnel, allowing one to have long historical family trees and create large backstories as desired. This is part of creating a family tree setup for dynastical games. They will have significantly less information saved for them, and are solely mentioned in family trees and other historical information. Further, they will not be included in general processing, and would thus be tracked separately from our current personnel.

As for dependents, it would be based on the primary role. As part of adding them I would add a dependent primary role, and thus make it so dependents are truly dependent and not working for the force. I'm not fully certain on how that conversion will work, but that's also future work. This is open because the baseline is ready to go, and the backend work to start on ancestors is getting close to being ready once I finish up working on portraits.

Windchild292:
Converting to draft to handle feedback from https://bg.battletech.com/forums/megamek-games/dev-looking-for-feedback-ancestors/

NickAragua (approved):

HammerGS:
Given the age and conflicts closing.

PR #1891, opened 2020-08-01
Add founder on personnel filter groups

### LINKED ISSUE #1888: Personnel Type of founders on Personnel tab
### Environment
Ver 0.47.7

### Description
You can check the list of dependent, then why not founders?

binaryspica:
Resolves #1888. I hope that it works....


binaryspica:
> Because of current implementation the changes will also require implementation in the PersonnelMarketDialog file

Isn't it only affects market? Else it is just for the consistency?

Windchild292:
Consistency and because of how the current implementation is (although it is on my list to fix this issue)

binaryspica:
> Consistency and because of how the current implementation is (although it is on my list to fix this issue)

Got it!

Windchild292:
Looking better, but missing the strings under PersonnelMarketDialog::getPersonnelGroupName

binaryspica:
Done. Is there anything left?

And it seems the reason why the market's personnel type have two last sections with '?'.

Windchild292:
Nothing else I can see, and that would be why.

PR #1893, opened 2020-08-03
Personnel Status: Fixing ChangeStatus, Standardizing Boolean Comparisons, and Implementing Death Options 

Windchild292:
This fixes ChangeStatus by standardizing the one in Person and the one in Campaign into a single method. Furthermore, it standardizes enum comparisons into using boolean methods in the enum itself and implements four new personnel statuses for various forms of death.

Review with whitespace differences off, as I've fixed a bunch of assorted spacing issues and tab spacing.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -863,6 +867,103 @@ public PersonnelStatus getStatus() {
...
+            setDateOfDeath(null);
+        }
+
+        switch (status) {
+            case ACTIVE:
+                if (getStatus().isMIA()) {
+                    ServiceLogger.recoveredMia(this, campaign.getDate());
+                } else if (getStatus().isDead()) {
+                    ServiceLogger.resurrected(this, campaign.getDate());
+                } else {
```
NickAragua, line 889: Should we make this explicit? Might look weird if the person is in a status where it doesn't make sense for them to be re-hired.

MekHQ/src/mekhq/gui/dialog/PersonnelMarketDialog.java
```diff
@@ -443,15 +442,18 @@ public boolean include(Entry<? extends PersonnelTableModel, ? extends Integer> e
...
-                        return person.getStatus() == PersonnelStatus.RETIRED;
-                    } else if(nGroup == PersonnelTab.PG_MIA) {
-                        return person.getStatus() == PersonnelStatus.MIA;
-                    } else if(nGroup == PersonnelTab.PG_KIA) {
-                        return person.getStatus() == PersonnelStatus.KIA;
+                        return person.getStatus().isActive();
+                    } else if (nGroup == PersonnelTab.PG_RETIRE) {
+                        return person.getStatus().isRetired();
+                    } else if (nGroup == PersonnelTab.PG_MIA) {
+                        return person.getStatus().isMIA();
```
NickAragua, line 449: Why would these "isMIA/Retired" checks even be necessary? This is the personnel market, for hiring new personnel? Unless I'm missing something.

Windchild292: This is required because of how the tabs on this dialog are currently set up (otherwise we end up with '?' selection options). I'm planning a refactor in the near future to fix this requirement.

MekHQ/src/mekhq/campaign/personnel/Person.java
Windchild292: The above handle all of the potential statuses that would not be a rehire, plus this would make the most sense for any status being changed to active that is not MIA or a version of death. (unless I'm missing a potential future addition, but this should be a solid future proofed method to do so)

PR #1899, opened 2020-08-03
1808: Adding an Option for the Player to Control All Attached Units

### LINKED ISSUE #1808: Request for Against the Bot: Take control of the integrated/house units?
### Environment
Ver 0.47.6

### Description
The liaison command is fine, but they are making headache in the most times. Although they will accompany to us, but they never lead my command(because the bot controls the unit), and most of them are only use fragile and faster light battlemech, makes most of them rush to the enemy formation and make the officer tip of the spear. Even if I want to make a defensive position, such a rushing VIP makes the move impossible. Their fragility is also the reason to losing them on the battle.

So, can't we take control of the unit? I know reducing the aggression may helps, but it is not the solution for the most times.

Windchild292:
This fixes #1808.


NickAragua:
I gotta ask - can't this just be done in the MegaMek lobby?

Windchild292:
It can, this just automates it.

PR #1910, opened 2020-08-07
retain force structure in destination campaign

NickAragua:
Two changes in this one:

First, the campaign export wizard now tries to re-construct force structure for exported units. So, if a mech was in Alpha Company/Fire Lance in the source campaign, it'll either create that force structure in the destination or place the mech there if it exists already.

Second, the personnel selection list in the export wizard displays the call sign of a person in addition to the role.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CampaignExportWizard.java
+++ b/MekHQ/src/mekhq/gui/dialog/CampaignExportWizard.java
@@ -648,12 +648,12 @@ public class CampaignExportWizard extends JDialog {
     
     private void attemptToAssignToForce(Unit unit, int sourceForceID, Campaign sourceCampaign, Campaign destinationCampaign) {
         Force sourceForce = sourceCampaign.getForce(sourceForceID);
-        if(sourceForce == null) {
+        if (sourceForce == null) {
             return;
         }
         
         // this indicates a unit assigned to the root-level force
-        if(sourceForce.getParentForce() == null) {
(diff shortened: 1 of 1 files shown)


NickAragua:
Found a bug, whee

NickAragua:
In all of my testing, the source campaign's root force was named "My Campaign" and the destination campaign's root force was named "My Campaign". So it worked fine. 

Turns out that's not always the case when I tried to export a substantial number of units from a "live" campaign.

MekHQ/src/mekhq/gui/dialog/CampaignExportWizard.java
```diff
@@ -644,31 +648,36 @@ private boolean exportToCampaign(File file) {
...
         
         // first thing we will try is to identify a force with the same name and tree structure in the destination campaign
         // if we find one, add the unit to it
-        // otherwise,
-        Force destForce = findForce(sourceForce.getFullName(), destinationCampaign.getForces());
-        if(destForce != null) {
+        // otherwise, chain-add forces
+
+        // hack: the root forces are irrelevant, so we replace the source root force name with the destination root force name
+        String sourceForceFullName = getDestinationFullName (sourceForce, sourceCampaign, destinationCampaign);
```
Windchild292, line 665:
```suggestion
        String sourceForceFullName = getDestinationFullName(sourceForce, sourceCampaign, destinationCampaign);
```

PR #1917, opened 2020-08-11
Fixing the Award Tracker post-Ancestors Removal

Windchild292:
This removes the Ancestors column, which was causing the Awards Tracker to not be able to process the data


NickAragua:
We use an excel file for actual data somewhere?

Windchild292:
This is a helper file in documents, but it does load data from the campaign file

BullseyeSmith:
@Windchild292, I never meant for the AwardTracker to limit you from making whatever changes need to be made to campaign files in order to make things better.  Were you able to find a way to add the Ancestor data without causing an error in the AwardTracker?  I can change it up to fit a new format if needed.  I have to change it anyways because of the surname tables.

Windchild292:
I was able to change from specifying which to remove to specifying which to keep, which largely allows us to make changes to the XML without causing any issues with the file.

PR #1928, opened 2020-08-16
790: Implementing Non-Combat TO&E Forces

### LINKED ISSUE #790: [Feature Request] Mark TO&E Force as Non-Combat
I'd like to mark a Force in the TO&E as a "Non-Combat" (or other similar label) which would remove them from the "Current Assignments" in "Briefing Room". I have a number of non-line elements (e.g. combat support and support units) that aren't relevant in that section, and being able to exclude them would be wonderful.

Windchild292:
This fixes #790.

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -2,6 +2,7 @@
  * Force.java
  *
  * Copyright (c) 2011 Jay Lawson <jaylawson39 at yahoo.com>. All rights reserved.
+ * Copyright (c) 2020 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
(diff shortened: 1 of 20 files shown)


MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -197,12 +213,19 @@ public void addSubForce(Force sub, boolean assignParent) {
     }
 
     /**
+     * @param combatForcesOnly to only include combat forces or to also include non-combat forces
      * @return all the unit ids in this force and all of its subforces
      */
-    public Vector<UUID> getAllUnits() {
-        Vector<UUID> allUnits = new Vector<>(units);
+    public Vector<UUID> getAllUnits(boolean combatForcesOnly) {
```
NickAragua, line 219: This is where your dislike of overloads comes back to bite you in the ass, this PR would have been about five instead of twenty files, with a lot less diffs if you keep the existing method and default that value to true.

Windchild292: I'd far rather have this in the future to improve readability. More work now means less work in the future, which is a trade I'd take any time.

NickAragua: a) I'm not convinced that readability is improved in any way here (it's very much in the eye of the beholder) and b) the surface area of code changes is greater, which means a greater chance of introducing defects.

Windchild292: The other side is by doing this I actually caught a few bugs that would have otherwise gone unseen, because I needed to go through every single call to the method to verify it was using the correct true/false value.

Windchild292: However, I have really bloated my recent PRs with minor formatting fixes and what not while not doing them in separate commits, which I apologize for.

NickAragua: In this case, it's not really worth it to roll all that work back anyway.

PR #1931, opened 2020-08-17
1922: CrewSkillUpgrader Infinite Loop Fix

### LINKED ISSUE #1922: 47.9 Freeze when advancing from Sunday to Monday
#1328 # Environment
Version 47.9
Windows 10 
Java version 1.8.0_261

### Description
Issue is intermittent, but the attached save will consistently reproduce the freeze (at least on my machine) by loading the campaign and pressing Advance Day. Issue has been encountered multiple times. In previous cases where the issue was persistent on reload from autosave, it could be resolved by "doing things" in MekHQ (purchase parts, ransom prisoners, etc), although this could take several tries before Advance Day was successful. 

No errors show up in log files, although they are still attached. 

### Files
[Autosave-2-Coppercoats-30270930.cpnx.gz](https://github.com/MegaMek/mekhq/files/5066583/Autosave-2-Coppercoats-30270930.cpnx.gz)
[megameklog.txt](https://github.com/MegaMek/mekhq/files/5066586/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/5066587/mekhqlog.txt)

No custom units or other files.

Windchild292:
This fixes #1922, which is a stallout caused by an infinite loop in assigning SPAs.

To review: The first commit is a formatting cleanup, the second is the code change for this branch.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
+++ b/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
@@ -115,7 +115,9 @@ public class CrewSkillUpgrader {
         // every time we generate an SPA, we reduce the max available XP
         // this logic also prevents attempting to assign an SPA when there are no SPAs
         // that cost less XP than the remaining cap
-        for (int x = 0; (x < spaCap) && (xpCap > minAbilityCost); x++) {
+        // We also want to ensure this only happens a maximum of a thousand times, to prevent an
+        // infinite loop
+        for (int x = 0, y = 0; (x < spaCap) && (xpCap > minAbilityCost) && (y < 1000); x++, y++) {
             int spaCost = addSingleSPA(entity, xpCap);
             // if we didn't assign an SPA, let's reroll it.
             if (spaCost == 0) {
--- a/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
+++ b/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
@@ -1,17 +1,21 @@
 /*
- * MegaMek - Copyright (C) 2019 Megamek Team
+ * Copyright (C) 2019, 2020 - The MegaMek Team. All Rights Reserved.
  *
- *  This program is free software; you can redistribute it and/or modify it
- *  under the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
```diff
@@ -85,34 +88,36 @@ public void upgradeCrew(Entity entity) {
...
         // every time we generate an SPA, we reduce the max available XP
         // this logic also prevents attempting to assign an SPA when there are no SPAs
         // that cost less XP than the remaining cap
-        for(int x = 0; (x < spaCap) && (xpCap > minAbilityCost); x++) {
+        // We also want to ensure this only happens a maximum of a thousand times, to prevent an
+        // infinite loop
+        for (int x = 0, y = 0; (x < spaCap) && (xpCap > minAbilityCost) && (y < 1000); x++, y++) {
             int spaCost = addSingleSPA(entity, xpCap);
             // if we didn't assign an SPA, let's reroll it.
             if (spaCost == 0) {
```
NickAragua, line 123: I think this logic is the main problem, it drops the x counter back one if we fail to generate an SPA. I think it needs to be taken out and the addSingleSPA modified, as described in the comment there.

MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
```diff
@@ -121,76 +126,76 @@ public void upgradeCrew(Entity entity) {
...
      * Upgrade an entity with a single SPA
      * @param entity
      * @return the xp cost of the added SPA
      */
-    public int addSingleSPA(Entity entity, double xpCap) {
+    private int addSingleSPA(Entity entity, double xpCap) {
         int unitType = entity.getUnitType();
-        
+
         List<SpecialAbility> choices = coalescedSPAList(unitType, xpCap);
```
NickAragua, line 138: The main problem is likely that after the list of SPAs is built, the only SPAs left are ones that wind up ineligible. So I would have a loop in here that terminates either once we've picked an SPA or run out of SPAs to pick. Any time an SPA is eliminated below for whatever reason, remove it from the "choices" list and go through the loop again. That way, we're guaranteed to eventually reduces "choices" to an empty list or pick one that's valid.

Issue #1271, opened 2019-08-27
Graphic issue with prisoners who are vehicle drivers 


MarcJC:
![offscreen](https://user-images.githubusercontent.com/20606833/63736857-8a9e1480-c852-11e9-87b7-699ccbbeeb7f.jpg)

On version 47.0  
The screen for prisoners who are vehicle drivers goes of the screen for some reason.  
My vehicle drivers are fine, and all my other prisoners who have other jobs are fine, the issue only occurs for the given subset.

Windchild292:
Closing as this appears to have been fixed during the changes to the panel in question.

PR #1932, opened 2020-08-17
Fixing Prisoners Displaying Total Earnings

Windchild292:
This fixes a minor display issue I noticed while testing #1271.


Windchild292:
It is a completely valid workflow, but was never how this was intended to work. This was just fixing it for how I intended it to work initially.

Windchild292:
It is a completely valid workflow, but was never how this was intended to work. This was just fixing it for how I intended it to work initially.

Windchild292:
Sleeping on it, I decided it was better to just do the proper implementation now and thus fixed the miss dependents check and a current total earnings for the display.

PR #1933, opened 2020-08-17
1481: Adding Option to Disable Mothball Information Save

### LINKED ISSUE #1481: Unable to assign mech to force in TO&E 0.47.4
### Environment
MekHQ 0.47.4
Windows 10
JAVA 1.8.0_241

### Description
After upgrading a Marauder via MML within MHQ and assigning both a tech and pilot I was unable to add the Mech to a force in the TO&E. The mech did not even show. It’s in the hanger and is undamaged. 

I did us the GM tools to complete the upgrade. 

Further investigation: 
When I originally mothballed the unit it had been assigned to force. This force id was carried over into the mothballed state. When the unit was activated it was "reassigned" to the original force with no crew. Due to the unit already being assigned to a force it could not be "reassigned" which is by design. 

This is less of a bug and more of an enhancement request that if a unit is mothballed when assigned to a force that original force id be stripped from the unit.

Windchild292:
This fixes #1481 by adding an option to disable mothball info save. I also fixed two other minor issues relating to mothballing at the same time, namely a pair of option names on the unit model and the indent of the mothball info save.


NickAragua:
Think we ought to move this into MekHQ options like that other setting?

Windchild292:
I think it fits with the save options, so I've moved it over @NickAragua.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4508,8 +4508,8 @@ public void startMothballing(UUID id) {
      * @param isGM A value indicating if the mothball action should
      *             be performed immediately by the GM.
      */
-    public void startMothballing(UUID id, boolean isGM) {
-        if (!isMothballed()) {
+    public void startMothballing(Campaign campaign, UUID id, boolean isGM) {
+        if (!isMothballed() && MekHQ.getMekHQOptions().getSaveMothballState()) {
```
NickAragua, line 4512: Why add campaign as a parameter here? We already have a pointer to the campaign in the Unit class, so it seems unnecessary.

Windchild292: Better code style, and unit should not include a pointer to campaign for OOP.

NickAragua: Eye of the beholder (I personally find it highly annoying to pass campaign objects around every time I call a method), and not sure object-oriented programming principles recommend one way or the other in terms of keeping pointers to frequently-referenced objects. 

You can make the argument that "startMothballing" doesn't necessarily belong in the Unit class and instead should be moved to some kind of static(?) method in a worker class (e.g. "UnitActionProcessor.startMothballing"), if you want to say that we should avoid mixing data and "business logic" in the same class. As it is, this particular change does nothing to either simplify or improve the code in any meaningful way.

Windchild292: I actually agree with you that it is annoying, but it is part of how to handle Objects. Namely, you are supposed to pass the reference in methods when they do not need to be stored with the data for that Object and are entirely external.. and this decoupling may make some future code easier (including a few bits of a longer project I'm working on with person and unit)
The rest is arguable from my experience, and not something I would say.

sixlettervariables: If this isn't going to get a different Campaign object than the unit's, then it shouldn't be added, just adds an argument without benefit.

Windchild292: Reverted.

NickAragua (approved):

PR #1947, opened 2020-08-19
1817: Implementing option to not save customs to XML

### LINKED ISSUE #1817: Please allow to remove the custom refit data
### Environment
Ver 0.47.6

### Description
Everytime I load the campaign data, it automatically generates the custom unit I have made through the gameplay. The problem is, it is forced to be listed on the unit cache, and there is no possible way to remove it from automatically generated. I am even tried to edit it by notepad, but it only result the error. Even if I no longer possess the unit, it still lasts and always haunt for me. The only way to get rid of this is just make a whole new campaign, abandon the former one. 

So, can't we have an option to remove the customs made in MekHQ? Also, prevent the campaign-generated customs from unit cache as well.

Windchild292:
This fixes #1817


Windchild292:
Good point, actually... This and #1933 should probably both be in a grouping on the MekHQ option side of things.

Windchild292:
Swapped this over to using MekHQ options instead of Campaign Options.

PR #1976, opened 2020-08-30
259: Implementing Customizable MekHQ Display Date Formats

### LINKED ISSUE #259: Inconsistent date/time formats

Windchild292:
This implements the final stage of my LocalDate switchover, and by thus fixes #259.

There are two format types, the longer format used by the title and report, and the shorter style used everywhere else.


MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -66,7 +66,7 @@
     private boolean loadAssignmentFromMarket;
     private boolean groupByUnit;
 
-    private final MekHqColors colors = new MekHqColors();
+    private static final transient MekHqColors colors = new MekHqColors();
```
NickAragua, line 69: ?

NickAragua (approved):
One question, but looks fine otherwise.

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
Windchild292: Was testing something out, flipped back.
```suggestion
    private final MekHqColors colors = new MekHqColors();
```

PR #2009, opened 2020-09-16
Moving the Misc Tab to MekHQOptions, Fixing UnitMarketReportRefresh, Properly Localizing lblOpforLocalForceLikelihood

Windchild292:
This moves the Campaign Options Misc Tab to MekHQOptions.

I also fixed the UnitMarketReportRefresh option and the localization string for lblOpforLocalForceLikelihood, as I noticed both while removing the Campaign Options Misc Tab.


NickAragua (approved):
I can't immediately think of a way to reasonably do it, but do we want to try to carry this setting over from campaigns as they are loaded?

Windchild292:
This already does so. On load of the campaign options the MekHQ option is set from the Campaign Option.

PR #2013, opened 2020-09-16
Adding MekHQ option to prefer GZIP in campaign saves

Windchild292:
This is mostly for certain forms of dev testing, but I thought it was best to expose this to the end user as an option and actually have it saved instead of just transient.


NickAragua (approved):
With the ongoing expansion of MekHQ options, we may consider reorganizing the UI in some way going forward.

Windchild292:
That is already in the cards, just adding a few options first.

Issue #630, opened 2018-01-24
Heat sinks not tracked correctly by Mek Lab and Warehouse


RexPearce:
To reproduce:

1. Load the attached campaign which has two mechs ready to be customized
2. Note in Warehouse that 12 single heat sinks are available
3. In Hanger, select the Night Hawk NTK-2Q to be customized in Mek Lab
4. In Mek Lab, change the heat sink type from double to single and assign the critical locations
5. **Bug:** Auto Fill will not assign the heat sinks at this point - need to assign manually
6. **Bug:** Note that Mek Labs list 8 single heat sinks required for the refit instead of none (i.e the warehouse is not being checked for the engine-free ones)
7. begin the refit
8. **Bug:** Note in Warehouse that only 4 heat sinks are reserved for the refit instead of 12 (and the remainder have not been ordered)
9. Advance the day until refit is complete
10. **Bug:** Note that Warehouse shows 8 single heat sinks available instead of none.

Upgrading from single to double heat sinks has the same result:

1. Note in Warehouse that 12 double heat sinks are available (and 8 single)
2. In Hanger, select the Warhammer WHM-6R to be customized in Mek Lab
3. In Mek Lab, change the Tech Level to advanced
4. Change the heat sink type from single to double and reduce the count to 12
5. Add 4 jump jets and 2 tons of armor to add weight
6. Assign the critical locations
7. **Bug:** Auto Fill will not assign the heat sinks at this point - need to assign manually
8. **Bug:** Note that Mek Labs shows 11 heats required for the refit instead of none (12 are available in warehouse)
9. begin the refit
10. **Bug:** Note in Warehouse that only 1 heat sink is reserved for the refit instead of 12 (and the remainder have not been ordered)
11. Advance the day until refit is complete
12. **Bug:** Note that Warehouse shows 11 double heat sinks available instead of none.

[My Campaign30250101-heatsinks.zip](https://github.com/MegaMek/mekhq/files/1660687/My.Campaign30250101-heatsinks.zip)

RexPearce:
I looked into this, and the problem is basically that integrated heat sinks can be added or changed without requiring any parts. (Removing integrated heat sinks works fine though, they will be added to the warehouse.)

Eclipse keeps freezing on me when I set any breakpoints in the refit code, so I've been struggling with this one, but here's what I've found:

This is the main block that tracks changes to integrated heatsinks:

[...]

PR #2018, opened 2020-09-17
Improved tracking of integrated heat sink changes during refit

RexPearce:
Fix for #630


Windchild292 (commented):
Just some formatting comments first. Anything relating to refit requires more thought to review because of how the system is done.

RexPearce:
I split the changes into multiple commits to make it easier to follow.

94523d9a51ce8027b5c00976cbeb4c59ada89040 is a bugfix to the original code. The original code was adding HeatSink as Parts to the shoppingList which caused them to be later discarded in .acquireParts(), preventing them from being ordered from the market. With this change, new integrated heat sinks are ordered from the market as expected, then added to the unit being refitted.

However, we don't want actually these heat sink parts added to the unit - they're supposed to represent heat sinks not tracked via Parts. Adding these parts to the unit causes two problems: (i) we bloat units with useless parts and (ii) if this 'bloated' unit is refit again, those extra heat sinks are added back to warehouse resulting in more free/unaccounted heat sinks (the problem we're trying to solve).

1bd5efc05ca97a538a068f3d324daa0dea1d6745 solves the second problem by stripping these extra heat sinks from the unit in the final stage of refit, after the market bookkeeping is complete. It detects them by looking for heat sinks attached to the unit without an assigned location.

518eff0aabb071820c8226d20465dbc5ff7c6050 fixes another problem where the warehouse was never checked for spare heat sinks before ordering from the market.

45f407c020714ed45a4a98d068aa42ea6350b5e3 fixes a couple problems with the first cleanup routine (i) Aero units weren't cleaned up because they use AeroHeatSinks instead of HeatSinks, and (ii) the extra heatsinks in tanks were causing Utilities.unscrambleEquipmentNumbers() to log errors. So now heat sinks are stripped from Tanks and Aeros before unscrambleEquipmentNumbers() is called, and Mechs are done afterwards so the locations are set properly.

Windchild292:
Sorry, badly worded on my part. I meant that changes to this system from anyone can cause unintended issues because of how it was originally coded, and thus any changes can take a while to think over. However, that breakup makes it easier to review.

Furthermore, may I suggest joining us on our Slack? It is easier to reach any of the team and to ask any questions you might find there.

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -881,14 +878,16 @@ public void calculate() {
                     oldIntegratedHS.add(oldHS.clone());
                 }
                 for (int i = 0; i < newCount - oldCount; i++) {
-                    newIntegratedHS.add(oldHS.clone());
+                    // Heat sink added for supply chain tracking purposes and removed from refit later
+                    newIntegratedHS.add(newHS.clone().getMissingPart());
```
Windchild292, line 882: Can't this just be newHS.getMissingPart() ?

Windchild292 (approved):
Approved, with the one comment remaining.

RexPearce:
No worries, just trying to save you some time on the review, and explain my reasoning in case I made any mistakes.

(I actually joined the Discord awhile back under a different name, but I haven't used it in awhile. I'll see if I can find the account details.)

PR #2020, opened 2020-09-18
1265: Option to Subtract Salvage Overages from Final Contract Payment

### LINKED ISSUE #1265: RFE: Subtract salvage overages from final contract payment
From what I understand, there's no penalty for going over your salvage limit on your last battle in a contract.

As a house rule, I've been applying a debit at the end of a contract equal to however much I went over in salvage. Instead, this amount could be subtracted from the final payment (either stopping at the final payment being zero, or even making the final payment be negative).

Windchild292:
This fixes #1265 

I also did a rework of money to use a getter instead of the local variable when I was trying to figure out why my percentage wasn't working as I expected, and a few more light changes on the CampaignOptionsDialog.


MekHQ/src/mekhq/campaign/finances/Money.java
```diff
@@ -42,6 +41,10 @@ private Money(BigMoney money) {
         this.wrapped = money;
     }
 
+    public BigMoney getWrapped() {
```
sixlettervariables, line 44: Does this need to be public?

Windchild292: It does not, just a normal access level for a getter

sixlettervariables: Okay, when @VicenteCartas worked on this BigMoney was meant as an implementation detail not to be exposed.

sixlettervariables (approved):

PR #2022, opened 2020-09-19
GM Tools Name and Callsign Generation

Windchild292:
This adds name and callsign generation to the GM Tools dialog, handling both loading from a person and from the standard option.

As part of this I reordered and regioned the file to improve readability. I also reworked the pre-existing RAT Roller to fix a few issues I discovered it mostly relating to data load following the preferences changes.


MekHQ/resources/mekhq/resources/GMToolsDialog.properties
```diff
@@ -0,0 +1,38 @@
+# GM Tools Dialog Resource Strings
+# Dice Panel
+dicePanel.text=Dice Roller
```
sixlettervariables, line 3: Weird line endings need to be resolved.

Windchild292: Carriage returns, apparently. Will clean them out once I finish my current option swapover.

Windchild292: Resolved, but had to add the Bloodname Panel start to get it to be picked up (I'm planning on handling that immediately, so no issues with it being left behind)

Windchild292:
This has stopped working because of a casting exception and I'm not sure why it is happening now and not before. Converting to draft as I work that out.

NickAragua (approved):

PR #2028, opened 2020-09-20
476: Log Entry for Personnel Moved In/Out of a ToE Force

### LINKED ISSUE #476: RFE: Add new log entry for Personnel moved in/out of a ToE Force
If a person is moved out of a given ToE Force please add an entry to the Personnel Log. Often this occurs when that person's mech/asf/vehicle is destroyed but other things could cause it.

This would make managing large forces easier by easily identifying which "Force" a person came from after after having their "unit" blown to hell after a battle.

Example
"01/01/3025 Removed from Command Lance"

Windchild292:
This fixes #476


sixlettervariables (changes requested):
This adds a lot of complexity to the Force class without much benefit. I would prefer events be raised, or the reporting be handled by the code adding/removing forces (i.e. TOEMouseAdapter).

Windchild292:
I originally had located the code in TOEMouseAdapter, and realized it needed to be here instead because this catches places that does not (Mothballing, for example, will cause these to occur). I could make it event based, but... this is the most logical place for the code to be. However, I could add some comments to explain the logic there?

However, I could remove it being based on the transfer option which would simplify the code a bit.

For implementation this follows the existing logic behind being assigned and removed from a unit logging, but with the addition of requiring addition or removal of a unit from a Force.

NickAragua (approved):

PR #2030, opened 2020-09-21
AmmoBin from warehouse can rarely come with free ammo

### LINKED ISSUE #2029: When refitting, ammo bins taken from the warehouse rarely can be full
Minor issue I noticed while testing:

Tricky to reproduce:
1. Add (GM) an OmniUnit with ammo bins (e.g. Badger Tracked Transport A)
2. Add (GM) a variant of the above with no ammo bins (e.g. variant F)
3. Salvage the bin and ammo from the first variant.
4. Refit the second variant to a third variant that uses same ammo as the first (e.g. variant Prime)
5. Note the refit to the third variant corrently won't order new ammo/bins, however after the refit, the ammo is still in warehouse

RexPearce:
Fixes #2029

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -1171,15 +1171,11 @@ public class Utilities {
                             if (m.getType().equals(epart.getType()) && !m.isDestroyed()) {
                                 epart.setEquipmentNum(equipNum);
                                 AmmoBin bin = (AmmoBin) epart;
+                                // Ensure Entity is synch'd with part
+                                bin.updateConditionFromPart();
                                 // Unload bin before munition change
-                                // TODO: if part's unit is set, both .unload() and .changeMunition check entity's
-                                // TODO: bin instead of unit's bin. The bins can be out of sync here,
-                                // TODO: resulting in full bins replacing empty bins
-                                // TODO: work-around is to clear then restore part's unit
-                                bin.setUnit(null);
                                 bin.unload();
                                 bin.changeMunition(m.getType());
-                                bin.setUnit(unit);
                                 found = true;
                                 break;
                             }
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/Utilities.java
```diff
@@ -1170,7 +1170,16 @@ public static void unscrambleEquipmentNumbers(Unit unit, boolean refit) {
...
+                                AmmoBin bin = (AmmoBin) epart;
+                                // Unload bin before munition change
+                                // TODO: if part's unit is set, both .unload() and .changeMunition check entity's
+                                // TODO: bin instead of unit's bin. The bins can be out of sync here,
+                                // TODO: resulting in full bins replacing empty bins
+                                // TODO: work-around is to clear then restore part's unit
+                                bin.setUnit(null);
+                                bin.unload();
+                                bin.changeMunition(m.getType());
+                                bin.setUnit(unit);
```
Windchild292, line 1182: This should be able to use updateConditionFromEntity to fix the mismatch first.

RexPearce:
This is an odd one. Only occurs when a refit pulls a bin from the warehouse that is compatible, but did not come off the same unit/variant. (EquipNum mismatch?). e.g. It doesn't occur if the bin was ordered from the market or if the bin came off the same unit/variant.

MekHQ/src/mekhq/Utilities.java
RexPearce: In this case, the Entity has the wrong value though.  Is there a function to update the entity from the unit/part?

RexPearce: .updateConditionFromPart looks promising

Windchild292 (approved):

PR #2045, opened 2020-09-23
Aero Fighter refit heat sink tracking improvements

RexPearce:
First commit fixes the way untracked/integrated heat sinks are calculated for Aero Fighters - should be based on engine type (not engine size/rating)

Second commit make heat sink cleanup for Aero Fighters more careful not to strip valid heat sinks.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Refit.java
+++ b/MekHQ/src/mekhq/campaign/parts/Refit.java
@@ -861,7 +861,6 @@ public class Refit extends Part implements IPartWork, IAcquisitionWork {
          * the heat sink type or heat sinks required for energy weapons for vehicles and
          * conventional fighters.
          */
-        // TODO: handle any special cases for BattleArmor, Protomechs, SupportVehicles, SmallCraft, DropShips, etc.
         if ((newEntity instanceof Mech)
                 || ((newEntity instanceof Aero) && !(newEntity instanceof ConvFighter))) {
             Part oldHS = heatSinkPart(oldUnit.getEntity());
@@ -1433,6 +1432,14 @@ public class Refit extends Part implements IPartWork, IAcquisitionWork {
                 suit.setUnit(oldUnit);
             }
         }
+        int expectedHeatSinkParts = 0;
+        if ((newEntity instanceof Aero) && ((newEntity.getEntityType() &
+                (Entity.ETYPE_CONV_FIGHTER | Entity.ETYPE_SMALL_CRAFT | Entity.ETYPE_JUMPSHIP)) == 0)) {
+            // Only Aerospace Fighters are expected to have heat sink parts (Mechs handled separately)
+            // SmallCraft, Dropship, Jumpship, Warship, and SpaceStation use SpacecraftCoolingSystem instead
+            expectedHeatSinkParts = ((Aero)newEntity).getHeatSinks() - ((Aero)newEntity).getPodHeatSinks() -
+                    untrackedHeatSinkCount(newEntity);
+        }
         for(int pid : newUnitParts) {
             Part part = oldUnit.getCampaign().getPart(pid);
             if(null == part) {
@@ -1440,12 +1447,21 @@ public class Refit extends Part implements IPartWork, IAcquisitionWork {
                         "part with id " + pid + " not found for refit of " + getDesc()); //$NON-NLS-1$
                 continue;
             }
-            if (((part instanceof HeatSink) || (part instanceof AeroHeatSink)) &&
-                    ((newEntity instanceof Tank) || (newEntity instanceof Aero))) {
-                // Tanks and Aeros do not track heat sink parts (Mechs handled later)
+            if (part instanceof HeatSink && newEntity instanceof Tank) {
+                // Unit should not have heat sink parts
                 // Remove heat sink parts added for supply chain tracking purposes
                 oldUnit.getCampaign().removePart(part);
                 continue;
+            }
+            else if (part instanceof AeroHeatSink && newEntity instanceof Aero && !part.isOmniPodded()) {
+                if (expectedHeatSinkParts > 0) {
+                    expectedHeatSinkParts--;
+                } else {
+                    // Unit has too many heat sink parts
+                    // Remove heat sink parts added for supply chain tracking purposes
+                    oldUnit.getCampaign().removePart(part);
+                    continue;
+                }
             }
             part.setUnit(oldUnit);
             part.setRefitId(null);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -1433,20 +1432,37 @@ else if (part instanceof MissingPart) {
                 suit.setUnit(oldUnit);
             }
         }
+        int expectedHeatSinkParts = 0;
+        if ((newEntity instanceof Aero) && ((newEntity.getEntityType() &
+                (Entity.ETYPE_CONV_FIGHTER | Entity.ETYPE_SMALL_CRAFT | Entity.ETYPE_JUMPSHIP)) == 0)) {
```
NickAragua, line 1437: The determination  of "is aero but not conventional fighter/smallcraft/jumpship" is used multiple times in this file, here and on line 2480. And probably in other places. Let's place it into a separate function for ease of reuse.

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -1433,20 +1432,37 @@ else if (part instanceof MissingPart) {
...
             Part part = oldUnit.getCampaign().getPart(pid);
             if(null == part) {
                 MekHQ.getLogger().log(getClass(), METHOD_NAME, LogLevel.ERROR,
                         "part with id " + pid + " not found for refit of " + getDesc()); //$NON-NLS-1$
                 continue;
             }
-            if (((part instanceof HeatSink) || (part instanceof AeroHeatSink)) &&
-                    ((newEntity instanceof Tank) || (newEntity instanceof Aero))) {
-                // Tanks and Aeros do not track heat sink parts (Mechs handled later)
+            if (part instanceof HeatSink && newEntity instanceof Tank) {
```
NickAragua, line 1450: What about conventional fighters? Those behave the same way as tanks do as I recall.

NickAragua (changes requested):
On change request, one question.

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -1433,20 +1432,37 @@ else if (part instanceof MissingPart) {
                 suit.setUnit(oldUnit);
             }
         }
+        int expectedHeatSinkParts = 0;
+        if ((newEntity instanceof Aero) && ((newEntity.getEntityType() &
+                (Entity.ETYPE_CONV_FIGHTER | Entity.ETYPE_SMALL_CRAFT | Entity.ETYPE_JUMPSHIP)) == 0)) {
+            // Only Aerospace Fighters are expected to have heat sink parts (Mechs handled separately)
+            // SmallCraft, Dropship, Jumpship, Warship, and SpaceStation use SpacecraftCoolingSystem instead
+            expectedHeatSinkParts = ((Aero)newEntity).getHeatSinks() - ((Aero)newEntity).getPodHeatSinks() -
```
Windchild292, line 1440:
```suggestion
            expectedHeatSinkParts = ((Aero) newEntity).getHeatSinks() - ((Aero) newEntity).getPodHeatSinks() -
```

MekHQ/src/mekhq/campaign/parts/Refit.java
Windchild292, line 1450:
```suggestion
            if ((part instanceof HeatSink) && (newEntity instanceof Tank)) {
```

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -1433,20 +1432,37 @@ else if (part instanceof MissingPart) {
...
-            if (((part instanceof HeatSink) || (part instanceof AeroHeatSink)) &&
-                    ((newEntity instanceof Tank) || (newEntity instanceof Aero))) {
-                // Tanks and Aeros do not track heat sink parts (Mechs handled later)
+            if (part instanceof HeatSink && newEntity instanceof Tank) {
+                // Unit should not have heat sink parts
                 // Remove heat sink parts added for supply chain tracking purposes
                 oldUnit.getCampaign().removePart(part);
                 continue;
             }
+            else if (part instanceof AeroHeatSink && newEntity instanceof Aero && !part.isOmniPodded()) {
```
Windchild292, line 1456:
```suggestion
            } else if ((part instanceof AeroHeatSink) && (newEntity instanceof Aero) && !part.isOmniPodded()) {
```

Windchild292 (commented):
Additionally, some formatting changes.

MekHQ/src/mekhq/campaign/parts/Refit.java
RexPearce: I wasn't a fan of this either. I searched around to see if there was an operator similar to instanceof that didn't match on sub-classes. There doesn't appear to be one, however I ran across a post that suggested doing the equivalent of "newEntity.getClass() == Aero.class", which seems like it would work. In the end, I just went with the method previously used (I think the one you mentioned at line 2480).

"newEntity.getClass() == Aero.class" is technically more precise in that it would exclude TeleMissile and FighterSquadron as well (not sure how though are used).

Would you prefer I move the original to a function or use the "newEntity.getClass() == Aero.class" method?

MekHQ/src/mekhq/campaign/parts/Refit.java
RexPearce: ConvFighters (and all other subclasses of Aero) should enter line 1457 with expectedHeatSinkParts = 0, and have all heat sink parts removed (except OmniPodded ones, but they'll only be present for Aero Fighters).

For readability, I could restore the check for ConvFighters back to line 1450, but I'd have to restore the check for AeroHeatSink as well. Either way works - let me know what you prefer.

MekHQ/src/mekhq/campaign/parts/Refit.java
NickAragua: Let's give that getClass() == Aero.class thing a try. Just make sure to beat up on that code with other non-ASF unit types as well.

MekHQ/src/mekhq/campaign/parts/Refit.java
NickAragua: I don't think it's necessary, let's leave it as is.

MekHQ/src/mekhq/campaign/parts/Refit.java
RexPearce: Tested and works well.

NickAragua (approved):

PR #2072, opened 2020-09-29
Issue #2017: Keep focus on unit during repairs

### LINKED ISSUE #2017: Losing focus during Repair
That one is bothering me for a long time:

When you are at Repair Bay, and you have everything sorted by STATUS, when you repair components - Mech can suddenly advance to better category, because it is no longer that much damaged. Mech advances by moving, but focus remains the same, so you are suddenly looking at another Mech and have to search for previous one. OK, that actually bugs for all things: Vehicles, Mechs, Planes etc.

sixlettervariables:
This fixes #2017 by remembering the focused unit when refreshing the serviceable units table.

### DIFF
--- a/MekHQ/src/mekhq/gui/RepairTab.java
+++ b/MekHQ/src/mekhq/gui/RepairTab.java
@@ -778,10 +778,7 @@ public final class RepairTab extends CampaignGuiTab implements ITechWorkPanel {
     }
 
     public void refreshServicedUnitList() {
-        UUID uuid = null;
-        if (null != getSelectedServicedUnit()) {
-            uuid = getSelectedServicedUnit().getId();
-        }
+        UUID uuid = (getSelectedServicedUnit() != null) ? getSelectedServicedUnit().getId() : null;
 
         ignoreUnitTable = true;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/RepairTab.java
```diff
@@ -766,21 +772,22 @@ public void focusOnUnit(UUID id) {
...
         }
+        return false;
     }
 
     public void refreshServicedUnitList() {
-        int selected = servicedUnitTable.getSelectedRow();
+        UUID uuid = null;
+        if (null != getSelectedServicedUnit()) {
+            uuid = getSelectedServicedUnit().getId();
+        }
```
Windchild292, line 784:
```suggestion
        UUID uuid = (getSelectedServicedUnit() != null) ? getSelectedServicedUnit().getId() : null;
```

Windchild292 (approved):
Looks good, just one idea

MekHQ/src/mekhq/gui/RepairTab.java
sixlettervariables: Sure, was simply keeping the source in line with existing code.

Issue #2066, opened 2020-09-28
java.lang.NumberFormatException when changing equipment on a small Support Vehicle


RexPearce:
Tested on 0.47.11-Dev (current master)

Steps to reproduce:
1) start a new campaign
2) add a small support vehicle, e.g. Pitbull Medium Truck (Standard)
3) Customize it in the Mek Lab and try to add or remove equipement

Example error from log:
```
Exception in thread "AWT-EventQueue-0" java.lang.NumberFormatException: Character N is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:518)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:401)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:834)
	at java.base/java.math.BigDecimal.valueOf(BigDecimal.java:1304)
	at org.joda.money.BigMoney.of(BigMoney.java:113)
	at mekhq.campaign.finances.Money.of(Money.java:46)
	at mekhq.campaign.finances.Money.of(Money.java:50)
	at mekhq.campaign.parts.Refit.calculate(Refit.java:855)
	at mekhq.campaign.parts.Refit.<init>(Refit.java:171)
	at mekhq.gui.MekLabTab.refreshRefitSummary(MekLabTab.java:254)
	at mekhq.gui.MekLabTab$SupportVehiclePanel.refreshStatus(MekLabTab.java:1055)
	at megameklab.com.ui.tabs.EquipmentTab.lambda$new$0(EquipmentTab.java:204)
	at java.desktop/javax.swing.table.AbstractTableModel.fireTableChanged(AbstractTableModel.java:297)
	at java.desktop/javax.swing.table.AbstractTableModel.fireTableDataChanged(AbstractTableModel.java:199)
	at megameklab.com.util.CriticalTableModel.refreshModel(CriticalTableModel.java:104)
	at megameklab.com.ui.tabs.EquipmentTab.fireTableRefresh(EquipmentTab.java:674)
	at megameklab.com.ui.tabs.EquipmentTab.actionPerformed(EquipmentTab.java:655)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
[...]

PR #2074, opened 2020-09-29
Issue #2066: Do not add zero or NaN costs for ammo during Refits

sixlettervariables:
If an ammo bin has zero shots per ton then the refit cost calculation code will throw a `NumberFormatException` due to `NaN` being passed to `Money.of`. This is a "fix" for the exception seen in #2066, but may not be a fix in the sense that the Refit calculation logic does not have any handling for `InfantryAmmoBin`s (and probably should, a la `LargeCraftAmmoBin`s).

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Refit.java
+++ b/MekHQ/src/mekhq/campaign/parts/Refit.java
@@ -848,7 +848,9 @@ public class Refit extends Part implements IPartWork, IAcquisitionWork {
         for (AmmoType type : ammoNeeded.keySet()) {
             int shotsNeeded = Math.max(ammoNeeded.get(type) - getAmmoAvailable(type), 0);
             int shotsPerTon = type.getShots();
-            cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double)shotsNeeded/shotsPerTon)));
+            if (shotsNeeded > 0 && shotsPerTon > 0) {
+                cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double)shotsNeeded/shotsPerTon)));
+            }
         }
 
         /*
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -848,7 +848,9 @@ public void calculate() {
         for (AmmoType type : ammoNeeded.keySet()) {
             int shotsNeeded = Math.max(ammoNeeded.get(type) - getAmmoAvailable(type), 0);
             int shotsPerTon = type.getShots();
-            cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double)shotsNeeded/shotsPerTon)));
+            if (shotsNeeded > 0 && shotsPerTon > 0) {
+                cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double)shotsNeeded/shotsPerTon)));
```
Windchild292, line 852:
```suggestion
                cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double) shotsNeeded / shotsPerTon)));
```

MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -848,7 +848,9 @@ public void calculate() {
         for (AmmoType type : ammoNeeded.keySet()) {
             int shotsNeeded = Math.max(ammoNeeded.get(type) - getAmmoAvailable(type), 0);
             int shotsPerTon = type.getShots();
-            cost = cost.plus(Money.of(type.getCost(newEntity, false, -1) * ((double)shotsNeeded/shotsPerTon)));
+            if (shotsNeeded > 0 && shotsPerTon > 0) {
```
Windchild292, line 851:
```suggestion
            if ((shotsNeeded > 0) && (shotsPerTon > 0)) {
```

Windchild292 (approved):

MekHQ/src/mekhq/campaign/parts/Refit.java
sixlettervariables: My IDE reports those as extraneous parentheses and removes them on save (I wouldn't add them either given the simplicity of the line).

PR #2076, opened 2020-09-30
Add Show BV Calculation to Unit context menu

sixlettervariables:
While debugging https://github.com/MegaMek/megamek/issues/2273 I created this action for the Unit context menu to make my life easier. The code is taken straight from MegaMek and produces an identical popup to the one found in the unit selector.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/actions/ShowUnitBvAction.java
+++ b/MekHQ/src/mekhq/campaign/unit/actions/ShowUnitBvAction.java
@@ -1,0 +1,68 @@
+/*
+ * ShowUnitBvAction.java
+ *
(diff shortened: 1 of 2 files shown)


Windchild292 (approved):

Windchild292:
Forgot to add my comment: Any way we could standardize the code between the two?

sixlettervariables:
I didn't bother given the small amount of code, but we certainly could add a common dialog in MM.

Windchild292:
That's probably wise, even though it is minimal... it can always expand from that bit (and it may be three times we repeat it, cause IIRC there's a right click action I added for testing in MegaMek)

PR #2102, opened 2020-10-02
Cleaning up unit XML Write

Windchild292:
Following on from #2101, I've added the same checks for Unit to further reduce XML size.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -30,6 +30,7 @@ import java.util.stream.Stream;
 import megamek.common.*;
 import megamek.common.InfantryBay.PlatoonType;
 import mekhq.campaign.finances.Money;
+import mekhq.campaign.force.Force;
 import mekhq.campaign.log.ServiceLogger;
 import mekhq.campaign.mission.Scenario;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1675,135 +1678,141 @@ public Money getBuyCost() {
...
+
+        if (mothballTime > 0) {
+            MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "mothballTime", mothballTime);
+        }
+
+        if (mothballed) {
+            MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "mothballed", true);
+        }
+
+        if (!fluffName.isEmpty()) {
```
sixlettervariables, line 1792: Can this be null?

sixlettervariables (commented):
One question.

MekHQ/src/mekhq/campaign/unit/Unit.java
Windchild292: Good catch. I've pushed a fix where the fluff name can't be null but is now only an empty string.

sixlettervariables (approved):

PR #2131, opened 2020-10-08
Move logic to find personnel of a given type out of Campaign

sixlettervariables:
Moves logic to find personnel by their roles out of Campaign and into its own class.


Windchild292:
I feel this would be best in a Barracks class akin to the new Hangar class. Thoughts?

sixlettervariables:
I'm working on that next bit here (getting rid of personnel IDs throughout unit/etc). I can hold off on this until after that lands.

PR #2146, opened 2020-10-16
2139: Fixing Award Tracker Data Manipulation

### LINKED ISSUE #2139: 47.11 Award Tracker Excel sheet query issue
### Environment
47.11
Windows 10

### Description
In its as-downloaded state (uploaded here for ease of use), the Award Tracker Excel sheet will fail if used on a 47.11 campaign (or, I believe, campaigns from several previous versions, but in any case the fix is the same). 

Reproduction steps:
1. Download the supplied sheet and campaign file. 
2. Set up the file path and filename on the Settings sheet as usual.
3. Open the Queries and Connections viewer, then press Refresh All. 
4. Observe the error (see below)
image
After hitting "OK", all queries except for "Units-Drivers" will complete, but the table will not update properly.
image

A simple modification can fix this. Open the "Units-Drivers" query and delete the steps named "Expanded driverId" and "Renamed Columns1". Close and save the query, and after refreshing once more the data should load properly. 

### Files
[Coppercoats.cpnx.gz](https://github.com/MegaMek/mekhq/files/5360508/Coppercoats.cpnx.gz)
[AwardTracker.xlsx](https://github.com/MegaMek/mekhq/files/5360509/AwardTracker.xlsx)

Windchild292:
This fixes #2139 and modifies a few of the queries to reduce the maintenance for the file.


sixlettervariables (approved):
I don't use this, but it's an ancillary tool so I'll approve given your statement it fixes the bug.

Issue #1537, opened 2020-03-01
Unable to right click on Meteor-U Fighter


MarcJC:
Playing ATB in HQ 47.4 completed a mission where I shot down and salvaged 2 Heavy Strike Fighters. 
One was a Inseki II and the other was a Meteor-U.  When I went into the hangar to assign techs the Inseki had no issue but the game refuses to right click on the Meteor. 

The Log shows the following error.

Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
   at mekhq.Utilities.getMunitionsFor(Utilities.java:258)
   at mekhq.gui.adapter.UnitTableMouseAdapter.maybeShowPopup(UnitTableMouseAdapter.java:673)
   at mekhq.gui.adapter.UnitTableMouseAdapter.mouseReleased(UnitTableMouseAdapter.java:604)
   at java.awt.AWTEventMulticaster.mouseReleased(Unknown Source)
   at java.awt.Component.processMouseEvent(Unknown Source)
   at javax.swing.JComponent.processMouseEvent(Unknown Source)
   at java.awt.Component.processEvent(Unknown Source)
   at java.awt.Container.processEvent(Unknown Source)
   at java.awt.Component.dispatchEventImpl(Unknown Source)
   at java.awt.Container.dispatchEventImpl(Unknown Source)
   at java.awt.Component.dispatchEvent(Unknown Source)
   at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
   at java.awt.LightweightDispatcher.processMouseEvent(Unknown Source)
   at java.awt.LightweightDispatcher.dispatchEvent(Unknown Source)
   at java.awt.Container.dispatchEventImpl(Unknown Source)
   at java.awt.Window.dispatchEventImpl(Unknown Source)
   at java.awt.Component.dispatchEvent(Unknown Source)
   at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
   at java.awt.EventQueue.access$500(Unknown Source)
   at java.awt.EventQueue$3.run(Unknown Source)
   at java.awt.EventQueue$3.run(Unknown Source)
   at java.security.AccessController.doPrivileged(Native Method)
   at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
   at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
[...]

PR #2149, opened 2020-10-16
1537: Fixing Null AmmoType NPE in Hangar Tab

Windchild292:
This fixes the display error from which #1537 was noticed, but does not fix the underlying issue with the transition of bombs from MegaMek.

Commit 1 is a refactor/code cleanup, commit 2 prevents the NPE, and commit 3 rolls back one of my refactor changes.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
@@ -41,7 +41,6 @@ import javax.swing.UIManager;
 import javax.swing.event.MouseInputAdapter;
 
 import megamek.client.ui.swing.UnitEditorDialog;
-import megamek.client.ui.swing.util.MenuScroller;
 import megamek.common.*;
 import megamek.common.loaders.BLKFile;
 import megamek.common.util.EncodeControl;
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -891,14 +885,14 @@ private void maybeShowPopup(MouseEvent e) {
...
-                    if (techsFound > 0) {
-                        JMenuHelpers.addMenuIfNonEmpty(menu, menuElite, 20);
-                        JMenuHelpers.addMenuIfNonEmpty(menu, menuVeteran, 20);
-                        JMenuHelpers.addMenuIfNonEmpty(menu, menuRegular, 20);
-                        JMenuHelpers.addMenuIfNonEmpty(menu, menuGreen, 20);
-                        JMenuHelpers.addMenuIfNonEmpty(menu, menuUltraGreen, 20);
 
-                        popup.add(menu);
+                    if (techsFound > 0) {
+                        JMenuHelpers.addMenuIfNonEmpty(menu, menuElite);
```
NickAragua, line 890: Why did the 20 get removed?

Windchild292: Because I overrode the method it is calling to make 20 a default value. Better to use that instead of a magic number

NickAragua: Got it.

NickAragua (approved):

PR #2151, opened 2020-10-16
MissionStatus Enum

Windchild292:
This PR swaps MissionStatus from magic numbers to an enum.

As part of this I reworked the CompleteMissionDialog, which is the first commit.


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3064,7 +3064,7 @@ public void readNews() {
     }
 
     public int getDeploymentDeficit(AtBContract contract) {
-        if (!contract.isActive()) {
+        if (!contract.getStatus().isActive()) {
```
NickAragua, line 3067: I would suggest retaining the isActive() method, but having it call getStatus().isActive(). That way you can reduce the footprint of this PR quite a bit while keeping the spirit.

NickAragua (commented):
One suggestion, looks fine otherwise.

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: That is purposeful. I'd rather have the code in the enum in question to prevent future problems... This is especially for tracing, as I've seen what happens when coders go over-the-top with putting methods that do this and the sheer amount of effort that it adds to tracing [especially for enums], and thus would prefer to avoid it in our code base.

sixlettervariables: My preference would be Contract::isActive as well. It is more intuitive and benefits from intellisense (etc). Much like Person::isDeployed which wraps a similar check for convenience.

sixlettervariables: A compromise could be `Contract::isActiveOn(LocalDate date)` which would handle a lot of these call sites.

Windchild292: I'm actually planning on implementing contact end date nag and removing most uses of isActive in the near future. This originally contained them, but it got too busy for the PR so I reverted it.

Futher, Person::isDeployed isn't an enum check but rather a unit value check. Because it is an enum check is why this is implemented as it is.

As for isActiveOn, that's a nice way to handle it that will simplify some of the logic for my next changes, but it should be handled by getActiveContract if it is simply an active state check that does not currently need to be handled based only on start date.

MekHQ/src/mekhq/gui/dialog/ContractMarketDialog.java
```diff
@@ -104,16 +103,15 @@ public ContractMarketDialog(Frame frame, Campaign c) {
      */
     private void countSuccessfulContracts() {
         HashMap<String, Integer> successfulContracts = new HashMap<>();
-        for (Mission m : campaign.getMissions()) {
-            if (m.isActive() || !(m instanceof AtBContract)
-                    || ((AtBContract) m).getEmployerCode().equals(campaign.getRetainerEmployerCode())) {
+        for (AtBContract contract : campaign.getAtBContracts()) {
+            if (contract.getStatus().isActive()
```
NickAragua, line 107: Looks like you missed a .isActiveOn() call here

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -2939,11 +2998,12 @@ public void readNews() {
...
+    /**
+     * @param contract an active AtBContract
+     * @return the current deployment deficit for the contract
+     */
     public int getDeploymentDeficit(AtBContract contract) {
-        if (!contract.isActive()) {
-            // Inactive contracts have no deficits.
-            return 0;
-        } else if (contract.getStartDate().compareTo(getLocalDate()) >= 0) {
+        if (contract.getStartDate().compareTo(getLocalDate()) >= 0) {
```
NickAragua, line 3006: Is this a good place for isActiveOn()?

NickAragua (changes requested):
Two minor potential code changes, good to go otherwise.

MekHQ/src/mekhq/gui/dialog/ContractMarketDialog.java
Windchild292: That's a bug

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: It would be

Windchild292:
Fixed both issues and have now pushed the changes

NickAragua (approved):

PR #2162, opened 2020-10-23
MekHQ PortraitChooser

Windchild292:
This moves MekHQ to using the PortraitChooser and starts some of the icon implementation for portraits.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -692,20 +685,21 @@ public void setCallsign(String n) {
...
-        return portraitCategory;
+    //region Portrait
+    public AbstractIcon getPortrait() {
+        return portrait;
     }
 
-    public String getPortraitFileName() {
-        return portraitFile;
+    public void setPortrait(AbstractIcon portrait) {
+        this.portrait = portrait;
```
sixlettervariables, line 694: Can this be null? If not lets require it to be non-null here.

sixlettervariables (approved):
One suggestion, otherwise looks good.

MekHQ/src/mekhq/campaign/personnel/Person.java
Windchild292: No code using AbstractIcon should be nullable outside of the GUI. That is the base design for AbstractIcon.

Windchild292: However, I've added an assert cause here and will open a PR or perhaps PRs to ensure this is the case in MM  too.

PR #2177, opened 2020-10-28
See if GitHub code scanning is simpler than LGTM

sixlettervariables:


sixlettervariables:
It takes a *lot* longer than LGTM, but has the benefit of being able to do "composite builds" if I set that up. Thoughts @Windchild292 @NickAragua @neoancient

Windchild292:
Honestly, is there any reason to not use both? Because adding a new format to scan the code should help prevent more code issues in the long run.

sixlettervariables:
I'm not sure how it differs from LGTM in it's analysis features. We'd have to run it in production for a bit to get the annotations.

Windchild292 (approved):

PR #2179, opened 2020-10-29
Remove more Unit related code from Campaign

sixlettervariables:
This removes more unit related code from Campaign and adds it to either:

a. Its own class (i.e. `HirePersonnelUnitAction`)
b. or, into `Unit`

I've added tests for the hire personnel action as well.


NickAragua (approved):
Not seeing any major issues with this one.

Issue #1801, opened 2020-06-12
Removing the unit while attached to a Dropship does not changes remaining bays of the Dropship


binaryspica:
### Environment
Ver 0.47.6 or less.

### Description
As title says. If you assign a unit to a dropship, and remove the unit, the dropship does not empty its space and its remaining bay is unchanged.

For example, if a dropship have 10 mech bays, and you put a mech to it, it have 9 remaining bays. And if you remove the mech by sell it, the dropship still have **9** remaining bays despite it must be returned to 10 remaining mech bays.

PR #2185, opened 2020-11-03
Remove use of unit IDs for transports

sixlettervariables:
This removes the use of unit IDs for "transports" and cleans up how we tracked where units were being transported. Previously we used a Map data structure, except it could only have one entry. This has been instead switched out for a more formal class, `TransportShipAssignment`. Tests have been added for the basic transport logic as well.


Windchild292:
Started to look over it and it got me thinking: is there any chance that it was a map in preparation for units assigned to units assigned to units? For example, infantry assigned to an APC assigned to a DropShip?

sixlettervariables:
The map tracked who contained the unit and in which bay. Having multiple entries would be confusing as you would not know which actually contained the unit without enumerating the map and checking the keys (transports) to see which contained the unit (getTransportedUnits).

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1569,20 +1611,23 @@ public double calcInfantryBayWeight(Entity unit) {
...
-        }
-        for (UUID id : u.getTransportShipId().keySet()) {
-            int bayNumber = u.getTransportShipId().get(id);
-            updateBayCapacity(unitType, unitWeight, true, bayNumber);
-        }
-        removeTransportedUnit(u.getId());
-        u.getTransportShipId().clear();
+        // Only unload and update the transport assignment for a unit
+        // if we have actually transported them.
+        if (removeTransportedUnit(u)) {
```
Windchild292, line 1616: What about the case whereby a unit is assigned to this transport according to the unit but not according to the transport ship?

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1569,20 +1611,23 @@ public double calcInfantryBayWeight(Entity unit) {
...
+            double unitWeight;
+            if (u.getEntity().getUnitType() == UnitType.INFANTRY) {
+                unitWeight = calcInfantryBayWeight(u.getEntity());
+            } else {
+                unitWeight = u.getEntity().getWeight();
+            }
```
Windchild292, line 1622: Trinary candidate.

Windchild292 (commented):
A few minor things first

sixlettervariables:
I've added the fix for #1801 to this as well.

Windchild292 (approved):
Just wanted to check that there isn't any logic in TOEMouseAdapter that should be removed given the logic movement for #1801? Otherwise, approved.

Windchild292 (approved):

PR #2198, opened 2020-11-06
Issue #785 SLDF Ranks are Wrong

### LINKED ISSUE #785: SLDF Ranks Are Wrong
The ranks for the SLDF are wrong. They didn't have Lieutenant JG and Lieutenant SG. They just had Lieutenant. Warrant Officer is also missing from the list of ranks.

jschmetzer:
Fixes #785. Adds 2nd Star League rank set, edits default Star League rank set to remove two lieutenants, add warrant, per FM: SLDF

### DIFF
--- a/MekHQ/data/universe/ranks.xml
+++ b/MekHQ/data/universe/ranks.xml
@@ -109,7 +109,7 @@
 			<payMultiplier>1.0</payMultiplier>
 		</rank> <!-- E20 -->
 		<rank>
-			<rankNames>-,-,-,-,-,-</rankNames>
+			<rankNames>Warrant Officer,-,-,-,-,-</rankNames>
 			<officer>false</officer>
 			<payMultiplier>1.0</payMultiplier>
 		</rank> <!-- WO1 -->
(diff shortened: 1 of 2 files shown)


HammerGS:
The data part looks good to me and the code does as well.

Windchild292:
This won't migrate properly.

jschmetzer:
Whoops. Figures I'd mess up my first one. 👍

Windchild292:
Just edit in progress is fine. There's also a type on the static constraint :)

jschmetzer:
Doh. There it is right there, glowing at me in Atom. 

Would that have caused the fail or should I rearrange as requested, too?

Windchild292:
That caused the fail, the rearrange will likely cause issues on existing games if it is not done

Windchild292:
This will cause a few other issues due to internal use of RS_SL... you could swap RS_SL to RS_FSL and keep the SSL as SL, or you'd need to do internal fixes in five or six locations.

jschmetzer:
Does the name string need to remain original as well, or could RS_SL = Second Star League?

Windchild292:
Not localized, and we currently handle SL as Second Star League. I'd say it is fine?

Windchild292 (approved):

PR #2209, opened 2020-11-07
Random Portrait Generator

Windchild292:
This PR moves the portrait generation code out of campaign and into a random portrait generator.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/generator/RandomPortraitGenerator.java
+++ b/MekHQ/src/mekhq/campaign/personnel/generator/RandomPortraitGenerator.java
@@ -100,8 +100,8 @@ public class RandomPortraitGenerator {
             if (temp.length == 2) {
                 return new Portrait(temp[0], temp[1]);
             } else {
-                MekHQ.getLogger().error("Failed to generate portrait for " + p.getFullTitle()
-                        + ". ");
+                MekHQ.getLogger().error("Failed to generate portrait for " + p.getFullTitle() + ". "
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/personnel/generator/RandomPortraitGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * This is a helper method that determines what possible unassigned portraits can be generated
+     * based on the supplied subdirectory
+     *
+     * @param existingPortraits the list of existing portraits that have already been assigned
+     * @param subDir the subdirectory to search
+     * @return a list of all possible unassigned random portraits
+     */
+    private static List<String> getPossibleRandomPortraits(List<String> existingPortraits, File subDir) {
```
sixlettervariables, line 122: Existing Portraits should be a Set to avoid bad runtimes on large campaigns.

sixlettervariables (changes requested):
Need to switch to a Set for adequate performance of `existingPortraits.contains`

MekHQ/src/mekhq/campaign/personnel/generator/RandomPortraitGenerator.java
Windchild292: Pre-existing and I've never heard any issues with its performance before even on the largest campaigns. However, given it is for contains it does make sense to swap over, so I've done so.

sixlettervariables (approved):

PR #2211, opened 2020-11-07
AbstractIcon: MekHQ Code Implementation Phase 1

Windchild292:
This PR is the first of a few that will move MekHQ to using AbstractIcon instead of icon strings. The focus was on removing any non-AbstractIcon portrait generation setups, although I also added some baselines for Camouflage and removed two duplicated constants as part of this.


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -2640,10 +2640,6 @@ public JFrame getFrame() {
         return frame;
     }
 
-    public CampaignGUI getCampaignGUI() {
```
sixlettervariables, line 2643: Good find.

sixlettervariables (approved):

PR #2221, opened 2020-11-10
2220: Implementing AwardFileFactory to handle loading pngs that aren't handled by Toolkit

### LINKED ISSUE #2220: 47.13 Certain awards cause personnel information to blank
47.13
Linux
Java 11.0.8

Clean install with no modifications or customization, there appear to be some awards which cause information (Mechwarrior, Status, Origin, Age, Skill Names, etc.) to disappear from within the personnel display on the upper right side of the screen. 

I noticed this issue when attempting to import my custom award set to 47.12 and, in addition to the blank personnel screen detailed above, some of the ribbons from my custom set also failed to display. (log and custom awards file from that attempt have not been included here)  

The issue can be recreated using various awards from the 'standard' set as demonstrated here using the 'Medal of Honor' award featured in the attached screenshots.

[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/5518231/mekhqlog.txt)

AwardScreen1

AwardScreen2

Windchild292:
This fixes #2220, but will need a revisit after the Java 11 migration as it may be one of the png features implemented in Java 11.

### DIFF
--- a/MekHQ/src/mekhq/MHQStaticDirectoryManager.java
+++ b/MekHQ/src/mekhq/MHQStaticDirectoryManager.java
@@ -26,6 +26,7 @@ import megamek.common.util.fileUtils.DirectoryItems;
 import megamek.common.util.fileUtils.ImageFileFactory;
 import mekhq.campaign.force.Force;
 import mekhq.gui.enums.LayeredForceIcon;
+import mekhq.io.AwardFileFactory;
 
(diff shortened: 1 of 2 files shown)


sixlettervariables (approved):
Looks like the CI failure is because I didn't do spider builds for CQL. I think CodeCov is showing failed because of a config option I need to tweak.

PR #2223, opened 2020-11-11
Upgrade to Gradle 6.7

sixlettervariables:
This upgrades to Gradle 6.7 and resolves the deprecations noted during builds. This also removes the unused (and likely unfixable) Mac OS X bundle.

When including MegaMek and MegaMekLab under Gradle 6+ you'll need to give the build a name, like in your `settings_local.gradle`:
```groovy
includeBuild('../megamek') {
    name = 'mhq-mm-composite'
}
includeBuild('../megameklab') {
    name = 'mhq-mml-composite'
}
```
No two `GradleBuild` tasks or includes can share the same name. Under Gradle 6.3+ it uses the directory name by default, leading to these issues.


Windchild292 (approved):
For immediate merge post-stable

neoancient (changes requested):
I can't get it to build a release. It fails at compileMM with the message "Included build ~/mekhq/MekHQ/build/repo/megamek has build path :mekhq:megamek which is the same as included build ~/mekhq/MekHQ/build/repo/megamek." I don't see what makes HQ different. MML is able to complete the compileMM task, which effectively identical.

dtsosie:
OS X User here: I installed gradle with "brew install gradle." Installed latest JDK from oracle (15.0.2) That currently yields 6.3 version. Updating gradle-wrapper.properties to point to gradle-6.3-bin.zip has allowed me to gradlew build/run from stable.

PR #2232, opened 2020-11-15
Issue #2231: Don't give back free parts when fixing a missing part

### LINKED ISSUE #2231: 47.14-DEV Locations being added back to Warehouse after being used for replacement?
### Environment
47.14-DEV ([build link](https://github.com/MegaMek/mekhq/actions/runs/362562451))
Windows 10
Java 11.0.8

### Description
Replacing locations does not seem to result in the number of locations stored in the Warehouse decreasing _once the task is completed_. It will actually decrease when the task is assigned, but then increase again once completed, resulting in no net decrease in locations available in the Warehouse. 

Reproduction steps:
1. Load supplied campaign. Note that the Warehouse contains 2 Mech Head (45 tons) replacement locations. 
2. Navigate to the Repair Bay, and on the damaged PHX-1, assign the Replace Mech Head task to the Elite technician with 480 minutes available at Extra Time x4. 
3. Note that the number of Mech Head (45 Tons) replacement locations in the Warehouse has now decreased to 1. 
4. Advance Day; the task should complete. Note that the number of Mech Head (45 Tons) replacement locations in the Warehouse has now increased back to 2.

This issue may only apply to locations - my testing showed that replacements properly consume items from the Warehouse for weapons, armor, cockpits, and actuators, using the same save. 

### Files
[GreyscalesRepairTest.cpnx.gz](https://github.com/MegaMek/mekhq/files/5541584/GreyscalesRepairTest.cpnx.gz)
No custom units or other files. 
No errors in megameklog or mekhqlog.

sixlettervariables:
When we reserved a part for overnight work, we would give back an extra part in two possible situations:
1. The work was for a missing part.
2. The work on a missing part failed, destroying the part.

In both of the cases the cause is simple: we were not checking if the part had any quantity left before finding a spare and incrementing its quantity.

This fixes #2231 by doing two things:
1. Not manually finding spares when fixing/reserving and incrementing their quantities.
2. Only giving parts back to the Quartermaster if they have a non-zero quantity.
    a. The Quartermaster class actually handles zero quantities, but this seemed like the more defensive solution.

I've also added tests for the reservation logic in MissingPart.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MissingPart.java
+++ b/MekHQ/src/mekhq/campaign/parts/MissingPart.java
@@ -22,18 +22,14 @@
 package mekhq.campaign.parts;
 
 import java.io.PrintWriter;
-import java.io.Serializable;
-import java.util.UUID;
 
 import megamek.common.ITechnology;
 import megamek.common.TargetRoll;
-import mekhq.MekHqXmlSerializable;
 import mekhq.MekHqXmlUtil;
 import mekhq.campaign.Campaign;
 import mekhq.campaign.finances.Money;
 import mekhq.campaign.personnel.SkillType;
 import mekhq.campaign.work.IAcquisitionWork;
-import mekhq.campaign.work.IPartWork;
 import mekhq.campaign.work.WorkTime;
 
 /**
@@ -41,7 +37,7 @@ import mekhq.campaign.work.WorkTime;
  * task needs to be performed
  * @author Jay Lawson <jaylawson39 at yahoo.com>
  */
-public abstract class MissingPart extends Part implements Serializable, MekHqXmlSerializable, IPartWork, IAcquisitionWork {
+public abstract class MissingPart extends Part implements IAcquisitionWork {
 
     /**
      *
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/parts/MissingPart.java
```diff
@@ -419,18 +421,19 @@ public void reservePart() {
...
-                    spare.incrementQuantity();
-                    campaign.getWarehouse().removePart(replacement);
+        if (hasReplacementPart()) {
+            Part replacement = getReplacementPart();
+            if (replacement != null) {
+                replacement.setReservedBy(null);
+
+                // Only return the replacement part to the campaign if we have one
+                if (replacement.getQuantity() > 0) {
+                    campaign.getQuartermaster().addPart(replacement, 0);
```
NickAragua, line 431: Remind me, what's the second argument here?

sixlettervariables: `transitDays`, should probably name the method better and add an immediate version.

Something like:
```java
quartermaster.addInTransitPart(part, transitDays);
quartermaster.addPart(part);
```

NickAragua (approved):

PR #2241, opened 2020-11-23
Fix bug where PartsStore may list INT_MAX as cost and tons for BA ammo

sixlettervariables:
Certain BattleArmor ammos (e.g. Micro Bomb Ammo, Clan Torpedo/LRM5 Ammo) do not state how many kg per shot they are, which `PartsStore` required to calculate the number of shots per ton (and subsequent cost). This led to a situation where you couldn't afford to buy the ammo, but if you GM added it you'd have INT_MAX shots available for your BattleArmor. Sell it back and you can print money!

This "fixes" the bug on the MHQ side by only recalculating shots per ton if it is BattleArmor ammo and `AmmoType::getKgPerShot()` is greater than zero. A more appropriate fix would be to update the kg per shot on the MM side for these ammo types.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/PartsStore.java
+++ b/MekHQ/src/mekhq/campaign/market/PartsStore.java
@@ -171,7 +171,7 @@ public class PartsStore implements Serializable {
             try {
                 newEntity = new MechFileParser(summary.getSourceFile(), summary.getEntryName()).getEntity();
             } catch (EntityLoadingException e) {
-                MekHQ.getLogger().error(getClass(), "stockBattleArmorSuits", e);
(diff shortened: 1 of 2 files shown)


NickAragua (approved):
Do we really not know what the weight for battle armor micro-bombs is?

sixlettervariables:
@NickAragua MM has logic that says, "if kg/shot is less than zero, calculate based on the number of shots." However, in all of the cases I found, those did not have kg/shot at all.

Other interesting cases were non-BA ammo that listed kg/shot, but you'd get less ammo per ton than the math would calculate you should. I've opened a PR on the MM side to track the missing kg/shot.

PR #2250, opened 2020-12-01
Fixing Load Last Save for .cpnx.gz files

Windchild292:
This fixes a bug raised on the BattleTech forums, namely that load last save only works for .cpnx files. Further, it only worked on the base campaign directory, not the preference campaign directory, which I've also fixed.


sixlettervariables (approved):
Pending CI.

PR #2265, opened 2020-12-05
Implementing Manual Bulk Pregnancy

Windchild292:
I've been finding this an irritant with some of my testing for personnel improvements as part of random death, and have added it here. This also fixes a bug where not all of the pregnancy rules are followed on manual pregnancies.


Windchild292:
Merged into #2493, as the changes here are required on that end.

Issue #2141, opened 2020-10-13
[47.11-] Subcontract Offers Revert to Standard Contracts on Load


NickAragua:
When saving a game with subcontract offers present in the contract market, the subcontract offers revert to standard contracts when the game is loaded.

To reproduce: 
- Take a garrison contract
- Wait until subcontract offer is generated (to help simulate it, make the subcontract roll always result in a subcontract)
- Save game
- Quit
- Load game: subcontract offer will now be a standard contract

Should be reproducible with any AtB-enabled campaign.

Windchild292:
We don't write parentContract to file.

PR #2282, opened 2020-12-13
Issue #2141: Subcontracts don't reference parent contract after campaign load

sixlettervariables:
AtBContract does not serialize its `parentContract` member, causing subcontracts to be made into full contracts after a save-load cycle. This resolves the issue by using a dummy reference type during deserialization and then restoring the reference after the missions are loaded from the campaign file, fixing #2141.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -1193,6 +1193,12 @@ public class AtBContract extends Contract implements Serializable {
                 +"<nextWeekBattleTypeMod>"
                 +nextWeekBattleTypeMod
                 +"</nextWeekBattleTypeMod>");
+        if (parentContract != null) {
+            pw1.println(MekHqXmlUtil.indentStr(indent+1)
+                    + "<parentContractId>"
+                    + parentContract.getId()
+                    + "</parentContractId>");
+        }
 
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -1193,6 +1193,12 @@ protected void writeToXmlBegin(PrintWriter pw1, int indent) {
                 +"<nextWeekBattleTypeMod>"
                 +nextWeekBattleTypeMod
                 +"</nextWeekBattleTypeMod>");
+        if (parentContract != null) {
+            pw1.println(MekHqXmlUtil.indentStr(indent+1)
+                    + "<parentContractId>"
+                    + parentContract.getId()
+                    + "</parentContractId>");
```
Windchild292, line 1200: Use the MekHqXmlUtil.writeSimpleXmlTag instead, it handled UUID with a null check (And doesn't write otherwise)

Windchild292 (approved):

MekHQ/src/mekhq/campaign/mission/AtBContract.java
sixlettervariables: The ID is an integer in this case.

Windchild292 (approved):

PR #2284, opened 2020-12-13
MHQ Colours: Adding Customizability as part of MHQ Options

Windchild292:
This PR moves MHQ Colours to instead be part of the MHQ Options, and lets one customize the options. This requires https://github.com/MegaMek/megamek/pull/2508.


MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
...
+    public Color getMothballingForeground() {
+        return new Color(userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(MekHqConstants.MOTHBALLING_FOREGROUND, Color.BLACK.getRGB()));
+    }
+
+    public void setMothballingForeground(Color value) {
+        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(MekHqConstants.MOTHBALLING_FOREGROUND, value.getRGB());
+    }
+
+    public Color getMothballingBackground() {
+        // new Color(153, 153, 255)
```
NickAragua, line 184: Random commented out code

MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
...
+    public Color getMothballedForeground() {
+        return new Color(userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(MekHqConstants.MOTHBALLED_FOREGROUND, Color.BLACK.getRGB()));
+    }
+
+    public void setMothballedForeground(Color value) {
+        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(MekHqConstants.MOTHBALLED_FOREGROUND, value.getRGB());
+    }
+
+    public Color getMothballedBackground() {
+        // new Color(204, 204, 255)
```
NickAragua, line 201: Random commented out code

MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
...
+    public Color getNotRepairableForeground() {
+        return new Color(userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(MekHqConstants.NOT_REPAIRABLE_FOREGROUND, Color.BLACK.getRGB()));
+    }
+
+    public void setNotRepairableForeground(Color value) {
+        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(MekHqConstants.NOT_REPAIRABLE_FOREGROUND, value.getRGB());
+    }
+
+    public Color getNotRepairableBackground() {
+        // new Color(190, 150, 55)
```
NickAragua, line 218: cleanup

MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
...
+    public Color getNonFunctionalForeground() {
+        return new Color(userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(MekHqConstants.NON_FUNCTIONAL_FOREGROUND, Color.BLACK.getRGB()));
+    }
+
+    public void setNonFunctionalForeground(Color value) {
+        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(MekHqConstants.NON_FUNCTIONAL_FOREGROUND, value.getRGB());
+    }
+
+    public Color getNonFunctionalBackground() {
+        // new Color(205, 92, 92)
```
NickAragua, line 235: cleanup

MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
...
+    public Color getNeedsPartsFixedForeground() {
+        return new Color(userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(MekHqConstants.NEEDS_PARTS_FIXED_FOREGROUND, Color.BLACK.getRGB()));
+    }
+
+    public void setNeedsPartsFixedForeground(Color value) {
+        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(MekHqConstants.NEEDS_PARTS_FIXED_FOREGROUND, value.getRGB());
+    }
+
+    public Color getNeedsPartsFixedBackground() {
+        // new Color(238, 238, 0)
```
NickAragua, line 252: cleanup

MekHQ/src/mekhq/MekHQOptions.java
```diff
@@ -100,6 +105,255 @@ public void setPersonnelFilterOnPrimaryRole(boolean value) {
         userPreferences.node(MekHqConstants.DISPLAY_NODE).putBoolean(MekHqConstants.PERSONNEL_FILTER_ON_PRIMARY_ROLE, value);
     }
     //endregion Personnel Tab Display Options
+    //endregion Expanded MekHQ Display Options
+
+    //region Colours
+    public Color getDeployedForeground() {
```
NickAragua, line 111: This seems like it could be done with a lot less code using a "getColor(enum areaBeingColored, boolean Foreground/Background)" call to reduce it from fifty thousand methods to two, and have the underlying mechanism be a hashMap.

NickAragua (changes requested):

MekHQ/src/mekhq/MekHQOptions.java
Windchild292: This is purposefully left there so that one can easily tell the RGB colour code of the default value.

MekHQ/src/mekhq/MekHQOptions.java
Windchild292: First, there are 60 methods, not 50k... extremely hyperbole.

Second, it would be extremely inefficient to convert from a String saved HashMap each time they are accessed. I could do an extremely large switch statement that would get larger as I start fixing further colour accessibility issues in MekHQ, but... I'd rather read a bunch of simple getters/setters.

The benefits of this implementation are in its persistence. However, the backend stores String, boolean, double, float, int, and long. Thus, it may require more methods and code in MekHQ Options, but the user benefit is worth one a longer file.

If we were using a regular write to file XML implementation instead of the Preferences implementation, that would be exactly how this would have been implemented. However, that is not how this system works nor the benefits of it.

NickAragua: Looking into it further, I'll buy the argument that changes to the "back end" aren't necessary, since it appears to already be organized that way more or less. But it seems to me that you can still reduce the 60 methods to 2 by passing the constant (WHATEVER_COLOR_BACKGROUND/FOREGROUND) in as an argument. In terms of human readability, there's not much difference between getCrippledUnitBackgroundColor and getColor(CRIPPLED_UNIT_BACKGROUND), but the latter lets you avoid having to copy+paste new color settings as you add them.

Windchild292: Each one has a unique foreground and background default and store value. These require unique keys and individually specified default return values. The only way I can see this can be done in your suggested way would be an extremely long switch statement. I could simplify the setters, but... honestly I'd find it more confusing going forwards.

NickAragua: public void setColor(string colorTypeConstant, Color value) {
        userPreferences.node(MekHqConstants.DISPLAY_NODE).putInt(colorTypeConstant, value.getRGB());
    }

public void getColor(string colorTypeConstant) {
        userPreferences.node(MekHqConstants.DISPLAY_NODE).getInt(colorTypeConstant, defaultColors.get(colorTypeConstant));
    }

private Map<String, Color> defaultColors = new HashMap<>();
[for each thing where we need a default color]
defaultColors.put(colorTypeConstant, defaultColor)

NickAragua: Either that or put the actual defaults into the default config file.

Windchild292: First, that is not how the base system works or is designed to work. Please take a look at how the rest of the setup works.

Second, I'm not going to change how the base system is designed to decrease the number of getters/setters or move off this setup. This file was designed to prevent errors by using simple getters/setters to access the base store instead of anything more complicated, as otherwise you can end up with duplicated default values (which was a bug in the initial implementation).

Finally, a default config file is exactly what this system is designed to prevent. All default config is inline, but easily user settable... and once set holds between implementations. That is a far better and more modern approach than having easily a half dozen separate default config files.

Windchild292: Further, that is also an illegal HshMap. Preferences and thus the defaults only handle primitives values. Colours store as integers.

Windchild292:
@NickAragua is there a reason other than style differences that this is still open?

NickAragua (approved):
Well, probably won't make much of a difference either way. Let 'er rip.

Issue #2283, opened 2020-12-13
47.15 Contracts Cannot Be Generated


jschmetzer:
### Environment
47.15
Java version 1.8.0_251
Platform Mac OS X 10.16 (x86_64)
Total memory available to MegaMek: 1,864,192 kB

Fresh campaign, 3030, on Galatea, stock AtB Default campaign presets.

1. No contracts available on campaign first load.
2. GM Generate yields error: Unable to GM-generate contract from the current location. Increase contract search radius or travel to a border world for better odds of success.
3. Log shows recurrent: WARN [mekhq.campaign.market.ContractMarket] {AWT-EventQueue-0} 
generateAtBContract(), line 323 : Could not generate an AtB Contract because we ran out of retries!

[contract_gen_fail.zip](https://github.com/MegaMek/mekhq/files/5684805/contract_gen_fail.zip)
cpnx, logs attached

NickAragua:
Confirmed, I've encountered this as well in my campaign.

sixlettervariables:
Were there other warnings? Usually it spits out a Why it had to retry.

NickAragua:
Nope, nothing else.

sixlettervariables:
FFS: https://github.com/MegaMek/mekhq/blob/3d6f5a5c9539967bee6ffd9b8ffa0f18dc5ec54f/MekHQ/src/mekhq/campaign/market/ContractMarket.java#L301-L302

If you pass 0, you get 0.

NickAragua:
Heh, doesn't leave much room for a workaround.

sixlettervariables:
Yeah didn't see this bug in my last push to fix before RC1.

PR #2291, opened 2020-12-16
Issue #2283: Ensure contracts are generated with the correct number of retries

sixlettervariables:
When I fixed a bug with contract generation hanging due to retries not being respected, I introduced another bug in contract generation. Now, when retry handling is done outside the `generateAtBContract` method, rather than `0` being sent for retries we send `1`, fixing #2283.

I'm still evaluating how (if at all) the current code can be unit tested, so that all three legs of contract generation get tested.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/ContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/ContractMarket.java
@@ -70,6 +70,11 @@ public class ContractMarket implements Serializable {
     public final static int CLAUSE_NUM = 4;
 
     /**
+     * An arbitrary maximum number of attempts to generate a contract.
+     */
+    private final static int MAXIMUM_GENERATION_RETRIES = 3;
+
+    /**
      * An arbitrary maximum number of attempts to find a random employer faction that
      * is not a Mercenary.
      */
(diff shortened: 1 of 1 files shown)


NickAragua (approved):

MekHQ/src/mekhq/campaign/market/ContractMarket.java
```diff
@@ -69,6 +69,11 @@
     public final static int CLAUSE_TRANSPORT = 3;
     public final static int CLAUSE_NUM = 4;
 
+    /**
+     * An arbitrary maximum number of attempts to generate a contract.
+     */
+    private final static int MAXIMUM_GENERATION_RETRIES = 3;
```
Windchild292, line 75: We should probably make this a campaign option.

sixlettervariables: I'm working on a follow-up with unit tests, I'll expand that there.

NickAragua: Not really sure why. From a user perspective, there's no clear relationship between it and program behavior.

sixlettervariables: I guess to help with odd issues.

PR #2302, opened 2020-12-20
Fixing Log Display Spacing

Windchild292:
This fixes a minor display issue whereby the dates are sometimes cut off.


sixlettervariables (approved):
Could you get rid of the arbitrary constant by adding two characters to the string?

Windchild292:
We can, and done.

sixlettervariables (approved):

Issue #2332, opened 2020-12-29
Refitting Thunderbolt 5S with Artemis launchers results in customization error


mjbroekman:
### Environment
What version of MekHQ does your issue apply to?  0.47.15
What operating system are you using?  Windows, Mac, or Linux? Mac
What java version are you using?  If unsure it will state your java version at the top of the mekhqlog.txt file.
	Compiled on 2020-12-15T00:28:16Z
	Today is 2020-12-28
	Java vendor N/A
	Java version 15.0.1
	Platform Mac OS X 10.15.7 (x86_64)
	Total memory available to MegaMek: 4,194,304 kB

### Description
I am attempting to refit my Thunderbolt 5S with upgraded equipment (artemis FCS, double heat sinks, endo steel). When I click on "Begin Refit", it gives a customization error:
![Screen Shot 2020-12-28 at 11 25 47 PM](https://user-images.githubusercontent.com/5141448/103258735-56fe5d80-4964-11eb-83ad-ed672352093e.png)

Equipment layout:
![Screen Shot 2020-12-28 at 11 24 55 PM](https://user-images.githubusercontent.com/5141448/103258753-70070e80-4964-11eb-9bea-a4b48011bc08.png)

At first, I only had one Artemis FCS unit (in the head), but after seeing the construction rules, I added a second unit and placed them in the torsos with the launchers.

### Files
Provide the .cpnx campaign file that this issue occurs in.
[Vidarion30390225.cpnx.gz](https://github.com/MegaMek/mekhq/files/5749611/Vidarion30390225.cpnx.gz)

And provide the megameklog.txt and mekhqlog.txt files for any NPE's or errors that MekHQ might be reporting.
[megameklog.txt](https://github.com/MegaMek/mekhq/files/5749613/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/5749614/mekhqlog.txt)

sixlettervariables:
Error in megameklog suggests the name is already taken (or not a valid filename):
```
23:25:39,076 ERROR [mekhq.campaign.parts.Refit] {AWT-EventQueue-0} 
saveCustomization(), line 1584 : A file already exists with the custom name data/mechfiles/customs/Vidarion/Thunderbolt TDR-5V Mk II.mtf. Please choose a different name. (Unit name and/or model)
[...]

PR #2334, opened 2020-12-29
Issue #2332: Delete refit file if unable to read it back when saving

sixlettervariables:
When we begin a custom refit, we save the file to disk and then load it back into the cache. However, if we encounter a fatal error at this point and cannot read the file we simply alert the user. The file itself is never removed from disk. This causes an odd situation where you are allowed to make a mech with the same name, but it won't save it back to disk and each attempt at fixing or recreating the mech will fail unless you rename it.

This changes the behavior of Refit::saveCustomization slightly, in that if the refit file cannot be read back by MHQ the file itself is deleted and the cache is refreshed. The downside is that we won't have the files to look back at to see why it failed on load. The upside is that users do not have to take manual actions to restore their campaign.

Fixes part of #2332.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Refit.java
+++ b/MekHQ/src/mekhq/campaign/parts/Refit.java
@@ -1561,7 +1561,7 @@ public class Refit extends Part implements IAcquisitionWork {
         String fileNameCampaign;
         try {
             if (newEntity instanceof Mech) {
-                //if this file already exists then don't overwrite it or we will end up with a bunch of copies
+                // if this file already exists then don't overwrite it or we will end up with a bunch of copies
                 String fileOutName = sCustomsDir + File.separator + fileName + ".mtf";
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/Refit.java
```diff
@@ -1558,41 +1558,67 @@ public void saveCustomization() throws EntityLoadingException {
...
-        if (null == summary) {
-            throw(new EntityLoadingException());
-        }
 
-        newEntity = new MechFileParser(summary.getSourceFile(), summary.getEntryName()).getEntity();
+        try {
+            MechSummaryCache.getInstance().loadMechData();
+
+            //I need to change the new entity to the one from the mtf file now, so that equip
+            //numbers will match
```
Windchild292, line 1594: Space after //, and equip should be the full word

Windchild292 (approved):

Windchild292 (approved):

Issue #2335, opened 2020-12-30
47.15 Campaign options changes after clicking "New Campaign" and canceling


gsparks3:
### Description
This seems to be equivalent to #1939, possibly a revival of the same issue somehow. I accidentally clicked "New Campaign" while playing my current campaign, hit "Cancel" a bunch of times because I didn't actually want to create a new campaign, and then checked my campaign settings (remembering what had happened previously). Said settings had in fact changed (again, particularly noticeable on the Experience and Skills tabs), and the changes were saved to the campaign file upon saving (as a _different_ file than my previous manual save, which still has the original options for comparison, since I was lucky enough to remember the issue and check first ...)

### Files
No custom units or other files in either campaign.

Save 1: original campaign settings.
[Coppercoats.cpnx.gz](https://github.com/MegaMek/mekhq/files/5752744/Coppercoats.cpnx.gz)

Save 2: altered campaign settings after "New Campaign" and canceling. 
[CoppercoatsTest.cpnx.gz](https://github.com/MegaMek/mekhq/files/5752747/CoppercoatsTest.cpnx.gz)

Nothing in the log files.

IllianiCBT:
This was resolved in 2024 during a complete rewrite of Campaign Options. Closing.

PR #2337, opened 2021-01-01
Shopping List No Longer Writes to XML While Empty and Cleanup

Windchild292:
While investigating #2335 I found that we write an empty shopping list to file. This PR removes that and cleans up spacing and formatting in the file in question.

To review: Wait until stable, then this can be merged whenever.


MekHQ/src/mekhq/campaign/market/ShoppingList.java
```diff
@@ -37,58 +36,66 @@
...
+    public ShoppingList(List<IAcquisitionWork> shoppingList) {
+        setShoppingList(shoppingList);
     }
-    
+
     public ShoppingList(IAcquisitionWork item) {
-        shoppingList = new ArrayList<IAcquisitionWork>();   
-        shoppingList.add(item);
+        setShoppingList(new ArrayList<>());
+        getShoppingList().add(item);
```
sixlettervariables, line 82: Do we want to have a mutable list returned? We've been paring those down in other places.

Windchild292: Missed my last commit cleaning up the old code seen here. There were no actual uses of this, so I've removed the code in question.

NickAragua (approved):

Issue #2342, opened 2021-01-04
Custom Units in XML with reserved characters in their name cannot be written to disk


sixlettervariables:
### Environment
What version of MekHQ does your issue apply to? 0.46.1+
What operating system are you using?  Windows, Mac, and/or Linux.
What java version are you using?  Java 8 and Java 11

### Description
Exception seen loading campaign from #2341.
```
14:50:01,200 INFO [mekhq.campaign.io.CampaignXmlParser] {SwingWorker-pool-1-thread-1} 
processCustom(), line 1129 : Loading Custom unit from XML...

14:50:01,200 ERROR [mekhq.campaign.io.CampaignXmlParser] {SwingWorker-pool-1-thread-1} 
processCustom(), line 1135 : data\mechfiles\customs\Brothers\Atlas AS7-D Mk II "Snipey McGee".mtf (The filename, directory name, or volume label syntax is incorrect)
java.io.FileNotFoundException: data\mechfiles\customs\Brothers\Atlas AS7-D Mk II "Snipey McGee".mtf (The filename, directory name, or volume label syntax is incorrect)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(Unknown Source)
	at java.io.FileOutputStream.<init>(Unknown Source)
	at java.io.FileOutputStream.<init>(Unknown Source)
	at mekhq.campaign.io.CampaignXmlParser.processCustom(CampaignXmlParser.java:1130)
	at mekhq.campaign.io.CampaignXmlParser.parse(CampaignXmlParser.java:195)
	at mekhq.campaign.CampaignFactory.createCampaign(CampaignFactory.java:83)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:193)
	at mekhq.gui.dialog.DataLoadingDialog$Task.doInBackground(DataLoadingDialog.java:125)
	at javax.swing.SwingWorker$1.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at javax.swing.SwingWorker.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)

```

### Files
[...]

PR #2343, opened 2021-01-04
Issue #2342: Customs in CPNX with invalid file name chars do not save to disk

sixlettervariables:
Found this while debugging another campaign issue, but basically we were not escaping the name of the custom before writing it to disk. This brings the code over from Refit and unblocks this problem for customs that may not have valid file names for certain operating systems.


Windchild292:
If the code is used twice it would best be moved to MekHQXMLUtil, or MegaMekXMLUtil if it is used in MM too

sixlettervariables:
It's not identical, no. This code writes the string MTF/BLK to disk, the refit logic serializes the Entity to disk.

NickAragua (approved):

PR #2351, opened 2021-01-07
PlanetaryAcquisitionFactionLimit enum

Windchild292:
This moves the planetary acquisition faction limit to an enum.


MekHQ/src/mekhq/campaign/enums/PlanetaryAcquisitionFactionLimit.java
```diff
@@ -0,0 +1,111 @@
...
+    public boolean isSelf() {
+        return this == SELF;
+    }
+
+    public boolean generateEnemy() {
+        return isAll();
+    }
+
+    public boolean generateNeutral() {
+        return generateEnemy() || isNeutral();
```
sixlettervariables, line 68: Should this instead be `isAll() || isNeutral()`? Seems off that asking to generate enemies would also bring along neutrals (not speaking to the logic in question, just the words/concepts used).

MekHQ/src/mekhq/campaign/enums/PlanetaryAcquisitionFactionLimit.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public boolean generateAllied() {
+        return generateNeutral() || isAllied();
```
sixlettervariables, line 72: Likewise, I'd expect this to be `isAll() || isAllied()` based on the name of the method.

MekHQ/src/mekhq/campaign/enums/PlanetaryAcquisitionFactionLimit.java
Windchild292: The logic is that if you allow planetary acquisitions from an enemy-held planet, you also allow acquisitions from neutral planets. This carries forwards so that if you allow generation on a neutral planet, you also allow it on allied planets. Maybe changing the name to generateOnNeutralPlanets would be better?

Windchild292: I've changed the names of the methods in question. The logic is the same, but naming it as generateOn{{Type}}Planet should prevent the confusion in future.

MekHQ/src/mekhq/campaign/enums/PlanetaryAcquisitionFactionLimit.java
```diff
@@ -0,0 +1,111 @@
...
+                    return SELF;
+                case 1:
+                default:
+                    return NEUTRAL;
+            }
+        } catch (Exception ignored) {
+
+        }
+
+        MekHQ.getLogger().error("Unable to parse a PlanetaryAcquisitionFactionLimit from " + text
```
NickAragua, line 100: Why not put this and the return into the catch? An empty catch looks weird.

NickAragua (changes requested):
Minor quibble, looks fine otherwise.

MekHQ/src/mekhq/campaign/enums/PlanetaryAcquisitionFactionLimit.java
Windchild292: It's been the style I've been using for all of these, mostly just to keep the first and second trys looking the same. It would look odd if this enum was a different style to the rest of them.

NickAragua: On that subject, why do we have valueOf followed up by a parseInt? Under what conditions would valueOf fail to return a meaningful value but a parseInt succeed?

Windchild292: Legacy parsing. We currently save it as an integer. This replaces it with an enum, which saves using the name.

NickAragua: How about enclosing both parse attempts in the same try/catch block, since it doesn't really add anything to have two separate try/catch blocks?

Windchild292: Actually, the two are required. ValueOf exceptions out if it isn't valid, which then goes to the integer try catch

NickAragua (approved):
Thinking about it some more, since java doesn't have a TryParse construct, there's not really a way to gracefully handle the situation in PlanetaryAcquisitionFactionLimit.parseFromString() without the empty exceptions, so let's get this merged.

PR #2381, opened 2021-01-19
Repair Tab: Adding better information blockers for Null/Void Signature System and Chameleon Shield

Windchild292:
This adds more informative blockers for the Null Signature System, Void Signature System, and Chameleon Shield. I'm not sold on how this is done, but it does work properly.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MissingMekLocation.java
+++ b/MekHQ/src/mekhq/campaign/parts/MissingMekLocation.java
@@ -27,6 +27,7 @@ import megamek.common.EquipmentType;
 import megamek.common.IArmorState;
 import megamek.common.LandAirMech;
 import megamek.common.Mech;
+import megamek.common.MiscType;
 import megamek.common.TechAdvancement;
 import mekhq.MekHqXmlUtil;
 import mekhq.campaign.Campaign;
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/parts/MissingMekLocation.java
```diff
@@ -265,7 +266,19 @@ public String checkFixable() {
                         }
                     }
                 }
+            } else if (slot.getType() == CriticalSlot.TYPE_EQUIPMENT) {
+                if ((slot.getMount() != null) && !slot.getMount().isDestroyed()) {
+                    EquipmentType equipmentType = slot.getMount().getType();
+                    if (equipmentType.hasFlag(MiscType.F_NULLSIG)) {
+                        return "Null-Signature System must be salvaged or scrapped first. It can then be re-installed.";
```
NickAragua, line 273: Not sure that the english here parses that great. You can't re-install a system if you scrap it.

NickAragua (changes requested):
Text adjustment. Also, I don't really see a point to changing EquipmentPart.java if all you're doing is whitespace changes.

MekHQ/src/mekhq/campaign/parts/MissingMekLocation.java
Windchild292: I agree, but wanted to keep it consistent with the other messages. We could rewrite them all or leave it like this.

Windchild292:
The EquipmentPart changes are an automatic IDE cleanup after I reverted some changes, as I originally thought the issue was there

MekHQ/src/mekhq/campaign/parts/MissingMekLocation.java
NickAragua: Let's get rid of the re-installed bit. It's pretty redundant.

Windchild292: Pushed that up, with it being removed from all of the messages with it.

NickAragua (approved):

PR #2409, opened 2021-02-06
Issue #2396: Resolve Infantry part and other part type bugs

### LINKED ISSUE #2396: Unable to load saved 47.16 campaign
After saving a campaign in 47.16, I cannot load it, or most of the auto-saves.  The only thing I can point to, is

- a change to a conventional infantry platoon, which seems to show up in the log.
- an exception being thrown about battle armor

[Brilliant Light III-Gamma30741228.cpnx.gz](https://github.com/MegaMek/mekhq/files/5861613/Brilliant.Light.III-Gamma30741228.cpnx.gz)
[megameklog.txt](https://github.com/MegaMek/mekhq/files/5861614/megameklog.txt)
[BIIIG Customs.zip](https://github.com/MegaMek/mekhq/files/5861615/BIIIG.Customs.zip)
[BIIIG-BA.zip](https://github.com/MegaMek/mekhq/files/5861618/BIIIG-BA.zip)

sixlettervariables:
When we initialize a unit, there are certain assumptions about where a part came from. If a refit went "bad" and picked the wrong type of part, you can end up in a situation where the unit cannot be initialized. Two approaches are taken here to resolve this issue:
1. If the entity type for a BA part isn't BattleArmor, remove the part from the unit.
2. For all of the sub-classes and extendable parts, make `isSamePartType` use strict class equality rather than `instanceof`.

The first approach directly fixes #2396.

The second approach hopes to make this a non-issue going forward. We shot ourselves in the foot a bit where we have base classes that are not abstract. Ideally, EquipmentPart would be abstract (as would Armor and AmmoBin), and the various type or entity specific classes would inherit from them. At that point `instanceof` would be a valid check.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
@@ -19,6 +19,7 @@
 package mekhq.campaign.parts.equipment;
 
 import java.io.PrintWriter;
+import java.util.Objects;
 
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
@@ -121,7 +122,8 @@ public class MissingLargeCraftAmmoBin extends MissingAmmoBin {
         // than an LargeCraftAmmoBin. Subclasses should use a similar check, which
         // breaks Composability to a degree but in this case we've used
         // subclasses where they're not truly composable.
-        if ((part instanceof LargeCraftAmmoBin) && (part.getClass() == LargeCraftAmmoBin.class)) {
+        if ((part instanceof LargeCraftAmmoBin)
+                && Objects.equals(part.getClass(), LargeCraftAmmoBin.class)) {
             LargeCraftAmmoBin ammoBin = (LargeCraftAmmoBin) part;
             return getType().equals(ammoBin.getType())
                     && (getFullShots() == ammoBin.getFullShots());
--- a/MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
@@ -122,8 +122,7 @@ public class MissingLargeCraftAmmoBin extends MissingAmmoBin {
         // than an LargeCraftAmmoBin. Subclasses should use a similar check, which
         // breaks Composability to a degree but in this case we've used
         // subclasses where they're not truly composable.
-        if ((part instanceof LargeCraftAmmoBin)
-                && Objects.equals(part.getClass(), LargeCraftAmmoBin.class)) {
+        if (Objects.equals(part.getClass(), LargeCraftAmmoBin.class)) {
             LargeCraftAmmoBin ammoBin = (LargeCraftAmmoBin) part;
             return getType().equals(ammoBin.getType())
                     && (getFullShots() == ammoBin.getFullShots());
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -2330,23 +2330,23 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
                 if (!(entity instanceof BattleArmor)) {
                     partsToRemove.add(part);
                 } else {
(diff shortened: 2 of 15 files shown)


MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
```diff
@@ -121,7 +122,8 @@ public boolean isAcceptableReplacement(Part part, boolean refit) {
         // than an LargeCraftAmmoBin. Subclasses should use a similar check, which
         // breaks Composability to a degree but in this case we've used
         // subclasses where they're not truly composable.
-        if ((part instanceof LargeCraftAmmoBin) && (part.getClass() == LargeCraftAmmoBin.class)) {
+        if ((part instanceof LargeCraftAmmoBin)
+                && Objects.equals(part.getClass(), LargeCraftAmmoBin.class)) {
```
Windchild292, line 126: Why the instanceof usage here? Wouldn't that be covered in the Objects.equals?

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -2325,19 +2327,27 @@ public void initializeParts(boolean addParts) {
...
-                    parts = new Part[((BattleArmor)entity).getSquadSize()];
+                if (!(entity instanceof BattleArmor)) {
+                    partsToRemove.add(part);
+                } else {
+                    Part[] parts = baEquipParts.get(((MissingBattleArmorEquipmentPart)part).getEquipmentNum());
+                    if (null == parts) {
+                        parts = new Part[((BattleArmor)entity).getSquadSize()];
+                    }
+                    parts[((MissingBattleArmorEquipmentPart)part).getTrooper()-BattleArmor.LOC_TROOPER_1] = part;
+                    baEquipParts.put(((MissingBattleArmorEquipmentPart)part).getEquipmentNum(), parts);
```
Windchild292, line 2349:
```suggestion
                    Part[] parts = baEquipParts.get(((MissingBattleArmorEquipmentPart) part).getEquipmentNum());
                    if (null == parts) {
                        parts = new Part[((BattleArmor) entity).getSquadSize()];
                    }
                    parts[((MissingBattleArmorEquipmentPart) part).getTrooper() - BattleArmor.LOC_TROOPER_1] = part;
                    baEquipParts.put(((MissingBattleArmorEquipmentPart) part).getEquipmentNum(), parts);
```

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -2325,19 +2327,27 @@ public void initializeParts(boolean addParts) {
...
-                    parts = new Part[((BattleArmor)entity).getSquadSize()];
+                if (!(entity instanceof BattleArmor)) {
+                    partsToRemove.add(part);
+                } else {
+                    Part[] parts = baEquipParts.get(((BattleArmorEquipmentPart)part).getEquipmentNum());
+                    if (null == parts) {
+                        parts = new Part[((BattleArmor)entity).getSquadSize()];
+                    }
+                    parts[((BattleArmorEquipmentPart)part).getTrooper()-BattleArmor.LOC_TROOPER_1] = part;
+                    baEquipParts.put(((BattleArmorEquipmentPart)part).getEquipmentNum(), parts);
```
Windchild292, line 2338:
```suggestion
                    Part[] parts = baEquipParts.get(((BattleArmorEquipmentPart) part).getEquipmentNum());
                    if (null == parts) {
                        parts = new Part[((BattleArmor) entity).getSquadSize()];
                    }
                    parts[((BattleArmorEquipmentPart) part).getTrooper() - BattleArmor.LOC_TROOPER_1] = part;
                    baEquipParts.put(((BattleArmorEquipmentPart) part).getEquipmentNum(), parts);
```

Windchild292 (commented):
One query first before I approve.

MekHQ/src/mekhq/campaign/parts/equipment/MissingLargeCraftAmmoBin.java
sixlettervariables: I originally only had the get-class check on the base classes and just had instanceof in the subclasses, however I must not have doubled back and removed that here.

Windchild292 (approved):

PR #2420, opened 2021-02-11
Issue #2418: Ensure context menu shortcut works on our tables

### LINKED ISSUE #2418: No Way to Reliably Trigger Right-click Context Menu from Keyboard
### Environment
What version of MekHQ does your issue apply to? 0.47.17
What operating system are you using?  WIndows
What java version are you using?  1.8.0_281

### Description
User Impact
It is impossible to reliably trigger the right click context menu from the keyboard. This makes certain tasks such as assigning personnel difficult to impossible with a screen reader.

Repro Steps:
1. Activate MekHQ.
2. Load a campaign with an existing roster for simplicity.
3. Move to the personnel tab.
4. Move into the personnel table with Tab and go down a few rows as appropriate.
5. Attempt to open the context menu from the keyboard. Press shift+F10 and the applications menu key.

Actual Results
Shift+f10 opens the Windows "System," menu, with "minimize," "restore," etc. The application key appears to do nothing on my system, though this may be specific to my setup. In any case both shortcuts should ideally be supported.

Expected Results
A way should be provided to open this menu from the keyboard reliably across all platforms.

Possible Resolutions:
Add a shortcut such that shift+f10 opens the context menu on the selected item in keyboard focus, much as right click does for the mouse.

### Files

NA

sixlettervariables:
This adds a new abstract class `JPopupMenuAdapter` to help connect the keyboard chord SHIFT+F10 to all of our context menus on JTable's and JTree's. This fixes #2418.

Apparently SHIFT+F10 sometimes works, but not always. This removes all doubt.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
@@ -19,13 +19,13 @@
 package mekhq.gui.adapter;
 
 import java.awt.event.ActionEvent;
+import java.util.Optional;
 
 import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
 import javax.swing.JTable;
 
-import megamek.common.annotations.Nullable;
 import mekhq.MekHQ;
 import mekhq.campaign.event.TransactionChangedEvent;
 import mekhq.campaign.event.TransactionVoidedEvent;
@@ -74,16 +74,10 @@ public class FinanceTableMouseAdapter extends JPopupMenuAdapter {
     }
 
     @Override
-    protected boolean shouldShowPopup() {
-        return financeTable.getSelectedRowCount() > 0;
-    }
-
-    @Override
-    @Nullable
-    protected JPopupMenu createPopupMenu() {
+    protected Optional<JPopupMenu> createPopupMenu() {
         int row = financeTable.getSelectedRow();
-        if (row < 0) {
-            return null;
+        if ((row < 0) || !gui.getCampaign().isGM()) {
+            return Optional.empty();
         }
 
         JPopupMenu popup = new JPopupMenu();
@@ -94,15 +88,13 @@ public class FinanceTableMouseAdapter extends JPopupMenuAdapter {
         JMenuItem deleteItem = new JMenuItem("Delete Transaction");
         deleteItem.setActionCommand("DELETE");
         deleteItem.addActionListener(this);
-        deleteItem.setEnabled(gui.getCampaign().isGM());
         menu.add(deleteItem);
 
         JMenuItem editItem = new JMenuItem("Edit Transaction");
         editItem.setActionCommand("EDIT");
         editItem.addActionListener(this);
-        editItem.setEnabled(gui.getCampaign().isGM());
         menu.add(editItem);
 
-        return popup;
+        return Optional.of(popup);
     }
 }
--- a/MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
@@ -19,15 +19,13 @@
 package mekhq.gui.adapter;
 
 import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
-import java.awt.event.MouseEvent;
 
 import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
 import javax.swing.JTable;
-import javax.swing.event.MouseInputAdapter;
 
+import megamek.common.annotations.Nullable;
 import mekhq.MekHQ;
 import mekhq.campaign.event.TransactionChangedEvent;
 import mekhq.campaign.event.TransactionVoidedEvent;
@@ -36,16 +34,20 @@ import mekhq.gui.CampaignGUI;
 import mekhq.gui.dialog.EditTransactionDialog;
 import mekhq.gui.model.FinanceTableModel;
 
-public class FinanceTableMouseAdapter extends MouseInputAdapter implements ActionListener {
+public class FinanceTableMouseAdapter extends JPopupMenuAdapter {
     private CampaignGUI gui;
     private JTable financeTable;
     private FinanceTableModel financeModel;
 
-    public FinanceTableMouseAdapter(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
-        super();
+    protected FinanceTableMouseAdapter(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
         this.gui = gui;
         this.financeTable = financeTable;
         this.financeModel = financeModel;
+    }
+
+    public static void connect(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
+        new FinanceTableMouseAdapter(gui, financeTable, financeModel)
+                .connect(financeTable);
     }
 
     @Override
@@ -72,38 +74,35 @@ public class FinanceTableMouseAdapter extends MouseInputAdapter implements Actio
     }
 
     @Override
-    public void mousePressed(MouseEvent e) {
-        maybeShowPopup(e);
+    protected boolean shouldShowPopup() {
+        return financeTable.getSelectedRowCount() > 0;
     }
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
-
-    private void maybeShowPopup(MouseEvent e) {
-        JPopupMenu popup = new JPopupMenu();
-        if (e.isPopupTrigger()) {
-            int row = financeTable.getSelectedRow();
-            if (row < 0) {
-                return;
-            }
-            JMenu menu = new JMenu("GM Mode");
-            popup.add(menu);
-
-            JMenuItem deleteItem = new JMenuItem("Delete Transaction");
-            deleteItem.setActionCommand("DELETE");
-            deleteItem.addActionListener(this);
-            deleteItem.setEnabled(gui.getCampaign().isGM());
-            menu.add(deleteItem);
-
-            JMenuItem editItem = new JMenuItem("Edit Transaction");
-            editItem.setActionCommand("EDIT");
-            editItem.addActionListener(this);
-            editItem.setEnabled(gui.getCampaign().isGM());
-            menu.add(editItem);
-
-            popup.show(e.getComponent(), e.getX(), e.getY());
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        int row = financeTable.getSelectedRow();
+        if (row < 0) {
+            return null;
         }
+
+        JPopupMenu popup = new JPopupMenu();
+
+        JMenu menu = new JMenu("GM Mode");
+        popup.add(menu);
+
+        JMenuItem deleteItem = new JMenuItem("Delete Transaction");
+        deleteItem.setActionCommand("DELETE");
+        deleteItem.addActionListener(this);
+        deleteItem.setEnabled(gui.getCampaign().isGM());
+        menu.add(deleteItem);
+
+        JMenuItem editItem = new JMenuItem("Edit Transaction");
+        editItem.setActionCommand("EDIT");
+        editItem.addActionListener(this);
+        editItem.setEnabled(gui.getCampaign().isGM());
+        menu.add(editItem);
+
+        return popup;
     }
 }
--- a/MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2021 - The MegaMek Team
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MekHQ is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MekHQ. If not, see <http://www.gnu.org/licenses/>.
+ */
+package mekhq.gui.adapter;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.event.InputEvent;
+import java.awt.event.KeyEvent;
+import java.awt.event.MouseEvent;
+
+import javax.swing.AbstractAction;
+import javax.swing.JComponent;
+import javax.swing.JPopupMenu;
+import javax.swing.KeyStroke;
+import javax.swing.event.MouseInputAdapter;
+
+import megamek.common.annotations.Nullable;
+
+/**
+ * Provides a popup menu adapter for a component which also ensures that
+ * the accessibility chord SHIFT+F10 opens the popup as well.
+ */
+public abstract class JPopupMenuAdapter extends MouseInputAdapter implements ActionListener {
+    public static final String COMMAND_OPEN_POPUP = "SHIFT_F10";
+
+    /**
+     * Connect the popup menu adapter to the component. Implementations should call this
+     * to connect the popup menu to both right click and the SHIFT+F10 accessibility chord.
+     * @param component The component to trap context menu actions.
+     */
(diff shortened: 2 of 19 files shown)


MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
```diff
@@ -72,38 +74,35 @@ public void actionPerformed(ActionEvent action) {
...
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        int row = financeTable.getSelectedRow();
+        if (row < 0) {
+            return null;
+        }
```
Windchild292, line 87: I'm pretty sure this is not necessary with the check in shouldShowPopup.

MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
```diff
@@ -72,38 +74,35 @@ public void actionPerformed(ActionEvent action) {
     }
 
     @Override
-    public void mousePressed(MouseEvent e) {
-        maybeShowPopup(e);
+    protected boolean shouldShowPopup() {
+        return financeTable.getSelectedRowCount() > 0;
```
Windchild292, line 78:
```suggestion
        return (financeTable.getSelectedRowCount() > 0) && gui.getCampaign().isGM();
```

Windchild292: The popup is only a GM menu, and thus shouldn't show up if they aren't a GM.

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Override
+    public void mousePressed(MouseEvent e) {
+        maybeShowPopup(e);
+    }
+
+    @Override
+    public void mouseReleased(MouseEvent e) {
+        maybeShowPopup(e);
+    }
```
Windchild292, line 82: Could this cause it to show up twice?

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                // Immediately return if there are no units selected
+                if (!shouldShowPopup()) {
+                    return;
+                }
+
+                JPopupMenu popup = createPopupMenu();
+                popup.show(component, component.getX(), component.getY());
```
Windchild292, line 64: Missing null check

MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
```diff
@@ -72,38 +74,35 @@ public void actionPerformed(ActionEvent action) {
...
     }
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        int row = financeTable.getSelectedRow();
```
Windchild292, line 84: Why is the nullable like this instead of 
```
protected @Nullable JPopupMenu createPopupMenu()
```

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public void mouseReleased(MouseEvent e) {
+        maybeShowPopup(e);
+    }
+
+    private void maybeShowPopup(MouseEvent e) {
+        if (e.isPopupTrigger()) {
+            // Immediately return if there are no units selected
```
Windchild292, line 86: The comment doesn't make sense anymore

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+    protected abstract boolean shouldShowPopup();
+
+    /**
+     * Creates the popup menu, or returns {@code null} if the
+     * popup menu should not be shown.
+     * @return A {@link JPopupMenu} to show, or {@code null} if no
+     *         popup menu should be shown.
+     */
+    @Nullable
+    protected abstract JPopupMenu createPopupMenu();
```
Windchild292, line 110: Same nullable question

MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
```diff
@@ -36,18 +34,22 @@
...
 
-    public FinanceTableMouseAdapter(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
-        super();
+    protected FinanceTableMouseAdapter(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
         this.gui = gui;
         this.financeTable = financeTable;
         this.financeModel = financeModel;
     }
 
+    public static void connect(CampaignGUI gui, JTable financeTable, FinanceTableModel financeModel) {
```
Windchild292, line 48: Could this being static cause issues with your goal of testing these components?

MekHQ/src/mekhq/gui/adapter/LoanTableMouseAdapter.java
```diff
@@ -89,49 +89,46 @@ public void actionPerformed(ActionEvent action) {
...
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        if (loanTable.getSelectedRowCount() == 0) {
+            return null;
+        }
```
Windchild292, line 101: This check isn't necessary given the check in shouldShowPopup

MekHQ/src/mekhq/gui/adapter/LoanTableMouseAdapter.java
```diff
@@ -89,49 +89,46 @@ public void actionPerformed(ActionEvent action) {
...
-            menuItem.setEnabled(gui.getCampaign().isGM());
-            menu.add(menuItem);
-            // end
-            popup.addSeparator();
-            popup.add(menu);
-            popup.show(e.getComponent(), e.getX(), e.getY());
-        }
+
+        int row = loanTable.getSelectedRow();
+        Loan loan = loanModel.getLoan(loanTable.convertRowIndexToModel(row));
```
Windchild292, line 106: could inline this

MekHQ/src/mekhq/gui/adapter/PartsTableMouseAdapter.java
```diff
@@ -250,170 +242,178 @@ public boolean areAllPartsPodded(Part[] parts) {
...
+    protected boolean shouldShowPopup() {
+        return partsTable.getSelectedRowCount() > 0;
+    }
+
+    @Override
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        if (partsTable.getSelectedRowCount() == 0) {
+            return null;
+        }
```
Windchild292, line 255: Should be removable.

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -137,20 +137,40 @@
...
+                    int size = splitPersonnel.getDividerSize();
+                    if ((width - location + size) < PersonnelTab.PERSONNEL_VIEW_WIDTH) {
+                        // expand
+                        splitPersonnel.resetToPreferredSizes();
+                    } else {
+                        // collapse
+                        splitPersonnel.setDividerLocation(1.0);
+                    }
+                }
+            }
```
Windchild292, line 170: Why are we overriding this here when it is only ever used here? Why not have it as part of the base class?

MekHQ/src/mekhq/gui/adapter/ProcurementTableMouseAdapter.java
```diff
@@ -1,145 +1,131 @@
...
-                        for (int curRow : rows) {
-                            if (curRow < 0) {
-                                continue;
-                            }
-                            model.getAcquisition(table.convertRowIndexToModel(curRow))
-                                    .ifPresent(ProcurementTableMouseAdapter.this::procureAllItems);
-                        }
+            }
+        });
+        menuItem.setEnabled(gui.getCampaign().isGM());
```
Windchild292, line 77: Noticed when reading through, this can be removed as the whole thing only shows if they are a GM

MekHQ/src/mekhq/gui/adapter/ProcurementTableMouseAdapter.java
```diff
@@ -1,145 +1,131 @@
...
-                            }
-                            model.getAcquisition(table.convertRowIndexToModel(curRow))
-                                    .ifPresent(a -> {
-                                        model.removeRow(row);
-                                        MekHQ.triggerEvent(new ProcurementEvent(a));
-                                    });
-                        }
+            }
+        });
+        menuItem.setEnabled(gui.getCampaign().isGM());
```
Windchild292, line 97: Same here, can be removed

MekHQ/src/mekhq/gui/adapter/ProcurementTableMouseAdapter.java
```diff
@@ -1,145 +1,131 @@
...
-            menuItem.setEnabled(gui.getCampaign().isGM());
-            menu.add(menuItem);
-            // end
-            popup.addSeparator();
-            popup.add(menu);
-            popup.show(e.getComponent(), e.getX(), e.getY());
-        }
+            }
+        });
+        menuItem.setEnabled(gui.getCampaign().isGM());
```
Windchild292, line 122: Same here, can be removed

MekHQ/src/mekhq/gui/adapter/ProcurementTableMouseAdapter.java
```diff
@@ -1,145 +1,131 @@
...
     }
 
-    @SuppressWarnings("serial")
-    private void maybeShowPopup(MouseEvent e) {
+    @Override
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        // GM Only (for now)
+        if ((table.getSelectedRowCount() == 0) || !gui.getCampaign().isGM()) {
+            return null;
```
Windchild292, line 45: This can be removed

MekHQ/src/mekhq/gui/adapter/ScenarioTableMouseAdapter.java
```diff
@@ -35,61 +33,62 @@
...
 
-    private void maybeShowPopup(MouseEvent e) {
+    @Override
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        int row = scenarioTable.getSelectedRow();
+        if (row < 0) {
+            return null;
+        }
+
```
Windchild292, line 66: Can't this be removed?

MekHQ/src/mekhq/gui/adapter/ServicedUnitsTableMouseAdapter.java
```diff
@@ -139,125 +140,121 @@ public void actionPerformed(ActionEvent action) {
...
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        if (servicedUnitTable.getSelectedRowCount() == 0) {
+            return null;
+        }
```
Windchild292, line 152: Can't this be removed?

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -146,13 +146,32 @@
...
+                    if ((width - location + size) < HangarTab.UNIT_VIEW_WIDTH) {
+                        // expand
+                        splitUnit.resetToPreferredSizes();
+                    } else {
+                        // collapse
+                        splitUnit.setDividerLocation(1.0);
+                    }
+                }
+            }
+        }.connect(unitTable);
```
Windchild292, line 172: Same as personnel, why are we not just overriding this is the class given it is only used there?

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -524,570 +543,565 @@ public void actionPerformed(ActionEvent action) {
...
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        if (unitTable.getSelectedRowCount() == 0) {
+            return null;
+        }
```
Windchild292, line 555: Same again, can be removed

MekHQ/src/mekhq/gui/adapter/TaskTableMouseAdapter.java
```diff
@@ -144,86 +146,88 @@ public void actionPerformed(ActionEvent action) {
...
-                return;
-            }
-            IPartWork partWork = taskModel.getTaskAt(taskTable.convertRowIndexToModel(row));
-            if (null == partWork) {
-                return;
-            }
+        int row = taskTable.getSelectedRow();
+        if (row < 0) {
+            return null;
+        }
```
Windchild292, line 167: Can be removed

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -590,539 +593,608 @@ public void actionPerformed(ActionEvent action) {
...
 
     @Override
-    public void mouseReleased(MouseEvent e) {
-        maybeShowPopup(e);
-    }
+    @Nullable
+    protected JPopupMenu createPopupMenu() {
+        if (tree.getSelectionPaths() == null) {
+            return null;
+        }
```
Windchild292, line 605: This can be removed

MekHQ/src/mekhq/gui/adapter/LoanTableMouseAdapter.java
Windchild292, line 101: Can be removed.

Windchild292 (commented):
Lots of comments, but largely they have to do with createPopupMenu having checks covered in shouldShowPopup, which is always called before it.

MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
sixlettervariables: I don't like eschewing preconditions if they're not both private; but I think I can simplify this to a single method.

MekHQ/src/mekhq/gui/adapter/FinanceTableMouseAdapter.java
sixlettervariables: No, I don't really have a good way to test Swing components unfortunately.

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
sixlettervariables: Maybe? Doesn't appear to.

sixlettervariables:
@Windchild292 I have changed the API of JPopupMenuAdapter to ensure it is used correctly from implementations. There no longer exists the possibility of a time lag between the pre-condition check and menu creation, ensuring implementations don't run into funny timing errors (e.g. menu takes a long time to create and in that space the user goes and right clicks something else).

I'll look into which mouse event to pop the context menu on.

MekHQ/src/mekhq/gui/adapter/JPopupMenuAdapter.java
sixlettervariables: This implementation is correct per:
https://docs.oracle.com/javase/tutorial/uiswing/components/menu.html#popup

Windchild292:
Just two questions left, both about the mouse click override being overridden during the construction instead of just being overridden in the adapter itself. Otherwise, the changes here look good

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
sixlettervariables: This is so I do not have to include the splitter as a variable in the class.

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
sixlettervariables: This is so I do not have to include the splitter as a variable in the class.

sixlettervariables:
The splitter is a separate behavior I did not want tied to the base details, so it is kept only in the connect logic. I tried to separate it out but it didn't seem to work the same...and I honestly gave up on figuring out why.

Windchild292 (approved):

PR #2438, opened 2021-02-23
Update AtB User Guide to new screens, add default presets.

dtsosie:
Added Presets matching guide and text to implement for new users.

Errata:
Pg 1: 
Add to ChangeLog: "V2.6 Eleventh published version (TBD): Add instructions to load campaign presets (*B*), update Personnel options (*B5*)"
Remove "B16: Miscellaneous..."

Pg 5: After first paragraph add paragraph.
"If you would like to load presets from a file. Hit the *"Confirm"* button. Your new campaign will load. Click *File->Import->Campaign Options from an xml file...*. This will load a saved campaign options file. If you would like the options preset loaded from the guide, they are located in your install directory within *"docs/AtB Stuff/Against_the_Bot_Starter_Guide_Presets.xml"*. You may continue with this guide by clicking *File->Campaign Options*.

Pg 10: 
Remove "Next, if you're fighting the Clans, you can swap the Default
Prisoner Status to Bondsman to indicate how Clanners take prisoners."
Personnel rearranged new pic included, sets both service and rank to checked to match the instructions.

Pg 11: 
Family Options rearranged new pic included

Page 16: Remove "B16: Misc" - No longer exists


Windchild292:
Don't we already have the presets in our folder? IIRC the Campaign Anon set is the starter document presets.

dtsosie:
This is a preset that matches the guide that I updated to match what is current in the AtB folder.

Windchild292 (changes requested):
I approve the PDF, however the preset needs to be removed from this PR. Campaign Anon is the origin author of this guide and thus that preset will hold as the one for the guide.

Issue #338, opened 2017-03-09
Feature Request: Sort refit technician list by type, skill and work hours


Tamren:
IIRC the refit system is being overhauled, but I figured I should mention this anyway.

When you select a refit kit or customize a vehicle the game pops up a menu that allows you to select which technician you want to perform the refit. However this list does not sort in any fashion so all of the techs from all three disciplines get jumbled together. If you pick a technician of the wrong type, such as an aero tech while refitting a mech (as I have done accidentally several times) it doesn't stop you from doing this and only tells you that the tech is the wrong type after the fact when the refit is complete.

So the optimal order to sort the list would be something like this, from High to Low.

1. The type of technician, with the proper type placed at the top.
2. The skill level of the technician.
3. The available work hours a tech has per day, minus the amount they spend on maintenance. So a tech with 0/480 minutes would be displayed higher than one with 180/180 minutes.

dtsosie:
Bug
Ordered list of refit technicians is sorted regardless of target refit unit.

Steps to reproduce:
1. New campaign
2. Enable GM Mode
3. Marketplace->Hire->Mech Tech; click "Add (GM)"
4. Marketplace->Hire->Mechanic; click "Add (GM)"
5. Marketplace->Purchase Unit; Heavy Tracked APC (MG); Add (GM)
6. Click Hanger; Right Click Heavy Tracked APC (MG); Customize->Choose Refit Kit
7. Click LRM Kit
Note Technician List.

Expected Result: Mechanic, Mech Tech
Actual Result: Mech Tech, Mechanic

Windchild292:
#2439 implements the first two. Sorting by available time has not been implemented as of yet.

PR #2439, opened 2021-02-23
Issue 338 sort technician

dtsosie:
https://github.com/MegaMek/mekhq/issues/338

Resolution of above issue. Progress over perfection. This change will notify players if they are about to use a wrong tech, and it will put the right techs per the Tech class (is right (refit target)) at the top.

I think this is best because tbh, I don't assign techs based on their skills. Some repairs take more or less long and I will assign like that. But a new player will assign a tech and only at the end discover they made a mistake. This fixes that issue.

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -1588,6 +1588,8 @@ public class CampaignGUI extends JPanel {
             int TimePerDay;
 
             List<Person> techs = getCampaign().getTechs();
+            int lastRightTech = 0;
+
             techs.sort(Comparator.comparingInt(Person::getPrimaryRole));
             for (Person tech : techs) {
                 if (getCampaign().isWorkingOnRefit(tech) || tech.isEngineer()) {
@@ -1611,7 +1613,11 @@ public class CampaignGUI extends JPanel {
                         + tech.getMinutesLeft() + "/" + TimePerDay
                         + " minutes";
                 techHash.put(name, tech);
-                techList.add(name);
+                if(tech.isRightTechTypeFor(r)) {
+                    techList.add(lastRightTech++, name);
+                } else {
+                    techList.add(name);
+                }
             }
 
             String s = (String) JOptionPane.showInputDialog(frame,
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -1628,7 +1628,20 @@ public class CampaignGUI extends JPanel {
                 return;
             }
 
-            r.setTech(techHash.get(s));
+            Person selectedTech = techHash.get(s);
+
+            if(!selectedTech.isRightTechTypeFor(r)) {
+                if(JOptionPane.NO_OPTION == JOptionPane
+                    .showConfirmDialog(
+                    null,
+                    "This tech is not appropriate for this unit. Would you like to continue?",
+                    "pending battle",
+                    JOptionPane.YES_NO_OPTION)) {
+                        return;
+                    }
+            }
+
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1611,7 +1613,11 @@ public void refitUnit(Refit r, boolean selectModelName) {
                         + tech.getMinutesLeft() + "/" + TimePerDay
                         + " minutes";
                 techHash.put(name, tech);
-                techList.add(name);
+                if(tech.isRightTechTypeFor(r)) {
```
Windchild292, line 1616:
```suggestion
                if (tech.isRightTechTypeFor(r)) {
```

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1622,7 +1628,20 @@ public void refitUnit(Refit r, boolean selectModelName) {
                 return;
             }
 
-            r.setTech(techHash.get(s));
+            Person selectedTech = techHash.get(s);
+
+            if(!selectedTech.isRightTechTypeFor(r)) {
+                if(JOptionPane.NO_OPTION == JOptionPane
+                    .showConfirmDialog(
+                    null,
```
Windchild292, line 1636:
```suggestion
            if (!selectedTech.isRightTechTypeFor(r)) {
                if (JOptionPane.NO_OPTION == JOptionPane.showConfirmDialog(null,
```

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1611,7 +1613,11 @@ public void refitUnit(Refit r, boolean selectModelName) {
                         + tech.getMinutesLeft() + "/" + TimePerDay
                         + " minutes";
                 techHash.put(name, tech);
-                techList.add(name);
+                if(tech.isRightTechTypeFor(r)) {
+                    techList.add(lastRightTech++, name);
```
Windchild292, line 1617: If this variable is removed and replaced with a 0, then you get the best techs to refit with first, while any other techs are sorted descending.

Alternatively, you could swap Line 1590 to:
```List<Person> techs = getCampaign().getTechs(false, null, true, true);```
And keep the lastRightTech, which would have the best of the proper type first, then the best of any other type, sorted by their time left.

Both changes allow one to remove the PrimaryRole sort, as why do we need to group the others by role?

Windchild292 (commented):
Some style issues with if statements, and a comment on and alternative order for implementation.

MekHQ/src/mekhq/gui/CampaignGUI.java
dtsosie: The idea was minimal change. Inserting in this order maintains the same order the user will see in the rest of the list, just with proper techs promoted to the front. I like the idea of removing PrimaryRole sort though. Back to the salt mines ...

Windchild292 (commented):
The remaining time doesn't appear to be sorted by, but otherwise this looks good.

dtsosie:
Well this is decision point time. #338 says they want to redo the whole thing anyway so is this going to be a a step toward that or a quick fix. We're already where it catches the common pain points and warns users if they are about to make a mistake. If we REALLY want to sort by time Person needs to know the skills involved. We will need to remove the primary secondary skill logic from the Tech loop in Campaign.java, incorporate it within Person.java so that the natural sort can know what times to generate. Perhaps overloading with Person.resetTimes(Refit r)? At that point we've signed ourselves up for including refit logic and this new functionality to CampaignTest.java. As this will take me some time to do I'd like feedback since I really don't know what the current course of the platform is and this will require some architecting unlike the quick bug fix I was attempting to do.

Windchild292 (approved):

PR #2440, opened 2021-02-24
Base Components

Windchild292:
This PR is divided into two groupings, new Preferences and Abstract Components. The new Abstract Components are designed to be used as a basis for MekHQ GUI components. All base components for these are accessible component, with the abstraction being only light modifications from those bases. The primary functionality of these components is implementing preferences, handling resources, and handling the frame. Some implement initialization, while others have it as an abstract method that needs to be implemented by child classes. This information is provided in a JavaDoc comment at the top of each file.

Base Components: AbstractDialog, AbstractButtonDialog, AbstractScrollPane, AbstractSplitPane, and AbstractTabbedPane
New Preferences: JSplitPanePreference, JTabbedPanePreference, JListPreference, and JDoubleNumberSpinnerPreference

To Review: The first four commits are refactoring, organizing, and standardizing preferences, 5-7 are the primary logic changes, 8-10 are bugfixes and remove ScrollPanes, and 11 re-adds the removed AbstractScrollPane (I couldn't get the preferences working, but forgot about the other use of having it standardized as abstract, namely in being able to reuse the preferences between places it is used), while 12 is the change that moves the base components and preferences to MegaMek.

All have been tested, but are left largely unimplemented as part of this PR to minimize the footprint.


MekHQ/src/mekhq/gui/baseComponents/AbstractMHQButtonDialog.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Inheriting classes must call initialize() in their constructors and override createCenterPane()
+ *
+ * The resources associated with this dialog need to contain at least the following keys:
+ * - "Ok.text" -> text for the ok button
+ * - "Ok.toolTipText" -> toolTipText for the ok button
+ * - "Cancel.text" -> text for the cancel button
+ * - "Cancel.toolTipText" -> toolTipText for the cancel button
+ */
+public abstract class AbstractMHQButtonDialog extends AbstractButtonDialog {
```
NickAragua, line 41: Gotta ask, what's different about a MekHQ button from a MegaMek button?

NickAragua (approved):
Not really much to say about this one.

MekHQ/src/mekhq/gui/baseComponents/AbstractMHQButtonDialog.java
Windchild292: Solely tracking on MHQ's preferencez instead of MM's.

Otherwise the dialogs are identical.

PR #2451, opened 2021-03-06
Issue #2446: Calculate the scrollable block increment correctly

### LINKED ISSUE #2446: Window panes not scrolling by page
### Environment
What version of MekHQ does your issue apply to? 0.47.17
What operating system are you using?  Windows
What java version are you using?  openjdk 11

### Description
Some of the detail screens are able to scroll by page when you select the bar of the scroll bar (normal Windows behavior), but many of them only scroll one line at a time instead. This makes scrolling tedious when a mouse wheel is not available.
Here are the ones I have found so far that only scroll one line at a time:

 TOE tab, Pilot and Unit detail tabs
 Briefing Room tab, Mission list
 Briefing Room tab, Scenario details
 Interstellar Map tab, System details
 Personnel tab, Personnel details
 Hanger tab, Unit details

sixlettervariables:
MekHQ uses `ScrollablePanel` as a base class for panels that should be scrolled. Its implementation uses 16 pixels for both the unit and block scrolling increment, which breaks the basic UI assumption that block scrolling should be in "pages". This restores the intended behavior of block scrolling [by adopting what is effectively the JTable implementation](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/swing/JTable.html#getScrollableUnitIncrement(java.awt.Rectangle,int,int)):
> Returns visibleRect.height or visibleRect.width, depending on this table's orientation.

Fixes #2446

### DIFF
--- a/MekHQ/src/mekhq/gui/view/ScrollablePanel.java
+++ b/MekHQ/src/mekhq/gui/view/ScrollablePanel.java
@@ -24,6 +24,7 @@ import java.awt.Rectangle;
 
 import javax.swing.JPanel;
 import javax.swing.Scrollable;
+import javax.swing.SwingConstants;
 
 /**
  * This is an extension of JPanel that implements scrollable so that all of our ViewPanel objects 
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/view/ScrollablePanel.java
```diff
@@ -47,11 +48,15 @@ public int getScrollableUnitIncrement(Rectangle visibleRect, int orientation, in
     }
 
     public int getScrollableBlockIncrement(Rectangle visibleRect, int orientation, int direction) {
-        return 16;
+        if (SwingConstants.VERTICAL == orientation) {
+            return visibleRect.height;
+        } else {
+            return visibleRect.width;
+        }
```
Windchild292, line 55: Ternary this

Windchild292 (approved):

PR #2464, opened 2021-03-12
2456: Incorrect Financial Export Date, Two Force Icon NPEs, Unit Market NPE

### LINKED ISSUE #2456: End of year finances not being output
### Environment
What version of MekHQ does your issue apply to? 0.48.0
What operating system are you using?  Windows 10
What java version are you using? AdoptOpenJDK 11.0.9.1

### Description
I've started a new campaign in 0.48.0, and for some reason, when I move into the year 3060, it isn't generating a financial report.  It did generate one at the end of 3059 (although it named it 3060 for some reason).  

I had this same issue in my 0.47.17 campaign.  I got a report at the end of 3059, but not at the end of 3060, and then I got another report at the end of 3061.

### Files
[Nightstalkers30601231.cpnx.gz](https://github.com/MegaMek/mekhq/files/6112780/Nightstalkers30601231.cpnx.gz)
[customs.zip](https://github.com/MegaMek/mekhq/files/6112783/customs.zip)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/6112787/mekhqlog.txt)
[megameklog.txt](https://github.com/MegaMek/mekhq/files/6112788/megameklog.txt)
The log files were saved after rolling over to Jan 1.

Windchild292:
This fixes #2456, which was caused by a Unit Market NPE. Further, this fixes financial date export and integrates it with the financial duration option, and fixes two Force Icon generation NPEs.

I'm going to be swapping the unit market to entirely run using faction instead of faction codes shortly.


MekHQ/src/mekhq/campaign/finances/enums/FinancialYearDuration.java
```diff
@@ -71,8 +75,38 @@ public boolean isEndOfFinancialYear(LocalDate today) {
         }
     }
 
+    /**
+     * @param tomorrow tomorrow's date. Because of how this is passed in, we are provided that
+     *                 instead of today
+     * @return the filename to use for the date during financial export
+     */
+    public String getExportFilenameDateString(final LocalDate tomorrow) {
+        final int year = tomorrow.getYear() - 1;
```
sixlettervariables, line 84: There is a lot of subtractions from year that I don't follow here, especially given the date is called "tomorrow".

So if I passed in 2021-03-14 (tomorrow), year would be 2020 and then "biannual" would be 2019-2020? Biannual means twice a year, at least where I come from, so I guess I don't follow why its two years and not two dates within the same year.

sixlettervariables (commented):
Most of this I have no problems with, but its lumped in with a change that I'm having a hard time following (unit tests would help that out).

This also isn't a good candidate for backporting for stable given the number of unrelated bugs lumped together.

NickAragua:
I'd have to agree about this not being a stable candidate. For something to be a stable candidate, it needs to be a relatively simple, isolated change, not a multi-file thing with formatting changes mixed in.

Windchild292:
Should have clarified what I meant for stable... the three NPEs are what I'd backport, not the financial date one (all three are together because they originate from the same campaign file and issue)

MekHQ/src/mekhq/campaign/finances/enums/FinancialYearDuration.java
Windchild292: When we hit the export, it will always be after we advance the day. Thus, the day passed in would be the new day, not the actual end of month.

So, how this works is you'll get the export as 2950-2959 when the provided date is 01-JAN-2960 and have set to DECENNIAL, instead of the previous export date of 2960... which makes no sense for finances exporting at the end of the decade. I agree it's confusing, but... it looks nice on the export.

Windchild292: BIANNUAL can mean either half a year or once every two years, whereas semiannual is always half a year. It is also specifically mentioned externally in the resource strings.

Further, this can only be hit when the previous is true... so you'll always know the value is the first day of the next year.

sixlettervariables: Biennial is every other year and should be used if that's what we mean. Biennial is unambiguous in that context. Semiannual and biannual are interchangeable.

sixlettervariables:
The individual commits may be candidates, but this PR is not. They'd have to be taken one at a time. That label is for a PR which can simply be rebased onto the stable branch.

MekHQ/src/mekhq/campaign/finances/enums/FinancialYearDuration.java
Windchild292: Swapped it over and added Unit Testing.

sixlettervariables (commented):
Quick question, otherwise looks good.

sixlettervariables:
Close/Reopen to kick off fresh CI.

MekHQ/unittests/mekhq/campaign/finances/enums/FinancialYearDurationTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertFalse(FinancialYearDuration.ANNUAL.isEndOfFinancialYear(LocalDate.of(3025, 12, 31)));
+
+        assertTrue(FinancialYearDuration.BIENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3026, 1)));
+        assertFalse(FinancialYearDuration.BIENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3025, 11)));
+
+        assertTrue(FinancialYearDuration.QUINQUENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3025, 1)));
+        assertFalse(FinancialYearDuration.QUINQUENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3026, 11)));
+
+        assertTrue(FinancialYearDuration.DECENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3030, 1)));
+        assertFalse(FinancialYearDuration.DECENNIAL.isEndOfFinancialYear(LocalDate.ofYearDay(3026, 11)));
```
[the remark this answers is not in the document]
Windchild292: Exactly.

sixlettervariables (approved):

PR #2484, opened 2021-03-27
Transfer Force info to MM, requires MM#2409 Lobby/GUI changes

SJuliez:
This requires the MegaMek lobby/force PR #2409!

Sends the force composition to MM when starting a scenario. Uses a String representation that is also used for loading/saving forces in MULs. Bot forces will receive a standard force based on the bot's name like "Rebels", see Screenshot (nvm the missing i18 entries, that's my Eclipse config).
Also replaces sending the entities individually with sending them as a block (which is required for the force setup to work in this way and, also, faster without the added thread sleep and if anything, safer, I guess).

image

I imagine that when forces need to transport more to MM than structure and name we'd send the forces as a MegaMek.Force object, but for now, this works and is simple.

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -281,6 +281,7 @@ public class AtBGameThread extends GameThread {
                     }
 
                     entity.setDeployRound(deploymentRound);
+                    entities.add(entity);
                 }
                 client.sendAddEntity(entities);
                 client.sendPlayerInfo();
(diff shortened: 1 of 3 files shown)


Windchild292 (changes requested):
Branch naming issues. This should use the same name as https://github.com/MegaMek/megamek/pull/2409 for the branch to properly handle CI. Please close and reopen with the same branch name.

Windchild292 (approved):

PR #2495, opened 2021-04-08
Base Components: JDisableablePanel

Windchild292:
This creates a new base component that allows one to enable or disable all components within a JPanel with a single selection. This does NOT include their child items, which will be handled as is regular for their setEnabled method.

### DIFF
--- a/MekHQ/src/mekhq/gui/baseComponents/JDisableablePanel.java
+++ b/MekHQ/src/mekhq/gui/baseComponents/JDisableablePanel.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2021 - The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MekHQ is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/baseComponents/JDisableablePanel.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+    //endregion Constructors
+
+    /**
+     * This override forces all child components to be the same value for enabled as this component,
+     * thus allowing one to easily enable/disable child components and panels.
+     * @param enabled whether to enable the child components or not
+     */
+    @Override
+    public void setEnabled(final boolean enabled) {
```
VicenteCartas, line 43: I feel that for something like this I would do some type of generic function that takes a java.awt.Container, and walks the list of getComponents() enabling/disabling them (and all their subcomponents). That would work for anything that inherits from that class, so you could disable other UI elements like Tabs or Trees.

It could be a recursive call, or just use a queue, and once you dequeue something to change their setEnable, you enqueue their sub-components.

Windchild292: I want this to only disable the components, not their subcomponents. If we want to disable the subcomponents, we just make that extend this panel (or another base component, but I cannot think of a good reason to have this being part of another base component currently)

VicenteCartas: I would probably then do something like setEnabled(component, state, recurse), instead of having JDisablePanel, JDisableTab... But at the same time, if you think only JPanel is going to need this, then that's it.

VicenteCartas (approved):

PR #2532, opened 2021-04-20
ProtoMech Bay Rating

Windchild292:
This fixes an issue raised on Discord where the bay rating for ProtoMechs is not taken into account. I also reorganized the checks so they are in the same order as they are in on the dialog in commit 2.

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
+++ b/MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
@@ -462,10 +462,10 @@ public abstract class AbstractUnitRating implements IUnitRating {
         setFighterBayCount(0);
         setSmallCraftBayCount(0);
         setProtoBayCount(0);
-        setBaBayCount(0);
-        setLightVeeBayCount(0);
-        setHeavyVeeBayCount(0);
         setSuperHeavyVeeBayCount(0);
+        setHeavyVeeBayCount(0);
+        setLightVeeBayCount(0);
+        setBaBayCount(0);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/rating/AbstractUnitRating.java
```diff
@@ -493,21 +492,24 @@ void updateBayCount(Entity e) {
...
-                } else if (bay instanceof HeavyVehicleBay) {
-                    setHeavyVeeBayCount(getHeavyVeeBayCount() + (int) bay.getCapacity());
-                } else if (bay instanceof SuperHeavyVehicleBay) {
-                    setSuperHeavyVeeBayCount(getSuperHeavyVeeBayCount() + (int) bay.getCapacity());
+                    setMechBayCount(getMechBayCount() + (int) bay.getCapacity());
                 } else if (bay instanceof ASFBay) {
                     setFighterBayCount(getFighterBayCount() + (int) bay.getCapacity());
                 } else if (bay instanceof SmallCraftBay) {
                     setSmallCraftBayCount(getSmallCraftBayCount() + (int) bay.getCapacity());
+                } else if (bay instanceof ProtomechBay) {
```
sixlettervariables, line 500: I made a quick check and it seems that none of these bays derive from each other, is that correct?

Windchild292: That is correct, I double checked before reordering.

sixlettervariables (approved):

Issue #2491, opened 2021-04-06
RFE: Allow Extra Time on Maintenance


ndilday:
Reading over _Strategic Operations_, it seems like the modifiers for taking extra time should also be allowed on maintenance checks. This could be accomlished by adding an additional menu options in the Hangar to modify the maintenance of a specific unit with extra time (and rush maintenance, I suppose).

NickAragua:
Looking at it in a little more detail, extra time is allowed for maintenance checks, but not 'rush'. 

This will make those maintenance rules a little less nightmarish to use, though especially for units just starting out with all green techs and such.

NickAragua:
Implemented with #2537

PR #2537, opened 2021-04-21
allow extra time for maintenance

NickAragua:
Adds an item to the unit popup menu allowing the user to set extra time (bulk select allowed, just ignores actions on invalid units) for maintenance. Also logic to store this multiplier and apply it to maintenance checks. Implements #2491.


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4711,74 +4717,87 @@ public boolean isPresent() {
     }
 
     public int getMaintenanceTime() {
+        int retVal = 0;
```
Windchild292, line 4720: Can we please call this time or maintenanceTime instead of retVal?

Windchild292 (commented):
Few comments first

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1835,6 +1837,8 @@ public void writeToXml(PrintWriter pw1, int indent) {
         if (astechDaysMaintained > 0) {
             MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "astechDaysMaintained", astechDaysMaintained);
         }
+        
+        MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "maintenanceMultiplier", maintenanceMultiplier);
```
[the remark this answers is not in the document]
NickAragua: This looks identical to the surrounding calls and makes no practical difference.

MekHQ/src/mekhq/campaign/unit/Unit.java
NickAragua: Really?

MekHQ/src/mekhq/campaign/unit/Unit.java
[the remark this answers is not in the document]
Windchild292: It will soon... the above will be legacy, the below will not.

MekHQ/src/mekhq/campaign/unit/Unit.java
Windchild292: Yes. I should be able to tell what a variable is from a glance. retVal tells me nothing as to what the value is, just that it is the one returned (and is a misnomer in this case, as it returns while being multiplied)

MekHQ/src/mekhq/campaign/unit/Unit.java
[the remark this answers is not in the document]
NickAragua: At which point, if you're planning to do some kind of mass reference update, you'll be able to use the same mechanism that you use to change the many other references.

MekHQ/src/mekhq/campaign/unit/Unit.java
NickAragua: Sounds like a personal preference. Pretty sure it's not going to make a difference in this method, since there are no other local variables anyway.

Windchild292: It's best practices for variable names in general, not a personal preference.

NickAragua: Best practices such as these are in the eye of the beholder, and you can easily argue this either way.

Windchild292: In that case... please tell me how retVal is better than time or maintenanceTime for a variable name.

NickAragua: "This is the value you will see coming back at the end of this method." has roughly same comprehension value as "This is the maintenance time." Therefore, it's not worth changing. If you want to name variables in code you write using a different naming convention, that's your call, but I can assure you that, unless it's actively misleading or unprofessional, nobody will give you a hard time about it.

Windchild292: It's a misnomer, as that's not the value you return.

As for naming conventions... I'd highly disagree. Making it understandable is key for future proofing. I'll agree to disagree though, to get this in.

Windchild292 (approved):

PR #2540, opened 2021-04-23
Factions Data Cleanup

Windchild292:
This PR cleans up a bunch of faction data, specifically doing the following:
- [x] Remove altfactions.xml
- [x] Swap Faction Hints majorPowers to use the factions.xml tags instead
- [x] Swap Faction Hints deepPeriphery to use the factions.xml tags instead
- [x] Make Playable a tag 
- [x] Standardize factions.xml data order
- [x] Remove duplicated factions.xml tags
- [x] Remove legacy faction ids


MekHQ/src/mekhq/campaign/market/ContractMarket.java
```diff
@@ -160,33 +160,27 @@ public void generateContractOffers(Campaign campaign) {
     public void generateContractOffers(Campaign campaign, boolean newCampaign) {
         if (((method == ContractMarketMethod.ATB_MONTHLY) && (campaign.getLocalDate().getDayOfMonth() == 1))
                 || newCampaign) {
-			Contract[] list = contracts.toArray(new Contract[contracts.size()]);
-			for (Contract c : list) {
-				removeContract(c);
-			}
+            Contract[] list = contracts.toArray(new Contract[0]);
```
NickAragua, line 163: Why are we sizing the array at 0?

MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -345,12 +345,7 @@ public static int getEffectiveNumUnits(Campaign campaign) {
 
     public static boolean isMinorPower(String fName) {
         Faction faction = Factions.getInstance().getFaction(fName);
-        if (null != faction) {
-            return !RandomFactionGenerator.getInstance().getFactionHints().isISMajorPower(faction) &&
-                    !faction.isClan();
-        } else {
-            return false;
-        }
+        return (faction != null) && !faction.isISMajorOrSuperPower() && !faction.isClan();
```
NickAragua, line 348: Honestly this seems like it belongs as a static method in the Faction class.

NickAragua (changes requested):
Couple of minor changes. Can't really speak to the accuracy of factions.xml.

MekHQ/src/mekhq/campaign/market/ContractMarket.java
Windchild292: IDE suggestion, because if the size isn't consistent it'll just create the array anyway. However, I'll change it to be a forEach implementation, which will be a single line of code.

MekHQ/src/mekhq/campaign/mission/AtBContract.java
Windchild292: This is specific to AtB Contracts, and is not equal to checking the minor power tag we have in the factions data, which is why I did not move it to factions. I think I'll move it to AtB Contract Market once I start that rework, but until then I'm going to leave it here.

NickAragua (approved):

PR #2545, opened 2021-04-28
add damaged heat sinks as missing parts when loading unit

NickAragua:
This fixes an issue I noticed where, if an aerospace fighter with 10 heat sinks had taken heat sink hits, you couldn't ever fix it without restoring the unit. Adding the appropriate number of missing heat sink parts to the unit seems to solve the issue.


sixlettervariables (approved):
Interesting find!

PR #2551, opened 2021-04-29
Objective effect text update

NickAragua:
This PR improves (somewhat) the human readability of the objectives section of the scenario resolution wizard, by specifying what each objective will accomplish when it succeeds/fails. Given the procedural nature of it, the language might be somewhat awkward but I tried to make it reasonable for the most common cases.

image

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioObjective.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioObjective.java
@@ -26,7 +26,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
 
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBElement;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/mission/ScenarioObjective.java
```diff
@@ -323,18 +323,70 @@ public String toShortString() {
...
+        StringBuilder result = new StringBuilder();
+        List<ObjectiveEffect> effectCollection = success ? getSuccessEffects() : getFailureEffects();
+        
+        if (!effectCollection.isEmpty()) {
+            result.append("<br/>");
+        }
+        
+        for (ObjectiveEffect effect : effectCollection) {
+            boolean scaledEffect = effect.effectScaling == EffectScalingType.Linear ||
+                    effect.effectScaling == EffectScalingType.Inverted;
```
Windchild292, line 352: Brackets, although implementing boolean comparisons (EffectScalingType::isLinear, EffectScalingType::Inverted) would be easier to read code. However, I'd say leave it as is for now and handle it in the future.

Windchild292 (approved):
One formatting comment. I'll leave adding these to resources and migrating the enums to be in their own classes with simplification boolean comparison methods for later mentions.

PR #2554, opened 2021-04-29
Objective update

NickAragua:
- Fixed an issue preventing reinforcement of scenarios with forces already on the strategic board.
- Fixed an issue causing "squished" rendering for the strategic board when loading saves in some states (by getting rid of the grid bag layout and using a more rigid grid layout).
- Bulletproofing


MekHQ/src/mekhq/gui/StratconPanel.java
```diff
@@ -117,7 +117,10 @@
      * Constructs a StratconPanel instance, given a parent campaign GUI and a pointer to an info area.
      */
     public StratconPanel(CampaignGUI gui, JLabel infoArea) {
-        campaign = gui.getCampaign();
+        if (gui != null) {
```
sixlettervariables, line 120: How does the rest of this survive without a `Campaign` instance? I see lots of other places where the code would die.

sixlettervariables (commented):
One question.

MekHQ/src/mekhq/gui/StratconPanel.java
NickAragua: To be honest, it's probably unnecessary. I wanted to re-test the StratconTab layout without having to reload the game, so I put in a call to InitTab in a couple of places, which re-initializes the StratconPanel - including once before the campaign GUI is loaded.

NickAragua: Reverted.

Windchild292 (approved):

PR #2581, opened 2021-05-04
Campaign Presets: Updating to modern version

Windchild292:
This updates the presets, partially so they are now nicely versioned.


NickAragua (approved):
I suppose it's fine as long as it doesn't prevent MHQ from loading.

Windchild292:
The presets were all modified within MHQ, so... no issue there. Just updates the internals based on the current setup, while adding set versions for future-proofing (the versions are already implemented and used internally too)

PR #2593, opened 2021-05-08
Stratcon - actually capture facilities

NickAragua:
This pull request fixes an issue I noticed where, when captured, facilities would pretend to be captured but still supply scenario modifiers associated with hostile versions of themselves.

We explicitly define what a facility becomes when captured to avoid messing around with trying to guess the "opposite" definition file or modifiers.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -1227,7 +1227,6 @@ public class StratconRulesManager {
      * modifier/type/alignment switches etc.
      */
     private static void switchFacilityOwner(StratconFacility facility) {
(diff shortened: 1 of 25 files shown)


MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -1225,6 +1221,36 @@ public static void processScenarioCompletion(ResolveScenarioTracker rst) {
             }
         }
     }
+    
+    /**
+     * Contains logic for what should happen when a facility gets captured:
+     * modifier/type/alignment switches etc.
+     */
+    private static void switchFacilityOwner(StratconFacility facility) {
+        if (facility.getCapturedDefinition() != null && !facility.getCapturedDefinition().isBlank()) {
```
Windchild292, line 1230: brackets

Windchild292 (approved):

PR #2619, opened 2021-05-18
fix multiple issues generating fixed objective scenarios

### LINKED ISSUE #2603: [49.1+390] Blank scenario created before Mission start
I  before finishing another one, and .
As I had just accepted a new Mission, an Assassinate  scenario immediately appeared in my Briefing Room, though it had not Date and no details (no OpFor, though it had condition sets).

The save is in the pre-start situation. But even when the mission start was reached and I was on-planet, it still was blank.

Note that this mission was accepted before the end of the previous one. ALso I did not reach the planet before mission start, in case it could be related.  

[Pending Scenario before mission start.cpnx.gz](https://github.com/MegaMek/mekhq/files/6453593/Pending.Scenario.before.mission.start.cpnx.gz)

NickAragua:
This fixes #2603 and a nightmarish tangle of issues discovered while testing that fix.

- Introduces a mechanism to specify whether a StratCon scenario should be displayed in the briefing tab.
- Introduces a mechanism to prevent a scenario generation report from being displayed in the campaign log.
- Fixes a logic error which would result in fixed objective scenarios being generated but then getting eliminated immediately upon contract start.
- Fixes an error preventing "add a force" modifiers from being applied to fixed objective scenarios due to missing mission ID
- Updated right click menu logic to prevent information leakage about unrevealed fixed objective scenarios

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -977,6 +977,13 @@ public class Campaign implements Serializable, ITechManager {
     //endregion Missions/Contracts
 
     /**
+     * Adds scenario to existing mission, generating a report.
+     */
+    public void addScenario(Scenario s, Mission m) {
+        addScenario(s, m, false);
+    }
+    
+    /**
      * Add scenario to an existing mission. This method will also assign the scenario an id, provided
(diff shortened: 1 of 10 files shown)


Windchild292 (approved):
Just one place for brackets

PR #2624, opened 2021-05-19
one more fix for the stratcon tab right click menu

NickAragua:
The intended logic is explained in the updated comments. Basically, I want the "assign force to this track" menu item to show up if there aren't any forces currently assigned to that hex and there aren't any visible scenarios on it. If there *is* a visible scenario on the hex, then we should always be getting the "manage scenario" menu item.

### DIFF
--- a/MekHQ/src/mekhq/gui/StratconPanel.java
+++ b/MekHQ/src/mekhq/gui/StratconPanel.java
@@ -151,9 +151,10 @@ public class StratconPanel extends JPanel implements ActionListener {
         
         StratconScenario scenario = getSelectedScenario();
         
-        // display "Manage Force Assignment" if
-        // there is not a force already on the hex
-        if (!currentTrack.getAssignedCoordForces().containsKey(coords)) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/StratconPanel.java
```diff
@@ -151,9 +151,10 @@ private void buildRightClickMenu(StratconCoords coords) {
         
         StratconScenario scenario = getSelectedScenario();
         
-        // display "Manage Force Assignment" if
-        // there is not a force already on the hex
-        if (!currentTrack.getAssignedCoordForces().containsKey(coords)) {
+        // display "Manage Force Assignment" if there is not a force already on the hex
+        // except if there is already a non-cloaked scenario here.
+        if (!currentTrack.getAssignedCoordForces().containsKey(coords) &&
+                (scenario == null || scenario.getBackingScenario().isCloaked())) {
```
Windchild292, line 157: brackets

Windchild292 (approved):

Issue #2633, opened 2021-05-24
[49.1] StratCon issues


Tecmes:
(these are for the 49.1 official release)

- [x] Maybe there's something I don't get, but after six days in the mission, and scouting two places, I had already "completed one objective". At this point I had  done zero scenarios.
![objective_complete](https://user-images.githubusercontent.com/80174634/119366661-328e7f00-bc7f-11eb-988e-e5529c36fbec.png)
The same, after a single scn. I just don't get what's going on. It was a convoy escort, so I didn't "locate a facility" and I don't see how I would have completed any objective there? Also the "1/0" is odd.
![objective_after_one_scn](https://user-images.githubusercontent.com/80174634/119409015-a8610d80-bcb4-11eb-8fa2-c391f6aa874a.png)

- [x] Don't know if that's normal: an auto-assignment assigned my whole force (i.e. "My campaign") for a Prevent Recon mission (with massive opfor), including the ASF lance. I should note I had a loose unit in the root level of my forces, so maybe that's a cause.
Yep. Going to leave it as is for now, will revisit if it becomes a consistent problem.
- [x] Prevent Recon: I started in deployement "Any". Normal?
Yes, that's one of the possible deployment scenarios.
- [x] Still having turrets trying to deploy in maps with no buildings (e.g. "AlliedTurrets", "Searchandrescue"), though I suspect there's no way right now to change the map appropriately from MHQ.
Let's stop bringing this up for now, it's not going to change until major changes happen on the MegaMek side.
- [x] Irregular Force scn often generates insane amounts of OpFor, both in terms of counts and BV.
This one's weird - it's "unit count scaled", with 2x infantry and 2x "civilian" units as your primary force. It may have picked your root level force as the "primary" force. I'll revisit it at some point.
- [x] Succeeding at a mission brings a single VP regardless of the totals put forward by the debrief pane. I guess this is intentional. In that case, it could be written at the top of the objective part of scns: "Complete some or all of the following objective to gain up to 1VP".
  Added some descriptive text to "required" scenarios in StratCon tab to indicate that they will provide/lose 1 campaign VP.
- [x] In the same vein, non-StratCon scn such as Ambush don't seem to affect VP though the debrief pane lists VP results. 
  This is intended - scenario VP isn't the same as campaign VP. Mostly I left those in so that people could still get the occasional Star League Cache mission for (what is likely) an intro-tech bug mech.
[...]

PR #2646, opened 2021-05-26
Small unit contract init fix

NickAragua:
Three changes:
- compilation fix
- convoy escort objective properly lists force
- ensure minimum objective count for small unit sizes (objective count rounds to 0 sometimes).

Addresses #2633


sixlettervariables (changes requested):
Failing unit test.

sixlettervariables (approved):

PR #2656, opened 2021-05-29
Cleaning up a few erroneous Campaign GUI properties

Windchild292:
The removed lines were unused, and the dialog tags were the only ones to load dialogs without the ...

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignGUI.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignGUI.properties
@@ -75,23 +75,19 @@ miCargoReport.text=Cargo Report...
 # View Menu
 menuView.text=View
 miShowHistoricalReportLog.text=Historical Daily Report Log
-miDetachLog.text=Detach Daily Report Log
-miAttachLog.text=Attach Daily Report Log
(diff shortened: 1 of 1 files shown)


MekHQ/resources/mekhq/resources/CampaignGUI.properties
```diff
@@ -75,23 +75,19 @@ miCargoReport.text=Cargo Report...
 # View Menu
 menuView.text=View
 miShowHistoricalReportLog.text=Historical Daily Report Log
-miDetachLog.text=Detach Daily Report Log
-miAttachLog.text=Attach Daily Report Log
```
NickAragua, line 79: We really ought to look at making the daily log detachable again. It's such a pain in the ass to have to keep going back to the command center tab for it.

NickAragua (approved):

MekHQ/resources/mekhq/resources/CampaignGUI.properties
Windchild292: I concur, and will open up an issue to open it in its own modeless window.

PR #2686, opened 2021-06-08
Prototype TSM mods

neoancient:
Implements the modifiers to maintenance, repair, and replacement for a mech with prototype TSM due to its unstable nature and tendency to burst into flames. See IO, p. 104.

1. All maintenance rolls for a mech equipped with P-TSM are made at +1.
2. All repair and replacement tasks take twice as long (I didn't apply this to reloading ammo bins, as this should not directly involve the myomer).
3. Any repair or replacement involving internal structure, location replacement, or actuators is made at +2.

Depends on MegaMek/megamek#2923

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1839,7 +1839,7 @@ public class Unit implements MekHqXmlSerializable, ITechnology {
         if (astechDaysMaintained > 0) {
             MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "astechDaysMaintained", astechDaysMaintained);
         }
-        
+
         MekHqXmlUtil.writeSimpleXmlTag(pw1, indent, "maintenanceMultiplier", maintenanceMultiplier);
 
         if (mothballTime > 0) {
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4595,6 +4595,20 @@ public void cancelMothballOrActivation() {
...
     }
 
+    /**
+     * Prototype TSM makes a unit harder to repair and maintain.
+     *
+     * @return Whether the unit has prototype TSM
+     */
+    public boolean hasPrototypeTSM() {
+        for (Mounted m : getEntity().getMisc()) {
+            if (m.getType().hasFlag(MiscType.F_TSM) && m.getType().hasFlag(MiscType.F_PROTOTYPE)) {
```
NickAragua, line 4605: One of these days, we'll need to come up with a mechanism for checking if a unit has some specific piece of equipment other than looping through all the damn equipment and looking for flags.

NickAragua (approved):

MekHQ/src/mekhq/campaign/unit/Unit.java
neoancient: For the record, I don't like the implementation here of increased repair time either, but I don't know a better way.

PR #2693, opened 2021-06-10
GM Tools Dialog Rewrite

Windchild292:
This PR is a complete rewrite of the GM Tools Dialog. It does the following:
- [x] Move to AbstractDialog
- [x] Divide into General and Names Tab
- [x] Modal only when editing a person
- [x] Expanding null protections
- [x] Expanding the options for rolling dice
- [x] RAT Unit Readout Display
- [x] Improve name generation display with large numbers of names


MekHQ/src/mekhq/gui/displayWrappers/ClanDisplay.java
```diff
@@ -0,0 +1,88 @@
...
+import mekhq.campaign.personnel.Clan;
+import mekhq.campaign.universe.Faction;
+
+import java.time.LocalDate;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * I only exist because of
```
NickAragua, line 32: I only exist because of....

MekHQ/src/mekhq/gui/displayWrappers/FactionDisplay.java
```diff
@@ -0,0 +1,84 @@
...
+import megamek.common.annotations.Nullable;
+import megamek.common.util.sorter.NaturalOrderComparator;
+import mekhq.campaign.universe.Faction;
+
+import java.time.LocalDate;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class FactionDisplay {
```
NickAragua, line 30: Let's put some documentation regarding what the point of this class is.

MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
```diff
@@ -42,887 +39,1256 @@
...
     private int bloodnameYear;
-    //endregion Bloodname Information
+    private Phenotype selectedPhenotype;
+    private String lastGeneratedBloodname;
+    //endregion Name Tab
+    //endregion GUI Variables
 
     //region Constants
     private static final String[] QUALITY_NAMES = {"F", "D", "C", "B", "A", "A*"};
     private static final String[] WEIGHT_NAMES = {"Light", "Medium", "Heavy", "Assault"};
```
NickAragua, line 126: Just in case we didn't have enough weight and quality names scattered around MegaMek/MekHQ, lol. No need to address since it's exsiting code.

MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
```diff
@@ -42,887 +39,1256 @@
...
      * Determine if a person's primary role supports operating a
      * given unit type.
      */
-    private boolean doesPersonPrimarilyDriveUnitType(int unitType) {
-        PersonnelRole primaryRole = getPerson().getPrimaryRole();
+    private boolean doesPersonPrimarilyDriveUnitType(final int unitType) {
         switch (unitType) {
             case UnitType.AERO:
-                return primaryRole.isAerospacePilot();
+                return getPerson().getPrimaryRole().isAerospacePilot();
```
NickAragua, line 1167: Not sure I get why we got rid of line 744 and added more code across the rest of the switch.

NickAragua (changes requested):
Couple of minor nitpicks, looks fine otherwise.

MekHQ/src/mekhq/gui/displayWrappers/ClanDisplay.java
Windchild292: ... I know I had a full line there at one point...

MekHQ/src/mekhq/gui/dialog/GMToolsDialog.java
Windchild292: Personal preference, if the line isn't too long then this allows me to tell the whole story in a single line.

NickAragua: Does the compiler know that?

Windchild292: There shouldn't be any difference in compilation. Actually, I believe this would execute extremely slightly faster because it doesn't need to create a temporary reference that then immediately needs to be garbage collected.

NickAragua: In this case, that's reasonable, since it's a switch and not a loop, so only one of these would get executed anyway.

NickAragua (approved):

PR #2698, opened 2021-06-11
adjust scenario template/modifier destination edges

NickAragua:
Since I changed the CardinalEdge enum, that caused all forces generated by StratCon to have the "nearest" destination edge and behave "weirdly" as a result. On the minus side, I had to go through all the templates and modifiers and modify the destination edges. On the plus side, I found a couple of inconsistencies along the way that I was able to fix. 

Destination edges, for reference: 5 = none, 6 = opposite deployment, 7 = random.

Bonus: boats modifier, which I somehow forgot to include in the initial release. Deploys water-naval units anywhere, if possible.


Windchild292 (approved):
Looks good, just checking that you meant to remove the Riot Suppression scenario template

NickAragua:
Yep, it got turned into "Irregular Forces" some time ago.

PR #2701, opened 2021-06-12
vtols off the ground; bombs expansion

NickAragua:
Three changes:

- generated VTOLs no longer start at elevation 0
- only put 1 TAG on bombers, hook it up with other types of bombs for the rest of the space instead
- wait a little longer to load transports; this seems to be a thread timing issue

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -318,7 +318,8 @@ public class AtBGameThread extends GameThread {
 
                     // we need to wait until the game has actually started to do transport loading
                     // This will load the bot's infantry into APCs
-                    Thread.sleep(MekHQ.getMekHQOptions().getStartGameDelay());
+                    // sometimes it needs more time?
+                    Thread.sleep(MekHQ.getMekHQOptions().getStartGameDelay() * 2);
                     if (scenario != null) {
                         AtBDynamicScenarioFactory.loadTransports(scenario, botClient);
                     }
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -25,8 +25,10 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 
@@ -45,7 +47,6 @@ import megamek.common.UnitType;
 import megamek.common.logging.LogLevel;
 import mekhq.campaign.force.Force;
 import mekhq.campaign.mission.AtBDynamicScenario;
-import mekhq.campaign.mission.AtBDynamicScenarioFactory;
 import mekhq.campaign.mission.AtBScenario;
 import mekhq.campaign.mission.BotForce;
 import mekhq.campaign.personnel.Person;
@@ -77,6 +78,8 @@ public class AtBGameThread extends GameThread {
     private static final String LOAD_FTR_DIALOG_TITLE = "Load Fighters on Transport?";
     private static final String LOAD_GND_DIALOG_TEXT = "Would you like the ground units assigned to %s to deploy loaded into its bays?";
     private static final String LOAD_GND_DIALOG_TITLE = "Load Ground Units on Transport?";
+    
+    public static final int CLIENT_RETRY_COUNT = 1000;
 
 
     @Override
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -318,7 +318,8 @@ public void run() {
 
                     // we need to wait until the game has actually started to do transport loading
                     // This will load the bot's infantry into APCs
-                    Thread.sleep(MekHQ.getMekHQOptions().getStartGameDelay());
+                    // sometimes it needs more time?
+                    Thread.sleep(MekHQ.getMekHQOptions().getStartGameDelay() * 2);
```
Windchild292, line 322: This makes no sense... the option exists so you can set the time based on your setup, and if it is having communication troubles you should just increase the option's value

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2311,6 +2314,15 @@ private static void setStartingAltitude(List<Entity> entityList, int startingAlt
                     ((IAero) entity).land();
                 }
             }
+            
+            // hack - set helis and WIGEs to an explicit altitude of 1
+            // currently there is no support for setting elevation for "ground" units 
+            // in the scenario template editor, but it looks dumb to have choppers
+            // start out on the ground
+            if (entity.getMovementMode() == EntityMovementMode.VTOL ||
+                    entity.getMovementMode() == EntityMovementMode.WIGE) {
```
Windchild292, line 2323: brackets

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2366,9 +2378,21 @@ private static void loadBombs(Entity entity, int[] validBombChoices, int year) {
...
+            weightModifier = 5;
+            bombChoices[bombIndex] = 1;
+            actualValidBombChoices.remove(randomBombChoiceIndex);
+            bombIndex = Utilities.getRandomItem(actualValidBombChoices);
+        }
+        
         // # of bombs is the unit's weight / (bomb cost * 5)
-        int numBombs = (int) (entity.getWeight() / (BombType.getBombCost(bombIndex) * 5));
+        int numBombs = (int) ((entity.getWeight() - weightModifier) / 
+                (BombType.getBombCost(bombIndex) * 5));
```
Windchild292, line 2395: Should it be 5.0, to ensure it isn't int / int division? Also, rounding?

Windchild292 (changes requested):
Bombs look good, the elevation hack is okay but there should be an issue opened to allow that to be set, but the thread sleep increase just goes entirely against the entire reason the option exists.

MekHQ/src/mekhq/AtBGameThread.java
NickAragua: Maybe I'm being excessively cynical, but I imagine most people don't touch the default setting for that, and have no idea what it means. Thus, the sequence of events I'm picturing is: 1. Person sees units not loading into transport. 2. Person goes onto github/discord/slack to complain about it. 3. Somebody has to waste their time explaining about the delay setting, maybe successfully maybe not. I'd like to skip parts 1-3 of this process.

Another option is to get rid of this, but set the default delay to 2x or even 3-4x higher. I'm currently leaning towards that.

A third option is to go back to sending entities over one at a time, which will most likely guarantee correct behavior, but will be considerably slower in the general case.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
NickAragua: All the numbers involved in the operation are integers. Rounding down (the default behavior) is correct because otherwise it would result in putting more bombs in than the unit has capacity.

MekHQ/src/mekhq/AtBGameThread.java
Windchild292: You probably aren't being cynical but instead realistic... I'd go with 4x the default as a base if this is as much of an issue as it sounds.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
Windchild292: To prevent misreads and IDE / CI issues, can I suggest this be written to use the double and a specified round?

MekHQ/src/mekhq/AtBGameThread.java
NickAragua: Narrowed down the actual cause of the problem, so now we basically wait for all the bot's units to load before attempting to load transports.

NickAragua:
Fixed a couple of other issues observed while testing:
- LAMS were deploying in air mech mode (that's no good, bot doesn't know how to drive them)
- Transport VTOLs were failing to load infantry because of the elevation check (irrelevant in lobby)
- Move loadTransports method to more appropriate location given when it's called.

Windchild292 (approved):

PR #2702, opened 2021-06-18
prevent double deployment of forces and units

NickAragua:
Prevent situations where a force or a unit may theoretically be selected more than once on the StratCon scenario management UI. This would basically cause a freeze when starting a scenario.

Now, if a force has any of its sub-units or sub-elements deployed, it cannot be deployed itself.


MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -951,16 +951,42 @@ public static boolean forceCompositionMatchesDeclaredUnitType(int primaryUnitTyp
...
         return retVal;
     }
+    
+    /**
+     * Returns true if any sub-element (unit or sub-force) of this force is deployed.
+     */
+    private static boolean subElementsDeployed(Force force, Campaign campaign) {
+        if (force.isDeployed()) {
+            return true;
+        }
```
sixlettervariables, line 972: Is this really `subElementsOrSelfDeployed` ?

sixlettervariables (approved):
A naming thing, but don't consider it blocking.

Issue #2717, opened 2021-06-23
49.2 Locks up when cancelling out of the 


mouldsgame:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
0.49.2
Windows
13.0.2
(Delete the above lines)

### Description
I was attempting to advance days after I accepted a contract.   When I did I recieved the warning that I didn't have enough Astechs.  I cancelled out of the advance days dialogue and once I did anywhere I clicked on the MekHQ screen only beeped and didn't register my click.   If I didn't know better I would think there was a dialogue box open behind the window.  (please let me know if that doesn't make sense)

### Files
[megameklog.txt](https://github.com/MegaMek/mekhq/files/6702336/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/6702342/mekhqlog.txt)
[Wraith Division30670101.cpnx.gz](https://github.com/MegaMek/mekhq/files/6702344/Wraith.Division30670101.cpnx.gz)

[Wraith Division CUstomers.zip](https://github.com/MegaMek/mekhq/files/6702354/Wraith.Division.CUstomers.zip)

HammerGS:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at mekhq.gui.adapter.UnitTableMouseAdapter.actionPerformed(UnitTableMouseAdapter.java:390)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI.doClick(BasicMenuItemUI.java:1020)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(BasicMenuItemUI.java:1064)
```

and 

```
Exception in thread "AWT-EventQueue-0" java.util.concurrent.CancellationException
[...]

PR #2718, opened 2021-06-23
2717: Fixing Bulk Refit Complete/Cancel

Windchild292:
This fixes the bulk refit code in the unit table mouse adapter, as noticed in #2717, and converts them into streams.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
@@ -88,6 +88,7 @@ import java.util.Optional;
 import java.util.ResourceBundle;
 import java.util.UUID;
 import java.util.Vector;
+import java.util.stream.Stream;
 
 public class UnitTableMouseAdapter extends JPopupMenuAdapter {
     //region Variable Declarations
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -379,17 +379,9 @@ public void actionPerformed(ActionEvent action) {
             ((MekLabTab) gui.getTab(GuiTabType.MEKLAB)).loadUnit(selectedUnit);
             gui.getTabMain().setSelectedIndex(GuiTabType.MEKLAB.getDefaultPos());
         } else if (command.equals(COMMAND_CANCEL_CUSTOMIZE)) {
-            for (Unit unit : units) {
-                if (unit.isRefitting()) {
-                    selectedUnit.getRefit().cancel();
-                }
-            }
+            Stream.of(units).filter(Unit::isRefitting).forEach(unit -> unit.getRefit().cancel());
```
sixlettervariables, line 382: Is this fix essentially that `Stream.of` skips null lists?

Windchild292: The problem is that selected unit is used instead of the unit... so it NPEs on the selectedUnit.getRefit() call after the first unit.

sixlettervariables: Hah, missed that.

sixlettervariables (approved):

PR #2719, opened 2021-06-23
2717: Advance Days Dialog no longer based on modality

### LINKED ISSUE #2717: 49.2 Locks up when cancelling out of the

Windchild292:
This fixes #2717 by removing the modality changes from the advance days dialog and instead tracking using a single variable.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/AdvanceDaysDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/AdvanceDaysDialog.java
@@ -41,6 +41,7 @@ import java.util.List;
 public class AdvanceDaysDialog extends AbstractMHQDialog {
     //region Variable Declarations
     private final CampaignGUI gui;
+    private boolean running;
 
     private JSpinner spnDays;
     private JButton btnStartAdvancement;
@@ -64,6 +65,14 @@ public class AdvanceDaysDialog extends AbstractMHQDialog {
     //region Getters/Setters
     public CampaignGUI getGUI() {
         return gui;
+    }
+
+    public boolean isRunning() {
+        return running;
+    }
+
+    public void setRunning(final boolean running) {
+        this.running = running;
     }
 
     public JSpinner getSpnDays() {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/AdvanceDaysDialog.java
```diff
@@ -294,7 +302,7 @@ public void startAdvancement(final ActionEvent evt) {
             getGUI().getCampaign().fetchAndClearNewReports();
         }
 
-        setModal(false);
+        setRunning(false);
```
sixlettervariables, line 305: Let's put that in a finally block in case we throw in the loop.

Windchild292: Better idea: put the internals in a try-catch with logging, then break. It would keep the current functionality and properly handle the remaining days.

sixlettervariables: Works for me.

PR #2720, opened 2021-06-23
MM 2199: MUL Parser Game Option-reliant Personnel Load

Windchild292:
This is the MHQ-side of https://github.com/MegaMek/megamek/pull/2964, but also standardizes the internals and removes a few unused and deprecated methods.


SJuliez (approved):
OK.

PR #2725, opened 2021-06-25
add tools to manage SP/VP; track intensity visibility

NickAragua:
Adds some tools for users to manage SP/VP, including GM mode tools; shows track scenario odds in GM mode.

image


MekHQ/src/mekhq/gui/StratconTab.java
```diff
@@ -56,6 +60,8 @@
     private JScrollPane expandedObjectivePanel;
     private boolean objectivesCollapsed = true;
 
+    CampaignManagementDialog cmd;
```
Windchild292, line 63: I can see no good reason to have this always initialized and not just recreate it as required.

MekHQ/src/mekhq/gui/StratconTab.java
```diff
@@ -115,6 +123,13 @@ private void initializeInfoPanel() {
 
         infoPanel.add(new JLabel("Current Campaign Status:"));
         infoPanel.add(campaignStatusText);
+        
+        JButton btnManageCampaignState = new JButton();
+        btnManageCampaignState.setHorizontalAlignment(SwingConstants.LEFT);
+        btnManageCampaignState.setVerticalAlignment(SwingConstants.TOP);
+        btnManageCampaignState.setText("Manage SP/VP");
```
Windchild292, line 130:
```suggestion
        JButton btnManageCampaignState = new JButton("Manage SP/VP");
        btnManageCampaignState.setHorizontalAlignment(SwingConstants.LEFT);
        btnManageCampaignState.setVerticalAlignment(SwingConstants.TOP);
```

MekHQ/src/mekhq/gui/StratconTab.java
```diff
@@ -394,6 +409,13 @@ private void repopulateTrackList() {
             stratconPanel.setVisible(false);
         }
     }
+    
+    private void showCampaignStateManagement(ActionEvent e) {
+        TrackDropdownItem selectedTrack = (TrackDropdownItem) cboCurrentTrack.getSelectedItem();
+        cmd.display(selectedTrack.contract.getStratconCampaignState(), selectedTrack.track, getCampaign().isGM());
+        cmd.setModalityType(ModalityType.APPLICATION_MODAL);
+        cmd.setVisible(true);
+    }
```
Windchild292, line 418: This is short enough that it can just be made inline, and the modality doesn't make any sense to me outside of needing to consider the current GM state.

MekHQ/src/mekhq/gui/stratcon/CampaignManagementDialog.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public class CampaignManagementDialog extends JDialog {
+    private StratconCampaignState currentCampaignState;
+    private StratconTab parent;
+    JButton btnConvertVPToSP;
+    JButton btnConvertSPtoBonusPart;
+    JButton btnGMAddVP;
+    JButton btnGMAddSP;
+    JLabel lblTrackScenarioOdds;
+
```
Windchild292, line 40: private...

Windchild292 (commented):
I'm not the biggest fan of how this is implemented, but think it in general looks okay for an alpha dialog subject to future change. Just a few comments first.

MekHQ/src/mekhq/gui/StratconTab.java
NickAragua: I expect there will be more stuff that needs to be GMed as time goes on, and I don't like being able to click somewhere else and have the dialog disappear on me.

MekHQ/src/mekhq/gui/StratconTab.java
NickAragua: On the other hand, I don't see a reason to keep creating/destroying a bunch of buttons and whatnot.

Windchild292: That's how we handle it in MekHQ, and... permanent memory vs clearable memory regularly matters when using a weaker computer.

Windchild292 (approved):
Looks good for an initial setup, albeit with the concerns voiced previously.

PR #2739, opened 2021-07-10
2736: Fixing Alternative Faction Code Generation by using correct ArrayList type

### LINKED ISSUE #2736: [49.2] Conflict alternativeFactionCodes (factions.xml) with use of traditional RATs
### Environment
MekHQ 49.2 (error occurs not with MekHQ 48.0)
Windows 10
Java SE Development Kit 16.0.1 (64-bit)
Java 8 Update 291 (64-bit)

### Description
Units for factions with the new implemented alternativeFactionCodes in factions.xml cannot generate units if "Use traditional Rats" is activated under Campaign Options> Against the Bot.

An example RAT for MOC and TH for the year 2580 is attached. The RAT for MOC works because there is no alternativeFactionCodes. This RAT does not work with TH because an alternativeFactionCodes is defined.

The error can be repeated for all corresponding factions in all RATs. The attached error code is output in mekhqlog.txt. 

### Files
[MekHQ 49-2.zip](https://github.com/MegaMek/mekhq/files/6776323/MekHQ.49-2.zip)

with logs, savegame, rats

Windchild292:
This fixes #2736

### DIFF
--- a/MekHQ/src/mekhq/campaign/universe/RATManager.java
+++ b/MekHQ/src/mekhq/campaign/universe/RATManager.java
@@ -333,7 +333,7 @@ public class RATManager extends AbstractUnitGenerator {
 
         final Faction faction = Factions.getInstance().getFaction(factionCode);
         if (faction.getAlternativeFactionCodes() != null) {
-            final List<String> alternativeFactionCodes = Arrays.asList(faction.getAlternativeFactionCodes());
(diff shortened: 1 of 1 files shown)


sixlettervariables (approved):
Arrays.asList just provides a List view into the array rather than an ArrayList.

Windchild292:
image

In the backend it is called an ArrayList (Arrays.java code for the image), hence my confusion @sixlettervariables.

PR #2731, opened 2021-06-30
251: Force Generator Unit Generation Empty Unit Table Parent Faction Fallback

Windchild292:
This fixes #251 and a common source of bug reports by adding a fallback system for force generator unit generation, with the goal of preventing empty OpFors. To do this I reserved two of the generation methods for OpFor generation and removed the two non-OpFor generation method uses.
To review: The second commit is the generation change, while the other three are a mixture of code cleanup and standardization.  (I was originally considering doing a full campaign options setup, but decided the bugfix should be merged in the nearer future.)

PR #2746, opened 2021-07-15
Fixing Unit Market Generation

Windchild292:
This fixes the generation order for the unit market after I misordered some values in https://github.com/MegaMek/mekhq/pull/2731


sixlettervariables (approved):
Makes a good case to ditch bare integers for enums when feasible.

Windchild292:
Hence my enuming work @sixlettervariables

sixlettervariables:
Just some encouragement :)

Issue #2748, opened 2021-07-16
No warning for pending scenarios on advance day


SuperStucco:
In the attached 49.2 campaign, using the advance day with a pending day does not give a warning.  The MekHQ settings do not disable the nag dialog, so it should be nagging.

[Autosave-1-Jansenn's Horde-30420212.cpnx.gz](https://github.com/MegaMek/mekhq/files/6831525/Autosave-1-Jansenn.s.Horde-30420212.cpnx.gz)

[Customs.zip](https://github.com/MegaMek/mekhq/files/6831592/Customs.zip)

Windchild292:
I'm going to need the following to properly check, as no log was included.

![image](https://user-images.githubusercontent.com/39067288/126919324-c87ead9a-706c-44ae-9bab-a7c8fba623db.png)

SuperStucco:
This should cover it.

[Customs.zip](https://github.com/MegaMek/mekhq/files/6904404/Customs.zip)

Unrelated but interesting, the Gazelle (Heavy Vehicle) is stock OOTB from ...\dropships.ZIP\TRO3057R.

Windchild292:
Duplicates #2709, fixed by #2711

Issue #2747, opened 2021-07-15
[BUG] StratOps fails to generate scenarios - Exception in thread "AWT-EventQueue-0"


Jamini:
### Environment
MekHQ-0.49.2
Ubuntu 20.04 (KDE Plasma 5.18.5)
Java version 11.0.11
Platform Linux 5.8.0-59-generic (amd64)

### Description
StratOps fails to generate scenarios
Logs attached.

### Files
Save
[Dumbcampaign.gz.cpnx.zip](https://github.com/MegaMek/megamek/files/6826193/Dumbcampaign.gz.cpnx.zip)

Logs here
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/6826191/mekhqlog.txt)

NickAragua:
Check your data\scenariotemplates folder. Does it have files called Allied Facility.xml and Hostile Facility.xml?

sixlettervariables:
Three distinct MekHQ NPEs:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at mekhq.campaign.stratcon.StratconScenarioFactory.getSpecificScenario(StratconScenarioFactory.java:124)
	at mekhq.campaign.stratcon.StratconContractInitializer.initializeObjectiveScenarios(StratconContractInitializer.java:294)
	at mekhq.campaign.stratcon.StratconContractInitializer.initializeCampaignState(StratconContractInitializer.java:101)
	at mekhq.campaign.mission.AtBContract.acceptContract(AtBContract.java:1196)
	at mekhq.gui.dialog.ContractMarketDialog.acceptContract(ContractMarketDialog.java:450)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
...
	at java.desktop/java.awt.Dialog.setVisible(Dialog.java:1005)
	at mekhq.gui.CampaignGUI.showContractMarket(CampaignGUI.java:1314)
	at mekhq.gui.CampaignGUI.lambda$initMenu$15(CampaignGUI.java:851)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
...
```
And:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at mekhq.campaign.stratcon.StratconScenarioFactory.getSpecificScenario(StratconScenarioFactory.java:124)
[...]

Issue #2709, opened 2021-06-20
MekHQ month advancement issue


theyoungster21:
### Environment
MekHQ unstable 0.49.2 also encountered on 0.49.1
Windows 10 home 64-bit
Java version 16.0.1

### Description
When the advance day button is clicked to advance a month during a contract, the game fails to advance to the 1st of the next month and only adds entries to the bottom of the daily activity log for the various markets updating. No other new month activity seems to occur (i.e. no money paid out, no exp, no events). Date will stay on the say before the end of the month, then after the advance day button is clicked a second time, it will change to the second day of the month.
![Example date failure](https://user-images.githubusercontent.com/86205253/122686147-c4859b00-d1dd-11eb-99e9-2523aae07fde.png)

Can be easily replicated by just advancing the day, no need to interact with the scheduled scenario. Whether it is completed or not the bug occurs.

### Files

[megameklog.txt](https://github.com/MegaMek/mekhq/files/6682981/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/6682982/mekhqlog.txt)
[Garett's Fusiliers in progress.zip](https://github.com/MegaMek/mekhq/files/6682996/Garett.s.Fusiliers.in.progress.zip)

sixlettervariables:
Two exceptions of note:
```
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at megamek.common.event.EventListener.trigger(EventListener.java:33)
	at megamek.common.event.EventBus.internalTrigger(EventBus.java:158)
	at megamek.common.event.EventBus.trigger(EventBus.java:147)
	at mekhq.MekHQ.triggerEvent(MekHQ.java:680)
	at mekhq.campaign.Campaign.newDay(Campaign.java:3406)
	at mekhq.campaign.CampaignController.advanceDay(CampaignController.java:82)
	at mekhq.gui.CampaignGUI.lambda$initTopButtons$42(CampaignGUI.java:1180)
...
[...]

PR #2750, opened 2021-07-17
fix NPE in outstanding scenario nag dialog

NickAragua:
For some reason, null scenarios exist in this list and probably cause #2748

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/OutstandingScenariosNagDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/OutstandingScenariosNagDialog.java
@@ -41,7 +41,7 @@ public class OutstandingScenariosNagDialog extends AbstractMHQNagDialog {
         return !MekHQ.getMekHQOptions().getNagDialogIgnore(getKey())
                 && campaign.getActiveAtBContracts(true).stream()
                         .anyMatch(contract -> contract.getCurrentAtBScenarios().stream()
-                                .anyMatch(scenario -> (scenario.getDate() != null)
+                                .anyMatch(scenario -> (scenario != null) && (scenario.getDate() != null)
                                         && scenario.getDate().isEqual(campaign.getLocalDate())));
     }
 }


MekHQ/src/mekhq/gui/dialog/OutstandingScenariosNagDialog.java
```diff
@@ -41,7 +41,7 @@ protected boolean checkNag(final Campaign campaign) {
         return !MekHQ.getMekHQOptions().getNagDialogIgnore(getKey())
                 && campaign.getActiveAtBContracts(true).stream()
                         .anyMatch(contract -> contract.getCurrentAtBScenarios().stream()
-                                .anyMatch(scenario -> (scenario.getDate() != null)
+                                .anyMatch(scenario -> (scenario != null) && (scenario.getDate() != null)
```
Windchild292, line 44: I'm not seeing how this is possible, given it should error out in Mission::getCurrentAtBScenarios during the filter if we have a null scenario in the list.

Namely, we process it with `.filter(scenario -> scenario.getStatus().isCurrent() && (scenario instanceof AtBScenario))`, which would NPE out on any null inclusions...

NickAragua: If you look at the logs in #2747, that's where the exception happens and that's the only way it's possible for a NPE to happen there.

sixlettervariables: The updates to that function to make it non-null occurred ~3 months ago:
https://github.com/MegaMek/mekhq/blame/v0.49.2/MekHQ/src/mekhq/campaign/mission/Mission.java#L164-L168

That fix is in 49.2 per git tags, so unless they were using a snapshot, I'm with Windchild in that I'm not quite sure how we got here.

Windchild292: Just opened #2748, and... the campaign file cannot be loaded because it's missing multiple campaign files. Further, there is no log included in that issue.

Windchild292: I'm pretty sure the error is a duplicate of #2709 and thus fixed by #2711

Windchild292:
Bug report duplicated #2709, the issue is fixed in #2711.
Closing this because of that.

PR #2751, opened 2021-07-18
Astech Divide By Zero Errors

Windchild292:
This prevents a few divide by zero issues raised on the Discord while improving the code for Campaign::getTargetFor and Campaign::getAvailableAstechs and adding a constant for Astech team size.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -4778,35 +4778,25 @@ public class Campaign implements Serializable, ITechManager {
         MekHQ.triggerEvent(new PersonChangedEvent(p));
     }
 
-    public TargetRoll getTargetFor(IPartWork partWork, Person tech) {
-        Skill skill = tech.getSkillForWorkingOn(partWork);
+    public TargetRoll getTargetFor(final IPartWork partWork, final Person tech) {
+        final Skill skill = tech.getSkillForWorkingOn(partWork);
         int modePenalty = partWork.getMode().expReduction;
-        if (null != partWork.getUnit() && !partWork.getUnit().isAvailable(partWork instanceof Refit)) {
-            return new TargetRoll(TargetRoll.IMPOSSIBLE,
-                    "This unit is not currently available!");
-        }
-        if ((partWork.getTech() != null) && !partWork.getTech().equals(tech)) {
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4847,37 +4842,38 @@ public TargetRoll getTargetFor(IPartWork partWork, Person tech) {
             target.addModifier(getFaction().getEraMod(getGameYear()), "era");
         }
 
-        boolean isOvertime = false;
+        final boolean isOvertime;
```
NickAragua, line 4845: What was wrong with the way it was before? Initializing a variable to its default state is a perfectly valid coding pattern.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4847,37 +4842,38 @@ public TargetRoll getTargetFor(IPartWork partWork, Person tech) {
...
-        if (null != partWork.getUnit() && partWork.getUnit().isSelfCrewed()) {
-            int hits;
-            if (null != partWork.getUnit().getEntity().getCrew()) {
-                hits = partWork.getUnit().getEntity().getCrew().getHits();
-            } else {
-                hits = 6;
-            }
-            helpMod = getShorthandedModForCrews(hits);
+        if ((partWork.getUnit() != null) && partWork.getUnit().isSelfCrewed()) {
+            helpMod = getShorthandedModForCrews((partWork.getUnit().getEntity().getCrew() == null)
```
NickAragua, line 4862: This looks familiar, didn't I see the same logic in the other pull request I just reviewed?

Maybe a better solution would be to unify that logic somewhere.

NickAragua (changes requested):

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: You did, and I've got a change that standardizes the logic in that PR.

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: This is a way to handle it that means you know the value is set here and never changes. I find it easier when debugging, as I don't have to trace down any other uses.

NickAragua (approved):

PR #2752, opened 2021-07-18
Fixing Campaign::getTargetForMaintenance IDE Potential NPE False Positive

Windchild292:
This fixes an IDE warning for a potential NPE (false positive, partWork.getUnit() is previously null checked, but better to clean up) and simplifies the code around the if statement.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -4860,13 +4860,7 @@ public class Campaign implements Serializable, ITechManager {
         }
         int helpMod;
         if (null != partWork.getUnit() && partWork.getUnit().isSelfCrewed()) {
-            int hits;
-            if (null != partWork.getUnit().getEntity().getCrew()) {
-                hits = partWork.getUnit().getEntity().getCrew().getHits();
-            } else {
-                hits = 6;
-            }
-            helpMod = getShorthandedModForCrews(hits);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4927,19 +4927,14 @@ public TargetRoll getTargetForMaintenance(IPartWork partWork, Person tech) {
...
-                int hits = 0;
-                if (null != partWork.getUnit().getEntity().getCrew()) {
-                    hits = partWork.getUnit().getEntity().getCrew().getHits();
-                } else {
-                    hits = 6;
-                }
-                helpMod = getShorthandedModForCrews(hits);
+            final int helpMod;
+            if (partWork.getUnit().isSelfCrewed()) {
+                helpMod = getShorthandedModForCrews((partWork.getUnit().getEntity().getCrew() == null)
```
NickAragua, line 4932: I have an extremely hard time parsing this expression. Are you passing the results of a ternary operator into a function and then assigning the resulting value to helpMod?

This really needs to be put back to the way it was or have a comment explaining exactly what it does.

NickAragua (changes requested):

MekHQ/src/mekhq/campaign/Campaign.java
Windchild292: Moved the logic into Campaign::getShorthandedModForCrews, as it was being duplicated and just makes more sense based on the name to be calculated there.

NickAragua (approved):

PR #2760, opened 2021-07-25
Rank fixes and improvements from HB: Major Periphery States

Windchild292:
This fixes a bunch of rank data issues based on HB:MPS.


NickAragua (commented):
I don't see anything wrong with this one although I have no idea what it actually does. Can you explain it in a little more detail?

Windchild292:
We had some ranks not properly lined up as per the sourcebook (which appears to be the most modern one for the topic in question).

The other fixes are adding in missing redirects so that lower tier ranks can be properly assigned as is expected

NickAragua (approved):

PR #2769, opened 2021-07-27
Fixed maps - MekHQ

NickAragua:
Bottom line - sometimes, at a user-configurable rate, StratCon will use an appropriately sized (and possibly terrain-typed) map from the data/boards directory instead of a generated map.

MekHQ component of the fixed maps feature, requires companion MegaMek PR.


MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -861,6 +864,32 @@ public static void setScenarioMapSize(AtBDynamicScenario scenario) {
...
+                MegaMekFile mapFile = new MegaMekFile(mapPath);
+                BoardDimensions dimensions = Board.getSize(mapFile.getFile());
+                
+                scenario.setMap(bc.getBoardPaths().get(mapPath));
+                scenario.setMapSizeX(dimensions.width());
+                scenario.setMapSizeY(dimensions.height());
+                scenario.setUsingFixedMap(true);
+                return;
+            }
+        }
```
Windchild292, line 888: We should really check if we should attempt (the compute check) before determining if there are potential maps for performance reasons.

Windchild292 (approved):

Windchild292 (changes requested):
Hit the wrong button, meant to do this earlier

Windchild292 (approved):

PR #2778, opened 2021-07-28
Making Money Serializable as it already has a read/write to XML setup

Windchild292:
This prevents regular IDE warnings for serialization differences.


VicenteCartas (approved):
Looks great!

PR #2781, opened 2021-07-30
Finances: Transaction Refactoring/Standardization

Windchild292:
Finances file two refactor: Transaction


NickAragua (approved):
This doesn't seem to do anything other than shuffle parameter order around, but fine.

PR #2808, opened 2021-08-14
2807: Removing EquipmentParts whose EquipmentTypes fail to parse

### LINKED ISSUE #2807: [0.47.1 -> 0.48.0 or 0.49.2] Import of MekHQ Campaign file fails
I am currently using a campagin file in version 0.47.1. I would like to upgrade but the campagin file does not load in version 0.48.0 or 0.49.2 of MekHQ. A dialogue appears that the issue is logged. However I cannot easily find out what I could fix. 

### Environment
Import from version 0.47.1 -> 0.48.0 or 0.49.2]
OS: Windows
JRE: 8 (Update 301)
Java version 15

### Files
Logs
[megameklog.txt](https://github.com/MegaMek/mekhq/files/6986634/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/6986635/mekhqlog.txt)
Campaign File
[Hunting Hellhounds30541122withoutLTtrailer.zip](https://github.com/MegaMek/mekhq/files/6986636/Hunting.Hellhounds30541122withoutLTtrailer.zip)
Custom Units
[customs.zip](https://github.com/MegaMek/mekhq/files/6986638/customs.zip)

Windchild292:
This (I believe... it works, but I'm not sure about unexpected behaviour) fixed #2807.

### DIFF
--- a/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
+++ b/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
@@ -1370,9 +1370,16 @@ public class CampaignXmlParser {
 
             Unit u = prt.getUnit();
             if (u != null) {
-                // get rid of any equipmentparts without locations or mounteds
+                // get rid of any equipmentparts without types, locations or mounteds
                 if (prt instanceof EquipmentPart) {
-                    EquipmentPart ePart = (EquipmentPart) prt;
+                    final EquipmentPart ePart = (EquipmentPart) prt;
+
+                    // Null Type... parsing failure
+                    if (ePart.getType() == null) {
+                        removeParts.add(prt);
+                        continue;
+                    }
+
                     Mounted m = u.getEntity().getEquipment(ePart.getEquipmentNum());
 
                     // Remove equipment parts missing mounts
@@ -1381,7 +1388,8 @@ public class CampaignXmlParser {
                         continue;
                     }
 
-                    // Remove equipment parts without a valid location, unless they're an ammo bin as they may not have a location
+                    // Remove equipment parts without a valid location, unless they're an ammo bin
+                    // as they may not have a location
                     if ((m.getLocation() == Entity.LOC_NONE) && !(prt instanceof AmmoBin)) {
                         removeParts.add(prt);
                         continue;


MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1370,9 +1370,16 @@ private static void postProcessParts(Campaign retVal, Version version) {
...
             Unit u = prt.getUnit();
             if (u != null) {
-                // get rid of any equipmentparts without locations or mounteds
+                // get rid of any equipmentparts without types, locations or mounteds
                 if (prt instanceof EquipmentPart) {
-                    EquipmentPart ePart = (EquipmentPart) prt;
+                    final EquipmentPart ePart = (EquipmentPart) prt;
+
+                    // Null Type... parsing failure
+                    if (ePart.getType() == null) {
```
NickAragua, line 1378: Let's log something about this if it happens.

NickAragua (changes requested):

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
Windchild292: I looked at it when doing this. There's not enough data to make a useful log... and the useful data is previously logged.

sixlettervariables: May need to move the check/log earlier in the call stack to catch the part type that was bad.

Windchild292: We already log it when we fail to restore the equipment type. Just found it can be null at this point and... it's better to remove than deal with it down the line

Windchild292: Is that good @NickAragua? Would like this done for 0.49.3 if possible

NickAragua (approved):

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
NickAragua: Might as well. Not like that piece of equipment is recoverable anyway.

PR #2812, opened 2021-08-18
Switching Person to be fully pass-by-reference with regards to Campaign

Windchild292:
This has been a long-term goal for me, and this finishes it by removing the last instances where Person's Campaign link is used.


SJuliez:
I have no obvious mistakes in the code but looking at it without having seen it before some of the calls look strange. If a person's salary can't be done by person.getSalary() (without using a person's own campaign copy) then it feels like the right call would be campaign.getSalary(person). I understand this would be a bigger change than simply passing in the campaign.

Windchild292:
Well... the goal is to remove the campaign object from a person, so they no longer have a cyclical reference to the campaign but instead rely on passed in campaign files. It'll especially make things easier when someone eventually gets to handling multiplayer campaigns.

Campaign is MekHQ's central file, and we pass it into most things, sometimes saved and sometimes not. However, we also want to avoid it growing too large and us ending up with another Server on our hand, although we're just around 6k-7k lines currently.

For salary, it's calculated for the individual person based on their various values (role and rank primarily) modified by the campaign's selected options, so person is the best fit for the logic. However, one of the second wave of personnel modules will be the salary module, which will isolate the code into it's own module and expand on the current setup.

SJuliez (approved):
All good.

PR #2827, opened 2021-08-21
2825: Individual Camouflage Bulk Assignment from Same Original Camouflage

### LINKED ISSUE #2825: RFE [0.48] Change Individual Camo for a selection for Hangar Objects
### Environment
0.48
Windows 10
OpenJDK 11

### Description
In order to add individual camos to a group of mechs each one has to be done individually.  Adding a grouping function that lets you apply camo for some or all of them would be less time consuming. 
When changing forces in megamek to other players, only the individual camo is respected otherwise the camos changes.

Windchild292:
This fixes #2825


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -419,17 +419,19 @@ public void actionPerformed(ActionEvent action) {
...
+            final CamoChooserDialog ccd = new CamoChooserDialog(gui.getFrame(),
                     selectedUnit.getUtilizedCamouflage(gui.getCampaign()), true);
             if (ccd.showDialog().isCancelled()) {
                 return;
             }
-            selectedUnit.getEntity().setCamouflage(ccd.getSelectedItem());
-            MekHQ.triggerEvent(new UnitChangedEvent(selectedUnit));
+            for (final Unit unit : units) {
+                unit.getEntity().setCamouflage(ccd.getSelectedItem());
+                MekHQ.triggerEvent(new UnitChangedEvent(unit));
```
NickAragua, line 433: So here's a thought: why are we triggering the unit changed event here but not for the remove camo command? Also, why are we triggering it at all?

NickAragua (approved):
One thought unrelated to the content of the PR itself, but looks fine otherwise.

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
Windchild292: Nice catch... I'll double check if it is required, and either add or remove if it is. IIRC it is supposed to force the icon to update

Windchild292: It's a bug with the remove camo command. I've updated it for the future.

PR #2831, opened 2021-08-22
2766: Preventing an exception whereby the processing entity has a -1 external id

### LINKED ISSUE #2766: 48.0 MekHQ Battle cannot resolve in HQ
So I have a HQ campaign Battle which does not completely resolve. I get the HQ window that asks the "Did your forces control the battlefield?" and when I say yes. The game creates the entitystatus and the salvage muls, but HQ does not bring up the summary windows to fully resolve the end of the battle. The megameklog.txt does not seem to update after that. Here is the last megameklog.txt file. The Game that I am trying to Resolve is Game 269 in the list of games. Thanks WIndchild!

### Environment
What version of MekHQ does your issue apply to?

48.0

What operating system are you using?  

Windows 10

What java version are you using?

using java 8  301 but plan to upgrade to java 11 soon

### Files
Provide the .cpnx campaign file that this issue occurs in.

[Master_Warrior_123430520428_Mission355_2nd_Batt_Whitworth_Post2.cpnx.gz](https://github.com/MegaMek/mekhq/files/6879907/Master_Warrior_123430520428_Mission355_2nd_Batt_Whitworth_Post2.cpnx.gz)

CUSTOM FILES
[MASTERS_CUSTOM_units.zip](https://github.com/MegaMek/mekhq/files/6879955/MASTERS_CUSTOM_units.zip)

Here is the gamesave.
[Motomara_Forrest_T27.sav.gz](https://github.com/MegaMek/mekhq/files/6879976/Motomara_Forrest_T27.sav.gz)

Windchild292:
This fixes #2766. I have no idea how this occurred, nor can I recreate it. However, it is clear what needed to change to prevent it from the logs.


MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -183,6 +183,10 @@ public void processGame() {
             if (!e.getSubEntities().isEmpty()) {
                 // Sub-entities have their own entry in the VictoryEvent data
                 continue;
+            } else if ("-1".equals(e.getExternalIdAsString())) {
```
NickAragua, line 186: I'd almost be tempted to say that if it doesn't parse to a UUID, then we reject it.

NickAragua: Not sure if -1 parses to a UUID, though.

Windchild292: It doesn't, hence the check if it is the external id and writing it as an error before continuing. The exception was trying to turn -1 into a UUID.

NickAragua: Right, in that case I'd suggest that we be a little more proactive about it and reject the incoming entity if the external ID doesn't parse to a UUID.

Alternately, maybe we want to generate a new UUID for it in that case. Otherwise, what'll happen is that entities added mid-game won't show up in the scenario resolution dialog at all, and I don't know whether that's worse or better than a "new" entity showing up.

NickAragua: Currently, I'm thinking it's some kind of deserialization scenario, because deserialization sometimes doesn't seem to call a constructor.

Windchild292: MegaMek should be generating a UUID for all external ids... I have no clue how it can possibly be -1 given that is the case, but it is.

That's my best guess too, but I can only take it based on the logs.

Windchild292: This just prevents it from being processed (hence the continue). I thought it better to just skip processing given our internal system has no idea what the unit is instead of processing a ghost unit

rjhancock (approved):

Windchild292:
Reverted in #2766.

PR #2855, opened 2021-08-31
New Campaign Project: Wave 1: Campaign Preset Rework

Windchild292:
This is a complete rework of Campaign Presets, the new name for MekHQ Game Presets. They have been heavily expanded in functionality, and the GUIs are new. I've also added the backend to handle preset management and editing, but I've not done the frontend design that would allow it.

Requires MegaMek/megamek#3127 and fixes MegaMek#2654.


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -641,10 +646,23 @@ private void initMenu() {
...
 
-        JMenuItem miExportOptions = new JMenuItem(resourceMap.getString("miExportOptions.text"));
-        miExportOptions.setMnemonic(KeyEvent.VK_C);
-        miExportOptions.addActionListener(this::miExportOptionsActionPerformed);
-        miExportXMLFile.add(miExportOptions);
+        final JMenuItem miExportCampaignPreset = new JMenuItem(resourceMap.getString("miExportCampaignPreset.text"));
+        miExportCampaignPreset.setName("miExportCampaignPreset");
+        miExportCampaignPreset.setMnemonic(KeyEvent.VK_C);
+        miExportCampaignPreset.addActionListener(evt -> {
+            final CreateCampaignPresetDialog createCampaignPresetDialog
```
NickAragua, line 653: Having a hard time figuring out how this inline method is more readable than a separate method, but don't care enough to make an issue out of it.

MekHQ/src/mekhq/campaign/CampaignPreset.java
```diff
@@ -0,0 +1,496 @@
...
+
+    public CampaignPreset(final Campaign campaign) {
+        this(campaign.getName(), "", true, campaign.getLocalDate(), campaign.getFaction(),
+                campaign.getCurrentSystem().getPrimaryPlanet(), campaign.getRankSystem(), 2,
+                campaign.getGameOptions(), campaign.getCampaignOptions(),
+                campaign.getRandomSkillPreferences(), SkillType.getSkillHash(),
+                SpecialAbility.getAllSpecialAbilities());
+    }
+
+    public CampaignPreset(final String title, final String description, final boolean userData,
```
NickAragua, line 106: Nothing says "readable" like a constructor with 13 parameters, but it's either this or set all the properties individually which has its own set of problems.

MekHQ/src/mekhq/campaign/CampaignPreset.java
```diff
@@ -0,0 +1,496 @@
...
+            MekHQ.getLogger().error(e);
+            final ResourceBundle resources = ResourceBundle.getBundle("mekhq.resources.Campaign", new EncodeControl());
+            JOptionPane.showMessageDialog(frame, resources.getString("CampaignPresetSaveFailure.text"),
+                    resources.getString("CampaignPresetSaveFailure.title"), JOptionPane.ERROR_MESSAGE);
+        }
+    }
+
+    public void writeToXML(final PrintWriter pw, int indent) {
+        pw.println("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
+        MegaMekXmlUtil.writeSimpleXMLOpenTag(pw, indent++, "campaignPreset", "version", MekHQOptions.VERSION);
```
NickAragua, line 301: I'm still kind of wondering why we're using this home-brewed serialization mechanism when you can use jaxb or some other third-party library that has already solved the problem. Some benefits include:
- we don't have to maintain a hand-rolled XML serialization library
- if we add a new property to the class, we don't have to update multiple different parsing/serialization routines

MekHQ/src/mekhq/gui/panes/RankSystemsPane.java
```diff
@@ -174,7 +174,6 @@ protected void initialize() {
                 BorderFactory.createTitledBorder(resources.getString("txtInstructionsRanks.title")),
                 BorderFactory.createEmptyBorder(5, 5, 5, 5)));
         txtInstructionsRanks.setOpaque(false);
-        txtInstructionsRanks.setMinimumSize(new Dimension(250, 120));
```
NickAragua, line 177: Seems a little random but ok.

NickAragua (approved):
Looks fine, other than some philosophical musing that I don't particularly care about either way.

MekHQ/src/mekhq/campaign/CampaignPreset.java
Windchild292: Some negatives include increased difficulty in debugging and especially in migration, and how new fields are handled when added.

Also, we should never need to update multiple parsing routings for a single element, just the one.

PR #2856, opened 2021-08-31
Random Origin Options Rework

Windchild292:
This moves the faction and planet selectors into their own folders, isolates their options into RandomOriginOptions, adds the ability to specify a central planet and whether to generate Clan-origin personnel for non-Clan factions, runs the generation based on the RandomOriginOptions, updates the backend to use WeightedDoubleMap and other modernization changes since being added, and moves them to being planet-based for their current location.

The new options and overall setup part of and required for Wave 2 of the New Campaign Project.


SJuliez (approved):
No complaints.

PR #2875, opened 2021-09-13
Proposed fix for #2873 - saving the part new flag correctly

### LINKED ISSUE #2873: Unit Values Change after Save and Load for GM Added Units
### Environment
MekHQ 0.49.3
Java vendor Eclipse Foundation
Java version 11.0.12
Platform Windows 10 10.0 (amd64)
Total memory available to MegaMek: 1,048,576 kB

### Description
When adding a unit through the 'Add (GM)' or GM Tools RAT generator, the unit value in the finance tab, hangar, and TO&E is approximately half the value of the list price of the mech.  Then doing a 'Save' and then 'Load' of the campaign file, the value changes after the load.  The value displayed after the load is approximately the unit list price but is off by about 1%.  The initial value displayed in the 'Purchase Unit' dialog appears to match the Master Unit List value but is slightly different than what is shown in the Hangar and Finance Tab.   The issue in this report is the fact that the value displayed is different after the save and load.

1. Create a new campaign
2. Click GM Mode
3. Marketplace->Purchace Unit
4. Search for Warhammer and select Warhammer 
5. Click 'Add (GM)' to add the unit.  The unit is added to the hangar.
6. Observe that the price in the hangar and in the finance tab is half of the list value.
7. Save the campaign file.
8. Load the campaign file.
9. The unit price is double the price prior to the save in the finance tab and hangar.
10. The value shown after the reload is off by ~1%?
11. If you add another unit after the save, the unit value is wrong until saved and reloaded.

### Versions
This issue has been reproduced in:
1. 48.0 Stable
2. 49.3 Dev
3. 49.4 Local

AddUnit
MasterUnitList
HangarTab
FinanceTab
FinanceTabPostSave
HangarPostSave

### Files
[logs.zip](https://github.com/MegaMek/mekhq/files/7148401/logs.zip)
[Test Merc - SaveGlitch30670101.cpnx.gz](https://github.com/MegaMek/mekhq/files/7148403/Test.Merc.-.SaveGlitch30670101.cpnx.gz)

HoneySkull:
When saving the parts list for GM added units, the tag is only saved
when the 'brandNew' property is TRUE and leave the tag out if the
property is FALSE.  Because the default constructor sets the property to
TRUE, false values re never saved and loaded.

Fixes #2873

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Part.java
+++ b/MekHQ/src/mekhq/campaign/parts/Part.java
@@ -681,10 +681,12 @@ public abstract class Part implements Serializable, MekHqXmlSerializable, IPartW
                 .append("</daysToArrival>")
                 .append(NL);
         }
-        if (brandNew) {
+        if (!brandNew) {
+        	//The default value for Part.brandNew is true.  Only store the tag if the value is false.
+        	//The lack of tag in the save file will ALWAYS result in TRUE.
             builder.append(level1)
                 .append("<brandNew>")
-                .append(true)
+                .append(false)
                 .append("</brandNew>")
                 .append(NL);
         }


Windchild292 (approved):
Thanks for the fix, and opened an issue to swap this to the standardized I/O setup

PR #2884, opened 2021-09-25
Briefing Room: Print Sheets now prints sheets for all units in AtB

Windchild292:
This has been raised on Discord a few times recently, so I've added it. I also cleaned up the code for the method and fixed any IDE warnings (protected makes no sense in a final class and unnecessary toString usages being the primary two).

To review: I put a comment over the actual functionality change, which was supposed to be in its own commit.

### DIFF
--- a/MekHQ/src/mekhq/gui/BriefingTab.java
+++ b/MekHQ/src/mekhq/gui/BriefingTab.java
@@ -25,12 +25,8 @@ import java.awt.GridBagLayout;
 import java.awt.GridLayout;
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
-import java.util.ResourceBundle;
-import java.util.UUID;
-import java.util.Vector;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/BriefingTab.java
```diff
@@ -517,51 +513,60 @@ protected void resolveScenario() {
...
         }
 
-        if (chosen.size() > 0) {
+        if (scenario instanceof AtBScenario) {
+            // Also print off allied sheets and all bot force sheets
+            chosen.addAll(((AtBScenario) scenario).getAlliesPlayer());
+            chosen.addAll(((AtBScenario) scenario).getBotForces().stream()
+                    .flatMap(botForce -> botForce.getEntityList().stream())
+                    .collect(Collectors.toList()));
+        }
```
Windchild292, line 562: This is the new functionality, where we also print off Player controlled allies and any bot force units

NickAragua (approved):

PR #2892, opened 2021-10-01
2877: Part XML File I/O using MekHqXmlUtil

### LINKED ISSUE #2877: Part I/O using MegaMekXMLUtil
Raised by reviewing #2875

Windchild292:
This fixes #2877


MekHQ/src/mekhq/campaign/work/WorkTime.java
```diff
@@ -102,38 +104,58 @@ private WorkTime(int id, String name, int mod, boolean isRushed, int expReductio
...
+
+        if (workTime != null) {
+            return workTime;
+        }
+
+        MekHQ.getLogger().error("Unable to parse " + text + " into a WorkTime. Returning NORMAL.");
+
+        return NORMAL;
+    }
+    //endregion File I/O
```
SJuliez, line 160: It feels to me like this would benefit from putting the whole code that follows the catch block into the catch block. Also the logger call and return normal should be in an else block.

SJuliez (changes requested):
Got a small formatting request but otherwise all good.

MekHQ/src/mekhq/campaign/work/WorkTime.java
Windchild292: First, you should always avoid putting further processing logic into a catch.

Second, this was just a quick improvement for how this is stored for a part, and... not the WorkTime cleanup/rework that will occur as part of the Enum All the Things project. This method follows the pattern used by all enums in that project for File I/O, however it looks weird because I used the legacy WorkTime.of setup instead of doing the full conversion.

SJuliez: Why (First)? Please provide me with a reasoning for such an assertive statement or cite a book.
Second: Yeah I know and we can let things be, but these are 7 lines of code, rearranged in 3 minutes. Shorter than arguing about them. I would have suggested a change but was on my tablet and typing that sentence already drove me half mad.

Windchild292: The first is because of how it was taught by my professors, and was drilled into us at that point. The logic is that exception handlers should be focused solely on the exception itself, as anything else distracts from the core cause. As the fallback is not handling the exception but is rather another (specifically legacy) parsing method, it's unrelated.

I will say that I'm pretty sure how I'm using exceptions as part of the flow control isn't a good way to handle code in general, but legacy migration is an exception to that.

For the second, it's not 7 lines of code, it's several thousand lines of code spread across the suite that follow this exact pattern. When working on enums I use a template, and copy/paste the sections I'm working on from that. This ensures every enum is extremely similar, but... any change that's based on the template's flow is not a minor change.

SJuliez:
Sorry had forgotten that this was waiting for me.

Windchild292:
No worries

PR #2851, opened 2021-08-29
Personnel Modules: Marriage: Modularization and Expanded Options

Windchild292:
This PR modularizes marriage, renames trying to marry to marriagable, makes it process its date separately from the campaign, and fixes #2895. Its partner is https://github.com/MegaMek/mekhq/pull/2493, which modularizes Procreation.

To Review: This was coded based on and heavily following the same format as #2493, and I'd recommend reviewing them both at the same time because of this similarity.

PR #2493, opened 2021-04-07
Procreation: Modularization, Expanded Options, and Bulk Manual Assignment/Removal

Windchild292:
This does:
- [x] Twin Probability Customization
- [x] Founder Tag Automatic Assignment
- [x] Bulk Manual Assignment and Removal
- [x] Date tracking separate from Campaign
- [x] Separate Random Procreation from Campaign
- [x] Modularize Randomization Methods
- [x] Optional Manual Procreation
- [x] Move Procreation out of Person
- [x] Renames Trying to Conceive to Conceivable
- [x] Clanner Marriage Enable/Disable Option 
- [x] Fixes #2896: Prisoner Marriage Enable/Disable Option

The twin probability customization was asked for in a comment to #2394, while the Founder tag options were suggested on Discord. The rest are things I wanted to add as part of standardizing the personnel modules of Procreation, Marriage, and Death.

PR #2908, opened 2021-10-18
Personnel Modules: Divorce: Modularization, Random Divorce, Expanded Options, and Bulk Manual Divorce

Windchild292:
This follows #2493 and #2851 by creating the Divorce Personnel Module, which uses the same design to ensure the same level of functionality and customization as the two previous modules. This module includes bulk manual divorce, weighted divorce surnames, and the percentage random divorce setup with the divorceable override flag.

Currently widowing is considered a divorce. I will be changing that as part of the Death module, but that is how it was coded originally and thus it works for now.


MekHQ/src/mekhq/campaign/personnel/divorce/AbstractDivorce.java
```diff
@@ -0,0 +1,270 @@
...
+    }
+    //endregion Getters/Setters
+
+    /**
+     * This is used to determine if a person can divorce.
+     * @param person the person to determine for
+     * @param randomDivorce if this is for random divorce or manual divorce
+     * @return null if they can, otherwise the reason they cannot
+     */
+    public @Nullable String canDivorce(final Person person, final boolean randomDivorce) {
```
SJuliez, line 129: Same as with Marriage I do think this would benefit from being a boolean and an additional method delivering a cause.

SJuliez (approved):
Only one thing that I'd prefer changed.

PR #2938, opened 2021-11-04
MHQ: Initial Localization Setup, Proper Date Localization

Windchild292:
This is a mixed bugfix and enhancement. The bug is missing localization with dates, the enhancement is to implement the backend required for date localization.

This will interface with the planned suite options setup.

To review:
This is a ton of files, but the changes are isolated to adding modifiers and the localization calls. Two locales are supplied, one for dates and one for the overall system.

Also, the plan is for this to not be guaranteed to apply without a restart, so this format will work as desired.


Windchild292:
This is now ready for review.

SJuliez (approved):
Fwiw, no complaints. This has been waiting long enough.

PR #2961, opened 2021-11-22
LocalDate: Removing Local Specification on Date Getters/Setters

Windchild292:
I forgot to do this when finishing the LocalDate changes, so this handles the renames as we now have a single standardized date setup.

For Review: This PR is solely variable/method renames in campaign, and has no other changes.


Windchild292:
Closing as this is useless without an immediate or rapid merge

PR #2964, opened 2021-11-23
actually replace VTOL rotor instead of eating the part

### LINKED ISSUE #2866: HQ: 49.3 Replacing a rotor on VTOL uses part but area is still destroyed.
[MMHQlog.zip](https://github.com/MegaMek/mekhq/files/7112537/MMHQlog.zip)

File attached with Logs save game customs folder and 2 screenshots.  

I had a VTOL that had the rotor blown off.  I ordered a new rotor and needed a 5 and rolled a 7.  However the section is still listed as X X for the IS and Armor.  The job under the repair tab will not allow me to replace the armor as it gives the message that the location is destroyed.  The part appears to be used up and the job tab under the repair bay for replacing the rotor is no longer there.  

I attempted to advance to next day after saving the game but that did not cause any change to the VTOL.  

GM editing the unit to put the rotor to full IS seems to solve the issue so this is not a major issue but something to look at when time permits

NickAragua:
Replacing a destroyed rotor needs to set the internal structure properly instead of just swallowing the replacement part and not doing anything. Fixes #2866

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MissingRotor.java
+++ b/MekHQ/src/mekhq/campaign/parts/MissingRotor.java
@@ -87,6 +87,16 @@ public class MissingRotor extends MissingPart {
 		// TODO Auto-generated method stub
 		return 0;
 	}
+	
+	@Override
+    public void fix() {
+	    if (null != unit && unit.getEntity() instanceof VTOL) {
+	        int maxIsVal = unit.getEntity().getOInternal(VTOL.LOC_ROTOR);
+	        unit.getEntity().setInternal(maxIsVal, VTOL.LOC_ROTOR);
+	    }
+	    
+        super.fix();
+    }
 
 	@Override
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/MissingRotor.java
```diff
@@ -87,6 +87,16 @@ public double getTonnage() {
 		// TODO Auto-generated method stub
 		return 0;
 	}
+	
+	@Override
+    public void fix() {
+	    if (null != unit && unit.getEntity() instanceof VTOL) {
```
Windchild292, line 93:
```suggestion
	    if ((null != unit) && (unit.getEntity() instanceof VTOL)) {
```

Windchild292 (commented):
Just missing brackets, otherwise looks good

PR #2967, opened 2021-11-23
Migrating to MavenCentral from JCenter

Windchild292:
JCenter is shutting down in February.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ allprojects {
         maven {
             url 'https://github.com/MegaMek/mavenrepo/raw/master'
         }
-        jcenter()
+        mavenCentral()
         maven {
             url 'https://jitpack.io'
         }


Windchild292:
I have no clue as to why this is failing when the MegaMekLab CI/CD is successful.

I'll investigate more when I'm less tired, but... it's just odd.

NickAragua (approved):

HoneySkull:
All my gradle commands fail unless I change this line back to jcenter().  Not sure why... but I get the same error from my local gradlew command locally as the nightly build on github.

NickAragua:
I wonder if it's because there's a dependency on the megamek/mml repos in MekHQ. Are you having the same problem building just plain megamek?

HoneySkull:
No - the MegaMek repo gradle commands are fine.  I'm mentioning it because it's complaining about the same line in the build.gradle file as I see in the nightly error logs.  I think it's the same error.

Windchild292:
MML builds fine both locally and online, MHQ build fine locally for me, but this apparently breaks in MHQ-integrated MML builds. I'm wondering if this is the same delayed response we saw with the version changes, but this time it breaks things rather than hides missing internal code changes.

I do have a fix idea, but can't test until tomorrow

PR #2976, opened 2021-11-27
Adding the ability to have CI/CD running on multiple Java versions

Windchild292:
It also fixes the naming of the CI/CD packages


Windchild292:
Replacing with general PR to handle all GitHub Action scripts

PR #2987, opened 2021-12-02
2292: Adding Maximum AtB Contract Generation Attempts Option

### LINKED ISSUE #2292: Maximum AtB Contract generation attempts should be a campaign option
Maximum AtB Contract generation attempts should be a campaign option.

_Originally posted by @Windchild292 in https://github.com/MegaMek/mekhq/pull/2291#discussion_r544445805_

Windchild292:
This fixes #2292.


NickAragua:
What possible value could this have to a user? There's basically no situation in which a contract generation will fail on one try but succeed on the next.

Windchild292:
I've experienced multiple failures for valid reasons (mostly relating to being unable to find a valid planet) since my original ask for this option, so it definitely isn't a no situation thing. I added it because I want it as an option, one that I will be immediately increasing to 10 retries.

I'd prefer to add a config file like EU4's defines.lua files, but this works until I have the time to create that setup.

NickAragua:
Looking at the code, it's theoretically possible that the subset of systems being searched changes. However, it's still as likely that this will fail on the 10th retry as it does on the 2nd or 3rd - sometimes, you're just not going to be able to generate a contract location no matter what.

A better solution would be to keep a list of factions for which contract location generation has failed and exclude those from the "random faction picker" logic. When you run out of factions, just go to "MERC" or "PIRATE". That way, you're guaranteed to generate a contract.

SJuliez:
I feel that with the merge conflicts, the time gone and the general disagreement if this is useful at all, this isn't going to make it in at any point. Therefore closing it.

PR #2992, opened 2021-12-04
Personnel Modules: Retirement / Dependent Stopgaps

Windchild292:
This expands on retirement and dependent randomization to add a few new options, and moves them to the personnel options tab. This is a functionality stopgap until I start working on the two actual modules.

- [x] Isolates Retirement and Dependent Campaign Options
- [x] Allows Random Retirement to be disabled
- [x] Allow Random Dependent Generation to be disabled

Note:
The setups are still isolated to AtB, and cannot be used outside of it.

For reviewers:
I made some changes in the retirement methods as I took a look into allowing it for all contracts... but it would be too much work to do nicely.


MekHQ/src/mekhq/campaign/personnel/enums/RandomDependentMethod.java
```diff
@@ -0,0 +1,76 @@
...
+        return this == NONE;
+    }
+
+    public boolean isAtB() {
+        return this == AGAINST_THE_BOT;
+    }
+    //endregion Boolean Comparisons
+
+/*
+    public AbstractDependent getMethod(final CampaignOptions options) {
```
SJuliez, line 61: Not used but want to keep it?

SJuliez (approved):
No complaints. There's commented out code in two classes where it isn't immediately clear why.

MekHQ/src/mekhq/campaign/personnel/enums/RandomDependentMethod.java
Windchild292: When I actually build these into full modules, this will be uncommented.

It means the enums are in the exact same style as the Marriage/Divorce/Procreation ones already merged, so it'll be faster in the future.

PR #3002, opened 2021-12-12
Log4j 2 Swapover

Windchild292:
This swaps MegaMekLab to Log4j 2 and by doing so improves our logging implementation in MegaMekLab.

The MekHQ::showInfo method has been standardized with MegaMek and MegaMekLab, with both called to set up the individual logs.

All three are used in MekHQ, with logs being split between MegaMek.log, MegaMekLab.log, and MekHQ.log depending on their origins.

CI is expected to fail due to certain resources issues. If it continues after merge I'll need to investigate it again.

### DIFF
--- a/MekHQ/build.gradle
+++ b/MekHQ/build.gradle
@@ -47,6 +47,7 @@ dependencies {
     implementation 'org.joda:joda-money:1.0.1'
     implementation 'org.apache.commons:commons-text:1.4'
     implementation 'org.apache.commons:commons-csv:1.4'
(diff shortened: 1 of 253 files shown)


SJuliez (approved):
OK!

SJuliez:
I hope it isn't a bad sign that all checks failed :)

Windchild292:
The only one that has a chance of passing is the CI/CD, and... it failed on uploading the release so it's not a true failure.

PR #3005, opened 2021-12-12
Adapt to MML package/class rename/move

SJuliez:
Adapts MHQ to MML's #978 rename & restructuring. I have unfortunately no working combined build so I have no way of testing this before the MML PR is merged, so there's certainly an error in here somewhere. My suggestion is merging MML's PR, then I'll correct any errors in here.

### DIFF
--- a/MekHQ/src/mekhq/gui/MekLabTab.java
+++ b/MekHQ/src/mekhq/gui/MekLabTab.java
@@ -27,7 +27,7 @@ import megamek.common.verifier.*;
 import megameklab.com.MMLConstants;
 import megameklab.com.MegaMekLab;
 import megameklab.com.ui.EntitySource;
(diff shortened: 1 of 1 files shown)


SJuliez:
OK now it should be correct. In order to get this to build with MML's current master the only solution that I found to work was going to the C:\users\me\.gradle\... directory and replacing the MML jar there with a hand-compiled from the local MML project and keeping the exact name of course. Without doing that building MHQ will always download a jar, call it 0.49-6-snapshot, but this will not be the current master (maybe its actually the 0.49-5 release). Therefore: when this is merged I don't know how to get MHQ to build correctly without hand-editing the MML jar until somehow MHQ can be convinced to dl the correct version or actually compile a local version.

Windchild292 (approved):

PR #3007, opened 2021-12-15
Skill Generation: Phenotype Bonus Applies to Proper Skills

Windchild292:
Taharqa found this issue, and raised it on Discord.

I also moved the non-clanner return to the top, as it should be an immediate return case.


MekHQ/src/mekhq/campaign/personnel/generator/DefaultSkillGenerator.java
```diff
@@ -55,7 +55,8 @@ public void generateSkills(Person person, int expLvl) {
             generateDefaultSkills(person, secondaryRole, expLvl, bonus, mod);
         }
 
-        bonus = getPhenotypeBonus(person);
+        // apply phenotype bonus only to primary skills
+        bonus = 0;
```
SJuliez, line 60: This feels like it would profit from getting rid of the bonus variable altogether and using the method and 0 directly as appropriate.

SJuliez (changes requested):
I do have a request for this.

MekHQ/src/mekhq/campaign/personnel/generator/DefaultSkillGenerator.java
Windchild292: It's a non-trivial method, so storing in a variable makes sense given it is used in two separate methods previously. I also think using the variable bonus is more readable than a bunch of inline magic numbers when it isn't related to graphical code.

SJuliez (approved):

PR #2946, opened 2021-11-07
AbstractIcon: Force Icon Rework

Windchild292:
This PR moves Force Icons to Abstract Icon, and divides them into three separate types. These are StandardForceIcon, LayeredForceIcon, and UnitIcon, all of which have differing functionality. Additionally, this adds LayeredForceIcon export to png, force icon TO&E icon copy/paste, and automated new day force icon operational status updates (based on a MekHQ option).

This PR also replaces the Force Icon images with the far more comprehensive and standardized Kailan pack, with full migration from the old setup and a new MekHQ extras repository containing the current pack and a pack of standard force icons created by Mjolnir06.

The full list of what this does, which I used to keep this organized:
- [x] Move Force and Layered Icons to Abstract Icon
- [x] Make the Unit Icon its own thing
- [x] Force Icon Data Swapover to Kailan Pack
- [x] Standard Force Icon migration to Layered Pieces
- [x] Layered Force Icon migration to new Layered Pieces
- [x] Unit Icon migration
- [x] New Standard Force Icon Dialog
- [x] New Layered Force Icon Dialog
- [x] New Unit Icon Dialog
- [x] GM Tools Force Icon Creation
- [x] Layered Force Icon Export
- [x] Create MekHQ Data Repository
- [x] Fixes #347: TO&E Force Icon Copy/Paste
- [x] Fixes #354
- [x] Fixes #2235
- [x] Fixes #2448
- [x] Fixes the flattened AbstractDirectory Setup by reworking it to no longer be flattened (required for the changes to work)
- [x] Removes the broken AbstractDirectory Zip Setup (required for the changes to work)
- [x] Swaps AbstractIconChooser to AbstractPanel
- [x] Adds Automated Force Icon Operational Status on New Day

For reviewers:
This file is large because of the ~7k lines of migration, and almost all of the file differences are because of the Force Icon pack swapover. Code-wise most of the work is on the dialog-side, to get the three icon types (with the fourth, internal type used in LayeredForceIcon) selectable and usable.

PR #3014, opened 2021-12-20
AbstractIcon: Force Icon Rework (Code Only)

Windchild292:
This is only the code for #2946, which is being opened to allow for it to be reviewed.

CI will fail on this PR.


MekHQ/src/mekhq/campaign/io/Migration/ForceIconMigrator.java
```diff
@@ -0,0 +1,6891 @@
...
+        migrateLogos(icon, layered);
+        migrateSpecialModifiers(icon, layered);
+        migrateTypes(icon, layered);
+        return layered;
+    }
+
+    private static void migrateAdjustments(final LayeredForceIcon original, final LayeredForceIcon migrated) {
+        for (final ForcePieceIcon piece : original.getPieces().getOrDefault(LayeredForceIconLayer.ADJUSTMENT, new ArrayList<>())) {
+            switch (piece.getFilename()) {
+                case "Air assault.png":
```
NickAragua: That is a very large amount of code to accommodate directory shuffling. I give it about a 0% chance of going through without a large number of  "where did my force icons go" complaints.

NickAragua (changes requested):
Looks fine, other than a complete lack of documentation on the newly introduced classes/methods and a few instances of magic numbers. Also, the migration routine is pretty rigid and I have my doubts about it.

MekHQ/src/mekhq/campaign/io/Migration/ForceIconMigrator.java
Windchild292: It's a complete internal pack swapover, not "directory shuffling". Kailan created and organized her pack based on the previous setup while taking advantage of the increased graphical functionality.

If you have a better and/or safer way to migrate the data from the old pack to the new one please feel free to write your own, but I've got faith that this will migrate the vast majority of users without issue, especially given most TO&Es use a small portion of the overall icons available.

MekHQ/src/mekhq/gui/model/PatientTableModel.java
```diff
@@ -69,7 +69,7 @@ public Component getListCellRendererComponent(
                 boolean isSelected,
                 boolean cellHasFocus) {
             Person p = getElementAt(index);
-            setPortrait(p);
+            setImage(p.getPortrait().getImage(54));
```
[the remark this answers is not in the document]
Windchild292: It's a graphical size scaling value, and is based on the individual graphical setup. That 54 tends to be the best working value in MekHQ is likely because of earlier code reuse, and is not because of anything else (portraits default to 72).

As it's an internal graphical value, like creating a dimension, insets, or anything else really graphical, I don't think it makes any sense to turn into a constant.

Windchild292:
I completely missed JavaDocing these where it makes sense, my bad.

Magic numbers are all graphical factors based on individual locations in MekHQ, which is why they aren't constants but instead  inline.

Migration for this is a necessary beast... it's 1k new and 750 removed icon files, after all.

If you're good with the replies to the other two, I'll fully JavaDoc the primary PR and then merge it in.

NickAragua:
Sure, whatever. I'd still explain the magic numbers in in-line comments, though.

PR #2864, opened 2021-09-04
2863: Adding Dates Personnel Tab Filter, Displaying Dates, and adding Pregnancy Colour

Windchild292:
This fixes #2863 and further expands the options to display all of the tracked dates in their own grouping.

PR #3017, opened 2021-12-21
PersonnelTableModelColumn Enum

Windchild292:
This swaps the PersonnelTableModel column magic numbers to the PersonnelTableModelColumn Enum, following up from #2864. This change let me standardize comparatators, fix any missing comparators, fix sort orders, update some resource keys, fix two old column setups (Batch XP and Personnel Market - First Name and Last Name are computed fields that replaced Given Name and Surname in most displays with the addition of the other name fields), and standardize everything for the three files.

I also added the Other view for the Personnel Table containing any columns that were previously export only, and fixed two bugs with the Retirement Table Model using Personnel Table Model keys.

I'm not certain about where the BatchXP and PersonnelMarket valid columns should be located, but I think it's fine where they are... and can't move the graphical data to the column enum.


SJuliez (approved):
Ok by me.

PR #3020, opened 2021-12-24
New Campaign Project: Campaign Options Pane

Windchild292:
Reviewers: Look at the bottom before starting on this PR.

This moves all core functionality for CampaignOptionsDialog to the new CampaignOptionsPane, leaving a shell dialog above the isolated pane. This nasty change is required for my startup project, where the new pane will be part of the integrated startup dialog.

Anything that I'm not certain is up to the standards of the modern changes is placed into the legacy regions, including some of my own code written over the past two years. All legacy setups will be looked at and replaced, refactored, or (unlikely) just moved to modern.

Reviewers:
This is a beast... but one that is 99% copy/paste. I've isolated everything that's not copy/paste into isolated commits, and would recommend reviewing them instead of going over the huge amount of copy/paste that just moves the legacy code instead of fixing it (I will be fixing it, but... this step is crucial for the New Campaign Project and needs to be handled faster than I can fix each individual tab).

The actual changes are:
1. 4eb928e : Bulk resources rename to work with the Base Components
2. 784e8fe : Setting up the new initialize method, with all of the methods to create the legacy tabs coded and ready to go.
3. 6284677 : Removing useless initialization calls and toString calls
4. f6143d6 : Fixing two PersonnelRole::toString calls that should be PersonnelRole::name calls
5. e3f89ee : Adding nicely organized tab titles, with regions setup for the future isolated tab rewrites
6. 260d4b9 : Fixing some panels and other variables that could just be made inline instead
7. aab0672 : Converting CampaignOptionsDialog to a Validation Dialog, albeit still only using the single validation step.
8. 5b6a141 : Converting long JPanels to JScrollablePanel for improved scrolling
9. 9e97de3 : Converting all JComboBox uses to MMComboBox
10. 6045c86 : Converting txtMaxAcquisitions to a JSpinner


Windchild292:
This has been finished and is ready for review.

Windchild292:
Converted to draft so this is ignored while PRs for 0.49.7 are open

SJuliez (approved):

PR #3021, opened 2021-12-25
prevent NPE when adding a new contract manually

NickAragua:
getSystem() in contract returns null if it's entered from the add mission -> new contract workflow.


Windchild292 (commented):
Wait, why are we comparing ids instead of just using the .equals override?

NickAragua:
'cause we know they're both systems? And .equals doesn't really help against them being null.

Windchild292:
That's not what .equals does? A basic implementation is same object, an override needs to check the instance and thus provides a null check.

NickAragua:
It doesn't help if the object on which .equals() is being called; in this case, though it's probably fine since we already check for that earlier.

Windchild292:
Yeah, and otherwise Objects.equals would work

Windchild292 (approved):

Issue #3023, opened 2021-12-26
After-action MUL file causiong MekHQ to lock up 0.49.5


BLR-IIC:
The attached after-action MUL file is causing MekHQ to lock up.  The only way I can get out of it is kill the process.  Log file and MUL file attached.
[After-action MUL file problem.zip](https://github.com/MegaMek/mekhq/files/7776500/After-action.MUL.file.problem.zip)

NickAragua:
Can we also get the campaign save/customs with the relevant scenario pending?

BLR-IIC:
Here's the relevant campaign save file:

[4th Jaguar Dragoons - Clan Smoke Jaguar30220222.cpnx.gz](https://github.com/MegaMek/mekhq/files/7776513/4th.Jaguar.Dragoons.-.Clan.Smoke.Jaguar30220222.cpnx.gz)

NickAragua:
There are about twenty-thirty customs that I don't have, can we supply those as well?

BLR-IIC:
Sure, none of them are a Sentinel as called out in one of the log files, if that's any help:
[4th Jaguar Dragoons Customs.zip](https://github.com/MegaMek/mekhq/files/7776523/4th.Jaguar.Dragoons.Customs.zip)

NickAragua:
![image](https://user-images.githubusercontent.com/29387808/147398299-975bd648-d8a6-4e73-9cac-f47bba12d867.png)

Still a few left. Recognize any of these guys?

BLR-IIC:
Here they are.  Hope this helps!

[Some missing customs.zip](https://github.com/MegaMek/mekhq/files/7776531/Some.missing.customs.zip)

PR #3025, opened 2021-12-26
Fix mul loading npe

NickAragua:
This fixes issue #3023 and probably many other similar errors for generating crews for units coming back to MekHQ for potential salvage. The previous approach was to use the monte carlo method, which breaks down when the skill is near the edge of the bell curve and leads to endless loops. Consolidated the logic instead to simply increment or decrement the skill level until it either matches the desired skill level or we can't change any more crew members' skills.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -838,8 +838,8 @@ public class Utilities {
             // if they would do so.
             boolean skillCannotChange = true;
             while (skillCannotChange) {
-                if ((skillLevel <= 0) && (skillIncrement == -1) || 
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/Utilities.java
```diff
@@ -709,46 +703,14 @@ public static int randomSkillFromTarget(int target) {
                     drivers.add(p);
                 }
 
-                // Regenerate as needed to balance
+                // Rebalance as needed to balance
                 if (drivers.size() != 0) {
```
Windchild292, line 707:
```
if (!drivers.isEmpty()) {
```
Nicer code than the current

Windchild292 (approved):
This looks great, just have an idea to improve a bit of pre-existing code

PR #3036, opened 2021-12-30
2993: Fixing Duplicated Small Craft and JumpShip Pilot Assignment

### LINKED ISSUE #2993: Doubled kill counts for vehicle crews
### Environment
I'm using 0.49.5 on Windows. The Java version is 11.0.13.

### Description
Kill records for vehicle crewmembers since switching to 0.49.5 (which happened around January 1, 3069 in game) are doubled.
image

This has happened for all crew I've looked at so far. Most have been first assigned to their vehicles since switching to 0.49.5 and Stratcon, but a few (including the one in the screenshot) have been assigned since before then. This is likely linked to the crew assignments being doubled in the TO&E, as shown here (and reported as a bug elsewhere):
image

### Files
[duplicatekillsbugfiles.zip](https://github.com/MegaMek/mekhq/files/7654673/duplicatekillsbugfiles.zip)

Windchild292:
This fixes #2993

### DIFF
--- a/MekHQ/src/mekhq/gui/menus/AssignUnitToPersonMenu.java
+++ b/MekHQ/src/mekhq/gui/menus/AssignUnitToPersonMenu.java
@@ -288,7 +288,12 @@ public class AssignUnitToPersonMenu extends JScrollableMenu {
                                     oldUnit.remove(person, !campaign.getCampaignOptions().useTransfers());
                                     useTransfers = campaign.getCampaignOptions().useTransfers();
                                 }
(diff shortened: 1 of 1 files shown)


SJuliez (approved):
It seems really strange that changing a menu item would resolve a double kill assignment but if you say so...

Windchild292:
I concur. It's because MHQ assigns a kill per person assigned to the unit. As this assigns a person twice, they get two kills assigned.

PR #3079, opened 2022-01-20
Removing the Abandoned Chat Setup

Windchild292:
This deletes a bunch of unused and unfinished code that has been abandoned since 2013 at the earliest.


AaronGullickson (approved):
Yeah, that was from me a LONG time ago. I adapted that code from somewhere with the idea of integrating it in at some point and then never did. I think there are much better ways to do this now.

PR #3084, opened 2022-01-21
Integrate 3151 SUCS

AaronGullickson:
This PR integrates changes to faction ownership of planetary systems from the 3151 [SUCS data](https://docs.google.com/spreadsheets/d/1x9bvFqSb4_or8JbvGj2LnezGkChWxEzRPf5FXvjonHE). The procedure for doing this is now automated from an R script that is available [here](https://github.com/AaronGullickson/update_factions). This is made possible by including a sucsId in the system_events data that allows us to easily link our planetary systems with theirs. 

The other complication here is the faction codes. The [SUCS faction codes](https://docs.google.com/spreadsheets/d/1x9bvFqSb4_or8JbvGj2LnezGkChWxEzRPf5FXvjonHE) do not always match ours although we have adopted their codes for any new factions. To address this issue, we created a [faction correspondence](https://docs.google.com/spreadsheets/d/117Mmhf7TtyumwCzB9bGKq05-SeTFKboZ7Ef2ZbyOEPk) sheet that includes a translation table for translating some of their codes to our codes. 

One final issue is the PIND code. That is our "Periphery Independent" code which is different from IND. SUCS just has an IND code. There is no good programmatic way to determine which one a system should be. In most cases, I can rely on their previous faction code, but if they have become newly independent, I have nothing to go on. My rough approach now is to base this on distance from Terra with 450 LY being the cutoff. however, because the inner sphere is not in fact a perfect sphere this is imperfect. So a few of these will need to be fixed by hand. 

Alternatively, we could just convert all of our PIND codes to IND to be consistent with SUCS. That would probably be preferable if does not impact things elsewhere, but we may use the PIND code specifically in generating random forces.


AaronGullickson:
Never mind. I stupidly branched this PR off of story arcs instead of master so let me redo that.

PR #3085, opened 2022-01-21
Integrate faction changes from 3151 SUCS data

AaronGullickson:
This PR integrates changes to faction ownership of planetary systems from the 3151 [SUCS data](https://docs.google.com/spreadsheets/d/1x9bvFqSb4_or8JbvGj2LnezGkChWxEzRPf5FXvjonHE). The procedure for doing this is now automated from an R script that is available [here](https://github.com/AaronGullickson/update_factions). This is made possible by including a sucsId in the system_events data that allows us to easily link our planetary systems with theirs. 

The other complication here is the faction codes. The [SUCS faction codes](https://docs.google.com/spreadsheets/d/1x9bvFqSb4_or8JbvGj2LnezGkChWxEzRPf5FXvjonHE) do not always match ours although we have adopted their codes for any new factions. To address this issue, we created a [faction correspondence](https://docs.google.com/spreadsheets/d/117Mmhf7TtyumwCzB9bGKq05-SeTFKboZ7Ef2ZbyOEPk) sheet that includes a translation table for translating some of their codes to our codes. 

One final issue is the PIND code. That is our "Periphery Independent" code which is different from IND. SUCS just has an IND code. There is no good programmatic way to determine which one a system should be. In most cases, I can rely on their previous faction code, but if they have become newly independent, I have nothing to go on. My rough approach now is to base this on distance from Terra with 450 LY being the cutoff. however, because the inner sphere is not in fact a perfect sphere this is imperfect. So a few of these will need to be fixed by hand. 

Alternatively, we could just convert all of our PIND codes to IND to be consistent with SUCS. That would probably be preferable if does not impact things elsewhere, but we may use the PIND code specifically in generating random forces.


HammerGS (approved):
Good to go.

Windchild292:
@Taharqa this has merge conflicts, and the faction correspondence sheet is currently marked as private.

AaronGullickson:
Yeah, trying to figure out whats going on as it should have pulled the latest system_events.xml straight from the repo. Give me a bit.

AaronGullickson:
Also, changed that sheet to public.

Windchild292:
I'm fine with removing PIND, it's just treated as an alias for IND internally anyway.

I've also got fixes for the JàrnFòlk and Fiefdom of Randis naming issues in #3033 (was adding some data to factions.xml when I noticed the two issues).

AaronGullickson:
Ok, I think the best option is to just remove PIND and replace with IND throughout so we are consistent with SUCS. I will add that change to this PR before merging.

AaronGullickson:
Conflicts are resolved, but not sure what caused them. Every faction change was flagged as a conflict even though it was working off the latest file.

Windchild292 (approved):

PR #2997, opened 2021-12-11
Story Arcs Basic Architecture

AaronGullickson:
This PR adds a major new feature, Story Arcs. I won't go over what story arcs are here as that is covered extensively on the [Project Details Page](https://github.com/orgs/MegaMek/projects/7?pane=info) for the GH project. Here I will discuss relevant information for reviewers.

This PR adds the underlying architecture for Story Arcs, and provides a test story arc named "Young Wolves" that players can use to try out the concept. It does not contain any ability to create story arcs apart from writing literal XML code. That is the next major planned PR in the project. 

I have tried to design the architecture of story arcs so that additional features are relatively easy to implement by simply extending the abstract classes of `StoryPoint` and `StoryTrigger` to new classes that can do particular things. In particular, as we develop story arcs further, we will likely see lots of requests to expand the use of `StoryTriggers` which can be used to make changes to the underlying campaign based on a certain result from a `StoryPoint`. Currently, I only built the triggers necessary for the example story arc, and one could imagine all kinds of other things that could be added. `StoryPoints` provide much more coverage of the kinds of things that are needed, but I can see the need to expand the number of "Check" StoryPoints which check a condition of something in the campaign.  

Although this PR changes/adds a lot of files, most of the work is contained within story arcs and should not affect base code. Any additional new features that I needed to get story arcs up and off the ground that involved significant changes to the base MHQ code have already been done as separate PRs and merged into master. Therefore, this code should touch existing code relatively lightly and any errors remaining will likely just mess up story arcs, not MHQ more broadly.

Given the extensiveness of the code, I am holding off on writing tests for the moment as that would delay this PR even further. On the project page, I do list adding tests for story arcs as one of the future goals, but it is prioritized lower than getting out an editor.

PR #3099, opened 2022-01-29
Remove bad logic from check on preventing breakthrough

AaronGullickson:
Ok, this is a very small PR intended to change the logic of the ScenarioObjectiveProcessor with regard to the preventing breakthrough objective. I am fairly certain that the existing logic makes no sense.

This came about because I was building a prevent breakthrough scenario for the demo story arc in PR #2997 and noticed that it was not assigning units correctly based on the outcome. Basically, the objectives was to prevent the breakthrough of 70% of the enemy forces (who were arriving in waves) for 12 turns before retreating. After running through the scenario and seemingly accomplishing the mission, the only enemy units that were shown as correct for the scenario objective were those units that I destroyed, even though I had prevented the breakthrough of most units. 

I tracked the issue back to two lines of code. The first problem is in `ScenarioObjectiveProcessor#UpdateObjectiveEntityState` which is called after the user makes changes to the status of a unit in the scenario resolution tracker. Specifically:

```java
case PreventReachMapEdge:
                        entityMeetsObjective = forceEntityDestruction ||
                            !forceEntityEscape && !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
```

The problem there is the `!forceEntityEscape` logic. By default any entity that is not destroyed or crippled is treated as escaping. The logic here assumes that if an entity is listed as escaping then it should be considered as having broken through despite the fact that it may have literally failed to breakthrough. This is bonkers. It effectively means that the only way to prevent an entity from breaking through is to destroy it and if that is the logic then this scenario objective is effectively no different than a forced withdrawal scenario objective.

The second problem regards who has control of the battlefield. In `ScenarioObjectiveProcessor#entityHasReachedDestinationEdge` we have:

```java
return ((entity.getRetreatedDirection() != OffBoardDirection.NONE) && (entity.getRetreatedDirection() == objective.getDestinationEdge())) ||
                // or the opponent has been routed and the entity hasn't been shot to hell
                (!entity.isCrippled() && !entity.isDestroyed() && !opponentHasBattlefieldControl);
```

The second part of the or statement is the problem. It assumes that if the other side did not have control of the battlefield and you weren't crippled and destroyed that you will eventually get there. But that is why assumptions are often problematic. It completely makes impossible the scenario I described above in which the goal is to hold off the enemy unit for a certain amount of time before pulling back. Battlefield control at the end of the scenario has nothing to do with it. 

In short, if the goal of the scenario is to prevent a breakthrough or to accomplish one, then what should matter is actually breaking through in the scenario.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
@@ -174,7 +174,7 @@ public class ScenarioObjectiveProcessor {
                         break;
                     case PreventReachMapEdge:
                         entityMeetsObjective = forceEntityDestruction ||
-                                !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
+                                !entityHasReachedDestinationEdge(entity, objective);
                         break;
                     case Preserve:
                         entityMeetsObjective = forceEntityEscape ||
@@ -182,7 +182,7 @@ public class ScenarioObjectiveProcessor {
                         break;
                     case ReachMapEdge:
                         entityMeetsObjective = forceEntityEscape ||
-                            !forceEntityDestruction && entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
+                            !forceEntityDestruction && entityHasReachedDestinationEdge(entity, objective);
                         break;
                     // criteria that we have no way of tracking will not be doing any updates
                     default:
@@ -215,11 +215,11 @@ public class ScenarioObjectiveProcessor {
                 case Capture:
                     return entityIsCaptured(entity, !opponentHasBattlefieldControl);
                 case PreventReachMapEdge:
-                    return !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
+                    return !entityHasReachedDestinationEdge(entity, objective);
                 case Preserve:
                     return !entityIsDestroyed(entity, opponentHasBattlefieldControl);
                 case ReachMapEdge:
-                    return entityHasReachedDestinationEdge(entity, objective, !opponentHasBattlefieldControl);
+                    return entityHasReachedDestinationEdge(entity, objective);
                 default:
                     return false;
             }
@@ -259,7 +259,7 @@ public class ScenarioObjectiveProcessor {
     /**
      * Check whether or not the entity can be considered as having reached the destination edge in the given objective
      */
-    private boolean entityHasReachedDestinationEdge(Entity entity, ScenarioObjective objective, boolean opponentHasBattlefieldControl) {
+    private boolean entityHasReachedDestinationEdge(Entity entity, ScenarioObjective objective) {
                 // we've reached the destination edge if we've reached an edge and it's the right one
         return ((entity.getRetreatedDirection() != OffBoardDirection.NONE) && (entity.getRetreatedDirection() == objective.getDestinationEdge()));
     }
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
@@ -174,7 +174,7 @@ public class ScenarioObjectiveProcessor {
                         break;
                     case PreventReachMapEdge:
                         entityMeetsObjective = forceEntityDestruction ||
-                            !forceEntityEscape && !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
+                                !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
                         break;
                     case Preserve:
                         entityMeetsObjective = forceEntityEscape ||
@@ -261,9 +261,7 @@ public class ScenarioObjectiveProcessor {
      */
     private boolean entityHasReachedDestinationEdge(Entity entity, ScenarioObjective objective, boolean opponentHasBattlefieldControl) {
                 // we've reached the destination edge if we've reached an edge and it's the right one
-        return ((entity.getRetreatedDirection() != OffBoardDirection.NONE) && (entity.getRetreatedDirection() == objective.getDestinationEdge())) ||
-                // or the opponent has been routed and the entity hasn't been shot to hell
-                (!entity.isCrippled() && !entity.isDestroyed() && !opponentHasBattlefieldControl);
+        return ((entity.getRetreatedDirection() != OffBoardDirection.NONE) && (entity.getRetreatedDirection() == objective.getDestinationEdge()));
     }
 
     /**
--- a/MekHQ/data/scenariotemplates/Harass.xml
+++ b/MekHQ/data/scenariotemplates/Harass.xml
@@ -128,12 +128,11 @@
                     <howMuch>1</howMuch>
                 </failureEffect>
             </failureEffects>
-            <additionalDetails>
-                <additionalDetail>Crippled units do not count towards this objective.</additionalDetail>
-            </additionalDetails>
-            <description>Prevent 50% of the following force(s) and unit(s) form reaching the opposite map edge:</description>
+            <additionalDetails/>
+            <description>Destroy 50% of the following force(s) and unit(s) before they can reach the destination
+                edge:</description>
             <destinationEdge>NONE</destinationEdge>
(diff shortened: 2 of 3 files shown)


MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
```diff
@@ -261,9 +261,7 @@ private boolean entityIsCaptured(Entity entity, boolean opponentHasBattlefieldCo
      */
     private boolean entityHasReachedDestinationEdge(Entity entity, ScenarioObjective objective, boolean opponentHasBattlefieldControl) {
```
Windchild292, line 262: The final parameter isn't necessary anymore, and should be removed.

Windchild292 (changes requested):
Changes are sound, just one useless parameter to remove

AaronGullickson:
I woul like to get @NickAragua's thoughts as well, since this will affect AtB.

MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
```diff
@@ -174,15 +174,15 @@ public void updateObjectiveEntityState(Entity entity, boolean forceEntityEscape,
                         break;
                     case PreventReachMapEdge:
                         entityMeetsObjective = forceEntityDestruction ||
-                            !forceEntityEscape && !entityHasReachedDestinationEdge(entity, objective, opponentHasBattlefieldControl);
+                                !entityHasReachedDestinationEdge(entity, objective);
```
NickAragua, line 177: The logic here was originally that the unit automatically counts as having reached the edge if the opposing force retreats. So if the player answers "no" to "did you have battlefield control", then this objective should automatically fail.

NickAragua (commented):
This may need some revision or discussion, as the change goes against the original intent of the functionality.

MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
AaronGullickson: I understand that and I am saying that this is bad logic. See the example in my OP. This is a fairly common example of what a breakthrough scenario might look like - hold off a numerically superior force for some time before retreating. Note an equally important case is also made impossible on the other side when the player is the breakthrough force - trying to break through a numerically superior force. In both cases, the player would not have battlefield control. 

My basic point is that the battlefield control bit basically assumes too much about the context of the scenario and as a result it reduces the flexibility of this objective. I think that is a poor design decision. Removing it simply makes the objective literally work as intended - the goal is to either exit your units or prevent the exit of other units to a certain map edge. 

This logic also make these two objectives largely superfluous. If you lose the objective because you do not have battlefield control, then in practice this objective is about establishing battlefield control, which makes it like most other objectives and diminishes the ability to give players a more diverse set of potential objectives.

NickAragua: Sounds like there needs to be two types of "Reach Edge" objective then - one affected by battlefield control and one not.

AaronGullickson: Maybe, but my guess is that the other type could probably be accomplished by existing scenario objectives. In this case, I think the Destroy objective in combination with a bot whose behavior is to exit via the player's home map edge and forced withdrawal set to false would basically accomplish this. 

In regards to tradeoffs, @NickAragua, for the PreventBreakthrough objective there is no tradeoff option. If you do not control the battlefield, then you will lose unless you destroy or cripple enough of the opposing force. Again, if that is the goal, it can be handled by other objectives.

NickAragua: I'm thinking of the following scenario:

You have bot mechs trying to cross over to the east side of a battlefield. If the player gives up control of the battlefield (by retreating or being annihilated), then any hostile mechs still on the field (barring immobilized and possibly crippled ones) should count has having reached the east side, even if they haven't done so at the point when the player ended the scenario. 

Unless I'm misreading the logic, this pull request makes it a requirement that the bot units *actually* cross the east edge even if the player's force has already been annihilated.

AaronGullickson: Well, thats exactly the problem. Whether the hostile mechs still on the field should count as having broken through depends on the context of the scenario. In my opening example, it should not if the player held out for the requisite number of turns before ending the scenario. In a scenario without a time limit, it s reasonable to assume those mechs will breakthrough. 

I think the missing feature that would allow us to distinguish the cases is really taking account of the time limits. Thats something I would like to do eventually, as I would like to improve scenario objectives in the long run, but that is a much bigger project. 

In the interim, I do not think the scenario you outline should be a reason to leave the logic. In some cases, until we work out a more complete system, players may have to correct things in the resolution wizard. But this scenario should be relatively rare as the bot's goal is not to destroy player units but to reach the other side. So, they are not sticking around to finish off heavily damaged units. And in situations where this occurs, the player can still get the correct outcome by continuing the game until the bot fully exits the map, which it will continue to do. Similarly in a breakthrough, a player who completely destroys the opposing force can continue the game and exit all of their units and I would argue that is precisely what they should do.

AaronGullickson: I should also note that if you are running a prevent breakthrough scenario without a time limit, then you are really better off to set the scenario objective as destroy rather than prevent breakthrough because this is actually what the player is being asked to do - destroy the units before they exit. In other words, I think the issue you are concerned with @NickAragua can mostly be handled by the existing architecture. It just requires some rethinking about how scenario objectives are assigned.

NickAragua: I don't think we're going to come to an agreement on this, but one other point is that there are already several scenarios that use these objective types and rely on them working the way they work currently, with no time limit involved. So if you want to modify one of those to work roughly the way it worked previously as an example of what you're thinking, then I'm fine letting this go.

Some examples (all in data/scenariotemplates)
"Breakthrough.xml"
"Prevent Breakthrough.xml"
"Pursuit.xml"
"Harass.xml"

AaronGullickson: Fixed almost all of these issues in commit  8d8e07121bad24. Nine templates involve either the Breakthrough or PreventBreakthrough objective but only two actually need to be changed:

* Breakthrough.xml - This works fine as is. If the player clears the board of opposition, they can walk off the map themselves and meet the objective. Note that if you don't want to make players walk off then you could also create another objective to ForceWithdraw 100% of the enemy force to win. If that is the preferred solution, I can add that code. Same applies to several similar templates below.
* Diversion.xml - This appears to be more or less the same as Breakthrough.xml with some slight tweaks to parameters, so the same logic applies. If the player clears the board, they can walk off to meet the objective.
* Pursuit.xml  - this is very similar to Breakthrough.xml but also includes logic where the player can win by breakthrough or by destroying/routing 50% of the OpFor. So this will work fine as is.
* PreventBreakthrough.xml and Harrass.xml - These are the two cases that I changed. The objective is now to destroy at least 50% of the opposing force before they reach the other map edge.  This is functionally equivalent to the prior objective but works with both old and new logic. It is literally the goal here.
* ConvoyEscort.xml - As with Pursuit.xml you can escort the units to the map edge or route the enemy force to win, so no issue here.
* Low Atmo - Escort.xml and Space - Blockade Run Escort.xml - As with Pursuit.xml, you can escort the units to the map edge or route the enemy force to win, so no issue here.
* ConvoyAttack.xml - This is the only one, I have not sorted out at the moment, because its logic is the most complicated, but I will work more on it later.

So, aside from the one remaining template which I need to fix, I think the overall impact here on existing ScenarioTemplates is actually relatively minimal and can easily be changed to actually better reflect what is being asked of the player in the two cases that need to be changed.

NickAragua (approved):

MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
NickAragua: All right, I think we've reached a reasonable compromise here.

Windchild292 (approved):

Windchild292:
Merging so this can join the new testing builds.

PR #3104, opened 2022-02-02
Creating Clamp Utilities

Windchild292:
These should be part of Java.Math, but aren't. This will allow us to simplify the code in MekHQ.


SJuliez (changes requested):
Request moving this to MM :)

Windchild292:
This was meant to be a shortterm bridge for that, but given it's been left open for a while I'm going to close this and create a nicer setup of constant code utilities.

PR #3105, opened 2022-02-02
Base Components: CompleteMissionDialog Rewrite

Windchild292:
This rewrites the CompleteMissionDialog so it implements AbstractMHQDialog, with all the UX and Accessibility improvements that originate from that.


SJuliez (changes requested):
Not really requesting changes, just something I couldnt figure out: Does the outcomeStatus Combobox in the dialog get any entries? It seems like the dialog gets created and shown but I cant find where the combo is filled.

Windchild292:
It's filled in the initial constructor.

SJuliez (approved):

PR #3115, opened 2022-02-13
Fixing MHQ's Javadoc Build

Windchild292:
This continues on from https://github.com/MegaMek/megamek/pull/3445 and https://github.com/MegaMek/megameklab/pull/1028 to fix the Javadoc build for MekHQ.

Akin to the previous two, there are still huge gaps in our Javadoc coverages, and tons of warnings.

This also changes a bunch of copyright headers relating to < and > emails in the copyrights, as it was far easier to just bulk rename than deal with it each of the hundred times it is used.


SJuliez (approved):
I will not look at this closely as it only concerns comments.

Issue #3111, opened 2022-02-11
RFE - Add 'Heroic' and 'Legendary' pilot skill brackets above Elite in MekHQ.


UlyssesSockdrawer:
### Environment
Any version - MekHQ, possibly MegaMek also?

### Description
MekHQ currently categorises pilots from Ultra Green up to Elite at 2/3. In longer campaigns many pilots can progress beyond this skill level.

Adding in the 'Heroic' (1/2) and 'Legendary' (0/1) skill brackets would help distinguish these pilots/crews in the hangar/personnel screen. 

Additionally these skill brackets are referenced in some of the Clan-focused module books (Battle of Tukayyid) so would bring MekHQ's skill ratings in line with the latest campaign books.

### Files
N/A

Windchild292:
MegaMek has recently had Heroic added, but I did miss Legendary.

This will be handled as part of the huge skill level rework that I'll be handling in the nearish future, phase one of updating skills (the rest of which is still far future)

Windchild292:
I've opened https://github.com/MegaMek/megamek/pull/3446 and https://github.com/MegaMek/mekhq/pull/3116 to add the Legendary SkillLevel to MegaMek. MekHQ still largely uses an older setup that needs a rework, as mentioned above.

Windchild292:
MegaMek now has Legendary. MekHQ will need the nasty rework to swap to the standardized enum.

PR #3116, opened 2022-02-13
Adding the Legendary SkillLevel

Windchild292:
This is the first part of handling #3111. MekHQ has only been partially swapped to use SkillLevel, as it previously used two setups. This is thus the only place that needs to change for now.

Requires https://github.com/MegaMek/megamek/pull/3446


SJuliez (approved):
Wow that line took me a minute to parse.

PR #3122, opened 2022-02-17
BLOCKER: Read the font directory for MHQ instead of ignoring it.

Windchild292:
This is required for resolutions above 1080p, and is just code copied from https://github.com/MegaMek/megamek/pull/3427.


HammerGS (commented):
I don't know the code, but can test it if you merge it. If it doesn't work we could revert it.

Windchild292:
Juliez approved the code in https://github.com/MegaMek/megamek/pull/3427

PR #3154, opened 2022-03-04
Properly refresh logistics target when handling the person update

Windchild292:
This bug was found by misclick... I kid you not.

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -2490,11 +2490,7 @@ public class CampaignGUI extends JPanel {
 
     @Subscribe
     public void handlePersonUpdate(PersonEvent ev) {
-        // only bother recalculating AtB parts availability if a logistics admin has been changed
-        // refreshPartsAvailability cuts out early with a "use AtB" check so it's not necessary here
-        if (ev.getPerson().hasRole(PersonnelRole.ADMINISTRATOR_LOGISTICS)) {
-            refreshPartsAvailability();
-        }
+        refreshPartsAvailability();
     }
 
     @Subscribe


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -2490,11 +2490,7 @@ public void handleMissionChanged(MissionEvent ev) {
 
     @Subscribe
     public void handlePersonUpdate(PersonEvent ev) {
-        // only bother recalculating AtB parts availability if a logistics admin has been changed
-        // refreshPartsAvailability cuts out early with a "use AtB" check so it's not necessary here
-        if (ev.getPerson().hasRole(PersonnelRole.ADMINISTRATOR_LOGISTICS)) {
-            refreshPartsAvailability();
-        }
+        refreshPartsAvailability();
```
NickAragua, line 2493: Is the efficiency shortcut that was removed present in refreshPartsAvailability()? Otherwise, that's going to be a lot of pointless refreshing.

NickAragua (commented):
Efficiency check.

MekHQ/src/mekhq/gui/CampaignGUI.java
Windchild292: The check itself is the bug as it is not based on the campaign options and thus ignores the proper availability checks.

NickAragua: The question still stands, does the parts availability refresh cut out early if the person doesn't have the correct role (whatever that is)?

Windchild292: It's filtered properly thanks to your code (from a quick blame check)

Windchild292: I misunderstood what you meant last night, my apologies for that.

I see what you mean now, and... yeah, my fix to handle the campaign option would definitely cause performance issues.

I'm going to convert this to draft as I'll need to investigate the best way to tell the correct roles to actually parse for.

Windchild292:
Replaced with #3320

PR #3155, opened 2022-03-04
Person Cleanup and Organization

Windchild292:
This PR handles the cleanup of Person. This class was the center of my focus when I first started to contribute to the suite, and has continued to be crucial to my code as I've expanded my focus. Thus, as part of this cleanup I've fixed a ton of mistakes I've made when working on the suite, primarily from my first few months of contributing.

I'll be following this cleanup with a complete reorganization PR so the class is nicely organized instead of its current spread out mess of methods with some grouped regions of sanity.


Windchild292 (commented):
Personal review changes after first evening

Windchild292:
This PR is ready for review

SJuliez (approved):
On some lines, only part of the " " or "\"" have been replaced with ' ' or '"'. Not a problem but I couldnt figure out why.

Windchild292:
Char is slightly faster performing than the string version.

PR #3183, opened 2022-03-21
Removing the MHQ NLS and i18n useless comments

Windchild292:


HammerGS (approved):
Not a fan of these monolithic refactoring but a quick scan looks like nothing code wise is changing.

Windchild292:
There isn't, it's just removing now useless comments.

PR #3187, opened 2022-03-23
3185: Fixing Ransom With None Skill Experience

### LINKED ISSUE #3185: 0.49.7 NPE Prisoner Ransom
Simple enough bug here.  When attempting to ransom a prisoner dependent, the game returns a NPE.  Obviously the reason for this is that there is no ransom value for a dependent.

Potential Solutions:
1) Assign dependents (and the unknown skill value) as 0.
2) Make children born to prisoners not prisoners
3) Don't allow prisoners to have children (although if more support is added for pirates, we'll still want slaves to be able to procreate)

[mekhq.log](https://github.com/MegaMek/mekhq/files/8336117/mekhq.log)

Windchild292:
This fixes #3185 by adding the none skill experience.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -218,6 +218,7 @@ public class Person {
     // initializes the AtB ransom values
     static {
         MECHWARRIOR_AERO_RANSOM_VALUES = new HashMap<>();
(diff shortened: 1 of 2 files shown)


NickAragua (approved):
I'd argue that you might get some money for those, but honestly it's a can of worms I don't want to touch, so this should be fine for now.

Windchild292:
I concur. It should be user customizable, which I have plans to do as part of the Prisoner module. The goal here is just to prevent it from breaking, which this PR handles.

PR #3219, opened 2022-04-12
Tearing out 0.45.4 Personnel log Migration

Windchild292:


NickAragua (approved):
Can't wait until the complaints start rolling in. "I want to upgrade my campaign directly from 37.5 to 45.4, but it keeps giving me errors!"

Windchild292:
It already doesn't work, as we've removed code for 0.45 migrations before this point... (surprisingly) with no comments over the past year.

PR #3224, opened 2022-04-14
Personnel Modules: Death

Windchild292:
This follows https://github.com/MegaMek/mekhq/pull/2493, https://github.com/MegaMek/mekhq/pull/2851, and https://github.com/MegaMek/mekhq/pull/2908 by creating the Death Module, which uses the same design to ensure the same level of functionality and customization as the three previous modules.

This PR does the following:
- [x] Abstract Random Death Baseline
- [x] Disabled Random Death Method Implementation
- [x] Percentage Random Death Method Implementation
- [x] Exponential Random Death Method Implementation
- [x] Age Range Random Death Method Implementation
- [x] Age Group Enum (Baby, Toddler, Child, Pre-Teen, Teen, Adult, Elder)
- [x] Ten Year Age Range Enum
- [x] Data-based Random Death Cause Generator
- [x] Random Death Cause Generator UserData Support
- [x] Random Death Cause Generator Data Refresh Menu Item


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -927,6 +929,13 @@ public void changeStatus(final Campaign campaign, final LocalDate today,
 
         switch (status) {
             case ACTIVE:
+                /*
```
NickAragua, line 932: I assume this is for future development - if so, then we don't need to clean it up.

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
```diff
@@ -0,0 +1,291 @@
...
+        return enableRandomDeathSuicideCause;
+    }
+
+    public Map<Gender, Map<TenYearAgeRange, WeightedDoubleMap<PersonnelStatus>>> getCauses() {
+        return causes;
+    }
+    //endregion Getters
+
+    /**
+     * This is used to determine if a person can divorce.
```
NickAragua, line 110: Though some people may view it that way, divorce != death

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
```diff
@@ -0,0 +1,291 @@
...
+            return PersonnelStatus.KIA;
+        } else if (person.hasInjuries(false)) {
+            final PersonnelStatus status = determineIfInjuriesCausedTheDeath(person);
+            if (!status.isActive()) {
+                return status;
+            }
+        }
+
+        if (person.isPregnant()) {
+            return PersonnelStatus.PREGNANCY_COMPLICATIONS;
```
NickAragua, line 185: Is this really always the case?

NickAragua (approved):
Looks pretty good  - one semantic question and one request to change documentation, but I don't see a need to re-review it unless it results in actual code changes.

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
Windchild292: I copied the base design from divorce during modernization, must have missed this comment.

MekHQ/src/mekhq/campaign/personnel/Person.java
Windchild292: It is, and is only included because I missed this in the second personnel status expansion PR... both of which were split off of the previous death PRs.

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
Windchild292: It is the design for now.

This should be a rare enough condition that this simplification should be fine, and I will be adjusting the cause generation with user feedback involved.

PR #3244, opened 2022-04-26
2696: Added coloring to the Unit List on the Personnell assignment menus

### LINKED ISSUE #2696: RFE: Coloured Unit Assignment in Personnel Tab
"Would it be possible to have the Unit Assignment column in the Personnel tab for a specific person be the same color the unit's line in the Hanger is due to damage? 

It would be helpful if the Unit Assignment column in the Personnel tab was color coded according to the assigned unit's status (yellow for light damage etc), and even better if the same color coding could apply to the right click submenu in the Assign to Unit-> As Pilot -> Mek -> Medium -> (coloring begins) damaged/not full ammo Wolverine is yellow, inoperable Shadow Hawk is red; as an example. Add Unit in the right click on ToE menu would also benefit from this, as that is also not colored for damaged units; though currently once in the ToE it's colored red if the unit is damaged. However this means having to inspect the Hangar tab to see what (potentially very similar named units) are damaged and what is not or removing and adding units you think aren't damaged until you get one that isn't red in the ToE.

It is sometimes takes some time to tell which out of your units is undamaged (such as when you pull Stinger number 4 out of mothballs/reserve lance while Stinger number 2 and Stinger number 5 are unpiloted due to battle damage) to assign crew to, as unless I'm missing something the Hanger right click on unit menu doesn't have an assign crew option, just an assign tech option. This might also work instead of coloring the Unit Assignment column for the purpose." - Comment in #2589 

I've split off the assign person to unit in the hangar tab to #2695, as this is not the first time I've heard this request.

swang300:
This resolves #2696 I was looking for a good place to put the coloring, but didn't see a common place any better than in Unit.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -3405,9 +3405,8 @@ public class Unit implements ITechnology {
         } else if (getActiveCrew().size() < getFullCrewSize()) {
         	return MekHQ.getMHQOptions().getUncrewedForeground();
         } else {
-        	return UIManager.getColor(type+".foreground");
+        	return UIManager.getColor(type + ".Foreground");
         }
-    	
     }
     
     public Color determineBackgroundColor(String type) {
@@ -3432,9 +3431,10 @@ public class Unit implements ITechnology {
         } else if (getActiveCrew().size() < getFullCrewSize()) {
         	return MekHQ.getMHQOptions().getUncrewedBackground();
         } else {
-        	return UIManager.getColor(type+".Background");
+        	return UIManager.getColor(type + ".Background");
         }	
     }
+    
     /**
      * Determines which crew member is considered the unit commander. For solo-piloted units there is
      * only one option, but units with multiple crew (vehicles, aerospace vessels, infantry) use the following
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -67,6 +67,8 @@ import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import javax.swing.UIManager;
+
 /**
  * This is a wrapper class for entity, so that we can add some functionality to it
  *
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3380,7 +3382,59 @@ public Image getImage(Component component) {
...
+        } else if (!isRepairable()) {
+        	return MekHQ.getMHQOptions().getNotRepairableForeground();
+        } else if (!isFunctional()) {
+        	return MekHQ.getMHQOptions().getNonFunctionalForeground();
+        } else if (hasPartsNeedingFixing()) {
+        	return MekHQ.getMHQOptions().getNeedsPartsFixedForeground();
+        } else if (getActiveCrew().size() < getFullCrewSize()) {
+        	return MekHQ.getMHQOptions().getUncrewedForeground();
+        } else {
+        	return UIManager.getColor(type+".foreground");
```
Windchild292, line 3408:
```suggestion
        	return UIManager.getColor(type + ".foreground");
```

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3380,7 +3382,59 @@ public Image getImage(Component component) {
...
+        } else if (!isRepairable()) {
+        	return MekHQ.getMHQOptions().getNotRepairableBackground();
+        } else if (!isFunctional()) {
+        	return MekHQ.getMHQOptions().getNonFunctionalBackground();
+        } else if (hasPartsNeedingFixing()) {
+        	return MekHQ.getMHQOptions().getNeedsPartsFixedBackground();
+        } else if (getActiveCrew().size() < getFullCrewSize()) {
+        	return MekHQ.getMHQOptions().getUncrewedBackground();
+        } else {
+        	return UIManager.getColor(type+".Background");
```
Windchild292, line 3435:
```suggestion
        	return UIManager.getColor(type + ".Background");
```

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3380,7 +3382,59 @@ public Image getImage(Component component) {
...
+        } else if (!isFunctional()) {
+        	return MekHQ.getMHQOptions().getNonFunctionalForeground();
+        } else if (hasPartsNeedingFixing()) {
+        	return MekHQ.getMHQOptions().getNeedsPartsFixedForeground();
+        } else if (getActiveCrew().size() < getFullCrewSize()) {
+        	return MekHQ.getMHQOptions().getUncrewedForeground();
+        } else {
+        	return UIManager.getColor(type+".foreground");
+        }
+
```
Windchild292, line 3410: No need for the extra line between braces here

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3380,7 +3382,59 @@ public Image getImage(Component component) {
...
+        	return MekHQ.getMHQOptions().getNonFunctionalBackground();
+        } else if (hasPartsNeedingFixing()) {
+        	return MekHQ.getMHQOptions().getNeedsPartsFixedBackground();
+        } else if (getActiveCrew().size() < getFullCrewSize()) {
+        	return MekHQ.getMHQOptions().getUncrewedBackground();
+        } else {
+        	return UIManager.getColor(type+".Background");
+        }	
+    }
     /**
```
Windchild292, line 3438: Missing line before next method

Windchild292 (changes requested):
Please swap this from tab spacing to 4 space spacing, and remove the SerialVersionUid fields as they are completely useless (only MegaMek save games use java.io.Serialization, and that is planned for removal in the near future).

There's also a few spacing change comments, but otherwise this looks good.

swang300:
Definitely fixed the spacing issues and the extra lines.  I think I fixed the tab-space issues.  Not too sure though with git commit transformations.

Windchild292 (changes requested):
This still has the serialVersionUID fields, which we don't accept.

Windchild292 (changes requested):
SerialVersionUID still exists in the two menus

Windchild292 (approved):
Thanks for the help!

swang300:
Sorry about that.  did a search just in case.  FYI  there's one more in CompanyGenerationOptions

Windchild292:
I know, I've just been waiting for the final MM removal to remove any missed remnants from changes that predate the start of the removal.

PR #3264, opened 2022-05-06
Connection Rework Phase 2.5: Adding Start Game Options for Low Resource Systems

Windchild292:
This is necessary for low resource systems and is thus a blocker.


Windchild292:
This blocker is ready for review

PR #3284, opened 2022-05-14
AbstractDeath: Fixing Cause Generation Typo

Windchild292:
This fixes the cause generation and doesn't load causes when random death isn't enabled (no need to waste the memory)

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
+++ b/MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
@@ -64,7 +64,9 @@ public abstract class AbstractDeath {
         setUseRandomPrisonerDeath(options.isUseRandomPrisonerDeath());
         this.enableRandomDeathSuicideCause = options.isUseRandomDeathSuicideCause();
         this.causes = new HashMap<>();
-        initializeCauses();
+        if (!method.isNone()) {
+            initializeCauses();
+        }
     }
     //endregion Constructors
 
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
```diff
@@ -64,7 +64,9 @@ protected AbstractDeath(final RandomDeathMethod method, final CampaignOptions op
         setUseRandomPrisonerDeath(options.isUseRandomPrisonerDeath());
         this.enableRandomDeathSuicideCause = options.isUseRandomDeathSuicideCause();
         this.causes = new HashMap<>();
-        initializeCauses();
+        if (!method.isNone()) {
```
rjhancock, line 67: Possible to add an opposite version of this for easier reading?

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
```diff
@@ -252,11 +254,12 @@ private void initializeCausesFromFile(final File file) {
             if (!wn.hasChildNodes()) {
                 continue;
             }
+
             try {
                 final Gender gender = Gender.valueOf(wn.getNodeName());
                 getCauses().putIfAbsent(gender, new HashMap<>());
                 final NodeList nl2 = wn.getChildNodes();
-                for (int j = 0; j < nl.getLength(); j++) {
+                for (int j = 0; j < nl2.getLength(); j++) {
```
rjhancock, line 262: Can we get better variable names other than `nl2` and `nl`? I know they mean `NodeList` bust still...

rjhancock (approved):

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
Windchild292: I personally find it to be harder to read, which is why I've coded it this way.

MekHQ/src/mekhq/campaign/personnel/death/AbstractDeath.java
Windchild292: It's a legacy naming setup, and one I don't want to fix currently. It is annoying, but standardization here is better for now.

rjhancock: Ugh. Maybe one day.

PR #3339, opened 2022-06-06
Fixing Try With Resources Resource Leaks

Windchild292:
Commit 1 moves personnel tab code outside of the file writing try catch and fixes missing null handling, Commit 2 fixes resource leaks where Closeable classes are improperly used in Try With Resources.

### DIFF
--- a/MekHQ/src/mekhq/campaign/universe/StarUtil.java
+++ b/MekHQ/src/mekhq/campaign/universe/StarUtil.java
@@ -624,7 +624,8 @@ public final class StarUtil {
 
     public static String getIconImage(Planet planet) {
         if (!planetIconDataLoaded) {
-            try (BufferedReader reader = new BufferedReader(new FileReader(new File("data", PLANET_ICON_DATA_FILE)))) {
+            try (FileReader fr = new FileReader(new File("data", PLANET_ICON_DATA_FILE));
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/universe/StarUtil.java
```diff
@@ -624,7 +624,8 @@ public static String getHPGClass(Integer hpg) {
 
     public static String getIconImage(Planet planet) {
         if (!planetIconDataLoaded) {
-            try (BufferedReader reader = new BufferedReader(new FileReader(new File("data", PLANET_ICON_DATA_FILE)))) {
+            try (FileReader fr = new FileReader(new File("data", PLANET_ICON_DATA_FILE));
+                 BufferedReader reader = new BufferedReader(fr)) {
```
NickAragua, line 628: How is this functionally different from the previous syntax?

NickAragua (commented):
The general idea looks good to me - one question.

MekHQ/src/mekhq/campaign/universe/StarUtil.java
Windchild292: Try with resources doesn't close inner declarations, so the new functionality actually closes the FileReader instead of letting it leak.

NickAragua (approved):

PR #3418, opened 2022-08-02
enable dynamic scenarios to use fixed MULs

NickAragua:
This pull request allows a force template to be loaded from a fixed MUL file instead of randomly generated.


MekHQ/src/mekhq/campaign/mission/ScenarioForceTemplate.java
```diff
@@ -621,6 +632,14 @@ public boolean isSubjectToRandomRemoval() {
...
         this.subjectToRandomRemoval = subjectToRandomRemoval;
     }
+    
+    public String getFixedMul() {
+        return fixedMul;
+    }
+
+    public void setFixedMul(String fixedMul) {
+        this.fixedMul = fixedMul;
+    }
```
Windchild292, line 642: Missing Nullable annotations

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -244,16 +246,55 @@ private static int generateForces(AtBDynamicScenario scenario, AtBContract contr
             effectiveUnitCount = calculateEffectiveUnitCount(scenario, campaign);
 
             for (ScenarioForceTemplate forceTemplate : currentForceTemplates) {
-                generatedLanceCount += generateForce(scenario, contract, campaign,
+                if (forceTemplate.getGenerationMethod() == ForceGenerationMethod.FixedMUL.ordinal()) {
+                    generatedLanceCount += generateFixedForce(scenario, contract, campaign, forceTemplate);
+                } else {
```
Windchild292, line 251: This should also null check ScenarioForceTemplate::getFixedMul, or the field needs to be validated to ensure it is not null

Windchild292 (changes requested):
The example is in the `fixedmuls` directory, while the constant uses `fixedMuls`. This will cause issues.

There's also a potential null issue as there's no validation to ensure that a fixed MUL isn't null. There's also no validation that the file actually exists.

PR #3426, opened 2022-08-08
Fix set quality npe

NickAragua:
Pretty straight forward - when you click 'cancel' on the 'gm -> set unit quality' dialog, it used to throw an exception.


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -181,6 +181,12 @@ public void actionPerformed(ActionEvent action) {
             String quality = (String) JOptionPane.showInputDialog(gui.getFrame(),
                     "Choose the new quality level", "Set Quality",
                     JOptionPane.PLAIN_MESSAGE, null, possibilities, "F");
+            
+            // showInputDialog returns null when cancel is clicked, so we set a default here
+            if(quality == null) {
```
Windchild292, line 186:
```suggestion
            if (quality == null) {
```

Windchild292 (approved):

PR #3459, opened 2022-09-18
Fixed Mul Scenarios added.  

Thom293:
Fixed Forces that can randomly generate as a battle in a Stratcon campaign. They have been tested and they work.  There are no code changes, just xml editing and some portraits and customs added.

This takes advantage of the new .mul battle system that Nick graciously added.

There are 3 fixed battles added.  1)  A combined arms unit of 1 mech company, 1 hover company, 4 ASF and some BA.  2) a command company of mechs, and 3) a single medium lance encounter.  They are unique "boss fights" of upgraded difficulty.  I do have two more to add from Phoenix, but they had some date inconsistencies which we are working to sort out.

The forces have portraits, ranks, callsigns, custom mechs, SPAs, Edge, and specialty ammo.  All forces are 100% canon according default MHQ settings.

They have been added to the scenario manifest and can generate as a random battle in a stratcon campaign.  New files are: OpFor-specific customs folder, portraits folder, the muls, the scenarios, and an AtB docs folder with documentation and screenshots.  All of the portraits and camo are unique and created by me (so no copyright issues) and are SFW. 

Each scenario has detailed instructions and warnings about difficulty and year.  Each is labeled as "unique" with a suggestion to delete and reroll if it is rolled a second time.  If course, like anything in MHQ, players can ignore them completely or play them more than once, as they see fit.  I have a number of other scenarios planned, but I need feedback first. 

For now, they work as designed, within the current limitations of the system.  In the future, it should be possible to automate more of the process, including year sorting, frequency, and the ability to pull from a list of mul forces, instead of the same one each time.  I plan to move them to a "Big Battle" type rarity event eventually, but for testing I have added them to the general pool because they will be more frequent.  That will be the next phase I will work on.


Thom293:
UPDATE:  I realized after I pushed this that because of the removal of the unofficial mechs folder, I had to add a one of the unofficial units to the customs folder.

Also updated the scenario objectives for the two large battles as it still listed The Bluecoats.

Also I added 2 new scenario modifiers which I created some time ago but never pushed.  Since it is related to this subject, I just pushed them now.  Elites - which is a 1 level upgrade of Veterans.  And Superior Equipment - which is a 1 level upgrade of Good Equipment.

HammerGS:
Hi, thanks for doing these up. But we are avoiding adding non-canon things to the releases which these unfortunately fit into. BUT! we'll look at adding them to the extra's downloads for MekHQ and going forward any more made will be added there. 

As such can you arrange to do up a zip file of them and attach here.

Thom293:
Yes.  But let me clarify.  This is a completely canon "random" mercenary force created under the canon rules as they exist in MHQ default settings.  I set out with the intention of keeping them completely canon.  Unless I messed something up, it is rules and year appropriate.  If I did mess something up,  I can fix it.  If the custom mechs are a problem, I can remove them.  But even they are canon rules creations and 2 of the 3 are in the current live MHQ unoffical list.  Canon Merc units can and will have rules-appropriate custom mechs.  Essentially this force could be generated randomly (except for the custom mechs) by the current MHQ for 3066.  They just have some human organization behind them.  This just removes the randomness and adds some order to the OpFor.  Nothing else.

I was eventually going to make them a rare-chance monthly random battle (Like a Big Battle or Ace Duel) to add some variety and challenge.  But for testing purposes they are just in the normal list.  

Anyway, if you still dont want to include them, that is fine.  Please let me know.

Thom293:
x

PR #3482, opened 2022-10-21
Infantry refactor adaptation from MM

SJuliez:
Requires MM #3944. Only changes method names as required.


NickAragua (approved):
Looks fine.

PR #3511, opened 2022-11-23
Contract Market: Adding Transit Time and Estimated Profit Columns

Windchild292:
The dialog's code needs a rewrite, so I just worked around it for now to add the functionality.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/ContractMarketDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/ContractMarketDialog.java
@@ -22,14 +22,18 @@ package mekhq.gui.dialog;
 
 import megamek.client.ui.preferences.*;
 import megamek.common.util.EncodeControl;
+import megamek.common.util.sorter.NaturalOrderComparator;
 import mekhq.MekHQ;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/ContractMarketDialog.java
```diff
@@ -242,11 +244,21 @@ private void initComponents() {
                 row.add("");
                 row.add(c.getType());
             }
+
+            final JumpPath path = campaign.calculateJumpPath(campaign.getCurrentSystem(), c.getSystem());
```
NickAragua, line 248: I feel like we already calculate the jump path for a contract at some point. Given that it's a fairly involved operation, maybe we should just pre-record this data at the time the contract is created instead?

Windchild292: That's probably a good idea, but definitely beyond the scope of this PR. I'll open an issue for it once this is merged, and probably handle it pretty quickly after.

NickAragua (approved):

PR #3538, opened 2022-12-17
Data Loading: Don't Override the Selected Date

Windchild292:
Found this while working on #3533


SJuliez (approved):
If presets were things like "Pirate Campaign in the 3080ies" a date would make sense. But I suppose presets are more a bunch of technical settings here.

Windchild292:
They still contain the date, it is just used earlier in the flow where a user can select and modify the date. This was overriding the user's input date.

PR #3539, opened 2022-12-18
3389: Personnel Table: Sort Age Using Birthdate

### LINKED ISSUE #3389: [RFE] [0.49.7] When sorting by Age, use birthdays to maintain consistent sorting
### Environment
MekHQ 0.49.7

### Description
When 'age' is one of the sorting keys, the real birthday should be getting used. This will prevent the sort order from changing when, all other sort keys are equal, a character's age increases.

I have had multiple occasions where, when I advance the day, my "age-sorted" list changes order because someone that was previously "younger" is now "the same age" as other people and, because the ages are the same, the sort key falls on to secondary keys.

Windchild292:
This fixes #3389

### DIFF
--- a/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
+++ b/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2021-2022 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
@@ -391,7 +391,8 @@ public enum PersonnelTableModelColumn {
             case CALLSIGN:
                 return person.getCallsign();
             case AGE:
-                return Integer.toString(person.getAge(campaign.getLocalDate()));
+            case BIRTHDAY:
+                return MekHQ.getMHQOptions().getDisplayFormattedDate(person.getBirthday());
             case PERSONNEL_STATUS:
                 return person.getStatus().toString();
             case GENDER:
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
```diff
@@ -871,14 +879,13 @@ public Comparator<?> getComparator(final Campaign campaign) {
...
-            case SPA_COUNT:
-            case IMPLANT_COUNT:
-                return new IntegerStringSorter();
+            case BIRTHDAY:
+            case RECRUITMENT_DATE:
+            case LAST_RANK_CHANGE_DATE:
+            case DUE_DATE:
+            case RETIREMENT_DATE:
+            case DEATH_DATE:
+                return new DateStringComparator();
```
SJuliez, line 888: I'm amazed that there would be a date string comparator that can deal with different sorts of date formatting. Thinking of European DDMMYYYY and American MMDDYYYY here...
Since the cell renderer replaces the getValueAt() cell value with a display string anyway it would be possible to have getValueAt() return the date itself to have an easier time sorting it. Just a thought, no real reason to change anything here.

SJuliez (approved):

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
Windchild292: MHQ has two date format options (short and long), and the display will always show it in the user's specified short format. That makes it rather easy to sort as we know the format it will use.

I don't want to make any changes to the table's data formats, so I'm going to leave it as a string instead of converting the setup to allow objects instead.

PR #3587, opened 2023-01-26
(DRAFT) Finishing Profession Unit Testing

Windchild292:
This is the last of the personnel enums to do


SJuliez (approved):
This is to not let this stew again.
I have about 0% XP on unit tests but I think I read you can prepend initialization to every test using something like @beforeall on an init method. The start of the tests looks very similar, maybe this is something to simplify, for what it's worth.

Windchild292:
The minor differences required for the tests unfortunately make that impossible. Also, this is still a draft

PR #3615, opened 2023-02-26
3605: News Properly Shows in Advance Days Dialog

### LINKED ISSUE #3605: Allow news blurbs links to appear when using "Advance Multiple Days"
When working on adding more news to make the game more lively, I noticed that the Advance Multiple Days menu ignores all the news, so if you advance the timeline, you don't get to see any of those.

The news do appear on the daily log as the days advance, so it's just missing them appearing on the advance days window.

Windchild292:
This fixes #3605

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2974,6 +2974,7 @@ public class Campaign implements ITechManager {
         for (NewsItem article : news.fetchNewsFor(getLocalDate())) {
             addReport(article.getHeadlineForReport());
         }
+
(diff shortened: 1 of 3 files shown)


rjhancock (approved):

SJuliez (approved):
I admit I don't understand the bug nor the solution. It looks like all reports would be added twice. (But obviously the reports werent added at all before) But then, I don't know where currentReportHTML is ever filled, it's only emptied in newDay(), so where does the report from line 298 come from? I know too little of this code :(

Windchild292:
I don't understand how reports are coded either, but this works so I'm fine with it for now.

Issue #2455, opened 2021-03-07
GEnerated scenario with thin atmosphere and fog


Tecmes:
According to TO57 (Light Fog section and others). this is not possible. 
I know this might be an AtB issue, not MekHQ, but I'm not sure.

Tecmes:
(this is the same, asyou may guess, with other incompatible conditions, such as snowfall+thin)

BullseyeSmith:
This reminds me of a Base Attack (Defender) battle I once had.  The weather said clear.  I had to keep all the civilians and allied vees & turrets alive.  Got all ready and launched the scenario.  First round everything (except mechs) dies.  I went back and looked again.  Weather: Clear. Wind: F4 Hurricane.  I just, I couldn't, I still can't.  GM deleted the scenario and pretended it never happened.

Windchild292:
Working as intended, AtB rules allow this.

HammerGS:
I disagree, with closing this. AtB rules especially for things like weather and planet conditions shouldn't be superseded by AtB.

NickAragua:
The specific combinations of "thin atmosphere" and some wind conditions is actually prohibited, so we should review that logic when we re-visit the weather tables for StratCon.

PR #3637, opened 2023-03-30
Enable weather restrictions/temperature passing to megamek

NickAragua:
Addresses #2455 by using the companion MegaMek pull request's weather restriction mechanism to filter out invalid weather/atmo/temperature combinations.

Also displays and passes the scenario temperature (currently pegged to the planetary temperature) to megamek when planetary conditions are enabled.

### DIFF
--- a/MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
@@ -79,6 +79,10 @@ public class AtBScenarioViewPanel extends JScrollablePanel {
     private JLabel lblWindDesc = new JLabel();
     private JLabel lblFog = new JLabel();
     private JLabel lblFogDesc = new JLabel();
+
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
```diff
@@ -174,7 +178,7 @@ private void initComponents() {
     }
 
     private void fillStats() {
-        ResourceBundle resourceMap = ResourceBundle.getBundle("mekhq.resources.AtBScenarioViewPanel",
+        ResourceBundle resourceMap = ResourceBundle.getBundle("mekhq.resources.ScenarioViewPanel",
```
NickAragua, line 181: The AtBScenarioViewPanel resource bundle is entirely a subset of the ScenarioViewPanel resource bundle. It seemed redundant, so I just used the latter.

NickAragua:
Build errors are due to having a dependency on a megamek PR that has since been merged

Issue #3652, opened 2023-04-27
IndexOutOfBoundsException Loading Save with Prison Break Special Scenario


cjdhein:
### Environment
What version of MekHQ does your issue apply to? Latest and Greatest (running off of master branch)
What operating system are you using? Mac
What java version are you using?  JDK 11.0.18

Getting an IndexOutOfBoundsException when loading save that has a Prison Break Special Scenario.

The save's XML has entries in the specialScenarioEnemies node for classes 0 through 5. The issue appears to be within AtBScenario.java's loadFieldsFromXmlNode function. It only creates 4 ArrayLists to hold entities in, so anything in the save for class 4 or 5 is never loaded (line 1672).

Unless I am missing something (entirely possible) - I believe that instead of:
`for (int i = 0; i < 4; i++)`
It should follow the same logic as we see in PrisonBreakBuiltInScenario.java:93 where we use the actual defined values for weight classes:
`for (int weight = EntityWeightClass.WEIGHT_ULTRA_LIGHT; weight <= EntityWeightClass.WEIGHT_COLOSSAL; weight++)`

Or at least the 4 should be 6. I tested the change locally and the scenario loads with no error.

[mekhq.log](https://github.com/MegaMek/mekhq/files/11347202/mekhq.log)
[bwr.cpnx.gz](https://github.com/MegaMek/mekhq/files/11347206/bwr.cpnx.gz)

Issue #3629, opened 2023-03-23
IndexOutofBoundsException - MekHQ after deploying single unit for special event


3Nugget:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
What version of MekHQ does your issue apply to?
0-49.12
What operating system are you using?  Windows, Mac, or Linux?
Windows 10 Home
What java version are you using?  If unsure, it will state your java version at the top of the mekhq.log file.
19.0.23
(Delete the above lines)

### Description
Describe the issue and provide screenshots ingame if able.
After selecting the following pilot for the Special Scenario: Ambush then received the Uncaught Java.lang.IndexOutofBoundsException . When going to the Briefing Room tab and selecting the same mission, the error repeats. May be that the mission they are being deployed to is outside the mission array Index, but that is just an educated guess.
[logs.zip](https://github.com/MegaMek/mekhq/files/11054046/logs.zip)
![image](https://user-images.githubusercontent.com/20347476/227293636-7794ff3c-98cd-4793-8cd5-11b93f81f03c.png)
![image](https://user-images.githubusercontent.com/20347476/227293948-4f23c7e3-474f-460d-95fd-ea42d9bd9c6b.png)

(Delete the above lines)

### Files
Provide the .cpnx campaign file that this issue occurs in.
[WrongWays Rough Riders30280108.cpnx.gz](https://github.com/MegaMek/mekhq/files/11054066/WrongWays.Rough.Riders30280108.cpnx.gz)

I am on the discord if further questions are needed, cand can be reached at @Yonkers#1362

PR #3657, opened 2023-05-01
Fix for IndexOutOfBoundsException in AtBScenario's loadFieldsFromXmlNode when loading a save with a special scenario

cjdhein:
Should allow for all weight classes of potential enemy forces to be loaded from the XML.


cjdhein:
This fix should solve #3652  and #3629

NickAragua (approved):

PR #3663, opened 2023-05-09
Fix CASE refit classes

Saklad5:
Adding CASE is a Class D refit. Removing CASE is not, so simply changing it does not necessarily make a refit Class D.

In addition, the current designation of Class E is never applicable.


Windchild292 (approved):

Saklad5:
I briefly tested this by opening a campaign save: it seems to work as expected.

- Adding CASE II to Clan 'Mechs is Class B (assuming a slot was removed too)
- Adding CASE II to IS 'Mechs that previously had CASE is Class B (always, since it replaces the CASE)
- Adding CASE *or* CASE II to IS 'Mechs in a location that previously had neither is Class D
- Removing CASE from IS 'Mechs is Class A

I only tested 'Mechs, admittedly, but given how CASE works in other units that shouldn't be relevant.

PR #3712, opened 2023-07-06
Fix Eclipse IDE Multiple Module dependency error - take 2.

HoneySkull:
Follow fix to MekHQ 'Multiple Module Dependency' errors when attempting to import MekHQ into Eclipse IDE.

This fix excludes the `xml-api` library from the Megameklab jar.  Additionally, it sets the `xml-api-ext` to be included in the runtime build as it is needed by Megameklab's xml graphics library and is not included in the Java Runtime library.

This gradle configuration allows MekHQ to be independently imported into the Eclipse IDE using Gradle integration while not breaking the scaled vector graphics functions provided by Megameklab.  (i.e. when printing record sheets while embarking on a scenario in the Briefing Room tab.

### DIFF
--- a/MekHQ/build.gradle
+++ b/MekHQ/build.gradle
@@ -37,8 +37,8 @@ dependencies {
         // We don't need the python and javascript engine taking up space
         exclude group: 'org.python', module: 'jython'
         exclude group: 'org.mozilla', module: 'rhino'
-        // Uncomment to work around Eclipse IDE Multiple Dependency errors. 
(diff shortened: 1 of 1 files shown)


NickAragua (approved):

HammerGS:
Going to merge this for testing with nightly builds.

HammerGS:
This seems to be working for MekHQ as well. Some of our QA people are testing and have had no issues

PR #3720, opened 2023-07-09
[Gradle] Add explicit dependencies on MM/MML (might not be correct)

sixlettervariables:
Trying to remove the transient nature of the build failures with GitHub actions and it seems like MM/MML do not consistently build before MekHQ, though I'm not sure why Gradle isn't detecting these dependencies when it works "fine" from my IDE.

This change consistently produces a deterministic build order between the three projects, but I'm not sure if it matters.

Thoughts?


sixlettervariables:
Nah this is wrong...

Issue #3759, opened 2023-09-13
Add weight to spare 'Mech locations


Saklad5:
### Environment
MegaMek Version: MekHQ v0.49.14 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description
'Mech locations (such as arms, legs, heads, center torsos, etcetera) currently have a weight of 0 in MekHQ. _Campaign Operations_ specifies the weight of each part on page 213 of the third printing:

> Each limb is considered to weigh 10 percent of the total internal structure weight. The center torso weighs 25 percent, the left and right torso weigh 15 percent each, and the head weighs 5 percent.

While provided in the context of FrankenMech construction, this would logically apply to spare parts as well.

Saklad5:
These weights apply to all 'Mechs, including tripods: their internal structure simply has 10% more weight due to having a fifth limb.

NickAragua:
Implemented with #3766

PR #3766, opened 2023-10-04
weight calculation for spare mech locations

NickAragua:
What it says on the tin. Requires MegaMek PR #4818. Implements #3759.


NickAragua:
It's pretty unlikely that we'll be able to easily implement tripod left/right leg weight calculation - there's absolutely no way to tell whether you're buying a left leg for a biped or a tripod.

gsparks3:
Aren't they the same weight anyway? For standard mechs you have 4 limbs at 10% each, side torsos at 15%, CT at 25%, and head at 5%, adding up to 100% structure weight. Tripods have 5 limbs, increasing the total fraction to 110%, exactly matching their increased structural weight during construction (i.e. the extra structure weight is entirely assigned to the extra leg).

PR #3771, opened 2023-10-24
Childrens filter in personnel screen

nutritiousemployee:
This is a simple added filter that shows any personnel whose age is lower than 13.
The aim is to have an easier way to sort out children without having to display the dates view, and order by birthdate.

The change for PersonnelFilter is in the filters list (row 80), and in the getFilteredInformation method. I had to pass the current date in the method as I couldn't find a way to retrieve that info from inside the method's switch.
The PersonnelTab only change should be on row 283, but the automated diff shows space changes in the whole file - not sure why as I'm using the suggested IDE settings.

Having changed getFilteredInformation there's two changes in PersonnelMarketDialog and RetirementDefectionDialog since they were calling the method as well. I believe these changes don't impact the functionality of the two classes.


MekHQ/src/mekhq/gui/PersonnelTab.java
```diff
@@ -51,377 +51,377 @@
...
+    changePersonnelView();
+  }
+
+  public void filterPersonnel() {
+    final PersonnelFilter filter = (choicePerson.getSelectedItem() == null)
+        ? PersonnelFilter.ACTIVE : choicePerson.getSelectedItem();
+    personnelSorter.setRowFilter(new RowFilter<>() {
+      @Override
+      public boolean include(Entry<? extends PersonnelTableModel, ? extends Integer> entry) {
+        return filter.getFilteredInformation(entry.getModel().getPerson(entry.getIdentifier()),getCampaignGui().getCampaign().getLocalDate());
```
NickAragua, line 283: Style change request: spaces between parameters in parameter lists for readability.

MekHQ/src/mekhq/gui/dialog/PersonnelMarketDialog.java
```diff
@@ -419,7 +419,7 @@ private void filterPersonnel() {
         sorter.setRowFilter(new RowFilter<>() {
             @Override
             public boolean include(Entry<? extends PersonnelTableModel, ? extends Integer> entry) {
-                return nGroup.getFilteredInformation(entry.getModel().getPerson(entry.getIdentifier()));
+                return nGroup.getFilteredInformation(entry.getModel().getPerson(entry.getIdentifier()),hqView.getCampaign().getLocalDate());
```
NickAragua, line 422: Style change request: spaces between parameters in parameter lists for readability.

MekHQ/src/mekhq/gui/dialog/RetirementDefectionDialog.java
```diff
@@ -590,7 +590,7 @@ public boolean include(Entry<? extends RetirementTableModel, ? extends Integer>
                         && !rdTracker.getRetirees(contract).contains(person.getId())) {
                     return false;
                 } else {
-                    return nGroup.getFilteredInformation(person);
+                    return nGroup.getFilteredInformation(person,hqView.getCampaign().getLocalDate());
```
NickAragua, line 593: Style change request: spaces between parameters in parameter lists for readability.

NickAragua (changes requested):
Looks good, just needs some whitespace changes for readability.

NickAragua (changes requested):
Sorry, one other thing - looks like you've changed all the tabs from 4 to 2 in PersonnelTab and PersonnelFilter, which we don't really want. Let's revert those, and then we'll be good to go.

nutritiousemployee:
This should be ready now.
There was also an extra space in PersonnelFilter for line 65 which I'm not sure if it was introduced by me by mistake or no, but it's gone either way

NickAragua (approved):
Looks good.

PR #3779, opened 2023-11-08
Made possible to set custom date for non AtB scenarios

p1mps:
Not sure if this is the right way to do it but as of now it's not possible to set a date for a custom scenario. What I did was to create a new ScenarioStatus so when the "Add Scenario" button is clicked a scenario that is not "current" is created (if you add a scenario and you want to set a custom date it cant be current). Maybe a bit more validation on the date should be done but I'm not an expert of the code base :)

### DIFF
--- a/MekHQ/resources/mekhq/resources/Mission_en_US.properties
+++ b/MekHQ/resources/mekhq/resources/Mission_en_US.properties
@@ -1,0 +1,2 @@
+MissionStatus.NEW.text=New
+MissionStatus.NEW.toolTipText=This mission/contract hasn't started yet
--- a/MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
@@ -88,6 +88,9 @@ public class CustomizeScenarioDialog extends JDialog {
             scenario = s;
             newScenario = false;
         }
+        if (newScenario) {
+            scenario.setStatus(ScenarioStatus.NEW);
+        }
         campaign = c;
         if (scenario.getDate() == null) {
             scenario.setDate(campaign.getLocalDate());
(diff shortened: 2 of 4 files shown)


AaronGullickson:
@p1mps Coming back to this because its relevant for story arcs. The way this works by default is that a scenario is not given a date until after it is completed. What would be the purpose of assigning the date before the scenario is resolved?

Thom293:
My belief is so that you can assign a scenario a date in the future.  As it is now it only assigns it the day you click the add button.  Would be useful for a number of reasons like SC only assigns scenarios on a weekly basis.  But if you could set it 1 month into the future you could have "You will be hot dropping the capital on this date and you need to have killed the Aux forces first and prepare yourself" etc.  I think that is what was discussed on discord when he posted this.

AaronGullickson:
Ok, got it. I think there are some issues with the code here though. It doesn't actually give the user the ability to set a custom date.

MekHQ/resources/mekhq/resources/Mission_en_US.properties
```diff
@@ -0,0 +1,2 @@
+MissionStatus.NEW.text=New
+MissionStatus.NEW.toolTipText=This mission/contract hasn't started yet
```
AaronGullickson, line 2: I don't see where this information is used.

MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
```diff
@@ -88,6 +88,9 @@ public CustomizeScenarioDialog(JFrame parent, boolean modal, Scenario s, Mission
             scenario = s;
             newScenario = false;
         }
+        if (newScenario) {
+            scenario.setStatus(ScenarioStatus.NEW);
+        }
         campaign = c;
         if (scenario.getDate() == null) {
             scenario.setDate(campaign.getLocalDate());
```
AaronGullickson, line 96: This will give status "Pending" to all new scenarios but the current date will still be assigned. Functionality would need to be added to the dialog to allow the user to pick a date. Furthermore, what is the distinction no between "Pending" and "Current?" Is the intent of "Current" that we are on the actual day the scenario should happen? If so, we would also need code that automatically switched from pending to current where appropriate  on newDay. The code would also need to check when executing the dialog to determine if its current or pending.

AaronGullickson:
The code in PR #3930 will now allow users to set the date of any scenario through the CustomizeScenarioDialog, which makes this PR redundant, so I am closing.

PR #3788, opened 2023-11-25
Adjust campaign creation dialogs to have correct jdialod owner for re…

gcoopercos:
Motivation:
In MekHq, the campaign creation user interface wasn't showing up.   After tracking this down, the modal campaign dialog was coming up behind the original dialog that spawned it.

Without rearranging things too much, a straight-forward fix appeared to be to make sure that the modal dialog had the owner of the JDialog that created it, not the owning JFrame.

I am guessing that this works ok on some platforms so it's not an issue for some. I'm also guessing that this is the case because which JDialog is rendered in front is arbitrary if they have the same JFrame. Regardless, this feels like a fairly benign change.

Change Description:

This is a companion PR to the one I just threw up for Megamek. That would have to go in first as this change depends on the function signature being there for the abstract dialog classes. 

These changes are isolated to creating the MekHQ Campaign dialogs with their proper JDialog owners so they don't show up behind the other dialogs.

(Please forgive if this doesn't follow the right standards, but we can work that out in the PRs if that works)

### DIFF
--- a/MekHQ/src/mekhq/gui/baseComponents/AbstractMHQButtonDialog.java
+++ b/MekHQ/src/mekhq/gui/baseComponents/AbstractMHQButtonDialog.java
@@ -28,9 +28,9 @@ import java.util.ResourceBundle;
  * This is the Base Dialog for a dialog with buttons in MegaMek. It extends Base Dialog, and adds a
  * button panel with base Ok and Cancel buttons. It also includes an enum tracker for the result of
  * the dialog.
- *
+ * <p>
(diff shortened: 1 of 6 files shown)


gcoopercos:
This is going to be dependent on:
https://github.com/MegaMek/megamek/pull/4923

SJuliez (approved):

HammerGS:
Thanks for the fixes, sorry it took so long to merge this one.

PR #3800, opened 2023-12-17
User files directory setting in MHQ

SJuliez:
This adds a setting to the misc MekHQ settings for the user directory of MegaMek/megamek#4961

Requires MegaMek/megamek#4961

As per requests, awards and rank systems are now read from the user directory.


HammerGS:
This looks like it was missed with the other User directory merges.

PR #3801, opened 2023-12-18
New StratCon Scenario Modifiers

Thom293:
This PR adds 23 StratCon modifiers that have been discussed in Discord and are a combination of ideas from various of our regulars.  Most "fill holes" for opfor types that we dont have. (EDIT: Added some beast infantry).  There were other suggestions too, but those will require a little more work.  These use the fixed mul system, so they are an attempt for Princess to sometimes get an actual organized force instead of just random units.  They are all canon introtech units that are always available.  Stock varients with no advanced tech.  I have tested them all and I believe they are balanced. If my math is right, there is approximately a 1/74 chance to draw any of them, or a 15/74 chance to draw one of the negative ones, so it is possible to go an entire contract and not draw any. (I did multiple times in testing, and I had their frequency quadrupled for testing purposes.  I "clicked through" about 20 6 month contracts to test their frequency and even with it quadrupled, they were suitably rare)  Played two full 6 month contracts with them and saw 2 per contract.

Note all units mentioned below are introtech versions and always available.  The modifiers added in this PR are:

Allied Officer Mech (1) - a generic randomly rolled Allied Mech Officer in a medium or heavy mech.  Added as a random modifier and a shared modifier from the Allied Base of Operations (the missing mirror of the Enemy Officer Mech added previously).

Allied Cavalry Lance with vet pilots (1) - 60 ton fast heavies that reinforce on turn 3.  Ostsol, Ostroc, Dragon, Champion.

Allied Horse Cavalry (1) - Allied beast infantry.  Into the valley of death rode the 600...

Allied Mech Ace with vet pilot (5) - a single unit, of different types (only one of these will appear): Ostscout, Trebuchet, Crusader, Guillotine, Goliath

Enemy Mech Ace with vet pilot (7) - a single unit, of different types (only one of these will appear): Firestarter, Phoenix Hawk, Grasshopper, Archer, Marauder, Awesome, Highlander

Enemy Air Cav (1) - Warrior VTOLs of two types that reinforce on turn 3.  

Enemy Air Mobile Infantry (1) - Strike Falcon VTOLs carrying Basic and Jump infantry that reinforce on turn 3. 

Enemy Beast Infantry (2) - Branth and Kangaroos with freaking laser beams that reinforce on turn 3.

Enemy Mercenary Themed Lances with vet pilots (4):

  Scouts - Phoenix Hawk, Vulcan, Firestarter, Javelin
  Skirmishers - Wolverine, Shadowhawk, Griffin, Phoenix Hawk
  Brawlers - Highlander, Orion, Flashman, Guillotine
  Fire Support - Archer, Awesome, Crusader, Thunderbolt

Enemy Veteran Pilots have 1-2 Edge and 1-2 SPAs (Most have 1), so if those options are enabled the pilots will have them.  Some of the Allied Veteran Pilots have 3 SPAs, since they are a boon to the player, and cannot be captured.  

If it helps, I have played with the enemy aces on my personal build for months, and they are a good addition.  I have combat tested against the VTOL and Merc Lances and fought both multiple times with a 3025 Company with only 4 veteran pilots in the whole company and no advanced tech.  These new modifiers add difficulty when you roll them, but are not overbearing or punishing or too common.  Because there is no way to sort muls by date and these are all introtech, during Post-Clan eras these will be trivialized.  But an Awesome-8Q with a good pilot is nothing to sneeze at in any era.

Last - it also moves the previous modifiers that were added from the "usermodifiermanifest" to the regular modifier manifest.  They should not have been put in the other file in the first place.  

I have tested all of them and they should work with no bugs (fingers crossed).  Let me know if any questions.  Thank you.


HammerGS:
I think these updates, but wondering if using fixed MUL's is the way to go. I mean in a 3150 game those base units are gonna get chewed up awful fast. 

I've barely dug into the scenario editor is there a way to generate using the RAT Generator?

JonhyBurner:
No way to choose RAT or even a year. You can use fixed MULs for forces like this, but then they are static.

I think ideally, the scenario list and the modifiers should have conditions and nested categories, right now they are simply randomly taken from a full list. If that was to be done, players could build several MULs of lances to be deployed according to year or faction. But that is a problem for the future. As it is, this is actually a good addition.

Just wondering what happens if a player is in a campaign before mechs exist in the timeline. I assume the mechs will appear anyway?

Thom293:
Yes.  So, with the current coding, there is no way to sort fixed MUL by year.  I do have a separate RFE for exactly that, but its not in the game at the moment.  But if it was, you make 1 mul per era, and then if you draw the "Fire Support" modifier it would pull the era appropriate one.  But as it is now the best option is to use "timeless" intro tech mechs that are always somewhat decent, whcih is what I tried to do.  (Nor is there a way to have the RAT generate a force with only the "Sniper" or "Brawler" role, etc  - although that would be cool too).

But yes in later timelines, these are going to be not as threatening.  But still, an Awesome with a 2/3 pilot that has Sniper SPA or an Archer that has Oblique Attacker SPA and all of them also having edge is still more of a threat than some randomly generated 4/5 pilot in a 3150 mech.

And yes on the early - pre-mech timeline - which it is my understanding is not supported because the dearth of RATs - the player would have to manully remove these.  But that is already the case with other modifiers too.  

I guess ultimately, these modifiers are good in some eras and a bit weaker in others - but in all cases they are at least equal or better than some randomly generated pilot in a similar bv mech.  i.e. they should not ever make the enemy force weaker by having them.

PR #3810, opened 2024-01-09
MekHQ part of fix for #4648

Sleet01:
Mek Mortars and IS BA Tube Artillery are excluded from the MekHQ "Spend XP" SPA menu because they have an AT2 Class of "NONE".  By rights this whole AT2 class thing should be torn out, but it seemed more sensible to massage the existing check and add classes to the missing weapons, rather than refactoring a lot of code across all three projects.

This patch restricts the range of AT classes that will be rejected to just the existing Naval / Large Ship weapons.
The full fix also relies on the MegaMek PR [here](https://github.com/MegaMek/megamek/pull/5027).

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/SpecialAbility.java
+++ b/MekHQ/src/mekhq/campaign/personnel/SpecialAbility.java
@@ -551,9 +551,10 @@ public class SpecialAbility {
             return false;
         }
 
+        // Should only apply to Large Ship-mounted weapons
         if (wt.getAtClass() == WeaponType.CLASS_NONE ||
                 wt.getAtClass() == WeaponType.CLASS_POINT_DEFENSE ||
-                wt.getAtClass() >= WeaponType.CLASS_CAPITAL_LASER) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/SpecialAbility.java
```diff
@@ -551,9 +551,10 @@ public static boolean isWeaponEligibleForSPA(EquipmentType et, PersonnelRole rol
             return false;
         }
 
+        // Should only apply to Large Ship-mounted weapons
         if (wt.getAtClass() == WeaponType.CLASS_NONE ||
                 wt.getAtClass() == WeaponType.CLASS_POINT_DEFENSE ||
-                wt.getAtClass() >= WeaponType.CLASS_CAPITAL_LASER) {
+                (wt.getAtClass() >= WeaponType.CLASS_CAPITAL_LASER && wt.getAtClass() <= WeaponType.CLASS_TELE_MISSILE)) {
```
NickAragua, line 557: <= ?

Sleet01: > <= ?

Yep.  That span of AT Class definitions all deal with Capital / Large Ship stuff; after `CLASS_TELE_MISSILE` there are a couple of newer classes that shouldn't be excluded anymore.

PR #3822, opened 2024-01-29
#3194 Awarding non-stackable medals to multiple people

### LINKED ISSUE #3194: RFE: Awarding non-stackable medals to multiple people should award to those that do not already have the medal
### Environment
MekHQ 0.49.7

### Description
When you are awarding non-stackable medals (such as Marksmanship, Galactic War on Pirating, etc) to a unit, the medal does not show up in the list of awards if _anyone_ in the selection has already received the medal. This means, if you have fought pirates, then had multiple personnel changes, and then fought pirates again, you need to go through and check each person's log to find the ones that have NOT received the medal so that you can award them the medal.

###Request for Enhancement
If you select a group of people in order to award them a medal, show all the medals. If the selected medal is non-stackable, only award it to those personnel that do NOT already have it.

jackreichelt:
This fixes #3194. Allows selecting multiple personnel and awarding a non-stackable award, which will only be added to those who do not already have it. When a single person is selected, the awards list will still be filtered as normal.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/PersonAwardController.java
+++ b/MekHQ/src/mekhq/campaign/personnel/PersonAwardController.java
@@ -108,12 +108,18 @@ public class PersonAwardController {
         final Award award;
         if (hasAward(setName, awardName)) {
             award = getAward(setName, awardName);
+
+            if (!award.canBeAwarded(person)) {
+                LogManager.getLogger().info("Award not stackable, returning.");
+                return;
+            }
         } else {
             award = AwardsFactory.getInstance().generateNew(setName, awardName);
             if (award == null) {
                 LogManager.getLogger().error("Cannot award a null award, returning.");
                 return;
             }
+
             awards.add(award);
         }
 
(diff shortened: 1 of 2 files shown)


jackreichelt:
Potentially this should still hide a non-stackable award from the list if everyone already has it. Can definitely implement that, but I'd love some feedback on if that's better UX. I am uncertain.

SJuliez (approved):

JonhyBurner:
Since single personnel awards menu hides the non-stackable that person already has, hiding it when choosing multiple personnel when everyone has it would be the expected behaviour for a user. In practice, it makes no difference, though.

jackreichelt:
After sleeping on it, I think that not hiding the award if all of a selected group is better. My rationale is that if you do hide it, people have to scroll, not see it, then double-check they haven't missed it. The list is quite long, so I think it's easy to miss things.

If you don't hide it, then when you apply it to a group, it's a simple process, with a guarantee at the end that all the selected personnel have the award (even if they had it all already).

A potential future change might be to merely disable an award in the menu, but still show it. This just helps preserve the list contents regardless of state/selection, which might make it more intuitive. But I think that's a very minor improvement, that might not be worth it.

rjhancock (commented):
Changes look sane.

PR #3825, opened 2024-01-31
#3740 Standardising on C-Bills

### LINKED ISSUE #3740: Consistent messaging in Daily Activity Log: ComStar bill vs. C-Bill
In 49.12, there is a difference in messaging between prisoner ransoms:

`25,000 ComStar bill has been credited to your account for prisoner ransoms following Assassinate.`

... and battle loss compensation:
`7,290 C-Bill in battle loss compensation (parts) for Shadow Hawk SHD-2H has been credited to your account.`

There may be other cases where this is different.  I recommend settling on "C-Bills" (1 C-Bill is going to be extremely uncommon, so plural will work in almost every case) as the single term, and a general review of Activity Log messages so that it is consistent across all uses.  Possibly switching to a constant or similar value in string building.

jackreichelt:
This fixes #3740.

### DIFF
--- a/MekHQ/src/mekhq/campaign/finances/Money.java
+++ b/MekHQ/src/mekhq/campaign/finances/Money.java
@@ -154,6 +154,7 @@ public class Money implements Comparable<Money> {
         return CurrencyManager.getInstance().getUiAmountAndSymbolPrinter().print(getWrapped().toMoney(RoundingMode.HALF_EVEN));
     }
 
+    @Deprecated
     public String toAmountAndNameString() {
         return CurrencyManager.getInstance().getUiAmountAndNamePrinter().print(getWrapped().toMoney(RoundingMode.HALF_EVEN));
     }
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/finances/Money.java
```diff
@@ -154,6 +154,7 @@ public String toAmountAndSymbolString() {
         return CurrencyManager.getInstance().getUiAmountAndSymbolPrinter().print(getWrapped().toMoney(RoundingMode.HALF_EVEN));
     }
 
+    @Deprecated
```
jackreichelt, line 157: I deprecated rather than removing to not cause issues with any potential in-flight PRs.

neoancient (approved):

rjhancock (commented):
Mostly whitespace removal (no objection) and method rename with a deprecation.

PR #3834, opened 2024-02-09
update planetary conditions chance logic

kuronekochomusuke:
requires https://github.com/MegaMek/megamek/pull/5203

- show changed temperature when it has been adjusted for weather
- fix issue where rerolls not using the correct terrain when there is a track
- fix issue when passing fixed maps to megamek, with path character
- use enums for planetary  conditions
- spilt out board and terrain
  -  board - space, atmosphere, and ground  
- use config files to determine terrain
  - filter by temperature
  - highlight terrain when it is set from a track
- use config files to determine generated map templates from terrain
- use config file to set weather percentages by terrain type 
  - light - standard, dark, 
  - wind - standard, high
  - weather - standard, wet, snowy, dry
  - fog - standard, heavy
   - blowing sand - standard, heavy
   - emi - standard, high
- log validations for the TerrainConditionsOddsManifest.xml config file
        

image


HammerGS (approved):
I've been testing and playing this and everything seems to work. Going to merge to allow the tools to run.

PR #3840, opened 2024-02-13
StratCon Heavy Battles, by PhoenixHeart. 

Thom293:
This PR is from files created by @PhoenixHeart512.  I am posting them for him because he doesnt have an IDE set up.

It adds "Heavy" versions of most of the existing ground battles.  Bascially it just sets the minimum weight class for OpFor units to Heavy, instead of light.  This greatly reduces the number of OpFor units, but increases their tonnage.  So it it allows you to roll battles that are not always filled by "vehicle spam" to fill BV, which can happen often once you get an advanced force.

Members of the discord have been using these as a separate download for months.  It was suggested to add them to the base game.  So here.  

It does not overwrite the previous versions - it just adds these as extra options to roll.   Screen below.

Screenshot 2024-02-12 171649


Thom293:
These errors appear unrelated to this PR, but if I need to change anything, let me know.  Thank you.

PR #3853, opened 2024-02-23
Partial fix for failing refit tests

Sleet01:
This will let some of the refit unit tests pass; we've seen that more tests need the EquipmentTypes initialized in recent months.  Note that this does not fix the issues related to Machine Gun Ammo bins.


Sleet01:
This now fixes all currently-failing MekHQ tests.

PR #3889, opened 2024-03-18
[Balance] [AtB/StratCon] Base Attack (Defender) Objective Changes

IllianiCBT:
### Current Implementation
Currently, when the players' base comes under attack (usually due to a difficult contract) they are presented with a special scenario. The failure of which results in a Contract Defeat.

Current objectives task players with keeping allied units alive, including their own. For each allied unit that is defeated the contract Victory Points (not Scenario) are reduced by 1. As allied losses are inevitable, contract loss is almost a given, even if the player plays well.

### Modification
This patch removes the Victory Point loss from allied defeats, but leaves all other objectives untouched. These are as follows:

- Ensure that at least 3 base units (turrets/civilians) survive
- Destroy, cripple or force the withdrawal of at least 50% of opposing forces
- Ensure that at least 50% of friendly forces survive

This still leaves margin for defeat, but removes the likelihood of suffering a 'failure in victory'.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/atb/scenario/BaseAttackBuiltInScenario.java
+++ b/MekHQ/src/mekhq/campaign/mission/atb/scenario/BaseAttackBuiltInScenario.java
@@ -166,8 +166,6 @@ public class BaseAttackBuiltInScenario extends AtBScenario {
         destroyHostiles.addForce(String.format("%s%s", contract.getEnemyBotName(), SECOND_ENEMY_FORCE_SUFFIX));
         ScenarioObjective keepFriendliesAlive = CommonObjectiveFactory.getKeepFriendliesAlive(campaign, contract, this,
                 50, false);
-        ScenarioObjective keepAttachedUnitsAlive = CommonObjectiveFactory.getKeepAttachedGroundUnitsAlive(contract,
-                this);
 
         ScenarioObjective preserveBaseUnits = null;
         if (!isAttacker()) {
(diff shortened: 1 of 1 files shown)


HammerGS:
Given our crazy implementation of turrets, and based on what I see in conversations on the Discord. I wonder if turrets should be modified in someway.

IllianiCBT:
> Given our crazy implementation of turrets, and based on what I see in conversations on the Discord. I wonder if turrets should be modified in someway.

As it stands, I believe the 'turrets need to survive' victory condition exists as a way to determine whether the base itself has been destroyed.

Changing that victory condition wouldn't be hard, though. Did you have something in mind?

HammerGS:
I would suggest taking the victory conditions conversation to the Discord, and see if the community has any ideas. Full support the PR in it's current form but wondering if more expansive options should be discussed.

IllianiCBT:
Discussed it with Discord. I'm going to keep this patch as is - leaving turrets alone. With a mind to revisiting them later, once potential turret RAT changes (from Thom) have a bit of play time. My thinking is that if the turret RAT is improved, it would remove the need for the civilian units and would improve overall scenario play-feel without the need for further Victory Condition changes.

If Thom's RAT change isn't accepted, I'll do another balance PR.

HammerGS:
After the chat on Discord, I don't see any issues with the Turret RAT frequencies being tweaked.

NickAragua (approved):

PR #3901, opened 2024-03-24
add lances to the force string when sending data to megamek

kuronekochomusuke:
- add lances to the force string when sending data to megamek for bot forces

image

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -21,16 +21,17 @@ package mekhq;
 import megamek.client.Client;
 import megamek.client.bot.BotClient;
 import megamek.client.bot.princess.Princess;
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
Looks good to me.

AaronGullickson:
@kuronekochomusuke The added method `GameThread.setupBotForces` depends on all missions being AtBContracts by recasting them as such. However, this is the `GameThread` not the `AtBGameThread`. This means it will lead to errors for anybody not running AtB. That needs to be corrected. I can't merge master into Story Arcs at the moment because this issue will render it non-functional.

You could do that by trying to remove the AtB dependency here, but I would rather you just move `GameThread.setupBotForces` to AtBGameThread and revert the GameThread back to its prior state. I think there may be better ways to handle this issue for the non-AtB case.

PR #3905, opened 2024-03-25
Switch BotForce XML writer

AaronGullickson:
This PR changes the XML writer to EntityListFile.WriteEntityList. Currently that method is private so this PR should only be merged after MegaMek/megamek#5282 which makes it public and accessible. 

Currently BotForce entities are written using the MHQXML writer. However, this writer is not up to date and misses important information like a lot of the deployment variables, as well as any assigned pilot portraits. From the notes on this method, it seems like there is a push to move all of this over to EntityListfile.WriteEntityList which is up to date. I need an XML writer that is up to date because the Story Arcs PR #2997 uses all kinds of entity information to set up better scenarios, particularly deployment information.


AaronGullickson:
Whoops, I branched this off of story arcs - Ignore this will send in a new PR.

PR #3907, opened 2024-03-25
[Enhancement] Split 'Experience' Modifier

IllianiCBT:
### Original Behavior
Retirement TN is 3 + Experience + Other Modifiers.

Experience is a combination of two modifiers: unit rating, and the experience rating of the employee.

The below image shows _TN 3 + Experience Modifier: 3 [Elite Pilot + 4, Unit Rating -1] + Officer: -1_

image

### Problem
This makes it difficult for users to identify what calculations are being made when determining Retirement TN. The above image for example features a +4 TN from an Elite Employee and -1 TN from Unit Rating. However, unless you know what the modifiers for each source already are, the average user cannot break this down mentally.

### Solution
I have split the 'Experience' modifier into Unit Rating (Unit Rating) and Experience (Employee Experience Rating). Calculations are unchanged.

image

Note: the above images are from separate campaign files, which is why Unit Rating is not identical across both images.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/RetirementDefectionTracker.java
+++ b/MekHQ/src/mekhq/campaign/personnel/RetirementDefectionTracker.java
@@ -173,8 +173,9 @@ public class RetirementDefectionTracker {
             }
 
             TargetRoll target = new TargetRoll(3, "Target");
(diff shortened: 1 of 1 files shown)


IllianiCBT:
I'm going to go ahead and close this, in favor of the more comprehensive changes made by https://github.com/MegaMek/mekhq/pull/3911

PR #3909, opened 2024-03-26
[Bandaid] Added Dialog to Confirm New Campaign

IllianiCBT:
### Current Implementation

### Problem

### Solution

### Update
Since the initial introduction of this PR I have significantly changed its implementation.

When selecting 'New Campaign' users will be presented with a dialog asking them to confirm that they wish to start a new campaign. Selecting 'no' or closing the dialog does nothing. Selecting 'yes' initiates the new campaign process (this process is identical to current implementation).

image

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -97,1610 +97,1615 @@ import java.util.zip.GZIPOutputStream;
  * The application's main frame.
  */
 public class CampaignGUI extends JPanel {
-    //region Variable Declarations
-    public static final int MAX_START_WIDTH = 1400;
-    public static final int MAX_START_HEIGHT = 900;
-    // the max quantity when mass purchasing parts, hiring, etc. using the JSpinner
-    public static final int MAX_QUANTITY_SPINNER = 10000;
+  //region Variable Declarations
+  public static final int MAX_START_WIDTH = 1400;
+  public static final int MAX_START_HEIGHT = 900;
+  // the max quantity when mass purchasing parts, hiring, etc. using the JSpinner
+  public static final int MAX_QUANTITY_SPINNER = 10000;
 
-    private JFrame frame;
+  private JFrame frame;
 
-    private MekHQ app;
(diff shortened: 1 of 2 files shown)


IllianiCBT:
> What would be the order with New Campaign moved down?

image

HammerGS:
What about on the splash, my understanding is that is another big source of issues. But I do think Load, Save, New might be a better order as it's easy to hit new campaign if you go for exit. Creating a new campaign that triggers the exit process

IllianiCBT:
> What about on the splash, my understanding is that is another big source of issues. But I do think Load, Save, New might be a better order as it's easy to hit new campaign if you go for exit. Creating a new campaign that triggers the exit process

The splash is unchanged, I'm not aware of any issues caused by that, but even if there were I don't think we could change the order without reducing usability.

image

I do see your point about placement by 'exit'. I didn't consider a situation where a user goes to click 'exit', hits 'new campaign', exits out of that dialog, clicks 'exit' and then saves their campaign.

Alrighty, with that in mind... I'm really struggling to find somewhere we can move 'new campaign' to where there is no risk of it being clicked accidentally. Any suggestions? Because my mind is blank.

HammerGS:
I think I'm changing my vote to remove it completely option.

SuperStucco:
Better isolation can be had by pushing the actual element into a flyout e.g. "New Campaign > Start New Campaign..."  This lets it live next to the save/load elements while stopping it from accidental use.

IllianiCBT:
### Update
Since the initial introduction of this PR I have significantly changed its implementation.

When selecting 'New Campaign' users will be presented with a dialog asking them to confirm that they wish to start a new campaign. Selecting 'no' or closing the dialog does nothing. Selecting 'yes' initiates the new campaign process (this process is identical to current implementation).

image

IllianiCBT:
I'm aware of the unusually large number of dirty edits on this PR. I just spotted them and will fix tomorrow.

IllianiCBT:
Too many dirty edits. I know how I did them, so easy to avoid in the future, but for the sake of my sanity I'm just going to redux this PR.

PR #3910, opened 2024-03-26
[Bandaid] Commented out 'New Campaign' to Remove From Toolbar [Option B]

IllianiCBT:
This is the sister PR to https://github.com/MegaMek/mekhq/pull/3909. Only one should be merged, not both.

### Current Implementation

### Problem

### Solution

Option B
I am presenting two possible bandaids to mitigate this issue.

Option B (this PR) removes 'new campaign' entirely. This is achieved by commenting out the 'new campaign' option. I opted to use the commenting out approach, as this enables easy restoration of the option at a later date - once the root issue has been investigated and resolved.

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -345,11 +345,11 @@ public class CampaignGUI extends JPanel {
         menuSave.addActionListener(this::saveCampaign);
         menuFile.add(menuSave);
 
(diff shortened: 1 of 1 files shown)


HammerGS:
I'm not sure I support the idea of removing the New Option, but do support the moving option depending on the order.

HammerGS:
Changing my vote and supporting the remove option the more I think on it.

IllianiCBT:
Closed due to superior solution being found.

PR #3915, opened 2024-03-27
Added Dialog to Confirm New Campaign (redux)

IllianiBird:
This is a redux of #3909 which was closed due to an unusually high volume of dirty edits.

### Current Implementation

### Problem

### Solution

### Update
Since the initial introduction of this PR I have significantly changed its implementation.

When selecting 'New Campaign' users will be presented with a dialog asking them to confirm that they wish to start a new campaign. Selecting 'no' or closing the dialog does nothing. Selecting 'yes' initiates the new campaign process (this process is identical to current implementation).

image

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/NewCampaignConfirmationDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/NewCampaignConfirmationDialog.java
@@ -23,7 +23,7 @@ import javax.swing.JOptionPane;
 public class NewCampaignConfirmationDialog {
   public int YesNoOption() {
     return JOptionPane.showConfirmDialog(null, "Are you sure you want to start a new campaign?\n" +
-            "Do not save your campaign, if you cancel after this point", "Start New Campaign?",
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
Looks good to me.

MekHQ/src/mekhq/gui/dialog/NewCampaignConfirmationDialog.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with MekHQ. If not, see <http://www.gnu.org/licenses/>.
+ */
+package mekhq.gui.dialog;
+
+import javax.swing.JOptionPane;
+
+public class NewCampaignConfirmationDialog {
+  public int YesNoOption() {
+    return JOptionPane.showConfirmDialog(null, "Are you sure you want to start a new campaign?\n" +
+            "Do not save your campaign, if you cancel after this point", "Start New Campaign?",
```
NickAragua, line 26:
```suggestion
            "Do not save your current campaign if you cancel after this point!", "Start New Campaign?",
```

PR #3921, opened 2024-03-28
Retirement Overhaul (Wave One)

### LINKED ISSUE #3817: Retirement checking not functioning as intended due to Campaign Operations rating system
As with ship searches (#3753) and contract type generation (#3729), retirement checks are using the campaign rating as a modifier for the target number.  When this is Campaign Operations the rating number is uncapped, so long term campaigns can get extremely large values compared to the limited level choices in the older Field Manual method.

This is currently in 49.12 but checking the current code base as of 49.17 the same problem will still be present.

IllianiCBT:
#### Module Name Change [UI]
- Currently the 'Retirement/Defection' system is called the 'Retirement/Defection' system.
- Currently Defection functionality is not yet implemented. Furthermore, there is no differentiation between employees retiring and employees resigning.
- I have changed references to 'Retirement/Defection' to 'Employee Turnover'. I have left any variable or class names intact. This is predominately a user-facing and/or comment-based change. 'Employee Turnover' better fits the currently implementation of Retirement/Defection and is a better fit for my Wave Two plans.
<p align=center>
  image
</p>

#### Payouts [Rebalance]
- Currently, payouts are based on 3 months' wages.
- Following discourse on Discord it became apparent that 3 months' wage is insufficient. Furthermore, there was a desire to have payouts linked to employee wages, rather than a fixed value.
- Increased payout values to 24 months' wages, based on Discord discussion this was felt to be a more balanced option. Furthermore, rather than using a fixed value, the system now fetches the exact wage of the employee and determines payout based on that value.

#### Unit Rating Modifiers [Rebalance / Bugfix]
- Currently, Unit Rating modifiers are uncapped and directly linked to the Unit Rating. CamOps unit rating has no minimum/maximum theoretical limit.
- Uncapped unit rating modifiers lead to scenarios where retirement TNs are undesirably low or high. Potentially to the point of becoming impossible to pass/fail.
- Rebalanced Unit Rating modifiers to no longer have such a large impact on TN. Unit Rating modifiers now range from + 2 to - 1, depending on unit rating. CamOps and FM:M unit rating systems now both result in the same TN adjustments. (closes #3817)

#### Experience Modifier [UI]
- Currently, the `Experience` modifier is applied to the retirement TN and is a combination of employee skill rating and company unit ranking.
- Having `Experience` be a combination of two modifiers makes it difficult for users to understand what calculations are going into the TN.
- Split `Experience` into `Unit Rating` and `Skill Rank`. `Skill Rank` is displayed using the employees' own skill rank (see below).
<p align="center">
  image
</p>

#### Shares Modifier Display [UI]
- Currently, when using Shares, if a contracts' share percentage is above 20% a modifier is applied to the retirement TN.
- The Shares modifier only displays once shares percentage is above 20%. As the default value for shares percentage is 20% this may lead users to not realize that the shares modifier exists.
- So long as Shares are being used, the Shares modifier will always display during the end of contract retirement check.

#### Empty Condition [Refactor]
- Currently, there is an empty conditional (lines 268-270).
<p align="center">
  image
</p>

- The empty conditional causes IntelliJ to yell at me every time I try to push a commit. It's annoying.
- I removed the empty conditional.

#### Faction Modifiers [Enhancement]
- Currently, companies with the 'Pirate' faction have a + 1 modifier to the retirement TN.
- Faction modifiers are limited only to the pirate faction and do not account for employee faction.
- Now the system checks whether the company Faction is 'Pirate' and if so, applies a + 1 modifier to the retirement TN. If the company Faction is _not_ 'Pirate', but the employee _is_ a Pirate, then the employee has a + 1 modifier to their retirement TN. Employees with the 'Mercenary' or 'Clan' origin have a + 1 / - 2 modifier to the retirement TN respectively, regardless of company faction.

#### Old Age Modifiers [Enhancement]
- Currently, employees 50 years old, or older, have a + 1 penalty to the retirement TN.
- The + 1 modifier is static, meaning a 50 year old has the same retirement desires as a 120 year old.
- Rebalanced and expanded age modifiers These offer a more granular range of modifiers, from + 1 to + 5. Included a condition for very young employees (20, or younger) which applies a - 1 modifier. I felt young pilots would be less likely to want to stray too far from their initial hire.
<p align="center">
  image
</p>

#### Bonus Modifiers (when not using Shares) [Rebalance]
- Currently, when not using Shares, users can pay a Bonus to employees to improve the TN by -1.
- The bonus modifier lacks impact, considering it is the only real way for a non-Shares using user to manipulate retirement checks.
- Improved the modifier from paying bonuses from - 1 to - 2.


IllianiCBT:
This PR is now feature complete. Over the weekend I’ll write the supporting documentation, but otherwise it’s good to go.

SJuliez (approved):

HammerGS:
@IllianiCBT This now has Conflicts that need to be fixed before merging.

IllianiCBT:
> @IllianiCBT This now has Conflicts that need to be fixed before merging.

Fixed :)

PR #3922, opened 2024-03-28
Add missing cockpit costs and weights

neoancient:
I noticed the costs and weights were incomplete when checking places the cockpit constants were being used.

Requires Megamek/megamek#5291.


Sleet01 (approved):
Changes look good, once the linked MM PR is pulled.

PR #3923, opened 2024-03-29
Hide Toughness When 0

IllianiCBT:
### Current Implementation
When enabled, via Campaign Options, Toughness will always be visible on the personnel panel.
<p align=center>
  image
</p>

### Problem
This does not follow the standard set out by other optional characteristics, such as 'Edge', which only display when having a non-zero value.

### Solution
Expanded the existing conditional to check that Toughness is enabled _and_ above zero.
<p align=center>
  image
</p>
<p align=center>
  image
</p>

### DIFF
--- a/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
@@ -1306,7 +1306,7 @@ public class PersonViewPanel extends JScrollablePanel {
             firsty++;
         }
 
-        if (campaign.getCampaignOptions().isUseToughness()) {
+        if ((campaign.getCampaignOptions().isUseToughness()) && (person.getToughness() > 0)) {
             lblTough1.setName("lblTough1");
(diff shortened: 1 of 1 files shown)


Thom293:
Just as an aside, I think Toughness is in a strange spot.  I dont think they can be "born" with any points in it, and I dont think you can buy or increase it in anyway.  Its just a non-stat.  

I think they should have a chance to be be able to be "born" with 1-2 toughness as it seems a "natural" trait like small pilot.

IllianiCBT:
> Just as an aside, I think Toughness is in a strange spot. I dont think they can be "born" with any points in it, and I dont think you can buy or increase it in anyway. Its just a non-stat.
> 
> I think they should have a chance to be be able to be "born" with 1-2 toughness as it seems a "natural" trait like small pilot.

I agree entirely, I even have this functionality built into my new pilot spreadsheet, but that would need to be a separate RFE.

Thom293:
I may have put one up before.  I at least thought about it.  ill check later and if not add.

Sleet01 (approved):
LGTM

Issue #3929, opened 2024-03-29
Prevent Advancing Day When Scenario is "Due"


AaronGullickson:
To prevent users from advancing the day when they should be completing a scenario with a "due" date of today, I think we should disable advancing the day when scenarios are due. Instead of advancing the day, they would get a pop-up that indicates they need to complete scenarios before advancing.

This should not be implemented until we can add the ability to change dates in the CustomizeScenarioDialog so that users can correct mistakes. This should be done by #3927

IllianiCBT:
Current implementation has the user presented with a dialog warning that scenarios are due and will be counted as a defeat if not attended to before advancing the day. The user can then decide to ignore the warning and advance anyways.

AaronGullickson:
Yeah, I found that code for AtB. But I am not talking about AtB. I would take all the code for AtB, dump it in a big pile and burn it if I could. I am talking about vanilla scenarios. I will make sure it doesn't conflict with AtB. This one will not be ignorable, but users will be able to change dates.

Issue #3927, opened 2024-03-29
Add the new Scenario features to CustomizeScenarioDialog


AaronGullickson:
As part of story arcs development, I have built a lot of new capabilities into the basic Scenario object. I also migrated some stuff from AtBScenario to basic Scenario. So you can now have the following in basic Scenario:

* Pre-specified board
* Pre-specified deployment.
* Pre-specified scenario objectives (ScenarioObjective).
* Pre-specified opponents and allies (BotForce). 
* Pre-specified planetary conditions.

However, it is not yet possible to access these capabilities from within MekHQ. I should add these new capabilities to CustomizeScenarioDialog. This is important prepatory work toward building a Story Arc editor. However, it can also be useful for players more generally because they can pre-specify all of this stuff before starting the scenario so that when they start it, they are just ready to go.

IllianiCBT:
I don’t know if #3779 is still a needed function, and if so, whether it could be covered by this?

Thom293:
Oh man these are awesome.  So many ways to use this.  I cant wait to see how this works.

AaronGullickson:
@IllianiCBT I have been thinking about the date issue as well, but I think thats a separate thing. Currently, the date is left empty until added after the scenario is completed, which makes sense. However, I think it would be useful to have the date pre-specified to prevent advancing the day until scenarios with the current date are completed.

AaronGullickson:
Ok, I think it makes sense to add a "Change Date" to the dialog as well, which I think would obviate the need for that PR.

I have another idea about preventing user from advancing days when they have scenarios "due" but that is a separate issue.

PR #3930, opened 2024-03-29
Prevent advancing day with pending vanilla scenarios

AaronGullickson:
This PR addresses issue #3929. It will prevent advancing the day when a scenario is due (is pending and has a due date of today or earlier). This only affects vanilla scenarios not AtB scenarios. This will make story arcs where scenarios are created by the arc easier to manage because players will be alerted when there is a scenario that requires attention. 

This is ready to go ~~but I am going to leave it in a draft state because I do not want to add this code until we give players a way to edit the date of a pending scenario, which I plan to do as part of addressing issue #3927.~~

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -2366,6 +2366,13 @@ public class CampaignGUI extends JPanel {
             return;
         }
 
+        if(getCampaign().checkScenariosDue()) {
+            JOptionPane.showMessageDialog(null, getResourceMap().getString("dialogCheckDueScenarios.text"),
+                    getResourceMap().getString("dialogCheckDueScenarios.title"), JOptionPane.WARNING_MESSAGE);
+            evt.cancel();
+            return;
+        }
+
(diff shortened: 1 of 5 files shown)


AaronGullickson:
I was going to wait on this until #3927 was done so that users can always edit the date of a scenario, but that PR is going to take awhile and the code for changing the date was already done. So I moved the code for changing the date in CustomizeScenarioDialog to this PR and it should now be ready to go. The game won't advance if a scenario has the current or earlier date, but the edit scenario dialog will now allow users to edit the date of pending scenarios, regardless of vanilla or AtB.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -2366,6 +2366,13 @@ public void handleDayEnding(DayEndingEvent evt) {
             return;
         }
 
+        if(getCampaign().checkScenariosDue()) {
```
Sleet01, line 2369: Should this be in a Nag Dialog with the other nags?

AaronGullickson:
Its not technically a nag because you can't dismiss it. Thats why I wanted to make sure to give people the ability to edit scenario dates.

PR #3931, opened 2024-03-29
Add new Scenario features to CustomizeScenarioDialog

AaronGullickson:
This PR addresses issue #3927. It adds a lot of new features to the `CustomizeScenarioDialog`. Because a lot of features were moved from `AtBScenario` to `Scenario` as part of the work on story arcs, you can now do a lot with `Scenario` but up until now there has been no way to access all of the goodies. With this PR, it should now be possible to set up a variety of things for your game before starting the game, including:

* maps, either fixed or random parameters from the `mapgen` directory
* other bot forces, including behavior, deployment, team, camo, and the ability to set a set of fixed units from a mul or randomized units.
* player's deployment
* scenario objectives
* scenario deployment limits

Now, you can of course, do most of this from the chat lounge as well, so why replicate that here? The first and foremost reasons is that I need this dialog for the story arc editor so that creators can create pre-generated scenarios chock full of good stuff. But I think it can be useful for anyone. I find that with big games, I often goof something up and have to reset and set it all up again in the chat lounge. Having all of this pre-specified in MHQ provides a QoL improvement. It also will integrate nicely with AtB, so that players can tweak their scenarios before entering the chat lounge.

Of course, I didn't want to just replicate the entire functionality of the chat lounge here because that would be insane. So, this dialog does not offer you as much richness in appearance as the chat lounge, but it can still do a lot. Whenever possible I tried to use existing methods from the chat lounge rather than duplicate code, but many of the dialogs there require a lot of assumptions about clients and games that made it difficult in some cases.

Some previews of the dialog. Here is one of the main dialog. I stuffed it pretty full. Future expansions may need to consider tabbing.

Screenshot 2024-05-06 at 2 40 17 PM

Next is a shot of the dialog for customizing deployment position. I based this on the one from MM and it includes the more advanced deployment options.

Screenshot 2024-05-06 at 2 40 22 PM

The map editor has two different looks. If you are using random maps, you will be given a choice of the different random generators in the `mapgen` directory:

Screenshot 2024-05-06 at 2 40 42 PM

If you are using fixed maps, then you get a display similar to the chat lounge where you can see thumbnail versions of each map (note Scenario does not have the ability currently to combine multiple fixed maps into a single map).

Screenshot 2024-05-06 at 2 41 01 PM
 
The BotForce dialog includes the ability to load a set of fixed entities from a MUL file. This allows players to use the ChatLounge if they want to get all the details just right and then load the entities here. You can also choose a random generation option using `BotForceRandomizer`. Note that you can also do both fixed and random.

Screenshot 2024-05-06 at 2 41 18 PM

Scenario objectives are available in the scenario objectives dialog. I based this on the one from the scenario template dialog with a good deal of improvement to its layout and design. 

Screenshot 2024-05-06 at 2 41 39 PM

Finally, you can limit scenario deployment as a percent of your total force or total number of units. You can also specify only certain types of units as allowed.

Screenshot 2024-05-06 at 2 41 56 PM


Thom293:
This is just a question and Im not trying to create additional work:  Is there a way here or already existing to pre-define the bot behaviour (e.g. Sniper, Berserk, Escape, etc.)?  If it already exists I havent seen how to do it other than to change manually in the lobby.  I presume it exists somehow because some scenarios they are set to flee first, but I dont see where it is actually set.

Again if it doesnt exist Im not trying to suggest additional work for you.  I can look at it myself or put up a separate RFE for later.  Thank you.

AaronGullickson:
It's there in bot force and I am planning to allow users to edit it.

AaronGullickson:
Screenshot of the first draft of the CustomizeBotForceDialog

Screenshot 2024-04-12 at 10 57 48 AM

SJuliez (approved):

AaronGullickson:
@HammerGS Conflicts dealt with - should be ready to merge.

PR #3936, opened 2024-03-30
Adjust Retirement Payouts for 49.19

IllianiCBT:
As discussed on Discord, this tweak adjusts payout values. It will be fully replaced by #3921 and should only be merged if 3921 is _not_ going to be merged into 49.19.


IllianiCBT:
No longer relevant

PR #3937, opened 2024-03-31
Added Nag for Wounded Personnel without Doctor

### LINKED ISSUE #3932: RFE:  Infirmary Nag for Untreated Patients
TItle.  Ran across this in story arcs. athough it is unrelated to that.

An Optional daily nag just like the "You have unmaintained units, are you sure you would like to advance the day?"  or the Pending Scenario Nag.  For when you have injured people but have not assigend them to a doctor yet.  I forget to do this all of the time.

So, something like:  "You have untreated patients in the infirmary.  Are you sure you would ike to advance the day?"

IllianiCBT:
### Current Implementation
None

### Problem
It's very easy to `Advance Day` without checking whether you have injured personnel in your roster, potentially causing their injuries to worsen.

### Solution
At day end this system will check to see if there are any active personnel with non-permanent injuries. If found, it will then check whether that person has been assigned a doctor. If no doctor has been assigned the system will return true prompting the nag dialog.

This nag can be suppressed through `MekHQ Options/Nag Options`.

<p align=center>
image
image
</p>

### Closes
Closes #3932


Sleet01 (approved):
Looks good to me!

HammerGS:
@IllianiCBT This now has merge conflicts that need to be fixed.

MekHQ/src/mekhq/gui/dialog/nagDialogs/UntreatedPersonnelNagDialog.java
```diff
@@ -0,0 +1,56 @@
...
+public class UntreatedPersonnelNagDialog extends AbstractMHQNagDialog {
+    public static boolean isUntreatedInjury (Campaign campaign) {
+        for (Person p : campaign.getActivePersonnel()) {
+            for (Injury i : p.getInjuries()) {
+                if (!i.isPermanent()) {
+                    if (p.getDoctorId() == null) {
+                        return true;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
IllianiCBT: Done - thanks for the advice, put me right in the ballpark! :)

MekHQ/src/mekhq/gui/dialog/nagDialogs/UntreatedPersonnelNagDialog.java
```diff
@@ -30,12 +30,16 @@
...
-                if (!i.isPermanent()) {
-                    if (p.getDoctorId() == null) {
-                        return true;
+            if (campaign.getCampaignOptions().isUseAdvancedMedical()) {
+                for (Injury i : p.getInjuries()) {
+                    if (!i.isPermanent()) {
+                        return p.getDoctorId() == null;
                     }
                 }
+            } else if (!campaign.getCampaignOptions().isUseAdvancedMedical()) {
```
AaronGullickson, line 39: No need for the if here. Just do `else {`.

IllianiCBT: So you'd think that, and I agree, that's how it _should_ work. However, when I was testing it was triggering the dialog twice whenever the 'else if' wasn't in place. It's beyond me as to why.

MekHQ/src/mekhq/gui/dialog/nagDialogs/UntreatedPersonnelNagDialog.java
```diff
@@ -0,0 +1,60 @@
...
+
+import javax.swing.*;
+
+public class UntreatedPersonnelNagDialog extends AbstractMHQNagDialog {
+    private static boolean isUntreatedInjury (Campaign campaign) {
+        for (Person p : campaign.getActivePersonnel()) {
+            if (campaign.getCampaignOptions().isUseAdvancedMedical()) {
+                for (Injury i : p.getInjuries()) {
+                    if (!i.isPermanent()) {
+                        return p.getDoctorId() == null;
```
AaronGullickson, line 36: This is wrong. You are going to return true/false on the first person in the loop and then the loop will stop processing. You should have left this how you had it before.  Same for the other case below.

```java
if (!i.isPermanent() && p.getDoctorId() == null) {
  return true;
} 
```

MekHQ/src/mekhq/gui/dialog/nagDialogs/UntreatedPersonnelNagDialog.java
AaronGullickson: Huh, that shouldn't happen.

AaronGullickson: I am not aware of how advanced medical works, but do you even need  both checks. Isn't it also hit dependent? i.e. if someone has zero hits, they have no non-permeanant wounds.

AaronGullickson: Found a better solution to all of this: `person.needsFixing()`. That will check both basic hits and advanced medical in one check and simplify your approach.

```java
for (Person p : campaign.getActivePersonnel()) {
    if(p.needsFixing() && p.getDoctorId() == null) {
       return true;
   }
}
```

that will cover both cases.

IllianiCBT: Done. I went ahead and used your suggested code and it works like a charm.

I opted not to check for permanent injuries, as personnel with permanent injuries still need to be assigned to a doctor, so I doubt anyone will complain.

AaronGullickson (approved):
Looks good now!

PR #3947, opened 2024-04-01
StratCon Air & Arty Modifiers

### LINKED ISSUE #3894: RFE: add a 'light artillery support' scenario modifier
At times the standard artillery scenario modifiers can be a bit overwhelming in opposition or overkill in support.  It would be nice to have a slightly less overwhelming version in play, using field artillery infantry rather than vehicles.  The lower round count makes them a little easier to fight past in opposition and not quite as obstructive when used as allies.

Thom293:
This PR addresses a recent RFE to add infantry field guns.  Its also adds ASF aces to fill the last hole of the most commonly used units.  I dont plan on adding any more at this time unless the ability to sort them by date happens.

Regardless, later I plan on updating all of the current .mul modifiers to have one for each era.  Even without date sorting, they can be added to the repo to their own folder with instructions that users can copy over depending on the era they are playing, or I can just put them in the files forum for people that want to use them.

Closes #3894


HammerGS (commented):
Looks good, brief testing and didn't see any issues.

Issue #3776, opened 2023-10-31
J17 - 49.16 - Custom Maps rolled from scenario are ignored in MM


Thom293:
Not exactly sure where this goes but I think its a MHQ issue...

If  a scennario rolls a custom map, when you go to MM, it is assigning a regular generic mapsheet.  This is the second time this has happened.  Both times have been in facility missions.  Both times it was a 25x25 map.  

And Just as an observation, 25x25 seems too small for a facility map.

![image](https://github.com/MegaMek/mekhq/assets/23645569/d1cf739f-e95b-4552-a1ca-946c46d3f5da)

This is the map it is supposed to be:
![image](https://github.com/MegaMek/mekhq/assets/23645569/1f2639c4-161b-4a79-94c7-e41c231300bf)
 
But this is what I got:

![image](https://github.com/MegaMek/mekhq/assets/23645569/3b368582-d22c-439c-aabf-61ea20dab3cc)

[mekhq.log](https://github.com/MegaMek/mekhq/files/13219273/mekhq.log)

[megamek.log](https://github.com/MegaMek/mekhq/files/13219278/megamek.log)

[Warpath Inc.30380110.cpnx.gz](https://github.com/MegaMek/mekhq/files/13219280/Warpath.Inc.30380110.cpnx.gz)
[Customs.zip](https://github.com/MegaMek/mekhq/files/13219285/Customs.zip)

IllianiCBT:
This appears fixed in the current version. Closing as resolved.

PR #3948, opened 2024-04-01
Capitalize deadborder directory for MM consistency

AaronGullickson:
In MekHQ, we have a directory  that is `data/boards/unofficial/deadborder` and in MegaMek we have a directory that is `data/boards/unofficial/Deadborder`. Since we use MegaMek's board folder for releases, this leads to a discrepancy when running MekHQ from source and from a release. This just capitalizes it in MekHQ to bring it in line with MegaMek. I checked other folders in that directory and they all matched.


Thom293:
Nice.  Based on your comments in Story Arcs, will this address and close #3776?

AaronGullickson:
No, its only going to affect maps in the `unofficial/deadborder` directory. I compared that map between MM and MHQ and the path appears to be identical, so I think #3776 is a different issue.

Thom293:
Ok.  thank you. Will edit out the number so it doesnt grab it by mistake.

HammerGS (approved):

PR #3950, opened 2024-04-01
Fix #3949: Custom ScenarioObjectives do not change ScenarioStatus in ResolveScenarioWizardDialog

### LINKED ISSUE #3949: Custom ScenarioObjectives do not change ScenarioStatus in ResolveScenarioWizardDialog
Currently, if you have a custom objective, you need to check manually in the ResolveScenarioWizardDialog whether you completed it or not. However, doing so does not change the current victory status in the final panel, like other objectives. There appears to be two reasons that I can see for this. First, in `ScenarioObjectiveProcessor.determineScenarioStatus`, you find the following code:

```java
            // some objectives aren't associated with units and their completion is set manually
            // in that case, we continue on
            if (!objectiveUnitCounts.containsKey(objective)) {
                continue;
            }
```

This kicks out custom objectives right from the get go despite the fact that this method reads in a `Map<ScenarioObjective, Boolean> objectiveOverrides` object that is supposed to handle this issue with code below this continue check:

```java
     boolean objectiveMet = (objectiveOverrides.containsKey(objective) && objectiveOverrides.get(objective) != null) ?
                objectiveOverrides.get(objective) : objectiveMet(objective, objectiveUnitCounts.get(objective));
```

This code is unreachable for custom objectives even though the objectiveOverrides object is explicitly designed for custom objectives. 

However, here we hit the second issue. The `ScenarioObjectiveProcessor.determineScenarioStatus` method is only called two places in ResolveScenarioWizardDialog and in both places it reads in an empty HashMap for that object. There is in fact no tracking of HashMap of the needed kind in ResolveScenarioWizardDialog. What is tracked is a Hashmap of check buttons that could be turned into the appropriate output. So it looks like an incomplete project?

AaronGullickson:
These fixes will enable checking custom ScenarioObjectives to affect the victory conditions.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
+++ b/MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
@@ -281,9 +281,9 @@ public class ScenarioObjectiveProcessor {
         }
 
         for (ScenarioObjective objective : scenario.getScenarioObjectives()) {
-            // some objectives aren't associated with units and their completion is set manually
-            // in that case, we continue on
-            if (!objectiveUnitCounts.containsKey(objective)) {
+
+            // if the scenario is not in our objectiveUnitCounts or objectiveOverrides, skip it
+            if (!(objectiveUnitCounts.containsKey(objective) || objectiveOverrides.containsKey(objective))) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
Looks good!

PR #3953, opened 2024-04-02
Add disabled option for Personnel Market

AaronGullickson:
For **years**, even aeons, MekHQ has included a personnel market that the user had no ability to disable, creating enormous frustration and breaking of immersion among those of who had no use for it. The unit market - you can disable it. The contract market - you can disable it. The personnel market - apparently it has been ordained by God.

But no longer! I bring good tidings. Despite an absolute byzantine coding scheme, I have managed against all odds in this PR to add a disabled option to the personnel market choices, without disrupting people's existing choices in saved games. Furthermore, if you choose disabled, you will not even be forced to look at the personnel market menu item. Hallelujah!


HammerGS (approved):
Tested and seems to work.

PR #3954, opened 2024-04-02
Added ATOW Salaries & Rank Multipliers

### LINKED ISSUE #3938: Rank
(Delete the above lines)

### Environment
Version 0.49.17
Windows 10
Java Version: 11.0.18

### Description
Pay Multiplier is set at 1 for all ranks.
That seem unrealistic and makes rank a less interesting decision than it otherwise could be.
Proposed solution: Have pay multiplier increase for higher ranks and have higher ranks have better retention on retirement rolls.

Rules reference: campaign ops pg 25 has a table that gives a rank multiplier of "(Rank Trait Points/3)+1, minimum 1" in the salary table.  Though I am unsure what that means  (I assume it to be RPG related)

### Files
Campaign options

IllianiCBT:
## Current Implementation
Currently, salaries appear to be mostly based on CamOps, with rank-based salary modifiers simplified to 'is this employee an officer or not'.

## Problem
While a perfectly working system, it lacks granulality. Assuming no external modifiers, under the current system a Corporal and a Sergeant Major would both be paid identically. As would a Lieutenant and the Star Lord.

### Rank Based Multipliers
Using ATOW as a basis, I went through and updated `ranks.xml` to include RAW salary rank multipliers.

Our rank system invents rank tiers, to ensure parity between different rank systems, so I used the ATOW:Companion SLDF rank listing to correlate across to mhq's rank tiers (E1, E2, E3, etc). Any gaps between rank tiers (where we invented additional tiers) I based the multiplier calculations on distance from the nearest RAW rank tier.

You can see my calculations [here](https://docs.google.com/spreadsheets/d/1GktLS7eJZB_772Jg9xsWzIRaqemydwetFKS9lqugHPk/edit?usp=sharing). Any Trait Point (TP) in red is one where I've extrapolated. Decimal TP means the distance between the last & next RAW rank tier is less than 1.

### Removing Current Modifiers
After discussing it with the team on Discord (see [here](https://discord.com/channels/458705327911731231/458767762219597825/1224435561226305546), it was decided that this new system should wholly replace the old.

Therefore, I've gone through and removed the old fixed Enlisted/Commissioned rank multipliers. I tested with old saves and nothing breaks if an old save is loaded. The system will just ignore the lines where Enlisted/Commissioned ranks are defined in the save file.

### Updating Base Salary, Expertise Modifiers, & Specialization Modifiers
I went through and updated base salaries, expertise modifiers, and specialization modifiers to match the values in ATOW. Where missing, I extrapolated values based on prior implementation.

The biggest change was made to LAM pilots. Previously, they were paid the sum of MechWarrior and Aero base pays (1,500 + 1,500 by default). I changed this to treat the Aero component as if it were a secondary role (i.e. the base pay is halved). This changes the calculation to 1,500 + 750 = 2,250c-bills (default values).

I also went through and removed where salary information is defined in the existing presets. This causes the system to pull from CampaignOptions.java and use the default values, instead.

Existing, custom presets, likely will not be affected by these changes and there is no way to resolve that. This will result in existing campaigns having salaries that don't quite match ATOW. This will be most evident with Elite employees, as CamOps gave them a *3.2 multiplier, whereas ATOW only gives them a *2.0. The actual pay difference will not be prohibitively large, so I considered that to be an acceptable sacrifice.

### Salary Calculations
The old salary calculations were based on CamOps, the ATOW system is very different. I have gone through and updated the calculations so that they're as close to ATOW as I could muster.

The biggest differences between the two systems are as follows:
- CamOps applies each multiplier sequentially. So if the base pay was 1,000c-bills and they had an expertise modifier of 1.6 and a specialization modifier of 1.28 the salary would be calculated as 1,000 * 1.6 * 1.28. ATOW applies each modifier additively. So the same employee would have their salary calculated as 1,000 * (1.6 + 1.28). This required the old system to be completely replaced with new calculations.
- Currently, we check whether an employee has the 'specialized infantry' and/or 'anti-mech' specializations both for primary and secondary roles. This would allow an employee with both specialized conventional infantry and anti-mech battle armor roles to double-dip, gaining the multipliers twice. Given that ATOW applies multipliers additively, this results in larger than intended salaries. I changed the system so that it checks whether either role qualifies and then applies only once.
- ATOW bases the expertise modifier on the mean skills used by both primary and secondary roles. Our current system applies the appropriate expertise modifier to the salary of the primary role and secondary role independently. This means that a Veteran MechWarrior/Elite MechTech would have the Veteran multiplier applied to their MechWarrior salary, and the Elite multiplier applied to their MechTech salary. Due to ATOW multipliers being additive, this would have resulted in double-dipping expertise multipliers. I also couldn't use ATOW's averaging of skill rating, at least not without rebuilding `getExperienceLevel()` which would likely have knock-on problems elsewhere. Instead, if the employee has a second role the system determines the mean of the two multipliers. So, using our Veteran MechWarrior/Elite MechTech as an example, the system would generate an expertise multiplier of *1.75 ((1.5 [veteran] + 2.0 [elite]) / 2).

## Closes
Closes #3938

### DIFF
--- a/MekHQ/data/universe/ranks.xml
+++ b/MekHQ/data/universe/ranks.xml
@@ -11,7 +11,7 @@ Rank:
 	Officer: Whether the rank is considered to be an officer.
 	Pay Multiplier: The pay bonus paid to those of this rank.
 -->
-<rankSystems version="0.49.0">
+<rankSystems version="0.49.19">
 	<rankSystem>
 		<code>SSLDF</code>
 		<name>Second Star League Defense Force</name>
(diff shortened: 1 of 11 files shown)


IllianiCBT:
Note: please don't be scared by the large number of changes GitHub is reporting. The overwhelming majority of those are from me using find & replace in `ranks.xml`.

IllianiCBT:
Closed pending redux due to Dev discussion

PR #3956, opened 2024-04-02
Change ransomed unit buttons to sold unit button in scenario resolve dialog

AaronGullickson:
The current ScenarioResolverDialog uses "Ransom" to refer to cases where you want to sell captured units. This is confusing language as you don't typically ransom units - I think it was borrowed from the code for personnel. I have changed the labeling to "Sold". To keep things consistent I also changed "Salvage" to "Salvaged" so all three labels are past tense : Salvaged, Sold, Escaped.

I also refactored all the names of objects in the code involving units with "ransomed" to "sold". Technically, I didn't need to do that to get the labels right, but it will make the code more consistent with the output and thus less likely to cause confusion among future developers. Hopefully, it won't break any pending PRs.


AaronGullickson:
Realized I forgot one refactoring.

AaronGullickson:
Ok, should be ready now.

IllianiBird (approved):
Looks good to me

SuperStucco:
Ransom may be the appropriate term here, as it wasn't just limited to personnel.  You aren't selling it on the open market, you are ransoming (trading) it back to the enemy in exchange for money, sometimes even just letting them walk/roll it off the field unmolested.  In lore, this could be money but could also be for an agreement to a supply of parts or return your own captured personnel.

AaronGullickson:
Who is to say you are not selling it on the open market? It's just a convenience function to avoid having to sell from the hangar.  People can have their own narrative about what is actually happening. The issue is using a label that makes it clear to users what is being done.

PR #3958, opened 2024-04-03
No selling units in scenario resolution if campaign disallows selling

### LINKED ISSUE #3955: Selling units in scenario resolution should not be possible if selling units is disallowed
Currently, a player who has disabled the selling of units in the campaign options can still sell captured units during scenario resolution with the "Sold" (formerly "Ransom") button. This button should be disabled as an option if selling has been disallowed.

AaronGullickson:
Very small PR. Closes #3955.


Sleet01 (approved):
LGTM

PR #3965, opened 2024-04-04
Unified Medal Language

IllianiCBT:
### Current Implementation
Currently, the term 'Mission' is used throughout the standard medal set to refer to kills scored across a Contract.

### Problem
Mission =/= Contract, insofar as mhq is concerned. Like mhq, the in-dev AutoMedal system treats Missions and Contracts as separate entities, therefore the language needs to be consistent. This is also confusing to users trying to identify which medals their employees are entitled to.

### Solution
I replaced all instances of 'mission' with 'contract'

### DIFF
--- a/MekHQ/data/universe/awards/standard.xml
+++ b/MekHQ/data/universe/awards/standard.xml
@@ -29,7 +29,7 @@
 	</award>
 	<award>
 		<name>House Distinguished Service</name>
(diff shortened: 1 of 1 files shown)


SJuliez (approved):

IllianiCBT:
Closed as no longer necessary.

PR #3968, opened 2024-04-06
Added Ability to Collapse/Expand Logs, Missions and Kills in Personnel Unit Screen

### LINKED ISSUE #3348: RFE:  Add ability to collapse Logs, Missions and Kills in Personnel Unit Screen
Was looking at one of my founder pilots and this became an issue.  He isnt even the one with the most entries.  This particular pilot has 64 log entries, 29 missions, an 19 kills.  And honestly he doesnt even have the most entries.

This could be in addition to or instead of #3349.  But I think I prefer the other one more.

So - just an option to collapse the catagories. And perhaps move kills to the top?  Right now its the last of the 3.

This gave me another idea which I will post in a similar but separate RFE.

IllianiCBT:
### Current Implementation
Currently, all personnel information is shown in an unbroken stream. Personal Log, Scenario Log, Kill Log, Permanent Injuries, etc.
<p align=center>
image
</p>

### Problems
As shown above, with even moderately experienced pilots this can very quickly be quite the list. Forcing users to scroll past all prior information before they can reach the information they do want. Further more, at least on my machine, the logs frequently bug out while being scrolled.
<p align=center>
image
</p>

### Solution
The biggest three offenders are Personal Log, Scenario Log, and (to a lesser degree) the Kills Log. I have implemented the ability to 'collapse' and 'expand' these two logs. To avoid the need to create a new collapse/expand class I used a slightly hacky method that otherwise achieves the same goal (insofar as the user is concerned).

When the employee is first viewed, the three logs are 'collapsed', as shown below.
<p align=center>
image
</p>

Clicking on the label doesn't expand the log, it actually swaps the 'header' with the full log. Clicking on the label of the full log swaps them out again.
<p align=center>
image
</p>

### Closes
Closes #3348


Thom293:
Thank you sir!

SJuliez (approved):
This is alright. If you want (and only then, it's no big deal), you could try reducing some of the repetition. For example, the clunky MouseListener. You could add:
```

private MouseListener getSwitchListener(JPanel current, JPanel switchTo) {
  return new MouseAdapter() {
    @Override
    public void mouseClicked(MouseEvent e) {
      if (current.isVisible()) {
        current.setVisible(false);
        switchTo.setVisible(true);
      }
   }
}
```
and then just
```
pnlLogHeader.addMouseListener(getSwitchListener(pnlLogHeader, pnlLog));
pnlLog.addMouseListener(getSwitchListener(pnlLog, pnlLogHeader));

same for the other panels...
```

(I have not tested this, I just wrote it in here)

IllianiCBT:
> This is alright. If you want (and only then, it's no big deal), you could try reducing some of the repetition. For example, the clunky MouseListener. You could add...

I actually had a very similar idea, last night, but I was already in bed by the time I thought of it.

Went ahead and refactored :)

PR #3972, opened 2024-04-06
remove atb code from gamethread

### LINKED ISSUE #3971: Lance name PR breaks vanilla GameThread
PR #3901, now merged, adds method GameThread.setupBotForces. But this method depends on all missions being AtBContracts by recasting them as such. However, this is the GameThread not the AtBGameThread. This means it will lead to errors for anybody not running AtB. That needs to be corrected. I can't merge master into Story Arcs at the moment because this issue will render it non-functional.

You could do that by trying to remove the AtB dependency here, but I would rather you just move GameThread.setupBotForces to AtBGameThread and revert the GameThread back to its prior state. I think there may be better ways to handle this issue for the non-AtB case.

kuronekochomusuke:
- remove atb code from gamethread

fixes #3971

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -21,6 +21,7 @@ package mekhq;
 import megamek.client.Client;
 import megamek.client.bot.BotClient;
 import megamek.client.bot.princess.Princess;
+import megamek.client.generator.RandomCallsignGenerator;
 import megamek.client.ui.swing.ClientGUI;
 import megamek.common.*;
 import megamek.common.planetaryconditions.PlanetaryConditions;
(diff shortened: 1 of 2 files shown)


AaronGullickson:
Ok, I am looking back through #3901, and I think there is another problem. You removed the call to `botForce.generateRandomForces()`. That currently has not usages. That needs to be called up prior to `botForce.getFullEntityList(campaign)` or the random units will not be generated. 

Also, while this technically solves the problem, it does so by forcing the bot in vanilla games (and story arc games) to be divided into lances, when that may not be the appropriate way to divide them up. If the goal here is to add a bit of nice fluff to the opfor, this issue may make the solution worse than the problem in some cases. For example, in the young wolves story arc, you will end up with clan opponents split into lances with names like "Monster Lance" which will totally break immersion. 

You could try working around this by adding a `lanceSize` variable to botForce directly that can be accessed by everyone. In that case, AtB should just assign that variable when it creates the BotForce. Then the GameThread just needs to look for the `lanceSize` that the BotForce would prefer. 

The other solution is the one I recommended in the issue. Move `setupBotForce` to AtBGameThread and return GameThread to its original state. This solution of randomization feels very AtB to me and thus more appropriate for AtB. I think there may be better solutions for other game modes. For example, I am working on an update to the CustomizeScenarioDialog that will allow people to manually create and edit BotForces (for use in Story Arcs but also vanilla play) and part of that would be the ability to directly specify formations and names for those formations.

PR #3973, opened 2024-04-07
Move new lance creation to AtBGameThread

### LINKED ISSUE #3971: Lance name PR breaks vanilla GameThread

AaronGullickson:
fixes #3971 

@kuronekochomusuke Please take a look at this as an alternate way to address #3971. It moves the force assignment stuff to AtBGameThread properly. I would like to think about how to do this as well for vanilla GameThread but would prefer to do it after I get story arcs integrated.


kuronekochomusuke (approved):
Looks good

Issue #3981, opened 2024-04-07
49.19-Snapshot AsTech Error in Log


IllianiCBT:
### Environment
Origin Project: MekHQ
Java Vendor: Eclipse Adoptium
Java Version: 11.0.22
Platform: Mac OS X 13.6.6 (x86_64)

### Description
I was trying to diagnose another issue when I noticed the below error occurring on Advance Day.

### Log
`02:50:44,277 ERROR [mekhq.campaign.Campaign] {AWT-EventQueue-0}
mekhq.campaign.Campaign.getAvailableAstechs(Campaign.java:5332) - Tried to getAvailableAstechs with 0 minutes. Returning 0 Astechs.`

[IllianiCBT_#3981.zip](https://github.com/MegaMek/mekhq/files/14896203/IllianiCBT_.3981.zip)

IllianiCBT:
I've identified the issue. `doMaintenance()` is attempting to get the tech for Large Vessels, but failing to see the tech officer assigned to that unit. I need to cook dinner shortly, so will see if I can't fix this tonight.

IllianiCBT:
Correction, the issue is caused by `getAvailableAstechs()` via `doMaintenance()`. As large vessels do not use Astechs, `doMaintenance()` is telling `getAvailableAstechs()` that remaining Astech time is being calculated, but 0 Astech time is actually being used.

As everything is working as intended, I don't think this error message is warranted.

PR #3985, opened 2024-04-07
Removed Unnecessary Error Log

IllianiCBT:
### Current Implementation
Currently, if 0 Astechs are assigned to a task 0 Astech minutes are used and an error message is popped into the log to confirm that this has happened:

`02:50:44,277 ERROR [mekhq.campaign.Campaign] {AWT-EventQueue-0} mekhq.campaign.Campaign.getAvailableAstechs(Campaign.java:5332) - Tried to getAvailableAstechs with 0 minutes. Returning 0 Astechs.`

### Problem
Large Craft do not use Astechs, causing this error to fire for each Large Craft in the Hanger. Furthermore, this appears to be working as intended: if 0 Astechs are assigned to the task, then 0 Astech minutes should be used.

### Solution
I removed the error message, in the event that 0 Astechs are assigned to a task 0 Astech minutes will be used silently.

closes... #3981

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -5329,7 +5329,7 @@ public class Campaign implements ITechManager {
 
     public int getAvailableAstechs(final int minutes, final boolean alreadyOvertime) {
         if (minutes == 0) {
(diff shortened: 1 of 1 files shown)


AaronGullickson (approved):
LGTM

PR #3988, opened 2024-04-10
Added Scenario & Mission Tracking to Kills, Added Ability to Assign Kills to Scenario and/or Mission

IllianiCBT:
## Forward
Throughout this PR the term 'Mission' can be read as referring to both Missions and Contracts interchangeably. AutoMedals is the internal name for the in-dev project that will automatically prompt users to optionally assign medals to those employees that have earned them.

### Current Functionality
Three data points are recorded for Kills: Killer, Killed, and Date.

### Problem
The in-dev AutoMedals project needs to be able to identify Kills that occurred over the course of a Mission, and those that occurred over the course of a single Scenario.

### Solution
I added functions to allow for the Scenario ID and Mission ID numbers to be baked into the stored Kill data. This information is logged and stored, alongside the existing Kill data, in the users' save file.

<p align=center>
image
</p>

These IDs are drawn from existing functionality, all I did was hook them into the `Kill` class. That means that any future system that calls this information (such as AutoMedals) will be able to fetch the scenario and mission data just by calling `getScenario()` or `getMission()`.

I also added the ability to fetch the internal names of all scenarios on the campaign to the `Campaign` class, (`getScenarios()`) as that functionality was missing and I needed for this PR. The `getScenarios()` function is basically identical to its Mission equivalent `getMissions()`.

### Testing
I ran a whole bunch of Human v. Princess, and Princess v. Princess scenarios. Somewhere around 20-30 and did not find any instances where kills failed to be assigned a Scenario or Mission ID. This testing was conducted with manually created scenarios, as well as AtB and StratCon scenarios. The testing also accounted for both automatically and manually assigned kills.

### Current Implementation
Currently, the Add/Edit Kill dialog allows you to assign Killer, Killed, and Date. This is true, also, for the Assign Kill dialog.

<p align=center>
image
</p>

### Problem
This renders a Kill's associated Mission and Scenario IDs completely invisible to users. Furthermore, this means there is no ability for users to assign Missions and/or Scenarios to Kills obtained outside of mm. Such as kills related to RPG-focused campaigns, or any other source of Kills.

Furthermore, as shown in the above image, the dialogs are titled as if they were Scenario dialogs and not Kill.

### Solution
I updated the dialogs so that they're a little more visually streamlined. I also renamed the dialogs to 'Edit Kill Entry' and 'Add Kill Entry'.

More importantly, I added the ability to state what Scenario and/or Missions the kill occurred on. This functionality is extended to all future and prior kills. Kills that do not currently have a Scenario or Mission ID will default to the index 0 combobox option, which is blank. Kills can be saved with blank Scenario or Mission IDs, at which point the ID is saved as '0'. This ensures we're not introducing new `null` values, although as ID 0 is not a valid Scenario or Mission ID, this will need to be factored into any future code.

There was the possibility of saving the values as `null`, but that would likely break a whole bunch of things so I figured it was best avoided. The only kills which will have a Scenario/Mission ID of `null` will be those from campaigns originating from prior versions, the frequency of which will decrease with time.

When the campaign is saved, any Kills with `null` Mission/Scenario IDs will be assigned IDs of 0.

(Assign Kill/Add New Kill dialog)
<p align=center>
image
</p>

(Same dialog, but with options selected)
<p align=center>
image
</p>

(Edit Kill dialog)
<p align=center>
image
</p>

### Limitations
Currently, mhq does not track what Mission a Scenario was spawned by. This removes the ability to filter Scenarios based on what Mission was selected. Instead, I sorted the list so that the most recent scenario will always be at the top.

Potentially, this functionality could be added through a follow-up PR, but it's unlikely this is something I'll personally do. At least, I'm not adding it to my current list of projects.

### Testing
I have accounted for Missions/Scenarios being deleted, after the kill is assigned an ID tied to that  Mission/Scenario. In those events, the system recognizes it's not a valid ID and just treats it as if it were ID 0.

I also accounted for users trying to add kills before a Scenario and/or Mission has been added to the campaign. In these cases users can only select the index 0 option for Scenario and/or Mission, which sets the relevant ID to 0.

Editing a kill that has a null Mission/Scenario ID simply treats it as if the ID were 0.

### Recommendation
Once we have a release version that includes this PR, I recommend we advise users to save their campaigns immediately. This will populate all prior Kills with default Mission/Scenario IDs and reduce the amount of `null` IDs the system needs to deal with.

I'll make sure to factor into AutoMedal the possibility that it's run before these default values have been assigned, so that will catch whatever left. But saving an existing campaign to the new version, when transferring saves across versions, should probably be encouraged anyway.

### Current Implementation
Currently, all kill information is exported and imported across campaigns.

### Problem
This means that Scenario/Mission IDs are also transferred across campaigns. This would cause conflicts with the destination campaign, with kills being assigned to the wrong Scenario & Mission.

### Solution
When exporting personnel information, kills have Scenario and Mission IDs reset to 0.

Any `null` IDs are also set to 0, as this is functionally identical to saving the campaign.

All this means is that the Kill is decoupled from its assigned Scenario and Mission, the Kill still remains in the employees' personnel file and can be manually reassigned, if desired.

While, when exporting, there is the option to export completed Missions, this would still result in Scenario ID conflicts, as well as conflicts with any re-used Mission IDs.

If the Active Mission's ID is ever reused (because it wouldn't have been used in the Destination Campaign), Kills from the Active Mission (in the Origin Campaign) would then be assigned to this new Mission. That functionality is undesirable, so I deemed the best approach was to wipe the slate clean.

### Testing
I tested this functionality with campaigns containing Kills both with default (0) Mission/Scenario IDs; non-zero Mission/Scenario IDs, and with `null` Mission/Scenario IDs. None caused issue.

### Current Implementation
In [AddOrEditKillEntryDialog.java](https://github.com/MegaMek/mekhq/pull/3988#diff-d39cb11c2632662fccf6252558d7edf068821a328a807d45a4186cc4814266c3) all new GUI elements in `initComponents()` are placed at the top of the function.

### Problems
This forces a lot of scrolling up and down to find what elements are being created and when.

### Solution
I moved the element initialization to the top of the relevant block of Swing code.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Kill.java
+++ b/MekHQ/src/mekhq/campaign/Kill.java
@@ -41,15 +41,19 @@ public class Kill {
     private LocalDate date;
     private String killed;
     private String killer;
+    private int missionId;
+    private int scenarioId;
 
     public Kill() {
     }
 
-    public Kill(UUID id, String kill, String killer, LocalDate d) {
+    public Kill(UUID id, String kill, String killer, LocalDate d, int missionId, int scenarioId) {
         pilotId = id;
         this.killed = kill;
         this.killer = killer;
         date = d;
+        this.missionId = missionId;
+        this.scenarioId = scenarioId;
     }
 
     public UUID getPilotId() {
@@ -72,6 +76,14 @@ public class Kill {
         return killer;
     }
 
+    public int getMissionId() {
+        return missionId;
+    }
+
+    public int getScenarioId() {
+        return scenarioId;
+    }
+
     public void setDate(LocalDate d) {
         date = d;
     }
@@ -82,6 +94,14 @@ public class Kill {
 
     public void setKilledByWhat(String s) {
         killer = s;
+    }
+
+    public void setMissionId(int id) {
+        missionId = id;
+    }
+
+    public void setScenarioId(int id) {
+        scenarioId = id;
     }
 
     public static Kill generateInstanceFromXML(Node wn, Version version) {
@@ -100,6 +120,10 @@ public class Kill {
                     retVal.killer = wn2.getTextContent();
                 } else if (wn2.getNodeName().equalsIgnoreCase("date")) {
                     retVal.date = MHQXMLUtility.parseDate(wn2.getTextContent().trim());
+                } else if (wn2.getNodeName().equalsIgnoreCase("missionId")) {
+                    retVal.missionId = Integer.parseInt(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("scenarioId")) {
+                    retVal.scenarioId = Integer.parseInt(wn2.getTextContent());
                 }
             }
         } catch (Exception ex) {
@@ -117,11 +141,13 @@ public class Kill {
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "killed", killed);
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "killer", killer);
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "date", date);
+        MHQXMLUtility.writeSimpleXMLTag(pw, indent, "missionId", missionId);
+        MHQXMLUtility.writeSimpleXMLTag(pw, indent, "scenarioId", scenarioId);
         MHQXMLUtility.writeSimpleXMLCloseTag(pw, --indent, "kill");
     }
 
     @Override
     public Kill clone() {
-        return new Kill(getPilotId(), getWhatKilled(), getKilledByWhat(), getDate());
+        return new Kill(getPilotId(), getWhatKilled(), getKilledByWhat(), getDate(), getMissionId(), getScenarioId());
     }
 }
--- a/MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
@@ -31,6 +31,7 @@ import org.apache.logging.log4j.LogManager;
 import javax.swing.*;
 import java.awt.*;
 import java.awt.event.ActionEvent;
+import java.text.ParseException;
 import java.time.LocalDate;
 import java.util.*;
 
@@ -54,11 +55,12 @@ public class AddOrEditKillEntryDialog extends JDialog {
     private JTextField txtKill;
     private JLabel lblKiller;
     private JTextField txtKiller;
+    private JLabel lblDate;
     private JButton btnDate;
     private JLabel lblMissionId;
     private JTextField txtMissionId;
     private JLabel lblScenarioId;
-    private JTextField txtScenarioId;
+    private JSpinner spnScenarioId;
     private Campaign campaign;
 
     public AddOrEditKillEntryDialog(JFrame parent, boolean modal, UUID killerPerson, String killerUnit, LocalDate entryDate, Campaign campaign) {
@@ -80,7 +82,7 @@ public class AddOrEditKillEntryDialog extends JDialog {
         this.missionId = this.kill.getMissionId();
         this.scenarioId = this.kill.getScenarioId();
         this.operationType = operationType;
-        initComponents();
+        initComponents(operationType, campaign);
         setLocationRelativeTo(parent);
         setUserPreferences();
     }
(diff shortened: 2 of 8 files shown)


MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
```diff
@@ -47,29 +48,44 @@ public class AddOrEditKillEntryDialog extends JDialog {
...
-    private AddOrEditKillEntryDialog(JFrame parent, boolean modal, int operationType, Kill kill) {
+    private AddOrEditKillEntryDialog(JFrame parent, boolean modal, int operationType, Kill kill, Campaign c) {
         super(parent, modal);
 
+        campaign = c;
+
         this.frame = parent;
         this.kill = Objects.requireNonNull(kill);
         this.date = this.kill.getDate();
+        this.missionId = this.kill.getMissionId();
```
Sleet01, line 87: It doesn't look like there is good handling for null values here, which looks like a possibility if the default constructor were used.

MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
```diff
@@ -47,29 +48,44 @@ public class AddOrEditKillEntryDialog extends JDialog {
...
+    private ArrayList<Integer> missionIdList;
+    private JLabel lblScenarioId;
+    private MMComboBox<String> cboScenarioId;
+    private ArrayList<Integer> scenarioIdList;
 
-    public AddOrEditKillEntryDialog(JFrame parent, boolean modal, UUID killerPerson, String killerUnit, LocalDate entryDate) {
-        this(parent, modal, ADD_OPERATION, new Kill(killerPerson, "?", killerUnit, entryDate));
+    public AddOrEditKillEntryDialog(JFrame parent, boolean modal, UUID killerPerson, String killerUnit, LocalDate entryDate, Campaign campaign) {
+        // We default missionId & scenarioId to 0 when adding new kills
+        this(parent, modal, ADD_OPERATION, new Kill(killerPerson, "?", killerUnit, entryDate, 0, 0), campaign);
```
Sleet01, line 72: If missionId and ScenarioId were initialized to 0, a simpler constructor could be used here.

Sleet01 (approved):
Overall looks good, just a minor nitpick about the new member initialization values.

MekHQ/src/mekhq/campaign/Kill.java
```diff
@@ -41,15 +41,19 @@ public class Kill {
     private LocalDate date;
     private String killed;
     private String killer;
+    private int missionId;
+    private int scenarioId;
```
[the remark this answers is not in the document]
IllianiCBT: The primitive 'int' is assigned a value of 0 by default

https://stackoverflow.com/questions/27211713/why-is-an-integer-variable-assigned-null-by-default

image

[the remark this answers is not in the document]
Sleet01: :see_no_evil:

MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
IllianiCBT: So, I looked into this and (assuming I'm not missing your point entirely, a non-zero chance) I don't think the functionality is improved by using a constructor here. This is the only situation where we're using the default value of `0` and that's to ensure that when the manual 'add kill' dialog triggers it's not at risk of encountering a `null` value for Scenario of Mission ID, before the new value is set. The moment the manual Kill is confirmed that default value is only used if no other value is set.

`Kill.java` already has `missionID` and `scenarioID` initializing at 0.

MekHQ/src/mekhq/gui/dialog/AddOrEditKillEntryDialog.java
IllianiCBT: In the event the default constructor is used, `missionID` and `scenarioID` will == 0

MekHQ/src/mekhq/campaign/Kill.java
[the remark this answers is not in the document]
IllianiCBT: Don't worry, I only discovered this by complete accident. But I rely on it a lot here, because it allows for easy handling of older saves with null `missionID` and `scenarioID` values

PR #3991, opened 2024-04-10
Add all deployment variables to BotForce

AaronGullickson:
Currently, `BotForce` only tracks the starting edge with the variable `start` despite the fact that we now have several more variables that can be used to basically set up any deployment zone on a map. This PR brings in the additional deployment variables from `Player` in MegaMek to `BotForce` and assigns them in the `GameThread` and `AtBGameThread`. Given that the only place to currently set these parameters is story arcs, it doesn't have any practical use at the moment, but will be very useful to story arcs. Furthermore PR #3931 will be able to allow users to set these new variables in an updated `CustomizeScenarioDialog`.  So, this is preparatory work.

I did decide to change the name of the starting edge variable from `start` to `startingPos` to make it consistent with the variable names used in `Player` and to reduce variable name ambiguity. I refactored the get/set methods appropriately and I added a legacy check to to the XML reader for reverse compatability.

I also added a `deployRound` variable to `BotForce` that allows for setting the entire `BotForce` to deploy on a later round. This is not technically part of the deployment variables in `Player` but it is a nice convenience function. Otherwise, each individual entity in `BotForce` has to be edited for deployment round. I actually already implemented this as part of the Story Arcs PR #2997, but it made more sense to include it here for record keeping and thematically. This deployRound variable is ignored in `AtBGameThread` because there are other ways that deployment round is determined there.

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -439,8 +439,17 @@ public class AtBGameThread extends GameThread {
                 LogManager.getLogger().error("Could not configure bot " + botClient.getName());
             } else {
                 botClient.getLocalPlayer().setTeam(botForce.getTeam());
-                botClient.getLocalPlayer().setStartingPos(botForce.getStart());
 
+                //set deployment
+                botClient.getLocalPlayer().setStartingPos(botForce.getStartingPos());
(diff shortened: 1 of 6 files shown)


Sleet01 (approved):
LGTM

PR #3996, opened 2024-04-15
Add new player deployment variables to Scenario

AaronGullickson:
I should have probably added this as part of PR #3991. That PR added all the new deployment variables to BotForce for use. However, the only deployment variable for players is still the starting edge. This now adds all of those same deployment variables to Scenario so that they can be applied to the player's forces. 

The changes are:

* Add the deployment variable to scenario along with getter/setter methods. I also refactored `Scenario.getStart `and `Scenario.setStart` to `Scenario.getStartingPos` and `Scenario.setStartingPos`, respectively, for greater consistency in naming. I put in a backwards compatibility check in the XML reader to still get cases where `start` was used.
* Changed `GameThread` and `AtBGameThread` to set the new deployment variables for the player's forces.

Several other files were touched by the PR but these are all AtB Scenario types that used the `getStart` and `setStart` methods and so were affected by the refactoring.

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -168,7 +168,15 @@ public class AtBGameThread extends GameThread {
                 client.sendPlanetaryConditions(planetaryConditions);
                 Thread.sleep(MekHQ.getMHQOptions().getStartGameDelay());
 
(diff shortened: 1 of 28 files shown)


Sleet01 (approved):
LGTM!

Issue #3995, opened 2024-04-15
[0.49.19 Nightly] RFE - Set MekHQ UI default to Flat Darcula to match MegaMek and MegaMekLab defaults


PhoenixHeart512:
### Environment
0.49.19 Nightly 4/15/2024

### Description
As it says in the title, can we set the mekHQ default to Flat Darcula so that it is consistent across all 3 programs in the suite?

HammerGS:
For Reference this PR was used in MM and MML

https://github.com/MegaMek/megamek/pull/5209

Sleet01:
Implementation has been merged.

PR #3997, opened 2024-04-15
Implements RFE #3995: Set default theme to overall suite default at startup... 

Sleet01:
plus allow DateChooser resizing if theme does not fit right (as is the case with Flat Darcula LaF).

This sets the default MekHQ UI theme to the GUIPreferences default theme at initial startup, which will be Flat Darcula currently.

New startup theme:
image

Testing:
- Started MekHQ with and without `mhq.preferences` in /mmconf.
- Tested changing and saving preferences with default theme loaded from GUIPreferences (currently this is Flat Darcula) at startup.
- Ran all three projects' unit tests


HammerGS:
Works for me.

PR #4009, opened 2024-04-17
Fix MHQ #4002: prevent infinite loop while assigning SPAs by removing unusable entries

### LINKED ISSUE #4002: 49.19 Nightly - Soft Freeze upon advancing day.
EDIT2: Ok - well - its not even the ghost pilot issue maybe.  What worked last night - deleting the pilot or mech - inexplicably does not work this morning with the exact same process.  Frustrated, I reloaded nightly from 4.14 and loaded same save and was able to advance to monday.  So I presume now, probably incorrectly, that this is something new that was introduced by the merges yesterday.  Regardless, Im going to let someone smarter than me figure it out. Im stumped.

EDIT:  Looks like this may be related to the infamous Ghost Pilot issue.  Normally it goes away upon reload, but I had mothballed the Warhammer that he died in so I guess that preserved whatever the bug is.  Illiani isolated it to a mech or Pilot and on a hunch I checked the recently departed and his mech.  Deleting either of them allowed it to advance to Monday.

Will close this and open a new issue.  Can re-open later if needed.

Title - was advancing day. Not sure if it was from a Saturday to Sunday or Sunday to Monday.  i would guess the latter, but not certain.

EDIT:  Looks like I got an autosave, so I guess it was advancing from Sunday to Monday when it froze.  So might be some weekly roll shenanigans.  I dont see anything in the log, but it is attached.  Will try to load that autosave after I post this.  

It did not give me a NPE or other Error pop up, it just became unresponsive.  I could move my mouse and switch to other windows outside of MHQ, but none of the menus or tabs or anything in MHQ could be clicked on.  Its almost as if there is an invisibile pop-up menu waiting to be clicked and rendering the rest unresponsive.  Alt tabbing and minimizing everything else to try to find a buried window did no good.  

Log and autosave and save before autosave and customs attached.

image

[mekhq.log](https://github.com/MegaMek/mekhq/files/14986571/mekhq.log)

[Autosave-1-Crazed Profanity-30700220.cpnx.gz](https://github.com/MegaMek/mekhq/files/14986593/Autosave-1-Crazed.Profanity-30700220.cpnx.gz)

[Crazed Profanity30700218a.cpnx.gz](https://github.com/MegaMek/mekhq/files/14986595/Crazed.Profanity30700218a.cpnx.gz)

[Customs.zip](https://github.com/MegaMek/mekhq/files/14986600/Customs.zip)

Sleet01:
Note: this _may_ fix #4002, or may only fix something that may be contributing to it.

Prevent infinite loop by reducing choices each time we `continue` the while loop.
Also remove a check for more elements in some SPAs that, honestly, I don't remember writing - it actively prevents some SPAs from being applied even if they are valid.

Testing:
- Tested with failing save from #4002;
- Add unit tests for large number of consecutive calls to the previously-failing function;
- Ran all 3 projects' unit tests as well.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
+++ b/MekHQ/src/mekhq/campaign/mission/CrewSkillUpgrader.java
@@ -171,8 +171,9 @@ public class CrewSkillUpgrader {
                             return 0;
                             // If we can't access the option, try a different one
                         } else if ((entity.getCrew().getOptions() == null) ||
-                                (entity.getCrew().getOptions(spa.getName()) == null) ||
-                                (!entity.getCrew().getOptions(spa.getName()).hasMoreElements())) {
+                                (entity.getCrew().getOptions(spa.getName()) == null)) {
+                            // Make sure to remove choices we can't use
+                            choices.remove(spaIndex);
                             continue;
                         }
 
@@ -182,6 +183,8 @@ public class CrewSkillUpgrader {
                             return spa.getCost();
                         } catch (NullPointerException e) {
                             LogManager.getLogger().warn("Attempted to assign SPA '" + spa.getName() + "' but SPA not found.");
+                            // Make sure to remove choices we can't use
+                            choices.remove(spaIndex);
                             continue;
                         }
                 }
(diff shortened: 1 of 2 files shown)


kuronekochomusuke (approved):
this fixed the freeze I was getting following IllianiCBT's steps to reproduce the issue

PR #4037, opened 2024-04-29
Better Randomness in BotForceRandomizer

AaronGullickson:
This PR makes two changes to `BotForceRandomizer`:

1. Adds a "Generic BV" option for balancing based on MegaMek/megamek#5431 (which will need to be merged first). This is now set as the default balancing option in `BotForceRandomizer`. 
2. Sets an upper and lower bounds on matching that will keep generated forces from overshooting.

The overshooting problem in (2) is a major problem. The current approach only implements a `while` loop that checks to see if the targeted number of "points" (e.g. BV, Generic BV, weight) have been reached but has no checks to see if the final generated force is too large. Because the while loop generates entities in chunks equal to the prespecified "lance" size, its very easy to go over. A common situation, would be that the player has a lance of mechs and the first time through the loop, the randomizer generates a lance of mechs that are just a little bit short of the target number of points. It then dutifully proceeds through another loop and generates a whole additional lance, which will almost guarantee it to overshoot its mark. 

Instead, this PR sets an upper and lower bounds on the target value of points. The default is +- 5% as a starting point but this is a variable in `BotForceRandomizer` that could be changed. Basically, it does this through a double while loop where it throws out generated forces that are outside the parameters. 

Since, we don't want to get caught in an infinite loop, the randomizer will also check the number of times it has iterated and when it hits multiples of 20 times, it will widen the lower and upper bounds by an additional 50%. So, if the bounds are set as +-5% and you get to 20 iterations, it will widen the bounds to +-7.5%. It will also decide to give up if it gets to 99 iterations with no success and the player will get an empty generation. This means that at the default starting bounds of +-5%, it will ultimately get to roughly +-25% before quitting ($5*1.5^4$). 

I have tested the results across a variety of parameters (using my scenario dialog update PR) and they do ensure that forces stay within a reasonable range and allow for significant variation in opfor when using generic BV.

Currently, this PR only affects Story Arcs, as AtB/StratCon still uses `AtBDynamicScenarioFactory` to generate OpFor. Ultimately, I want to also get AtB/StratCon using `BotForceRandomizer`, but that is a bigger project than I want to take on in this single PR.


AaronGullickson:
I am going to convert this back to draft as I think I will take this opportunity to also have BotForceRandomizer use one of the better randomizers from MegaMek (likely RAT Generator of Formation Builder).

AaronGullickson:
Ok, never mind. It looks like its already using RATGenerator and this would be a good bit of additional work that would likely mean new parameters, which would make my updates to the CustomizeScenarioDialog out of date. So I would rather get this  PR and #3931 merged and then look into changing the generator in a separate PR.

sldfgunslinger2766:
More options = better 💪

SJuliez (approved):

PR #4052, opened 2024-05-04
move modifiedTemperature to AtBScenario

### LINKED ISSUE #4051: Temperature is incorrectly set in Scenario
The default temperature setting in `Scenario` is 25. However all scenarios are defaulting to a temperature of zero instead. This is because planetary conditions are actually being set from a variable called `modifiedTemperature` rather than `temperature`. This `modifiedTemperature` variable however is never set in Scenario and so it gets the default integer value of zero. I believe this code is part of @kuronekochomusuke's work on varying planetary conditions in AtB scenarios. It is a classic example of AtB logic interfering with generic classes.

This could be easily fixed by just adding `modifiedTemperature = 25` to the `Scenario` constructor. However, I would like to think more about why we need both a `temperature` and a `modifiedTemperature` variable at all and whether there might be a better way to fix this issue. I find it confusing and non-intuitive and I don't know how it all works together. If temperature is being modified somehow, why not just write the modified temperature back onto the `temperature` variable?

kuronekochomusuke:
- move modifiedTemperature to AtBScenario
- save and load modifiedTemperature to xml

fixes #4051

### DIFF
--- a/MekHQ/src/mekhq/GameThread.java
+++ b/MekHQ/src/mekhq/GameThread.java
@@ -180,7 +180,7 @@ class GameThread extends Thread implements CloseClientListener {
                 planetaryConditions.setWind(scenario.getWind());
                 planetaryConditions.setFog(scenario.getFog());
                 planetaryConditions.setAtmosphere(scenario.getAtmosphere());
-                planetaryConditions.setTemperature(scenario.getModifiedTemperature());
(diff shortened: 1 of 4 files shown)


AaronGullickson (commented):
LGTM

HammerGS (approved):

AaronGullickson (approved):

PR #4059, opened 2024-05-07
Fix unit tests on Large Craft Bays

AaronGullickson:
It works! All green. I think some of this is correcting mistakes from my last PR plus a few other tests that were failing.

### DIFF
--- a/MekHQ/unittests/mekhq/campaign/parts/equipment/EquipmentPartTest.java
+++ b/MekHQ/unittests/mekhq/campaign/parts/equipment/EquipmentPartTest.java
@@ -1435,7 +1435,7 @@ public class EquipmentPartTest {
 
         int location = SmallCraft.LOC_HULL;
         int equipmentNum = 42;
-        Mounted mounted = mock(Mounted.class);
+        WeaponMounted mounted = mock(WeaponMounted.class);
         when(mounted.getLocation()).thenReturn(location);
         doAnswer(inv -> {
             when(mounted.isDestroyed()).thenReturn(true);
@@ -1449,8 +1449,7 @@ public class EquipmentPartTest {
         when(weaponBay.getLocation()).thenReturn(location);
         when(weaponBay.getType()).thenReturn(bayWeaponType);
         List<WeaponMounted> bayWeapons = new ArrayList<>();
-        //bayWeapons.addElement(33);
-        bayWeapons.add(weaponBay);
+        bayWeapons.add(mounted);
         when(weaponBay.getBayWeapons()).thenReturn(bayWeapons);
         doReturn(weaponBay).when(entity).getEquipment(eq(bayEqNum));
         doReturn(bayEqNum).when(entity).getEquipmentNum(eq(weaponBay));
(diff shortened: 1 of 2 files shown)


MekHQ/unittests/mekhq/campaign/parts/equipment/EquipmentPartTest.java
```diff
@@ -1538,7 +1537,7 @@ public void checkWeaponBayWeaponRemovedOthersOkayTest() {
         when(weaponBay.getType()).thenReturn(bayWeaponType);
         List<WeaponMounted> bayWeapons = new ArrayList<>();
         bayWeapons.add(otherMounted);
-        bayWeapons.add(weaponBay);
+        bayWeapons.add(mounted);
```
AaronGullickson, line 1540: In several cases, this was the problem - the weaponBay was being added as a weapon of itself, instead of `mounted` which is the key equipment being tested.

MekHQ/unittests/mekhq/campaign/parts/equipment/LargeCraftAmmoBinTest.java
```diff
@@ -208,11 +208,11 @@ public void getBayReturnsBayMatchingEqNum() {
         when(bay.getBayAmmo()).thenReturn(bayAmmo);
         when(entity.getEquipment(bayNum)).thenReturn((Mounted) bay);
 
+        ammoBin.setUnit(unit);
+
         // Set the bay as if we're in deserialization code
         ammoBin.setBay(bayNum);
 
-        ammoBin.setUnit(unit);
-
```
AaronGullickson, line 215: The unit has to be set before you add the bay number.

MekHQ/unittests/mekhq/campaign/parts/equipment/LargeCraftAmmoBinTest.java
```diff
@@ -276,6 +276,7 @@ public void getBayFallsBackToMatchingWeaponBayIfNoBayEquipment() {
         weaponBays.add(wrongWeaponBay);
         weaponBays.add(weaponBay);
         when(entity.getWeaponBayList()).thenReturn(weaponBays);
+        when(entity.whichBay(equipmentNum)).thenReturn((WeaponMounted) weaponBay);
```
AaronGullickson, line 279: I don't *love* this solution because `entity.whichBay` should already work because it calls `entity.getWeaponBayList` which is mocked in line 278. However, on the debugger, that method was returning null within the `whichBay` function. I think its ok?

MekHQ/unittests/mekhq/campaign/parts/equipment/EquipmentPartTest.java
```diff
@@ -1449,8 +1449,7 @@ public void checkWeaponBayOnlyWeaponRemovedTest() {
         when(weaponBay.getLocation()).thenReturn(location);
         when(weaponBay.getType()).thenReturn(bayWeaponType);
         List<WeaponMounted> bayWeapons = new ArrayList<>();
-        //bayWeapons.addElement(33);
-        bayWeapons.add(weaponBay);
+        bayWeapons.add(mounted);
```
AaronGullickson, line 1452: Not sure what 33 was supposed to be as this equipment number is not used in this test and its supposed to be testing a weapon bay with only one weapon.

SJuliez (approved):
If it works, it works.

PR #4066, opened 2024-05-08
if planetary conditions are disabled in options don't send conditions …

### LINKED ISSUE #4062: StratCon is generating missions with weather modifiers when they are disabled
### Environment
MekHQ: 49.19
Java Version: 11.0.16

### Description
I think these screenshots explain everything.
image
image

### Files
I can upload the campaign and my custom unit files if needed, but this seems to be happening all the time as I have seen it in two different campaigns.

kuronekochomusuke:
- if planetary conditions are disabled in campaign options, don't send conditions to megamek

fixes #4062

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -158,15 +158,21 @@ public class AtBGameThread extends GameThread {
                 Thread.sleep(MekHQ.getMHQOptions().getStartGameDelay());
 
                 PlanetaryConditions planetaryConditions = new PlanetaryConditions();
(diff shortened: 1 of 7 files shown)


Sleet01:
Looks like `testFactionBorderTrackerAllPlanets` is failing, but only under Temurin 17; this usually indicates a spy or mock issue due to Java 17's more restrictive reflection rules.

One option is to rewrite the test, but the other is to annotate it to be ignored _only_ under Temurin 17:

```@DisabledForJreRange(min = JRE.JAVA_17, max = JRE.OTHER)```

Sleet01 (approved):
Looks good to me.

PR #4082, opened 2024-05-15
Updated Tooltips for the Marriageable & Trying to Conceive Flags

### LINKED ISSUE #4080: RFE: Remove Child Marriage and Knocked Up Children Flags
Not sure it works this way, as I've not seen any pregnant and married 4 year olds, but it looks like they can be flagged for marriage and pregnancy by default. 
Screenshot 2024-05-14 122538
I have not changed any settings on this person. There may be other calculations which prevent the marriage calculation and the pregnancy calculation from picking these up. It does look like I cannot force them to be pregnant until they turn 14, which I guess is good. 
Maybe mask the flag until target age or just don't care since it appears to be only a flag problem.

IllianiCBT:
This PR modifies the 'Marriageable' and 'Trying to Conceive' flags to explicitly state that underage personnel will not be considered eligible for marriage or reproduction, even if these flags are enabled.

I also changed these tooltips to be multiline tooltips, as they're a little wordy.

closes #4080

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
@@ -2165,7 +2165,8 @@ public class PersonnelTableMouseAdapter extends JPopupMenuAdapter {
                 || !gui.getCampaign().getCampaignOptions().getRandomMarriageMethod().isNone())
                 && Stream.of(selected).allMatch(p -> p.isMarriageable() == person.isMarriageable())) {
             cbMenuItem = new JCheckBoxMenuItem(resources.getString("miMarriageable.text"));
-            cbMenuItem.setToolTipText(resources.getString("miMarriageable.toolTipText"));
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -2165,7 +2165,8 @@ protected Optional<JPopupMenu> createPopupMenu() {
                 || !gui.getCampaign().getCampaignOptions().getRandomMarriageMethod().isNone())
                 && Stream.of(selected).allMatch(p -> p.isMarriageable() == person.isMarriageable())) {
             cbMenuItem = new JCheckBoxMenuItem(resources.getString("miMarriageable.text"));
-            cbMenuItem.setToolTipText(resources.getString("miMarriageable.toolTipText"));
+            cbMenuItem.setToolTipText(MultiLineTooltip.splitToolTip(resources.getString("miMarriageable.toolTipText")
```
SJuliez, line 2168: The replaceAll("0") seems a bit artificial. There's an example for using a String.format in line 512

SJuliez (approved):
I'd suggest changing that call. Otherwise allgood.

PR #4087, opened 2024-05-16
Reintegrate Pay Multiplier into Salary Calculation

IllianiCBT:
At some point, I've no idea when, the 'rank multiplier' got snipped out of the salary calculation. This means that salary rank multipliers weren't applying.

This PR restores that functionality.


Sleet01 (approved):
LGTM

PR #4094, opened 2024-05-21
MegaMek code adaptation

SJuliez:
I created a problem with the "MegamekController" aka KeyDispatcher not being available from MHQ as MHQ did not create a MegaMekGUI. This corrects the problem. I was able to start a story arc mission without problems.


Sleet01 (approved):
Looks good to me.

PR #4095, opened 2024-05-21
Restored ranks.xml

IllianiCBT:
Restored `ranks.xml` in the MekHQ/userdata/data/universe path. The file was accidentally deleted in a prior PR.

This prevents an error in mhq log when mhq is first run.

Turns out 'delete' in GitHub doesn't mean 'discard this change'. Now I know!


Sleet01 (approved):
LGTM!

PR #4102, opened 2024-05-23
Fixed Award Tier Count Calculations in PersonViewPanel

IllianiCBT:
This commit updates the calculation of `awardTierCount` in the `PersonViewPanel` class.

The new logic determines `awardTierCount` as the minimum between the number of medal files and the maximum of either 1 or the number of awards divided by the `AwardTierSize` specified in the campaign options.

These changes affect ribbons, medals, and miscellaneous awards.

This commit fixes a bug reported on Discord, where the award image getters were counting from 1 instead of 0. Additionally, extra binders have been added to ensure the index is always within bounds.


Sleet01 (approved):
Looks good.

PR #4104, opened 2024-05-23
Fixed Post-Scenario Tracking System's Handling of Multiple Personnel in autoAwards

IllianiCBT:
This commit refactors the automatic awards tracking system to handle multiple personnel simultaneously, instead of one-by-one. This is achieved by changing the `person` parameter to a `HashMap` of UUIDs and injury counts in the `AutoAwardsController`'s `PostScenarioController`. Associated methods have been adjusted to handle the `HashMap` data, and the retrieval of kills has been consolidated within methods where necessary.

This resolves a bug discovered by our QA team, where faulty logic was causing auto-awards to trigger multiple times during post-scenario conditions.


Sleet01 (approved):
LGTM

PR #4105, opened 2024-05-23
Updated Post-Scenario Logging for Prisoners

IllianiCBT:
This commit introduces a new `ServiceLogger` method to post to the personnel log when a character is captured in a scenario, instead of logging to the scenario log. This ensures that prisoners do not receive scenario credit for the scenario in which they were captured. The change modifies the existing log behavior and includes a new log entry text in the relevant resources file.

Logging in the personnel log ensures there is still a record of the scenario and mission, this feature was requested on Discord when removing the scenario log was discussed.

This PR addresses an oversight reported by our QA team, where prisoners were incorrectly receiving award credit for the scenario in which they were captured.


Sleet01 (approved):
LGTM

PR #4106, opened 2024-05-24
Use tabs in customize scenario dialog

AaronGullickson:
The new CustomizeScenarioDialog was quite large which can cause problems on small monitors. So, I redid the layout using tabs to make the overall dialog smaller, as shown in the attached images.

Screenshot 2024-05-24 at 11 46 28 AM
Screenshot 2024-05-24 at 11 46 31 AM
Screenshot 2024-05-24 at 11 46 33 AM

I also fixed a small problem on CustomizeScenarioObjectivesDialog where I was using the wrong label.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
@@ -206,7 +206,11 @@ public class CustomizeScenarioDialog extends JDialog {
             setTitle(resourceMap.getString("title"));
         }
 
-        JPanel panMain = new JPanel(new GridBagLayout());
+        JTabbedPane panTabs = new JTabbedPane();
+
+        JPanel panBasic = new JPanel(new GridBagLayout());
+        JPanel panRewards = new JPanel(new GridLayout(2, 0));
+
         JPanel panInfo = new JPanel(new GridBagLayout());
         JPanel panWrite = new JPanel(new GridBagLayout());
         JPanel panBtn = new JPanel(new FlowLayout());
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
```diff
@@ -331,11 +336,8 @@ public Component getListCellRendererComponent(final JList<?> list, final Object
                 BorderFactory.createEmptyBorder(5,5,5,5)));
 
         initOtherForcesPanel(resourceMap);
-        panOtherForces.setBorder(BorderFactory.createCompoundBorder(
-                BorderFactory.createEmptyBorder(0, 0, 10, 0),
-                BorderFactory.createTitledBorder(resourceMap.getString("panOtherForces.title"))));
-        panOtherForces.setPreferredSize(new Dimension(600,250));
-        panOtherForces.setMinimumSize(new Dimension(600,250));
+        panOtherForces.setPreferredSize(new Dimension(600,300));
```
Sleet01, line 339: I think we'd talked previously about making the MekHQ UI more responsive; would it be possible to use a persistent size, or a % of screen size, or something other than magic numbers here?

Sleet01 (approved):
Looks good, although if we are going to try to make MHQ's UI more responsive, we could start with new UI elements...

MekHQ/src/mekhq/gui/dialog/CustomizeScenarioDialog.java
AaronGullickson: Looks like it already got merged, but I can think about the issue in future dialog stuff.

PR #4109, opened 2024-05-25
Fix Hyperspace Navigation & Small Arms Parsing in Education Module

IllianiCBT:
This fixes the incorrect hyperspace navigation and small arms parsing within academy.java


HammerGS (approved):
Tested and seems to be working.

PR #4114, opened 2024-05-26
Additional Education Module Bug Fixes

IllianiCBT:
This PR addresses a number of bugs that were picked up by QA.

- Corrected faulty logic in Max Jump Count check
- Fixed Max Jump Count setting failing to save user choices. This is why the above logic error went undetected.
- Addressed users not understanding why Trueborn students only have a travel time of 2 days (via Campaign Option to enable/disable this)
- Fixed formatting errors in clan washout text
- Fixed missed course index assignment causing problems with clan graduation and (potentially) skill increases.


Sleet01 (approved):
LGTM

PR #4115, opened 2024-05-26
Fixed Award Image Display in Person View

IllianiCBT:
This is another quick fix to the display of award ribbons, medals, and miscs in the person view panel. My previous fix appeared to resolve the issue, but instead it failed whenever 'person' had more than one issuance of the award. This adds better bounds by using the `MathUtility.clamp()`.


Sleet01 (approved):
LGTM

PR #4117, opened 2024-05-27
Even More Education Module Bug Fixes

IllianiCBT:
This PR resolves a number of bugs picked up by QA.

- Fixed string formatting for a number of cases that go missed previously. They should all have been caught now.
- Fixed faction discount not applying correctly for Local Academies
- Fixed out-of-range error when applying Skill Bonuses during graduation
- Streamlined and fixed the journey home portion of education processing
- Added better reporting of the journey home to Person View (this wasn't a bug, but improves UX)


Sleet01 (approved):
Looks good.  I think the missing `return` lines got added in multiple PRs, but Git should figure that out.

PR #4122, opened 2024-05-29
Removed crew fatigue setting from Unit

IllianiCBT:
The most recent batch of merges came with some oddities. Somehow an import got lost from EducationController.java and some rogue code snuck in from my Turnover and Retention module. That, at least, is probably mea culpa (even if I don't know _how_ those couple of lines snuck in).

- Removed crew fatigue setting from Unit.java
- Imported missing PlanetarySystem import in EducationController.java


Sleet01 (approved):
LGTM.  Merging to fix nightlies.

PR #4124, opened 2024-05-30
Fixed Award Eligibility Tracker Issuing Double Scenario Kill Awards... again

IllianiCBT:
Removed some extra code that was causing autoAwards to count personnel award eligibility twice for Scenario Kill Awards. I also added a missing line that would have caused Misc Awards to appear on the wrong award eligibility page.


Sleet01 (approved):
LGTM

PR #4134, opened 2024-06-03
Updated Labels and Tooltips for Better Clarity and Grammar

IllianiCBT:
- Removed trailing dots for a cleaner GUI
- Updated tooltips and text fields within a handful of ResourcesBundles
- All text grammatically reviewed.
- Removed a cuss from one of our in-code comments


SJuliez (approved):
I'd like to note that trailing dots indicate that a menu option is not a direct action but opens a dialog to input further settings. So it is a valid design choice. I have no strong feelings on that so I'll leave it as a comment.

IntelliJ IDEA uses that:
image
So does Notepad++
image

SJuliez:
Any thoughts on those dots?

IllianiCBT:
> Any thoughts on those dots?

Completely forgot to do this: all restored

SJuliez (approved):

PR #4029, opened 2024-04-24
Linked StratCon OpFor Generation to AtB Difficulty Options, Added OpFor Generation Balance Checks

IllianiBird:
### Current Implementation
StratCon generates opposing forces based on player force Battle Value (BV), ensuring a minimum percentage based on Unit Rating and a unit allowance (based on lances or regional equivalents). If the generated force exceeds the unit allowance, StratCon trims it down until it falls within the allowance.

The current system assumes that player-deployed units are roughly equivalent in value to those generated by StratCon. However, the minimum BV percentage, ranging from 50-100% based on company rating, often results in opposing forces surpassing the player's BV. Additionally, StratCon only adjusts the OpFor unit count if it exceeds the unit allowance, not the BV allowance, leading to challenges with Clan OpFor encounters.

Moreover, the lack of utilization of the "Against the Bot" (AtB) difficulty option means players have no control over the size of opposing forces they face. This issue is compounded by the absence of tools to mitigate the frequency of facing multiple opposing forces due to Scenario modifiers.

### Solutions
**1. BV Allowance Adjustment:** Tying StratCon's BV allowance to the AtB difficulty option introduces scalability.

**2. BV Minimum Reduction:** Reducing BV minimums with a bell curve ensures more varied opposing forces while preventing campaigns at higher company ratings from becoming overly challenging.

**3. Random Unit Trimming:** Introducing a step to randomly remove units ensures opposing forces remain balanced and within a reasonable margin above the BV allowance (125%).

### Campaign Option Customization
In Campaign Options, three new options have been introduced to enhance control over scenario modifiers:

<p align=center>
image
</p>

**1. Maximum:** This option determines the maximum number of random Scenario Mods that can be added to a scenario. Note that fixed modifiers, such as those from StratCon Facilities, remain unaffected by this option.

**2. Chance:** The Chance option represents the percentile chance of a scenario having a random modifier. If a random modifier is rolled, the system will attempt to roll another. This process continues until the system fails to roll a modifier or the maximum number of modifiers is reached.

**3. BV Cap:** The BV Cap option sets the BV Allowance for Scenario Modifiers to the specified percentage. Additionally, a new Scenario Modifier tag (overrideBvCap) has been introduced to override this setting where appropriate. If a scenario modifier overrides the BV Cap, a message is printed to `mekhq.log` to assist in troubleshooting, particularly if users believe their opposing forces are being over-BV'd.

### Closes
closes #4011
closes #4028

PR #4138, opened 2024-06-03
Added Scenario Modifier Options to Campaign Options

### LINKED ISSUE #4028: RFE:  Stratcon.  Make Minimum/Maximum # of Scenario Modifiers Configurable
Title.  There are people like me that want a difficult experience.  And others that want an less intense one.  Modifiers are one of the best ways to control difficultiy.  Right now, up to 3 can be rolled, but according to a code comment, it was always intended that this number be user configurable.  

So I would suggest allow it to be configurable betwee 1-5 in MHQ Campaign Options.

Also, Some modifiers are scenario objectives and should always be allowed (like destroy X base turrets) and some are not randomly rolled, but added by facilities, so they need to be allowed. (OR - also allow the % chance of a facility to add its modifier to a local battle could also be configurable (right now its a flat 75% chance I think).  Allowing people to change it to whatever number they want.

I believe this is the code for the number of random modifiers.  

This would partially address #4011. 

image

image

image

IllianiCBT:
This PR adds the scenario modifier options from the now defunct #4029.

## Closes
Closes #4028


SJuliez (approved):

IllianiCBT:
Conflicts Resolved

PR #4127, opened 2024-06-01
Fixed Clan Creche Graduation in Education Module

IllianiCBT:
This update resolves a typo in the clan creche graduation process. I also removed a number of 'TODO's that got missed by #4118.

This PR must be merged after #4118.

PR #4126, opened 2024-05-31
Added Ability to Manually Edit Education Level to Edit Person View & Fixed Minor Bugs

IllianiCBT:
Added Education Level to CustomizePersonDialog and CreateCharacterDialog. Users can now set the education level of characters through the edit person dialog.

I also applied a couple of minor corrections to the prestigious academies xml, and added a condition to a academy tooltip method call that should result in more robust tooltip creation for the education module.

PR #4139, opened 2024-06-03
Added Second Chance Caste to the Education Module for Warrior Caste Washouts

IllianiCBT:
Requires: #4127 & #4126

This PR adds the ability to select a 'second chance' Warrior Caste sub-caste for Warrior Caste washouts.

When Warrior Caste personnel washout of the Warrior Caste mhq will check that their course index does not match the index of the chosen second chance caste. If there is no match (and second chance is not set to `NONE`), then the aspirant washes into the chosen sub-caste.

If second chance is set to `NONE`, or the aspirant is washing out of the chosen sub-caste, they will washout into a civilian caste as normal.


MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
```diff
@@ -1610,6 +1677,27 @@ private static String graduationEventPicker() {
...
+    private static int getSecondChanceCasteIndex(Campaign campaign) {
+        switch (campaign.getCampaignOptions().getSecondChanceCaste()) {
+            case NONE:
+                return -1;
+            case BA:
+                return 4;
+            case INFANTRY:
+                return 5;
+            case VEHICLE:
+                return 6;
```
SJuliez, line 1695: This looks like it could be shortened by a default: return -1, right?

SJuliez (approved):

IllianiCBT:
Resolved Conflicts

SJuliez:
Yet again, conflicts :)
Also, see my comment. Can get rid of the NONE: case and end in default: return -1;

IllianiCBT:
Addressed feedback and resolved conflicts

PR #4142, opened 2024-06-03
Added No Commander Nag Dialog

IllianiCBT:
- Introduced new Nag dialog for when there's no overall assigned commander in the campaign
- Implemented checkbox option to ignore this reminder
- Updated GUI properties file with No Commander Nag related texts
- Optimized untreated injury check in UntreatedPersonnelNagDialog

Given how much of the turnover and retention module; and upcoming morale module; utilize the overall force commander, it seemed reasonable to have a nag that reminds users to set one.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/nagDialogs/NoCommanderNagDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/nagDialogs/NoCommanderNagDialog.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2021 - 2024 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/gui/dialog/nagDialogs/NoCommanderNagDialog.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2021 - The MegaMek Team. All Rights Reserved.
```
SJuliez, line 2: hehe might make this 2024

SJuliez (approved):

MekHQ/src/mekhq/gui/dialog/nagDialogs/NoCommanderNagDialog.java
IllianiCBT: Yeah, that would be helpful wouldn't it? Fixed,

PR #4143, opened 2024-06-04
Adjusted Founder Turnover Modifier & Updated Documentation

IllianiCBT:
Requires #4101

The turnover target number modifier for Founders has been buffed from +2 to +4.

Documentation has been updated to include this and the introduction of a 'Pay Bonus Default' setting that was accidentally pushed to #4101.


MekHQ/src/mekhq/campaign/personnel/enums/PersonnelStatus.java
```diff
@@ -224,6 +246,38 @@ public static PersonnelStatus parseFromString(final String text) {
...
+                case 17:
+                    return OLD_AGE;
+                case 18:
+                    return MEDICAL_COMPLICATIONS;
+                case 19:
+                    return PREGNANCY_COMPLICATIONS;
+                case 20:
+                    return UNDETERMINED;
+                case 21:
+                    return SUICIDE;
```
SJuliez, line 280: ever play Rimworld? There's also FOOD_BINGE and TANTRUM
:D

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4732,6 +4732,30 @@ public void cancelMothballOrActivation() {
     }
     //endregion Mothballing/Activation
 
+    /**
+     * Returns a list of all soldiers in a unit.
+     *
+     * @return a list of Person objects representing the soldiers.
+     */
+    public List<Person> getSoldiers() {
```
SJuliez, line 4740: I suggest giving these methods good titles and/or comprehensive comments. We have many such methods in MM for equipment (hasWorkingMisc...) and it's never quite clear from title/comment what they really do. Might e.g. place the // comments into the javadoc. And, maybe state that soldiers are Infantry personnel (which includes BA)
Also, I believe in a case like this, the @return text alone would be alright. No need to repeat it as in l.4736 unless there's additional info somewhere.

SJuliez (approved):

MekHQ/src/mekhq/campaign/personnel/enums/PersonnelStatus.java
IllianiCBT: Ironically, given I've become the MechSims guy, I don't actually _like_ playing games like Rimworld or Dwarf Fortress. Though I do enjoy listening to people recount their Dwarf Fortress misadventures.

IllianiCBT:
Addressed requested changes

Issue #2238, opened 2020-11-21
47.14 Graphical issues in Personnel Log


gsparks3:
### Environment
47.14
Windows 10
Java 11.0.8

### Description
When scrolling up and down in the Personnel Log on the Personnel tab, often entries will have significant graphical rendering issues. Example:
![image](https://user-images.githubusercontent.com/59574316/99883716-3e798580-2bf7-11eb-9909-a4fec61d7ce8.png)
This seems to clear up if an entry in each section (Personnel Log, Mission Log, Kill Log) is clicked on (each section will resolve the graphical issue individually as an entry is selected). However, switching away from the selected personnel and then back will cause the issue to recur upon scrolling once more. It can be resolved again the same way as before. 

Please check if this issue is related to the Java version in use - I seem to recall not seeing it while using Java 8, but I could be wrong. 

### Files
A campaign file is provided mostly so that one with a good number of personnel log entries is available. This issue has been present for quite some time, across multiple versions, and seems to be campaign-agnostic. 
[Greyscales.cpnx.gz](https://github.com/MegaMek/mekhq/files/5578104/Greyscales.cpnx.gz)

Windchild292:
I've seen this in Java 11 and 13.

IllianiCBT:
This is a java issue and not something we can resolve. Closing.

PR #4145, opened 2024-06-04
Reversed Order of Personnel Logs, Renamed 'Personnel Log' to 'Personal Log'

### LINKED ISSUE #4032: RFE 49.19: UX/UI - optionally display personnel, scenario and kill logs by date descending instead of ascending.
# Request
When going through logs for personnel, the current display lists dates ascendingly (oldest dates list at the top). Since when referring back to the log, I often find that I am interested much more in their recent accomplishments than their old ones, sorting dates descendingly (most recent dates at the top) would make much more sense, as I am a lot less likely to want to quickly refer back to old information. Checking new log info also requires a lot of scrolling on long-serving personnel, though the new collapsible log headers help with this already.

The ability to change sorting format would be useful and improve functionality minorly.

IllianiCBT:
This PR reverses the order of the three personnel logs (personal, scenarios, and kills) so that the order is newest to oldest; instead of oldest to newest. This will help reduce the amount of scrolling users need to do, whenever they're trying to review the personnel logs.

I also went ahead and renamed the 'personnel log' (the one that covers things like unit assignment, marriage, children, and injuries) to 'personal log'. This was a pet peeve of mine and I highly expect I am the only person bothered by this but having 'personnel logs' (as a collection of the three logs) and 'personnel log' (being only the personal log) made discussing the logs a right pain.

## Closes
Partially addresses #2238 by reducing the amount of scrolling required.
Closes #4032


SJuliez (approved):
Note: Leaving the initial comment like this *will* close both issues

PR #4149, opened 2024-06-04
Added Automatic Bonus Parts Exchange at Contract End, Added Bonus Parts Display to Mission Stats Panel

### LINKED ISSUE #3390: [RFE] Stratcon - Supply Point / Bonus Part conversion to C-Bills at the end of contract
### Description
I just wrapped up a short Stratcon mission that resulted in my having 20 Supply points and 2 Bonus parts "left over" on the last day of the contract with nothing to really spend them on. All my units were fully repaired and I had parts stockpiled in the warehouse.

However, it does seem to be a bit cheesy to convert the supply points to bonus parts, then use the bonus parts for the most expensive parts in the game, and then sell those for c-bills.

### Request
I'm not sure if there is an official conversion rate, but it would be nice if excess supply points (and bonus parts) were _automatically_ converted to some payout when a mission is concluded (sort of like the remainder of the contract pay) rather than simply lost if the player forgets to use them.

## Idea 1
The Chaos Campaign rules have a supply point concept and a conversion rate (I think it's 1 supply point converts to 10k c-bills). It may be too low for non-Chaos Campaign-based games, but it's a start (?)

## Idea 2
Flexible conversion rate

# Idea 2.1
Conversion rate = avg price of available tech for the player's faction. This is lots of adding, but is a relatively reasonable way to do it.

# Idea 2.2
Conversion rate = avg price of all parts in use

## Idea 3
Negotiable pricing... This would add an unofficial aspect to the contract negotiation where employers would set a value for excess supply points / bonus parts at contract end. _Potentially_ something that could be negotiated by an Admin/Logistics character.

IllianiCBT:
Requires: #4138

At the end of an AtBContract any remaining Bonus Parts (that would otherwise be lost) are automatically exchanged for c-bills at a rate defined by the user (default: 500,000 c-bills). An optional cap on the number of parts exchanged can also be defined by the user (default: 10).

This does not otherwise affect Bonus Parts usage. So parts can be gained and used in the same way they always have. This PR just adds a small bonus in the event the user forgets to use their Bonus Parts.

I also added a Bonus Part count to the Mission Stats panel, so users don't need to go hunting to determine how many Bonus Parts they have available. This count is refreshed on new day

## Closes
Closes #3390


IllianiCBT:
Resolved the surprisingly numerous conflicts

IllianiCBT:
Resolved conflicts

SJuliez (approved):

Issue #2085, opened 2020-10-01
RFE: AtB: Vary unit quality in the Unit Market


RexPearce:
It'd be cool to have some variety in the quality of the units offered for sale in the Unit Market (along with prices to match). So that  you have choice between a pristine mech from the factory that costs more, or a discounted option from the Mercenary Market but the parts are mostly A/B rating.

This could be extended furthered by have some units be randomly damaged (e.g. missing engine, gyro, weapons, limbs, etc.) for an even greater discount. (Could even be hidden: Surprise, that Black Market Marauder doesn't have an engine!)

purpledragon2:
I've always thought there should be an option for this.  Furthermore, it should extend to the equipment available to allies and opfor.  I suggest an option to vary unit/equipment quality for each based on location from and unit ratings where it applies to allies and opfor.

PR #4150, opened 2024-06-04
Added Unit Quality Parameter to addNewUnit Method and Related Tests

### LINKED ISSUE #2405: RFE: variable quality level on salvaged units
Currently, as of 47.16 all salvaged units have equipment with quality 'D'.  A-rated line units should have a fair share of units with higher-quality components, and F-rated units should have an abundance of much lower quality components.

What I would like to see is when quality rules are in play, salvaged units are assigned random quality to components based on the contract opposition rating.  An A-rated opponent is biased towards E/D quality components, while F-rated opponents are heavily biased towards A/B quality components.

Bonus points for having an option which skews the results downwards based on the time from the start of the contract i.e. an increasing reduction in salvage quality as the contract progresses.

IllianiCBT:
This PR adds a `int quality` parameter to our `addNewUnit()` method, allowing calls to assign quality ratings to the Unit being added.

This is needed for a PR I'm working on (to address #2085 & #2405) and to fix a bug/not fully implemented feature with the AtB Unit Market. Due to the number of files this change touched on, it made sense to push the quality parameter addition separately to make reviewing easier.

The Unit Market is _meant_ to sell Units at varying quality ratings, and is set up to do so, however as `addNewUnit()` didn't accept a quality parameter all added Units defaulted to quality rating 3 (D). To be clear: this PR only enables the ability to fully implement that functionality, it does not implement the functionality itself. That will be added in a follow-up PR.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Quartermaster.java
+++ b/MekHQ/src/mekhq/campaign/Quartermaster.java
@@ -45,7 +45,7 @@ public class Quartermaster {
         PlanetSpecificFailure,
         Success
     }
-    
+
     private final Campaign campaign;
 
     /**
@@ -470,13 +470,13 @@ public class Quartermaster {
             Money cost = new Unit(en, getCampaign()).getBuyCost();
             if (getCampaign().getFinances().debit(TransactionType.UNIT_PURCHASE, getCampaign().getLocalDate(),
                     cost, "Purchased " + en.getShortName())) {
-                getCampaign().addNewUnit(en, false, days);
+                getCampaign().addNewUnit(en, false, days, 3);
                 return true;
             } else {
                 return false;
             }
         } else {
-            getCampaign().addNewUnit(en, false, days);
+            getCampaign().addNewUnit(en, false, days, 3);
             return true;
         }
     }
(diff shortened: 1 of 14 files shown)


MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -470,13 +470,13 @@ public boolean buyUnit(Entity en, int days) {
             Money cost = new Unit(en, getCampaign()).getBuyCost();
             if (getCampaign().getFinances().debit(TransactionType.UNIT_PURCHASE, getCampaign().getLocalDate(),
                     cost, "Purchased " + en.getShortName())) {
-                getCampaign().addNewUnit(en, false, days);
+                getCampaign().addNewUnit(en, false, days, 3);
```
SJuliez, line 473: It might make sense to retain the old version like this. To avoid having to update all the old calls.
```
public Unit addNewUnit(Entity en, boolean allowNewPilots, int days) {
   return addNewUnit(en, allowNewPilots, days, 3);
}
```

IllianiCBT: Oh, that is such a good idea!

IllianiCBT:
Requested review actions resolved

SJuliez (approved):

PR #4153, opened 2024-06-05
Changed Warning Color from 'Red' to 'Orange'

IllianiCBT:
As per title, this PR has replaced instances of font color 'red' with font color 'orange'. The color red can be difficult to read on the darker background used by the Darcula theme.

The other colors we use (predominantly green) do not have this issue. The below screenshot illustrates this issue compared to font color orange.

image

### DIFF
--- a/MekHQ/resources/mekhq/resources/Finances.properties
+++ b/MekHQ/resources/mekhq/resources/Finances.properties
@@ -105,7 +105,7 @@ Overhead.text=Your account has been debited for %s in overhead expenses
 # Loans
 Loan.title=loan payment to %s
 Loan.text=Your account has been debited for %s in loan payment to %s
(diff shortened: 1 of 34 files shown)


SJuliez:
There are some configurable colors in MM's GUIPreferences that you could use:
image

I assume MHQ doesn't have its own version of those. But that's a bigger change obviously.

IllianiCBT:
> There are some configurable colors in MM's GUIPreferences that you could use: [image]
> 
> I assume MHQ doesn't have its own version of those. But that's a bigger change obviously.

Honestly, adding three new font color options *is* the better way to do this. I'll go ahead and make those changes. A job done properly is a job done once, after all. :)

IllianiCBT:
This PR now has reduced scope and functionality, as reflected by the updated title and description.

SJuliez (approved):

PR #4156, opened 2024-06-06
Replaced Uses of Font Color 'Red' with New MekHQ Font Color Option Call

IllianiCBT:
Requires: #4153

This PR replaces calls to font color 'red' with the 'negative event' font color assigned via the MekHQ Options added by #4153.


SJuliez (commented):
I'm not sure if leaving the closing /font and /span was on purpose. If so, I'd not do that. Better to either format in code or in resources, but not mixed.

IllianiCBT:
Resolved conflicts and addressed requested revisions

SJuliez (approved):

PR #4157, opened 2024-06-06
Replaced Uses of Font Color 'Green' and 'Orange' with New MekHQ Font Color Option Calls

IllianiCBT:
Requires: #4153 and #4156

This PR replaces calls to font color 'green' with the 'positive event' font color assigned via the MekHQ Options added by #4153.

It also does the same for font color 'orange', replacing the handful of instances where that is used with the 'warning event' font color.

It also corrects one or two mistakes made in #4156


IllianiCBT:
Conflicts resolved, also removed in-resource </font> closures. Feedback was added in #4156 but applied here, too, so I made sure to correct both.

IllianiCBT:
Fixed conflicts

PR #4158, opened 2024-06-06
Added Campaign Option to Randomize New Unit Quality

IllianiCBT:
Requires: #4150

This PR adds a new campaign option to have new Units generate with randomized Quality ratings. I based the chance for each rating on the AtB Mek Quality table.

This table is based on a `2d6 + modifier` roll, with results clamped within 2 and 12. Please note that Unit Quality is one of those cases where F is the best quality and A is the worst. The original AtB table appears to have unit quality incorrectly flipped around, so I factored for that.

image

Random Company Generator is affected by these changes, but applies a +2 modifier for the campaign commanders' roll, and a +1 modifier for officer rolls.

PR #4159, opened 2024-06-06
Added Variable Unit Quality to Salvaged Units

### LINKED ISSUE #2405: RFE: variable quality level on salvaged units

IllianiCBT:
Requires: #4150 and #4158

This PR extends the optional randomization of Unit Quality to units salvaged from scenarios. Furthermore, it adds the ability for OpFor Quality to influence the quality rating of salvaged units (if the contract is an instance of AtBContract).

If the scenario is spawned from an AtBContract the quality of the contract OpFor will influence the quality of units salvaged from scenarios. If random unit quality is enabled, this is applied in the form of a modifier. Otherwise, the quality is fixed as shown below:

image

It's important to note that while Dragoon Rating runs from F-A*, unit quality runs from A-F.

## Closes
Closes #2405


IllianiCBT:
Resolved conflicts

NickAragua (approved):

PR #4160, opened 2024-06-06
Added Variable Unit Quality to Unit Markets and Updated Unit Market Logic

### LINKED ISSUE #2085: RFE: AtB: Vary unit quality in the Unit Market

IllianiCBT:
Requires: #4150 and #4158

This PR enhances the functionality of the AtB Unit Market with the following objectives:

1. Streamline the process for adding new units for sale. Upon reviewing the code, I noticed numerous dice rolls that lacked clarity regarding their comparison to other market types. I standardized these rolls, introducing human-readable rarity levels to provide maintainers with clear insights into the rarity of units within any given market.

2. Introduce greater depth to the markets by considering campaign faction and contract state more effectively. Previously, many markets defaulted to the employer faction, potentially leading to difficulties for House Units in maintaining their faction's units between contracts, or problems when no employer is available.

3. Associate unit quality with market type. While initially, I incorrectly believed this had always been the intent, it seemed too good an idea to abandon. This is applicable for campaigns with and without random unit quality enabled. In the event random unit quality is enabled, the market types will apply different modifiers to the quality roll. In the event this option is disabled, the unit quality will be fixed based on market type.

Additionally, I have provided comprehensive documentation for unit markets and renamed the 'AtB Stuff' folder to 'Against the Bot'. You can access the documentation [here](https://docs.google.com/document/d/1yBzdo2cjESk7sPrfzC5kE59XJ9pd9XDV1P_nWLUlLPE/edit?usp=sharing).

## Update
This now includes a bug fix for the Unit Market that otherwise prevents Artillery from spawning in the unit market. This bug fix was merged into this PR to avoid conflicts.

## Closes
Closes #2085


IllianiCBT:
Conflicts resolved

IllianiCBT:
Resolved conflicts

MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
```diff
@@ -169,28 +170,27 @@ public String getName(LocalDate when) {
...
 
     public Set<Faction> getFactionSet(LocalDate when) {
-        Set<Faction> factions = new HashSet<>();
-        for (Planet planet : planets.values()) {
-            Set<Faction> f = planet.getFactionSet(when);
-            if (null != f) {
-                factions.addAll(f);
-            }
-        }
+        Set<Faction> factions = planets.values().stream()
```
NickAragua, line 183: I'm kind of curious to see a "how long does this take to run" comparison between this and the older code. In my experience, the syntactic sugar carries a performance penalty, but I'm happy to be proven wrong.

MekHQ/src/mekhq/gui/panes/UnitMarketPane.java
```diff
@@ -469,6 +482,50 @@ private void finalizeEntityAcquisition(final List<UnitMarketOffer> offers,
         getMarketModel().setData(getCampaign().getUnitMarket().getOffers());
     }
 
+    /**
+     * Returns the quality of a unit based on the given market type.
+     *
+     * @param market the type of market
+     * @return the quality of the unit
+     */
+    private int getQuality(UnitMarketType market) {
```
NickAragua, line 491: Do we want to have this "business logic" code in this UI class, or maybe move it to the UnitMarketType class instead?

MekHQ/src/mekhq/campaign/market/unitMarket/AtBMonthlyUnitMarket.java
```diff
@@ -150,8 +239,19 @@ public void addOffers(final Campaign campaign, final int num, UnitMarketType mar
             final Collection<MissionRole> missionRoles = new ArrayList<>();
             if (unitType == UnitType.TANK) {
                 movementModes.addAll(IUnitGenerator.MIXED_TANK_VTOL);
-                missionRoles.add(MissionRole.MIXED_ARTILLERY);
+
+                // should a special unit type be picked? This allows us to force MissionRole that would otherwise be filtered out
+                if (Compute.randomInt(30) == 0) {
```
NickAragua, line 244: How did we arrive at "1 in 30 tanks should be artillery or support"?

NickAragua (changes requested):
A couple of questions/maintainability requests.

MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
IllianiCBT: I was under the impression that streams were faster than loops, but that appears to not be as universally true as I was led to believe.

According to google, streams are faster when looping through large volumes of data, but slower with small volumes. This is definitely a case of 'small volume' so I'll go ahead and revert this change and be more judicious with their usage in future.

MekHQ/src/mekhq/campaign/market/unitMarket/AtBMonthlyUnitMarket.java
IllianiCBT: That was just the value that 'felt right' during testing.

There is an argument to be made for making this a campaign option, but given how roles now work it's not possible to use the old method of generation, so I had to pick a value.

After Stucco's role overhaul, the `MIXED_ARTILLERY` MissionRole now filters out anything that isn't mixed artillery (which just happens filter out all vehicles). Previously this role was unique in that it deliberately didn't filter anything out, it just told mhq to include mixed artillery in addition to other units.

Not including some kind of arbitrary system to force the unit market to add artillery (and support) units results in those units never being added to the unit market.  Potentially this is due to filtering occurring elsewhere in the code, but I couldn't find anything that would indicate this taking place.

NickAragua: Fair enough, we can always revisit it in a subsequent PR.

IllianiCBT:
Updated to include requested revisions.

I also moved the arbitrary 'special vehicle chance' value to campaign options. It's no less arbitrary, but this at least means the user can better adjust it to meet their needs.

Moving `getQuality` was handled in a second follow-up commit, as I forgot to include it in the initial revision. Apologies for the spam.

PR #4164, opened 2024-06-06
More detailed auto-selection of bombs

### LINKED ISSUE #2674: Stratcon AI heavy bombers are always have maximum possible bombs and doesn't care for its cruising speed
### Environment
Ver 0.49.1

### Description
Yes it may not a problem for many bombers, but it is very serious problem for heavy bombers.

r1

For example, I have two ally bombers on the mission, and one of them is Thunderbird TRB-D36. As you see, its basic movement is 5/8. It may have up to 20 bombs, sure, but it will reduce its movement to 1/2, which is not a good move. Although aerospace units CAN takeoff with flanking movement, but always work with flanking movement of up to 2 seems quite punitive for such bombers. Isn't better to adjust the number of bombs to keep at least cruising speed of 2?

Asides, what the hell on the idea of carry inferno bombs on here? Yes in the most times it is not an issue and is a very effective bomb, sure, but the mission's terrain is HEAVY **URBAN** and the ally is come to defend(actually, evacuate the ally) it rather than attack or raze it. I don't think that burn your own city is something helpful to defend your one. Is it better to make a new post for this - choose the right bombs for the missions, especially for depend missions on the terrain? Such as put Rocket Launchers instead of the other bombs, as my aero wings do on the same mission to avoid collateral damage. Yes it's only a computer game and destroying the building doesn't affects anything on here, but it's somewhat weird that what the defending wings have seems nothing but raze their own ground....

SuperStucco:
Randomly generated units that are cleared to carry bombs currently do so in a somewhat blunt manner - pick a random bomb type, load up to maximum based on carrying capacity until it's just out of thrust.  This PR adds more randomness to the selection process, puts in a few guards to maintain a minimum thrust level, and produces more role-focused selections.  It's a bit chunkier than I would like but should be a good start.

The new process looks like this:

1. Randomly determine how many of the fighters will carry external ordnance, somewhere around half to all
2. For each fighter, calculate a maximum load using minimum thrust values as well as carrying capacity.  Some additional randomness is included.
3. Randomly select a loadout such as "Normal", "Anti-Mech", or "Anti-Conventional".  Each loadout is a list of appropriate ordnance types with proportional values.  Pirates, including "Unidentified Hostiles" get their own selections of low tech/high chaos loadouts.
4. Bombers deployed for space or low altitude combat will only get air-to-air loadouts (no HE, cluster, inferno, etc.), while those for ground support will mostly be air-to-ground loadouts.  A few will be configured as CAP/combat air patrol rather than ground support.
5. Randomly generate ordnance using the proportions in the selected loadout, up to the calculated limit. Ordnance unavailable due to era restrictions is replaced with either HE bombs (ground only) or rocket launchers (ground and air). Lower rated forces (typically Dragoons rating D/F) will randomly have some advanced ordnance randomly replaced in the same way.
6. If any guided ordnance is selected (laser/TAG guided bombs, homing Arrow IV) some of the remaining fighters without ordnance will get an external TAG

Differences from the current system:
- overall lighter bomb loads (but the heavies will still hurt)
- fewer single-type loads, and more mixed loads e.g. HE and rockets, or cluster and inferno
- large ordnance like Arrow IV, anti-ship, ASEW will only be common with high (A/B) rated forces
- slight increase in HE bombs and rockets over other bomb types in general, particularly in earlier eras and lower rated forces

Should help resolve #2674 

Eventually, I think this should switch to reading values from an external data file rather than using hard-coded values, and probably be pulled into it's own class as well to clean up the current location.  That can wait until after the general process/framework is validated.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
@@ -66,6 +66,7 @@ import java.io.File;
 import java.time.LocalDate;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * This class handles the creation and substantive manipulation of AtBDynamicScenarios
@@ -2589,8 +2590,18 @@ public class AtBDynamicScenarioFactory {
                                          int quality,
                                          boolean isPirate) {
 
-        // Get all valid bombers
-        List<Entity> bomberList = entityList.stream().filter(Targetable::isBomber).collect(Collectors.toList());
+        // Get all valid bombers, and sort unarmed ones to the front
+        List<Entity> bomberList = new ArrayList<>();
+        for (Entity curEntity : entityList) {
+            if (curEntity.isBomber()) {
+                if (!curEntity.getIndividualWeaponList().isEmpty()) {
+                    bomberList.add(curEntity);
+                } else {
+                    bomberList.add(0, curEntity);
+                }
+            }
+        }
+
         if (bomberList.isEmpty()) {
             return;
         }
@@ -2602,13 +2613,16 @@ public class AtBDynamicScenarioFactory {
         int maxBombers = (int) Math.ceil(((Compute.randomInt(60) + 40) / 100.0) * bomberList.size());
         int numBombers = 0;
 
-        int[] generatedBombs = new int[BombType.B_NUM];
+        int[] generatedBombs;
         Map<Integer, int[]> bombsByCarrier = new HashMap<>();
 
         boolean forceHasGuided = false;
         for (int i = 0; i < bomberList.size(); i++) {
+            bombsByCarrier.put(i, new int[BombType.B_NUM]);
+
+            // Only generate loadouts up to the maximum number, use empty loadout for the rest
             if (numBombers >= maxBombers) {
-                break;
+                continue;
             }
 
             Entity curBomber = bomberList.get(i);
@@ -2661,36 +2675,25 @@ public class AtBDynamicScenarioFactory {
 
         }
 
-        // Load bombs onto entities. If there is guided ordnance present then randomly add some TAG
+        // Load ordnance onto units. If there is guided ordnance present then randomly add some TAG
         // pods to those without the guided ordnance.
         int tagCount = Math.min(bomberList.size(), Compute.randomInt(3));
         for (int i = 0; i < bomberList.size(); i++) {
             Entity curBomber = bomberList.get(i);
 
-            if (bombsByCarrier.containsKey(i)) {
-                generatedBombs = bombsByCarrier.get(i);
+            generatedBombs = bombsByCarrier.get(i);
 
-                if (forceHasGuided &&
-                        tagCount > 0 &&
-                        Arrays.stream(generatedBombs).sum() < (Math.floor(curBomber.getWeight() / 5)) &&
-                        !hasGuidedOrdnance(generatedBombs)) {
-                    generatedBombs[BombType.B_TAG]++;
+            // Don't combine guided ordnance with external TAG
+            if (forceHasGuided && tagCount > 0) {
+                if (addExternalTAG(generatedBombs, true,
+                        Math.min((int) Math.floor(curBomber.getWeight() / 5.0), (curBomber.getWalkMP() - 2) * 5))) {
                     tagCount--;
                 }
+            }
 
+            // Load the provided ordnance onto the unit
+            if (generatedBombs != null && Arrays.stream(generatedBombs).sum() > 0) {
                 ((IBomber) curBomber).setBombChoices(generatedBombs);
-
-            } else {
-
-                // If one of the entities was given guided ordnance then put a TAG on the ones
-                // that have nothing
-                if (forceHasGuided && tagCount > 0) {
-                    Arrays.fill(generatedBombs, BombType.B_NONE);
-                    generatedBombs[BombType.B_TAG]++;
-                    tagCount--;
-                    ((IBomber) curBomber).setBombChoices(generatedBombs);
-                }
-
             }
 
         }
@@ -2700,6 +2703,7 @@ public class AtBDynamicScenarioFactory {
     /**
      * Randomly generate a set of external ordnance up to the number of indicated bomb units. Lower
      * rated forces are more likely to get simpler types (HE and rockets).
+     * Because TAG is only useful as one-per-fighter, it should be handled elsewhere.
      * @param bombUnits   how many bomb units to generate, some types count as more than one unit so
      *                    returned counts may be lower than this but never higher
      * @param airOnly     true to only select air-to-air ordnance
@@ -2941,6 +2945,56 @@ public class AtBDynamicScenarioFactory {
             if (bombLoad[curHomingBomb] > 0) {
                 return true;
             }
+        }
+
+        return false;
+    }
+
+    /**
+     * Updates a bomb load to include an external TAG system. If this exceeds the provided
+     * maximum load (in bomb units i.e. Arrow IV counts as multiple units), then one of the basic
+     * one-slot types is removed and the TAG system is added in it's place.
+     *
+     * @param bombLoad    array of size BombType.B_NUM, suitable for setting bombs on IBomber
+     *                    entities
+     * @param skipGuided  true to only select external TAG for units without guided ordnance
+     * @param maxLoad     Maximum external ordnance load in total bomb units (NOT bomb count)
+     * @return            true, if TAG was added, false otherwise
+     */
+    private static boolean addExternalTAG (int[] bombLoad, boolean skipGuided, int maxLoad) {
+        if (bombLoad.length < BombType.B_NUM) {
+            throw new IllegalArgumentException("Invalid array length for bombLoad parameter.");
+        }
+
+        if (!skipGuided || !hasGuidedOrdnance(bombLoad)) {
+
+            // If there's enough room, add it
+            int totalLoad = IntStream.range(0, bombLoad.length).map(i -> BombType.getBombCost(i) * Math.max(bombLoad[i], 0)).sum();
+            if (totalLoad < maxLoad) {
+                bombLoad[BombType.B_TAG]++;
+                return true;
+            } else if (totalLoad == maxLoad) {
+
+                List<Integer> replaceableTypes = Arrays.asList(
+                        BombType.B_RL,
+                        BombType.B_HE,
+                        BombType.B_INFERNO,
+                        BombType.B_CLUSTER);
+                for (int i = 0; i < replaceableTypes.size(); i++) {
+                    if (bombLoad[i] > 0) {
+                        bombLoad[i]--;
+                        bombLoad[BombType.B_TAG]++;
+                        return true;
+                    }
+                }
+
+            } else {
+                // Already overloaded, don't bother
+                return false;
+            }
+
+            bombLoad[BombType.B_TAG]++;
+            return true;
         }
 
         return false;
--- a/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
@@ -2379,7 +2379,7 @@ public class AtBDynamicScenarioFactory {
     private static int calculateAtBSpeed(Entity entity) {
         int speed = entity.getWalkMP();
         if (entity.getJumpMP() > 0) {
-            if (entity instanceof megamek.common.Infantry) {
+            if (entity instanceof Infantry) {
                 speed = entity.getJumpMP();
             } else {
                 speed++;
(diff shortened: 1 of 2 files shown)


IllianiCBT:
How will this interact with @Sleet01 ’s special munitions distribution system?

Sleet01:
> How will this interact with @Sleet01 ’s special munitions distribution system?

@SuperStucco @IllianiCBT I think we can separate the two initially, then integrate them as necessary over the next release or two (after 0.49.20):

- I am pulling the existing AtB bomb randomizer code out of MML so it will be available in MM games, which means we can have this new code live in AtB for now and they shouldn't interfere with each other.
- I don't have this level of role definition so SuperStucco's work will likely be better for MHQ games, providing more immersion.
- My updates are built around the new file format that users can modify and use to define their own loadouts for chassis, model, or even specific unit, which could be integrated into this new randomizer / generator code fairly easily - although there is also an argument to be made for adding a role and weight class component to the file format, similar to this setup.

@SuperStucco I would say that moving code that you want to extract into a new class is better done sooner than later, so that you can set up unit tests early and validate basic functionality more easily.

I'll do a full code review when I'm back on broadband, but based on a quick scan I don't think there's any reason our work couldn't be integrated into one seamless system for outfitting units.

Sleet01:
@SuperStucco There's a thread on the MegaMek Discord server where we discussed some of the stuff I was doing in my work: https://discord.com/channels/458705327911731231/1222035892965736539

IllianiCBT:
I'm pleased to hear that, because I'm greedy and want both :D

SuperStucco:
> @SuperStucco I would say that moving code that you want to extract into a new class is better done sooner than later, so that you can set up unit tests early and validate basic functionality more easily.

I'm taking a more 'eating the elephant' approach with more small bites, as I get up to speed. While this gets reviewed I'm looking at a couple of other things - having field gun infantry generation respect the availability values, and potentially generating marine/XCT infantry in place of standard infantry in hostile environment conditions.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
-                // if this entity has no guns (e.g. is a Boeing Jump Bomber)
-                if (entity.getIndividualWeaponList().isEmpty()) {
-                    loadBombs(entity, validBombChoices, campaign.getGameYear());
-                    continue;
+        boolean hasGuided = false;
+        for (int i = 0; i < bomberList.size(); i++) {
+            if (numBombers >= maxBombers) {
+                break;
+            }
+
```
IllianiCBT, line 2603: Is it worth adding a check here? I'm guessing that if the unit is unarmed it's not going to need any bombs.

```
if (isUnarmed) {
    continue
}
```

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+                randomThreshold = 25;
+                break;
+            case IUnitRating.DRAGOON_D:
+                randomThreshold = 40;
+                break;
+            case IUnitRating.DRAGOON_F:
+                randomThreshold = 80;
+                break;
+            default:
+                randomThreshold = 99;
```
IllianiCBT, line 2899: I'm wondering whether we should throw an exception here, instead of assigning a default value to `randomThreshold`. There shouldn't be any point in which quality isn't within the F-A* bounds, so having an exception thrown would mean we get alerted to the problem rather than it failing silently.

IllianiCBT:
I was curious in seeing how everything worked, so I had a scroll through your code. I love the use of `Map` here, that's something I've only started using and it's great seeing just how powerful maps can be.

I did have a couple of points of feedback, nothing major, just thoughts that plonked into my noggin while I was reading. If Sleet comes along and disputes anything I've commented, assume they're right and I'm wrong. :D

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
gsparks3: Possible relevant: The Boeing Jump Bomber is specifically designed to drop bombs, but has no internal weapons.

SuperStucco: Carried this logic forward from the previous test.  If it's carrying bombs, it really should be loaded up as that's the only realistic contribution it can make.  Lower down in the code, they are specifically excluded from counting towards the total number of loaded bombers.

If anything they would be prioritized over armed units.  Would it be worth re-ordering the bomber subset List<> to put unarmed entities first so they always get loaded up?

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
SuperStucco: Currently the method is called correctly, but the quality parameter is a straight int type. So it could be called inappropriately with bad values (negative, large numbers, etc.).

So, using a
`throw new IllegalArgumentException("Unrecognized rating value: " + quality);`
statement, and let the thrown exception get handled upstream?

IllianiCBT: Yeah, my thinking is that if we throw an exception we know something has gone wrong; if we silently handle it with a default value the only way we know there was a bad call is if someone happens to notice an unusually large quantity of rockets being handed out.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
IllianiCBT: No, if that's the case then it's likely fine how it is. I hadn't spotted the previous test where it was defined :)

HammerGS:
Should this be considered a draft, or is it ready to go.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -76,11 +77,135 @@ public class AtBDynamicScenarioFactory {
...
 
-    // bomb types assignable to aerospace units on ground maps
-    private static final int[] validBotBombs = { BombType.B_HE, BombType.B_CLUSTER, BombType.B_RL,
-            BombType.B_INFERNO, BombType.B_THUNDER, BombType.B_FAE_SMALL, BombType.B_FAE_LARGE,
-            BombType.B_LG, BombType.B_ARROW, BombType.B_HOMING, BombType.B_TAG };
-    private static final int[] validBotAABombs = { BombType.B_RL, BombType.B_LAA, BombType.B_AAA };
+    /**
+     * Distribution of various external ordnance choices for non-pirate forces
+     */
+    private static final Map<String, Integer> bombMapGroundSpread = Map.ofEntries (
```
Sleet01, line 83: It would be useful to know what these numbers represent: relative weights, percentages, absolute counts?

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -457,10 +582,34 @@ public static int generateForce(AtBDynamicScenario scenario, AtBContract contrac
...
+                    default:
+                        ownerBaseQuality = quality;
+                        break;
+                }
 
-                populateAeroBombs(generatedLance, campaign, !isAeroMap);
+                // Load the fighters with bombs
+                populateAeroBombs(generatedLance,
+                        campaign,
+                        (mapLocation != MapLocation.Space && mapLocation != MapLocation.LowAtmosphere),
```
Sleet01, line 610: Is there a reason to not use `!isAeroMap` for the third argument here?
The new code will prevent bombers from being loaded in space; space bombing is allowed both in the rules and the game.
If there _is_ a reason to change this check, should it not also be updated in `AtBScenario.java`?

Sleet01: NVM, I see that the use of this argument has been changed further on.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+        // Load bombs onto entities. If there is guided ordnance present then randomly add some TAG
+        // pods to those without the guided ordnance
+        int tagCount = Math.min(bomberList.size(), Compute.randomInt(3));
+        for (int i = 0; i < bomberList.size(); i++) {
+            Entity curBomber = bomberList.get(i);
+
+            if (bombsByCarrier.containsKey(i)) {
+                generatedBombs = bombsByCarrier.get(i);
+
+                if (hasGuided &&
```
Sleet01, line 2666: Move this into a conditional above the tagCount declaration; then it can be removed from this line and line 2681, and this section can be avoided if no guided munitions have been loaded in the entire bomber fleet.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+                        Arrays.stream(generatedBombs).sum() < (Math.floor(curBomber.getWeight() / 5)) &&
+                        generatedBombs[BombType.B_LG] == 0 &&
+                        generatedBombs[BombType.B_HOMING] == 0) {
+                    generatedBombs[BombType.B_TAG]++;
+                    tagCount--;
                 }
 
-                if (numBombers >= maxBombers) {
-                    break;
+                ((IBomber) curBomber).setBombChoices(generatedBombs);
```
Sleet01, line 2675: I don't like this construction:
- We're mixing bomb configuration and bomb setting; this should be separate for cleanliness and ease of parsing.
-  We duplicate this call in two places, but neither will get called if `hasGuided` is false so if there are no guided munitions, no bombs will get set on any Aero unit.
- We have a hard-coded set of guided munitions that doesn't include Aero Guided Arrow IV Bomb munitions, ; this should probably be a set or list that we declare up with the other constants.
- It looks like the second option erases all bombs for the current unit and just gives it one single TAG pod, which I guarantee is not what we want.

I would actually suggest splitting the TAG-adding code out into a separate helper that just mutates a generatedBombs array based on the current bomb load on the bomber, the bomber's weight, and perhaps the bomber's loaded speed - that way we can avoid pushing a bomber from "fast enough" to "too slow" by adding the TAG pods only to units that have a slot to spare before dropping down one thrust rating.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+                    campaign.getGameYear());
+            // Whoops, go yell at the ordnance technician
+            if (Arrays.stream(generatedBombs).sum() == 0) {
+                continue;
+            }
+
+            // Set a flag if any of the ordnance guided by TAG
+            hasGuided = !hasGuided &&
+                    (generatedBombs[BombType.B_LG] > 0 ||
+                            generatedBombs[BombType.B_HOMING] > 0);
```
Sleet01, line 2644: See comment below; instead of two separate locations with different tests of hard-coded munitions, we should have a set or list of guided munitions that we can use to check for matches, and it should include ASF Homing Arrow IV bomb munitions as well.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+
+        double completeWeight = loadoutMap.values().stream().mapToDouble(curWeight -> Math.max(curWeight, 1.0)).sum();
+        double randomThreshold = (Compute.randomInt(100) / 100.0) * completeWeight;
+        for (String curMap : loadoutMap.keySet()) {
+            countWeight += Math.max(loadoutMap.get(curMap), 1.0);
+            if (countWeight >= randomThreshold) {
+                mapName = curMap;
+                break;
+            }
+        }
```
Sleet01, line 2738: This randomly chooses which entry loadout map type to apply to the unit by summing the relative weights in the loadout map, then checking a random value somewhere between 0.0 and (sum of weights) against a cumulative sum of the weights again?  And this works?

Just when I thought I couldn't hate arbitrary relative weighting any _more_...

My gut feeling is that this isn't going to work as planned, because bare Maps don't guarantee insertion-order ordering of entries.  So e.g. for `bombMapPirateGroundSpread`, a randomThreshold of <= 3.0 is equally likely to produce "Normal" or "Firestorm", and a value of 7.1+ will either select "Firestorm" or nothing; the chance of returning "Firestorm" is actually higher than the relative weighting of 7 and 3 would suggest.

Unfortunately I'm not an expert in this field so I'd need to go do some research to figure out the ideal approach, but I would say that switching the maps to LinkedHashMaps... unfortunately this requires using something other than Map.ofEntries() to populate them.  Perhaps a custom class that supports ofEntries()...

Sleet01: Actually, looking at the code below that uses this selected string, I don't see why we couldn't just create array lists that give us weighted xBombLoad mappings, without the role names that are only used for the switch cases.

e.g., instead of `bombMapGroundSpread` plus a bunch of complicated weighted-option-selecting code _plus_ a switch on `mapName`, just cut out the middleman and do:

```
    public static ArrayList<Map<Integer, Integer>> createWeightedList(Map<Map<Integer, Integer>, Integer> weightMap) {
        // For each map M, there will be N entries where N is the arbitrary weight (integer)
        ArrayList<Map<Integer, Integer>> entriesByWeight = new ArrayList<Map<Integer, Integer>>();
        
        for(Map.Entry<Map<Integer, Integer>, Integer> m: weightMap.entrySet()) {
            // Create a sublist of size N = relative weight
            ArrayList<Map<Integer, Integer>> subList = new ArrayList<Map<Integer, Integer>>(m.getValue());
            for (int i=0;i<subList.size();i++) {
                // Fill with references to the specified Map of BombType:Weight
                subList.add(m.getKey());
            }
            // Add all entries
            entriesByWeight.addAll(subList);
        }
        return entriesByWeight;
    }
```

Pass it a Map of Maps and weights, like:
```
    private static final Map<String, Integer> standardBombMapsMap = Map.ofEntries (
            Map.entry(normalBombLoad, 6),
            Map.entry(antiMekBombLoad, 3),
            Map.entry(antiConvBombLoad, 2),
            Map.entry(standoffBombLoad, 1),
            Map.entry(strikeBombLoad, 2)
    );

    private static final ArrayList<Map<Integer, Integer>> standardBombWeightedList = 
        createWeightedList(standardBombMapsMap);
```

Then a randomized integer between 0 and the length of the resulting ArrayList can be used to randomly select one of the maps:

```
    Map<Integer, Integer> curMap = standardBombWeightedList.get(Compute.randomInt(standardBombWeightedList.size());
```

Which is all kind of verbose, but now you can just get the maps and the probability of getting a given map equals (weight / sum of weights).

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+
+        Map<Integer,Integer> bombMap;
+        if (!isPirate) {
+            if (!airOnly) {
+                switch (mapName) {
```
Sleet01, line 2743: This can all be significantly simplified.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -2419,77 +2568,357 @@ private static void correctNonAeroFlyerBehavior(List<Entity> entityList, int boa
...
+
+        int selectedBombType = -1;
+        int loopSafety = 0;
+        for (int curLoad = 0; curLoad < bombUnits && loopSafety < 10;) {
+
+            // Randomly get the ordnance type
+            completeWeight = workingBombMap.values().stream().mapToDouble(curWeight -> Math.max(curWeight, 1.0)).sum();
+            randomThreshold = (Compute.randomInt(100) / 100.0) * completeWeight;
+            countWeight = 0.0;
+            for (int curBomb : workingBombMap.keySet()) {
```
Sleet01, line 2840: See comments on similar block that selects specific maps from the map of roles and weights.

Sleet01 (changes requested):
Apologies for making a bunch of corrections; I was re-acquainting myself with some of the base code while doing the review.

In general I like this approach, and it's similar to what I'm doing for generalized munition loadout creation.  But I have some concerns and suggestions.

At a high level:
1. the random selection system is suspect.  I don't believe it will actually select items using the weights specified, due to vagaries of the Java Map class.  I think we can get by with turning Maps of the various BombType: weight mappings into ArrayLists and selecting using random indexes.
2. the mixing of bomb load mutation and application rubs me the wrong way; it should probably be two separate functions, or at least two separate and distinct blocks of code.  There should really only be one call to `setBombChoices` for each unit.
3. some of the larger conditionals could be shrunk by creating collections of the Homing ammunition types, and checking for membership.

This looks like a significant improvement over the existing stuff, and I look forward to seeing it in action!

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
Sleet01: NM, superseded by later comments.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
SuperStucco: The TAG part might be cleaned up a bit. The reasoning on that is, you can't use a TAG pod on the same round you're dropping bombs. So can't self-designate for laser/TAG guided bombs, and while you can self-designate for a homing Arrow that you dropped last round, the minimum movement of fixed wing aircraft means the chances they can be over the target zone the very next round is slim.  So TAG designators get pushed to any fixed wing aircraft that did NOT get any external ordnance, which happens from the random limit on how many units to load ordnance on to. That means units with external TAG will only have the one unit and will always have sufficient thrust; it also means that if all units are loaded with ordnance any guided ordnance will need ground designators or built-in TAG.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
SuperStucco: I've stepped through the code several dozen times during the development and testing process, and it does seem to produce reliable results at least as far as is intended.

As for the proposed substitute, that's getting very complicated, to the point where I'm having trouble following it. The (switch) is wordier and less elegant but also less opaque.

Don't want to put too much effort into fiddling with this, as eventually those distribution numbers are going to get pushed out to external data so players can fiddle with it to suit their needs.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
SuperStucco: Not certain how. The test cannot be combined, as it needs to split between pirate and non-pirate loads, then it also needs to choose between air/ground and air/air loads - didn't feel that with only two choices of air/air loads, and the limited ordnance choices warranted an extra set of maps.

Granted, I could remove the negation, but the coding style is to put more likely choices at the top so when stepping through or manually reading code, large jumps to following lines are outliers.

Sleet01:
@SuperStucco Apologies, I totally missed the re-review request.  I will take a look as soon as I can.

Sleet01 (approved):
LGTM

Sleet01:
Check failures are for Gradle version mismatch, should be good to go.

PR #4172, opened 2024-06-07
Implemented More Education Module Bug Fixes

IllianiCBT:
As per title, this PR implements a number of additional bug fixes applied to the education module.

During prior testing, I added a line designed to deliberately skip the return home and graduation stages. It turns out, leaving that line in caused the return home and graduation stages to get skipped. Who could have guessed that would happen? Truly, I marvel at my own incompetence.

However, while hunting this bug down I was able to spot a couple other (relatively minor) issues, which have since been resolved. These were mainly visual errors, but crossing them off the list isn't a bad thing.


HammerGS (approved):
Tested and seems to work.

PR #4173, opened 2024-06-07
Fixed Multi-Person Original Unit Assignments in Personnel Market & Interactions with Turnover and Retention Module

IllianiCBT:
This bug was identified by our wonderful QA team and actually pre-dates the Turnover and Retention Module.

Basically, whenever personnel are hired from the personnel market and that person has an attached unit, our code was checking to see whether it was a single-person unit. If so, the unit is treated as being owned by that individual. When that individual departed the campaign, they tried reclaiming their original unit.

However, none of this took place if the unit was multi-crewed. The original Retirement Module addressed this by _always_ reclaiming whatever unit the soldiers or BA were attached to - whether it was their original unit or not. Something that came to light during testing of the Turnover and Retention module which is built around the original Retirement code.

Now, any units acquired through the personnel market are attached to the unit commander as their 'original unit'. When departing the campaign, the unit commander will reclaim their original unit as normal. This fix is applied to any multi-crewed unit acquired through the personnel market.

I also removed the special handler for soldiers and BA, from the retirement code. This means that the original unit commander (the owner) can be reassigned to new units without losing their original unit data; it also means that soldiers and BA won't steal whatever unit they happened to be assigned to, when leaving the campaign.


HammerGS (approved):
Tested and seems to work.

PR #4183, opened 2024-06-08
Added Multiple Turnover and Retention Module Improvements

IllianiCBT:
## Turnover Prompt
This PR addresses an issue where the turnover prompt would interrupt the 'advance multiple days' feature. Now, the prompt will still appear, and advancing multiple days will pause until the prompt is resolved.

If the turnover dialog is canceled (excluding cancelling the prompt), advance multiple days will also be canceled. This allows users to address the reason for canceling the dialog, such as taking out a loan to cover departure pay.

## Children
When parents depart, there is a chance they will take their children with them. I added a check to see if the child has been orphaned; in that case, the child will always remain with the unit. Users can then decide what to do with the orphaned child. I also added a personal log message stating when the child was orphaned.

## Stealing Units
In the original Retirement code, there was a "feature" where departing personnel would sometimes steal units. I thought I had removed all references to this functionality, but I discovered a use-case while addressing the issue below. The theft of units and money is better handled by the upcoming morale module.

## Sacking Personnel
In the original Retirement module, all departures were treated as retirements, including sackings. When I wrote the code for the Turnover and Retention module, I used much of the Retirement module's code.

Unfortunately, I didn't account for sackings, so before this PR, a sacking would have been treated as a retirement, defection, or resignation, depending on various factors. At no point was a sacking recognized as such, and "sacked" didn't exist as a personnel status.

I've added this status and implemented additional checks. I also improved how the Turnover and Retention module handles personnel "retiring" due to being KIA. Now, civilian spouses will depart the unit as originally intended.


HammerGS (approved):
Tested and seems to work.

PR #4185, opened 2024-06-08
Added Nag to Alert Users at Contract End

### LINKED ISSUE #4182: [RFE] Add End of Contract Nag
Title.

Something like:

"Your Contact is Over.  Manually End the Contract in the Briefing tab."

IllianiBird:
This PR adds a nag dialog that triggers if the user tries to advance day on the final day of a contract. This nag will only trigger once, if the user advances the day anyway the nag will not repeat for that contract. This is a necessary by-product of using the `getActiveContracts` method, instead of `getContracts`. This method was chosen as it reduces the amount of parsing needed to filter old contracts.

## Closes
Closes #4182


HammerGS (approved):
Tested and seems to work

PR #4186, opened 2024-06-09
Adjusted autoAwards Dialog Size to Account for Windows Taskbar

IllianiCBT:
This PR reduces the height of the autoAwards ceremony dialog to 94% of the user's screen size.

We received reports from Windows users that the dialog was appearing behind the taskbar. This issue was caused by the `getScreenSize()` method not accounting for the taskbar space. By reducing the dialog height and positioning it at the top of the screen, we avoid this issue while still maximizing screen usage for the dialog.


IllianiCBT:
Flipping this from 'draft' to 'live. Description has been updated accordingly.

This will need testing by QA due to the limitations of me not having a windows computer; so having to test via proxy (thanks for your help, Hammer).

HammerGS (approved):
Tested and working.

PR #4187, opened 2024-06-09
Restored Missing Code for Scenario Modifier Options & Campaign Option Layout Tweaks

IllianiBird:
This PR predominately restores missing code from my prior PR that added options to influence the number; frequency; and BV allowance of non-facility Scenario Modifiers. This code was lost due to an error between keyboard and chair on my side. Git is not a friend to the uneducated.

While I was at it, I also applied some minor formatting tweaks to campaign options. These tweaks were needed due to the large volume of new campaign options that were added over the past week.

Lastly, I updated my Finance Tab bug fix to include all the new finance tab options that were added over the week. These couldn't be added till the fix was merged, so were always going to need to be added in a separate PR, so why not now.


HammerGS (approved):
Tested and seems to work

PR #4188, opened 2024-06-09
Added Campaign Start Date Logging to Campaign

IllianiCBT:
This PR adds Campaign Start Date logging to the campaign class. Or, to put it another way, mhq now remembers what date the campaign started on.

I need this for the Turnover and Retention Module, but figured it would also be useful for Story Arcs and (if I ever do it) my RFE for campaign scoring. I also imagined there were other use-cases where knowing _when_ the campaign started might prove useful.

Existing campaigns with a `null` campaign start date have their start date set to whatever the current date is.


HammerGS (approved):
Looks good to me.

PR #4191, opened 2024-06-09
Prevented Infinite Loop from Bulk Hiring Civilian or Assistant Roles

IllianiCBT:
This PR prevents an infinite loop caused by the new 'fixed skill level' parameter added to the Bulk Hire Personnel dialog.

Basically, when picking roles without skill levels (civilians, or assistants), the fixed skill level can never be reached so mhq gets into an infinite loop of generating persons and then throwing them away (because they don't meet the required skill level).

In addition to fixing this, I also added a new `isAssistant()` method to `PersonnelRole.java` which allows us to more efficiently check whether someone has the `Astech` or `Medic` roles.


HammerGS (approved):
Tested and seems to work.

PR #4194, opened 2024-06-09
Add the word Dragoon when Dragoon rating is used.

repligator:
Adds the word "Dragoon" in front of the Dragoon rating, in order to prevent it from being confused for the tech rating, which also uses A-F, but in the opposite order. Also minor typo fixes.


HammerGS (approved):
Looks good to me.

IllianiCBT:
Oh man, I never knew I needed this, but now I won't be able to live without it!

PR #4196, opened 2024-06-10
Further Adjusted FM:M Dragoon Rating to CamOps Reputation Conversion

IllianiBird:
This further adjusts the Dragoon to Reputation conversion we use to allow users with CamOps enabled to still interact with AtB systems. In an attempt to bring them roughly close to one another I have reduced the amount the CamOps modifier is divided by from 3 to 2.5 (parsed as an int).

As with previous adjustments, this is only a bandaid. As we move to better support CamOps, eventually we're going to need to figure out a way to decouple AtB from FM:M entirely, but that is no small feat and outside the scope of this PR.


IllianiCBT:
I'm just going to make this a campaign option, tomorrow.

PR #4205, opened 2024-06-10
Fixed bug in Company Generator, add word Dragoon in front of Dragoon rating.

repligator:
Accidentally canceled an early PR, and then accidentally combined it with this one. Sorry for the mess.

Fixes a bug that caused officers and the company commander to never generate with leadership. Leadership had been placed at 0 in a randomization system that was indexed starting with 1.


IllianiCBT (approved):
LGTM

This is particularly important for the Turnover & Retention module that uses Leadership a lot.

PR #4208, opened 2024-06-11
Added Numerous Improvements to the Turnover and Retention Module

IllianiBird:
This PR is a letter of appreciation to our wonderful QA team who have spent hours of their game time testing, breaking, and offering feedback as to how the Retention & Turnover module could be improved. Without their feedback the module would have released in a sorry state.

Barring any emergent bugs, I believe this PR represents the module in a finished state.

_Documentation has been updated to reflect these changes._

### Hostile Territory Modifier
As suggested by QA, I've added a new modifier that applies a -2 reduction to turnover target numbers while the campaign is engaged in an offensive contract. The logic here is that personnel wouldn't want to depart if they had nowhere to depart to.

This modifier only really works if AtB is enabled. If AtB is disabled the modifier is applied whenever the campaign has an active contract.

As with all other modifiers, this modifier has full campaign option support.

### Default Setting Changes
`Turnover Target Number Method` has been changed from `NEGOTIATION` to `FIXED`. Based on testing, I was very tempting to remove non-fixed options entirely. I opted to keep the dynamic methods (Negotiation and Administration) for those users who want a more involved experience.

`Enable Default Bonuses` has been changed from `true` to `false`. While this option can be useful, with the change to `Turnover Target Number Method` bonuses shouldn't be needed as often. Keeping this enabled is a substantial cash drain on new companies.

`Hide Loyalty` has been changed from `true` to `false`. I like the idea of this option, but the reality leads to the user having no way to tell why some personnel are resigning more often than others.

### Management Skill
`Unskilled Penalty` has been changed from `-2` to `0`. This means even a single point of `Leadership` will see improvements to turnover target numbers, without there being a mallus for the user to overcome before seeing improvements.

Furthermore, the values have been flipped, so that a negative Management Skill modifier _reduces_ the target number, instead of increasing it.

### Loyalty
Loyalty has seen a number of changes: the values for loyalty have been flipped so that a negative loyalty _decreases_ the target number, while a positive loyalty increases it.

Furthermore, descriptive names have been added to each level of loyalty, to make it easier for users to understand how loyal someone is. These descriptive names have replaced the generic 'loyalty' modifier label in the turnover dialog.

The loyalty modifier for the character flagged as the campaign commander is improved by 1. This helps reduce the impact of a poor loyalty roll on the commander. This modifier is applying dynamically, so changing who is flagged as the commander will change who gets the benefits of this change.

Finally, I added the ability to directly edit loyalty in the Edit Person dialog.

### Text Changes
A number of text changes have been made across the module, moving towards better explaining the module through the existing dialogs. These text changes also include prompts encouraging users to read the documentation. This move was made to remind users that the documentation does exist and they should really read it.

### Retention Bonuses
Previously, Retention Bonuses (aka 'bonuses') were only available in campaigns that had Shares disabled. Paying a retention bonus reduces the turnover target number by 2, depriving Shares users of a substantial benefit. This was meant to be counteracted by Shares users being able to pay larger shares percentages of contract pay, however the two modifiers were not equivalent unless the Share user substantially tanked their contract pay.

This was a holdover from the old retirement system.

I went ahead and removed the 'non-shares only' caveat from the module ensuring all users have access to retention bonuses.

Furthermore, the cost of paying a retention bonus now factors in the frequency of turnover checks. The amount is divided by 12, if turnover frequency is set to `MONTHLY`; or 52, if turnover frequency is set to `WEEKLY`.

### Sacking Civilians
The module didn't know what to do when civilians were sacked, as they are technically not employees. This resulted in an NPE which has now been addressed.

### Default Dialog Size
I adjusted the turnover dialog default sizes from full screen to match the changes implemented in #4186. This will ensure a better experience for windows users. Furthermore, I reduced the default width of both the turnover dialogs and the autoAwards award ceremony. As these dialogs all based their size on the users' screen size, it made sense to mirror the changes.

### Suppressed Turnover Checks on Campaign Start
Previously, if the campaign started on the 1st of the month (and monthly checks were enabled) the turnover check prompt would appear at the end of the first day. This was undesirable, so I used the newly saved campaign start date to suppress the turnover dialog for the first 7 days of the campaign. This also highlighted a bug with the aforementioned campaign start date not persisting between saves. This has been addressed in #4202.

### Turnover Prompts
I received reports of an edge case that could cause the user to be hit with multiple turnover check prompts. I substantially simplified the code, which should remove this bug.

### Turnover Top Panel Display Bugs
There were numerous issues with the top panel of the turnover dialog (the part that shows total bonus payments due, or total shares). I went through and aggressively fixed these issues. It took a while, but I _think_ I have solved the problems reported by QA.

### Payout Load Bug
There was a small bug that prevented the turnover module correctly loading the saved c-bill value of failed turnover checks. This was a bug that existed in the old retirement system but one I wasn't aware of until recently.

### autoAwards Ceremony Appearing Behind MHQ
_This fix was included in the current PR to avoid conflicts with the next issue's resolution._

I received a report about an unusual bug that caused the autoAwards ceremony dialog to appear behind the MHQ window. This happened due to improper placement of the call that triggered the award ceremony. ~~Moving the call out of `ResolveScenarioTracker.java` _seems_ to have resolved the issue however further testing will be required.~~ While incorrect call placement wasn't great, the real reason seems to have been me failing to set ModalityType whenever the autoAwards dialog was triggered. This has since been resolved.

### Silent NPE When Resolving Scenario
When MHQ resolves a scenario, it follows a different process depending on whether the scenario was resolved manually or automatically. If the scenario was resolved automatically and nobody died, MHQ would silently throw a NPE.

This happened because there wasn't a check to ensure there were deceased individuals to process before attempting to process them. This issue did not affect manually resolved scenarios due to a conditional check in the code handling manual resolutions.

This was an old bug that I had completely forgotten about until I was troubleshooting the autoAward issue described earlier. Either way, I added a conditional which resolves the bug.

### Fixed Faction Modifiers
Some faction modifiers were incorrectly increasing the target number; they are now correctly decreasing the target number.


IllianiCBT:
Flipping this from draft to live. Description has been updated accordingly.

This is a big patch with a lot of improvements.

PR #4214, opened 2024-06-12
Add basic force generator role functionality to MekHQ scenario random unit generation

SuperStucco:
This is another chunky PR, touches lots of bits and pieces inside AtBDynamicScenarioFactory, some just for readability improvements (so I could follow along easily while testing).  There's a minor interaction with pending #4164 (improved bomb selection), in that the populateAeroBombs(...) method gets called towards the bottom of the generateForce(...) method, but the two are complete independent.

Side note: due to the fixed MUL loading issue from #4213 and using similar functionality, did additional checking specifically for fixed MUL affects.

All these changes allow roles (EDIT: these are the force generator roles, from .../docs/RAT Stuff/rat-generator.txt) to be passed down from the 'generated force' level to the individual random unit selection. This has two immediate effects: artillery, and infantry in hostile environments.

Previously, artillery was read from a flag in the scenario settings and passed down through multiple levels using a specific boolean method parameter. That parameter has been removed, and the scenario flag is now used to add the ARTILLERY role to the force selection.

Infantry for hostile environments is an entirely new feature.  When certain hostile conditions are present, either the XCT or MARINE roles are automatically added to the force selection for infantry, limiting what types of infantry are generated.  Because there are so few official units that fit these parameters, a fall-back position has been included - if no explicit hostile environment infantry is available, normal infantry is generated and the armor kit is swapped out for hostile environment suits.  As a bonus, under less hostile conditions that are still deadly to unprotected infantry, normal infantry will have their infantry kit swapped out for something appropriate e.g. snow/heatsuits for temperature, light environmental suits for low atmosphere pressure.  The role selection and armor kit changes are applied to base scenario forces and scenario modifier forces, including infantry that is loaded into transport bays of APCs.

This PR adds in hooks to to allow force generator role filters to be added to the scenario template and scenario modifier XMLs.  Each force can have an optional roleChoices element with one or more forceRole elements, using the same role formatting as the force generator XMLs.  A single forceRole element will always be used while multiple instances will result in random selection.  As an example, the HostileFacility.XML scenario template might look like this:

```
<scenarioForces>
   ...
   <entry>
            <key>Primary Opfor</key>
            <value>
               ...
               <roleChoices>
                   <forceRole></forceRole>
                   <forceRole>urban</forceRole>
                   <forceRole>urban</forceRole>
               </roleChoices>
               ...
            </value> 
   </entry>
   ...
</scenarioForces>
```

Two thirds of the time the Primary OpFor force will generate using the URBAN role, while the rest will be normal generation without roles (current behavior).

Similarly, the AlliedReinforcments.XML might look like this:

```
<forceDefinition>
    ...
    <forceName>Allied Ground Reinforcements</forceName>
    ...
    <roleChoices>
        <forceRole></forceRole>
        <forceRole>fire_support</forceRole>
    </roleChoices>
    ...
</forceDefinition>
```
Each time the modifier is used, it randomly chooses between standard generation and fire support.

Using an empty forceRole entry will select with no roles, as does not including the <roleChoices> block (which is current behavior).

Multiple roles can combined using commas and no spaces e.g. "support,cargo".  The more roles that are combined the smaller the number to choose from, so use with caution.  Roles may have restrictions that are incompatible with other force properties e.g. the RECON role should not be used on a force that may generate heavy or assault weight units.  Naturally these role selection elements have no affect on fixed MUL unit selection, but should still be omitted for clarity.


Thom293:
This looks amazing and allows for so many new possibilities.  

So if I read it correctly anytime X infantry is spawned it will check the weather and give them the appropriate kit?  Will it work for the fixed mul units?  E.g. the Horse Cavalary or the Air Mobile Infantry or Crack Field Artillery (which are always the same units given how the .muls work)?

I may bug you with lots of questions!

SuperStucco:
> This looks amazing and allows for so many new possibilities.
> 
> So if I read it correctly anytime X infantry is spawned it will check the weather and give them the appropriate kit? Will it work for the fixed mul units? E.g. the Horse Cavalary or the Air Mobile Infantry or Crack Field Artillery (which are always the same units given how the .muls work)?
> 
> I may bug you with lots of questions!

The fixed MUL units are generated in another method, so I don't think the infantry armor kit swaps will address those.  The method which handles the replacement process should be callable for the fixed MUL forces, although there might need to be some jiggery-pokery to get environmental conditions to pass to that method.  Not sure about the viability of a branth in parka, or breathing gear for horses though...

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -380,73 +411,223 @@ public static int generateForce(AtBDynamicScenario scenario, AtBContract contrac
...
+                    isTainted = true;
+                    break;
+                default:
+                    break;
+            }
+        }
+        if (scenario.getWind().isTornadoF1ToF3() || scenario.getWind().isTornadoF4()) {
+            allowsConvInfantry = false;
+            if (scenario.getWind().isTornadoF4()) {
+                allowsTanks = false;
```
Sleet01, line 443: Freakin' glorious, this is excellent!

IllianiCBT:
I see you added a new XML tag, would you mind Discord DMing me the process you took to do this? An improperly added XML tag was the cause of the incorrectly loading Fixed MUL issue and I'd like to not fall into that trap again.

HammerGS (approved):
Tested and seems to work.

HammerGS:
@SuperStucco This looks like it needs to be updated.

SuperStucco:
> @SuperStucco This looks like it needs to be updated.

Not really certain as to what needs to be changed here.

IllianiCBT:
Potentially it's as simple as hitting the 'update branch' button below this message. Hopefully that's all it is.

IllianiCBT:
Seeing a handful of missing `EntitySavingException` in the Gradle report. Might be worth reviewing this PR against `master`. I vaguely remember there were some changes to `EntitySavingException` at the end of last dev cycle, which might be causing issues here. Or it could just be a case of an import needing to be added. @Sleet01 might know more, I believe they were the one who made the changes.

Sleet01:
@SuperStucco it looks like the original issue resolved (it was likely just due to the old build system) but now there is a merge conflict from some of my changes in `AtBDynamicScenarioFactory.java`.

Would you be able to deal with this in your repo?  It should be relatively straightforward (as far as merge conflict resolution goes):

1. Checkout mekhq's master on your machine.
2. pull the latest MekHQ code from the MegaMek repo to your local Master.
3. Checkout your Roles-by-unit-type branch.
4. Use 'git rebase master' to move your branch up the MHQ repo history to be derived from the latest code.
5. When git prompts you to resolve the conflicts (if any), keep the Master lines in case of direct conflicts, and your new code where it doesn't directly conflict.
6. Follow the Git instructions (at least on the CLI) to `git add </path/to/changed/files` (or use the GUI)
7. Use `git rebase --continue` (or its GUI equivalent) until complete.
8. Push the rebased branch up.

If this sounds like too much trouble, I can try to resolve the conflicts directly in the GitHub editor, but I generally don't like to do this because it puts the remote branch ahead of your local branch, and fixing that state can be an even bigger hassle.

What do you think?

Sleet01:
@SuperStucco I'm not sure all the merges went well; it looks like some of the older CI files got pulled in as well.  I will examine more closely when I get the time.

IllianiCBT:
Are you using IDEA? If so, check out `Roles-by-unit-type`, then select `master`. Make sure `master` is fully updated, then right-click `master` and select `compare with 'roles-by-unit-type'`. This will show you all the areas where `master` and `roles-by-unit-type` differ and will allow you to selectively rollback the bad merge until you're left with the changes you want to keep. It's a pain in the bum, but is likely the easiest way to handle things at this point.

IllianiCBT:
Another option, would be to manually-copy paste your changes into a new branch and just chalk this one up as a loss. If there are still merge issues at that point, DM me on Discord and I'll walk you through resolving them.

HammerGS:
Wanted to check on this. Do you still need any help sorting this out?

SuperStucco:
> Wanted to check on this. Do you still need any help sorting this out?

Yeah.  Not sure what/how is getting messed up here, not exactly doing well with the build script (?) updates, and a number of related issues.  Not even sure what I've got here is everything that was originally submitted.

SuperStucco:
OK, I *think* I've gotten the issues worked out and it's building properly here.  There's no code updates to push, and the failed check looked to be the gradle version/build script update problem which I've hopefully updated.

IllianiCBT:
Hi Stucco, it looks like you need to update your MM and (presumably) your MML repos to current. I spoke with Tex and they confirmed that's the current cause behind the Gradle test failing. In their words...

> the CI scripts check for authors repo branch (with same name) then defaults to their master

SuperStucco:
> Hi Stucco, it looks like you need to update your MM and (presumably) your MML repos to current. I spoke with Tex and they confirmed that's the current cause behind the Gradle test failing. In their words...
> 
> > the CI scripts check for authors repo branch (with same name) then defaults to their master

Might be part of the issue - I'm not working off a specifically named 'master' branch as parent.  I've had more than a few issues in the past I consider the actual 'master' branch completely unreliable at this point.  So I'm working off a number of other branches which are named other-than-master e..g. master-v3 for MM and master-mhq for MHQ.

SuperStucco:
Someone else is going to have to take a whack at fixing this, because I'm just going in circles - SerialKiller error, then than goes away and is replaced with code errors (even though I haven't changed anything), then back again.

rjhancock:
Did you update your MM repo with the same branch name to the latest code of master?

IllianiCBT:
Closing as super-ceded by https://github.com/MegaMek/mekhq/pull/4418

PR #4215, opened 2024-06-12
Fixed Handling of autoAwards Post-Scenario Kill Tracking

### LINKED ISSUE #4211: [Nightly] Kills tracking broken
# Environment

V: 0.49.20 Snapshot-8138370
O: linux (Ubuntu 20.04.4)
J: openjdk 11.0.23+9-1ubuntu1~20.04.2
Window Manager: Xfwm4 4.14
Desktop Environment: Xfce 4.14

# Description

Fluff information "kills" will be overwritten with the value of the last battle.

## Before
before

## After
after

(Edit: added screenshots)

IllianiCBT:
This was a really nasty bug. I ended up having to go through the post-scenario autoAwards code line by line to find the culprit.

The problem is, while I've fixed the issue, I still don't know _why_ it was an issue in the first place.

If anyone has any insight as to why `removeIf` was writing to a list outside of scope I'd appreciate hearing an explanation.

## Closes
Closes #4211


MekHQ/src/mekhq/campaign/Kill.java
IllianiCBT, line 1: This isn't needed, but given how evil this bug was I got paranoid.

kernhe:
Just tested the changes with two different games and the bug is fixed. Thank you!

IllianiCBT:
> Just tested the changes with two different games and the bug is fixed. Thank you!

I am so freaking relieved to hear that

kernhe (approved):
Works as expected

MekHQ/src/mekhq/campaign/personnel/autoAwards/AutoAwardsController.java
```diff
@@ -787,11 +787,17 @@ private Map<Integer, List<Object>> ScenarioKillAwardsManager(List<UUID> personne
         for (UUID person : personnel) {
             Map<Integer, List<Object>> data;
 
-            List<Kill> kills = campaign.getKillsFor(campaign.getPerson(person).getId());
-            kills.removeIf(kill -> kill.getScenarioId() != scenarioId);
+            List<Kill> allKills = campaign.getKillsFor(campaign.getPerson(person).getId());
+            List<Kill> scenarioKills = new ArrayList<>();
+
+            for (Kill kill : allKills) {
```
NickAragua, line 793: This can get a little bit inefficient for long-serving pilots with hundreds or thousands of kills. Any thoughts on optimizing it?

IllianiCBT: A little rework might make it so that only kills from the scenario just concluded are passed in, which would remove the vast majority of the necessary filtering. Will depend on how scenario kills are distributed during the post-scenario handling in mhq.

IllianiCBT: Was an easy fix. Thanks for mentioning this

IllianiCBT:
> Would it be possible to run them all on a single screen? Right now it seems to run one screen per award?

It's one screen per award type. So all kill awards fire; then all theatre of war awards; and so on. This was done for performance reasons, as it helps stagger the amount of information autoAwards needs to calculate and display for very large campaigns. Prior to making this decision really big campaigns could cause lock-ups.

Thom293:
Ahh. Makes sense.  Ok thank you.

PR #4231, opened 2024-06-17
Added Nag for Invalid Faction & Updated Missing Federation Commonwealth Start/End Date

### LINKED ISSUE #4222: Invalid Faction
### Environment
     While in mekhq-windows-0.49.19.1
     Using Windows 10
     Java Vendor: Eclipse Adoptium
     Java Version: 11.0.22

### Description
     While playing MekHQ - Stratcon, I selected the year 3049 and the faction fed suns. For this reason, i was unable to get a contract because they are fedcom now.  Illiani, wrote, "I’ll try and get a ‘your faction no longer exists’ nag put in before 49.20 launches"

### Files
     N/A

IllianiBird:
This PR introduces a notification that appears when the user attempts to advance the day while the campaign faction is set to one that doesn't yet exist or no longer exists. This notification was needed as AtB will stop generating contracts for factions that no longer hold planetary systems.

It is not possible to check whether planetary systems are being held without parsing all systems in the universe, which I deemed too resource intensive for this purpose. So we rely on faction start/end data, instead.

A special handler has been added for the formation and separation of the Federated Commonwealth, as this accounts for most faction invalidations. This handler is necessary because both the Federated Suns and Lyran Alliance/Commonwealth still exist during this period (at least as far as MHQ is concerned) but are dormant and will not generate AtB Contracts.

I also added start and end dates to the Federated Commonwealth faction, as these were missing. Additionally, I added a missing faction change for the Yed Posterior system, which remained under Federated Commonwealth control long after the FC ceased to be an entity. It now correctly switches to Lyran Alliance control in 3063 (Sarna data, so marked as 'non-canon').

## Closes
Closes #4222


MekHQ/data/universe/factions.xml
```diff
@@ -1101,6 +1101,8 @@ successor - unimplemented tag describing another faction code as the specified f
     <faction>
         <shortname>FC</shortname>
         <fullname>Federated Commonwealth</fullname>
+        <altNamesByYear year='3028'>Federated Commonwealth Alliance</altNamesByYear>
```
HammerGS, line 1104: This date would be too early. Per Fed Com Civil War - April 1st/3042 is when the AFFC was founded, but FedCom didn't come into being until the Death of Melissa on June 19th 3055. I would recommend using the 3042 date as that is when you'd see the military hiring Mercs.

MekHQ/data/universe/factions.xml
```diff
@@ -1110,6 +1112,8 @@ successor - unimplemented tag describing another faction code as the specified f
         <layeredForceIconLogoFilename>Federated Commonwealth.png</layeredForceIconLogoFilename>
         <colorRGB>248,212,44</colorRGB>
         <tags>is,super,playable</tags>
+        <start>3028</start>
+        <end>3068</end>
```
HammerGS, line 1116: This also could be replaced with - 24 April 3067, when things got officially dissolved.

IllianiCBT: Here, I wasn't sure if it accepted LocalDate or just year; I couldn't spy any other uses of LocalDate, so my concern with using 3067 is that it would cause the faction to end on 01/01/3067. By using 3068 we ensure the faction still exists for the period of 3067 that the FedCon still existed.

MekHQ/data/universe/factions.xml
IllianiCBT: My understanding is that the Federated Commonwealth Alliance (as opposed to the Federated Commonwealth) came into being when the accords were signed in 3028. Then, in 3055 the alliance officially rebranded as the Federated Commonwealth in 3055, iirc, after things went poorly during the war of 3039.

Thom293:
This horrid political best has raised its head again!  Who dares poke the sleeping giant?!?  ;)

IllianiCBT:
@HammerGS what's the final verdict on the dates? :)

HammerGS (approved):
I'm good with the data for now. I think down the road we should looks at the dates again.

IllianiCBT:
We're of the same mind there.

Issue #4057, opened 2024-05-06
[0.49.19] NPE, "Could not find a mech summary"


repligator:
Edited - Still present in latest nightly

### Environment
Nightly 2024-05-23T01:11:31.212783451, Linux, Java 17.0.10

### Description
I used the 'Purchase Unit' option to purchase a new unit. Got an NPE, logs show "Could not find a mech summary". Save file and logs are attached.
[mekhq.log](https://github.com/MegaMek/mekhq/files/15225825/mekhq.log)
[3151Bears31510511.cpnx.gz](https://github.com/MegaMek/mekhq/files/15225829/3151Bears31510511.cpnx.gz)
[Autosave-1-GBD3098-31020730.cpnx.gz](https://github.com/MegaMek/mekhq/files/15225854/Autosave-1-GBD3098-31020730.cpnx.gz)
[Autosave-2-GBD3098-31020729.cpnx.gz](https://github.com/MegaMek/mekhq/files/15225855/Autosave-2-GBD3098-31020729.cpnx.gz)

repligator:
Woops, wrong save file.
[GBD309831020731.cpnx.gz](https://github.com/MegaMek/mekhq/files/15225868/GBD309831020731.cpnx.gz)

HammerGS:
From log

```
15:54:03,457 ERROR [mekhq.MekHQ] {AWT-EventQueue-0}
mekhq.MekHQ.lambda$main$0(MekHQ.java:242) - Uncaught Exception Detected
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:209)
	at mekhq.campaign.Quartermaster.buyUnit(Quartermaster.java:467)
	at mekhq.campaign.unit.UnitOrder.find(UnitOrder.java:184)
	at mekhq.campaign.Campaign.acquireEquipment(Campaign.java:2539)
	at mekhq.campaign.Campaign.goShoppingByPlanet(Campaign.java:2332)
	at mekhq.campaign.Campaign.goShopping(Campaign.java:2196)
	at mekhq.campaign.Campaign.newDay(Campaign.java:3511)
	at mekhq.campaign.CampaignController.advanceDay(CampaignController.java:82)
	at mekhq.gui.CampaignGUI.lambda$initTopButtons$59(CampaignGUI.java:1067)
```

repligator:
New save/logs from nightly
[...]

PR #4236, opened 2024-06-18
Fixes a bug that would prevent mechs with differing clan/IS names fro…

repligator:
Fixes a bug that would prevent mechs with differing clan/IS names from being purchased through the purchase unit menu. See issue #4057 .

`get.Chassis + get.Model` would output "Gladiator J" whereas `get.FullChassis + get.Model` outputs "Gladiator (Executioner) J".


Sleet01 (approved):
LGTM!
I thought we caught all the remaining calls in the last pass; nicely spotted.

PR #4237, opened 2024-06-19
Added Prestigious Academies (A-A)

IllianiCBT:
This PR adds the first batch of prestigious academies. Pending Hammer's approval, these are the final values and will replace the placeholder values previously merged.

I also made some minor changes to the academy code:

I updated local academy qualification start dates to match those used by prestigious academies.

I removed 'promotions' as I was not satisfied with their implementation. I plan to review the Ranks System down the road and re-implement it then.

Child graduations no longer use the same graduation text as adults.

Additionally, I removed the 'faction' field, as it was a relic from a previous iteration and is no longer necessary.


HammerGS (approved):
Can't speak to code, but data looks good.

PR #4281, opened 2024-06-23
Added Population Check to Education Module

IllianiCBT:
This PR adds a check to ensure a system is populated before allowing personnel to attend academies there.

Additionally, I added a check for total system population depletion after enrolling (as part of the new week checks). This ensures more accurate dates for the destruction of prestigious academies when that destruction results from system depopulation (assuming we have accurate data). If accurate data isn't available, mhq will default to the academy's static destruction year (defined in the academy XML) as a fallback.

This also allows extending academy destruction events to local academies, which wasn't possible previously.

Marking this as a bug, as personnel being able to attend academies with 0 system population is unintended behavior.


NickAragua (commented):
Why are we bothering making this an option? I guess it's fine, but I just don't see a situation where we'd want to have academies in uninhabited systems.

IllianiCBT:
> Why are we bothering making this an option? I guess it's fine, but I just don't see a situation where we'd want to have academies in uninhabited systems.

The option is to enable the message that tells users why there is no academy in the relevant system. Otherwise the academies will just be missing with no explanation as to why.

IllianiCBT:
I just spotted the cause of confusion: `enablePopulationConflict` should have been `enableShowPopulationConflict`. Pushing a fix.

NickAragua:
> > Why are we bothering making this an option? I guess it's fine, but I just don't see a situation where we'd want to have academies in uninhabited systems.
> 
> The option is to enable the message that tells users why there is no academy in the relevant system. Otherwise the academies will just be missing with no explanation as to why.

I guess in that case, I'm even more convinced that we don't really need it as an option.

IllianiCBT:
Would you be able to advise why you don't think the option is needed? I think it would be helpful if I better knew the direction you were coming from :)

NickAragua:
I think a screenshot of the affected area may be helpful to understand what we're talking about exactly. Right now, to me, it looks like you're adding a campaign config option to control the display of invalid locations of academies in a menu where academies are being assigned to a person? What's the scenario where you'd want to be able to say that a person is attending a blown up academy?

IllianiCBT:
The campaign option is part of this range of choices. Each option determines which academies will still be displayed when certain ineligibility criteria are met. Instead of simply disabling the 'show ineligible academies' option, which would hide all academies the individual is ineligible for, this option allows for more nuanced control.

When in use, academies will be shown like this:

[Image](https://github.com/MegaMek/mekhq/assets/103902653/65b3fc68-0028-4070-82fc-a855671d37ba)

These academies cannot be interacted with, but they inform the user why the academy can't be selected. However, the user may decide that they don't care, in which case the option can be disabled. But perhaps they do care, or maybe they only care *if* the system is depopulated. By removing the campaign option, we take away that choice from the user. Moreover, it could potentially lead to false bug reports with users claiming that academies aren't displaying. Or, worse, that academies aren't showing when they should, but the user assumes it's because their personnel are ineligible.

I introduced these options to prevent situations like this, where it's difficult for the user to easily see what academies they can send personnel to:

[Image](https://github.com/MegaMek/mekhq/assets/103902653/ecc0f76f-00c2-4981-984d-60cd5ec563d8)

But also to avoid situations like this, where no academies are visible, but the user has no way to tell why:

[Image](https://github.com/MegaMek/mekhq/assets/103902653/f1167ce7-64c2-4ac3-916f-b3b2ea7915c0)

I hope this clarifies things. I understand it's not always easy to envision how things will appear in-game when all you have is the GitHub changes list. :)

NickAragua (approved):

NickAragua:
Fair enough. I'm not really a giant fan of adding an option for everything as it tends to increase code complexity; I'm also not convinced that adding an extra option buried in campaign settings somewhere will help with user complaints. But since the work is done already, we may as well leave it in.

IllianiCBT:
I see your point. I know you just approved this, but I’m going to go ahead and reduce the options down to just ‘display ineligible academies’. I think having all these campaign options wolves a problem that doesn’t exist.

PR #4332, opened 2024-07-02
CI Updates for windows build and normalizing

rjhancock:


kuronekochomusuke:
should the memory also be increase for mekhq?  since when you launch megamek the memory is shared between the 2

IllianiCBT:
Honestly, that would be pretty rad

rjhancock:
Everything should be building correctly now. MekHQ was upped to 4G. MM set to 2G. MML set to 1G.

PR #4369, opened 2024-07-11
Added Option to Include Non-Combatants in Field Kitchen Capacity Calculations

### LINKED ISSUE #4329: Field Kitchens should not need to service support personnel
### Environment
MegaMek Version: MekHQ v0.49.20 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description

MekHQ 0.49.20 adds a rough implementation of Field Kitchens, diverging from _Campaign Operations_ primarily in not tracking actual deployment. The current approach instead compares the number of Field Kitchens on fully-crewed units against the total number of personnel in the force, providing the Fatigue bonus to everyone if there's enough capacity.

Putting aside the Field Kitchen-equivalents included with quarters on Small/Large Craft, this approach shouldn't be counting support personnel at all. Page 219 of the fourth printing of _Campaign Operations_ describes Field Kitchens thusly:

> **Field Kitchen**: Forces that have access to field kitchen reduce fatigue by 1 extra point. One field kitchen can supply 150 combat troops (and their support personnel) per Morale/ Fatigue Cycle. When determining which forces have access to the available field kitchen resources, do not split any units; that is, all 28 ground-pounders in that platoon are counted or none are.

While the logic of this is debatable, the meaning is not: players should only need one Field Kitchen for every 150 *combat* personnel, and the requirement should ignore support personnel entirely due to being included in the former.

---

This issue was opened at [the request](https://discord.com/channels/458705327911731231/458706545379639296/1256445791451484263) of @IllianiCBT.

IllianiCBT:
CamOps excludes support staff from field kitchen capacity calculations, which just doesn't feel right to me. Surely Bill the Admin/HR deserves to eat too!

The option added by this PR is enabled by default and prohibits non-combatants from eating at the field kitchen. This allows us to simulate CamOps, but still leave the option open for non-combatants getting served for users who want that (read: me).

Closes #4329

Addendum: while I was fiddling around in `Campaign.java` I figured I'd also do some light refactoring, to clean things up.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -330,6 +330,8 @@ chkUseInjuryFatigue.text=Injuries Increase Fatigue (unofficial)
 chkUseInjuryFatigue.toolTipText=If enabled, personnel gain fatigue whenever they suffer injuries during a scenario.
 lblFieldKitchenCapacity.text=Field Kitchen Capacity
 lblFieldKitchenCapacity.toolTipText=How many personnel can be served per Field Kitchen? Reduces effective Fatigue by 1.
+chkFieldKitchenIgnoreNonCombatants.text=Ignore Non-Combatants
+chkFieldKitchenIgnoreNonCombatants.toolTipText=If enabled, non-combatants will not count towards the number of personnel being served by a field kitchen.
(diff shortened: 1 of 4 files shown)


Saklad5:
> Surely Bill the Admin/HR deserves to eat too!

I think the idea is that Bill the Admin is back on the DropShip or something. I won't deny it's a bit odd, but until MekHQ is capable of only requiring Field Kitchens while deployed this is a better compromise.

HammerGS (approved):

PR #4377, opened 2024-07-11
Expanded Manual Assignment of Personnel Statuses to Include `PoW`, `On Leave`, and `AWOL` Statuses

IllianiCBT:
The PersonnelFilter for the MIA status now includes friendly Prisoners of War (PoWs). The GUI properties file has been updated to reflect this change in the relevant text and tooltips. This decision was made to avoid creating an additional category and to reduce option clutter.

Additionally, I adjusted the `PersonnelStatus.getImplementedStatuses()` method to no longer filter out the `On_Leave`, `AWOL`, and `PoW` statuses. These statuses were previously disabled as they were not implemented, but with ongoing expansions to the personnel modules, this is no longer the case. This allows users to manually assign these statuses to personnel.

Although the PoW status will not be fully utilized until the S&E module is released later this year, it is now available for users to incorporate into their role-playing scenarios.


HammerGS (changes requested):
Tested and seems to work.  (Ignore this mis-click)

HammerGS (approved):
Ignore the Requested changes.

PR #4382, opened 2024-07-12
Added In-Unit Education to Education Module

IllianiCBT:
This PR adds a new standard academy set to the education module: Unit Education. This set has five options (Unit Creche, Home Schooling, Apprenticeship, In-House Boot Camp, and In-House Basic Training) and is focused on providing users with options to train personnel in-house, rather than sending them to an academy for more formal training.

Training someone in-house produces inferior results than if personnel were to be sent to an academy, and is more dangerous, but it is free. So there is that. You get what you pay for.

This PR includes fixes from #4365 and should be merged after that PR.


HammerGS (approved):
Tested and seems to work

PR #4388, opened 2024-07-12
 Replaced CargoCapacityNagDialog functionality

IllianiBird:
This commit removed the CargoCapacityNagDialog functionality and all of its references. The cargo capacity report was redesigned to be more interactive, showing color-coded status directly in the Campaign Summary report. Users will no longer receive a dialog prompt when cargo capacity is exceeded, instead, they'll be visually prompted within the campaign statistics display, as shown below.

Within Capacity:
image

Cargo at Capacity:
image

Cargo Exceeds Capacity
image

(Actual colors dependent on MekHQ color options)

This implementation was my original intent, prior to writing the nag. However, at the time, I didn't have the knowledge necessary to get it to work so settled on a nag. That nag proved _incredibly_ annoying, so has been removed.


HammerGS (approved):
Tested and seems to work

PR #4395, opened 2024-07-13
Added Execution and Jettison Options for Prisoners

IllianiCBT:
Two new options for handling prisoners have been introduced: "execute" does what it says on the box, while "jettison" launches them into space.

The 'free' prisoner action has been limited to while the campaign is planet-side. To simulate that there isn't really anywhere to free people in the depths of space. Speaking of which...

'Jettison' is only available while the campaign is _not_ planet-side.


HammerGS (approved):
Tested and seems to work.

PR #4396, opened 2024-07-13
Changes some wording around to help avoid confusion about StratCon.

repligator:
Most of the discussion of StratCon and AtB is phrased as if they are mutually exclusive. 
"I think StratCon is better at X then AtB"
"I still like using AtB instead of StratCon",
This could be a point of confusion for new users who don't realize that the current implementation of StratCon relies on AtB to function. This PR is meant to preempt such questions as
"Where is the StratCon tab?"
"I turned AtB off, why can't I see any StratCon contracts?"
"Does X apply to just AtB or StratCon as well?"

I've marked this as a draft until it can be discussed by the userbase.


Sleet01 (approved):
LGTM!

IllianiCBT:
Bonus Parts only exist in Missions classes as AtBContract, so the options should probably still mention AtB (and StratCon, if you wish).

PR #4398, opened 2024-07-13
Added the Ability to Ransom Friendly PoWs

IllianiCBT:
As per title, this PR adds the ability to ransom friendly PoWs. The valuation of PoWs is identical to the valuation of enemy prisoners.

This PR is designed to work alongside #4377 to expand user role-play options.

PR #4400, opened 2024-07-13
Implemented the Capture of Missing Friendly Personnel During Scenario Resolution

IllianiCBT:
During scenario resolution, if the user does not control the battlefield, any MIA personnel are captured by the enemy and are assigned the POW personnel status. They can then be ransomed using #4398.


IllianiCBT:
This doesn't appear to be working as intended. Flipping to draft until I can investigate more.

IllianiCBT:
Nevermind, I'm just dumb. Flipping back to live.

HammerGS (approved):

PR #4401, opened 2024-07-13
Update news item for raid on Mararn

azhais:
News incorrectly lists it as Maram.

https://www.sarna.net/wiki/Mararn

Also verified this in the Tortuga Prime PDF


IllianiCBT:
Congratulations on your first PR! Here's to many more!

PR #4415, opened 2024-07-15
Refactored Award Tier Count Calculation in PersonViewPanel

IllianiCBT:
Awards would display incorrectly if award tier was set to 1, effectively skipping the first tier of award images. This PR addresses that.

### DIFF
--- a/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/PersonViewPanel.java
@@ -345,8 +345,14 @@ public class PersonViewPanel extends JScrollablePanel {
                 rowRibbonsBox.setBackground(Color.RED);
             }
             try {
+                int numAwards = person.getAwardController().getNumberOfAwards(award);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/view/PersonViewPanel.java
```diff
@@ -345,8 +345,14 @@ private Box drawRibbons() {
                 rowRibbonsBox.setBackground(Color.RED);
             }
             try {
+                int numAwards = person.getAwardController().getNumberOfAwards(award);
```
HoneySkull, line 348: Consider putting this code that is duplicated multiple times into method for readability.  Changes to one of these blocks will likely need the same change to all three similar logic.

IllianiCBT: You're absolutely right. I'll see it done. Thanks for pointing this out. :)

HoneySkull (approved):

PR #4418, opened 2024-07-15
Added Basic Force Generator Role Functionality to Scenario Random Unit Generation

IllianiCBT:
**All credit to SuperStucco. All I've done is duplicated their work to resolve git issues.**

This PR should not be merged until it has been approved by @SuperStucco as the chance of errors creeping in while I transcribed their work is non-zero.

> This is another chunky PR, touches lots of bits and pieces inside AtBDynamicScenarioFactory, some just for readability improvements (so I could follow along easily while testing).  There's a minor interaction with pending #4164 (improved bomb selection), in that the populateAeroBombs(...) method gets called towards the bottom of the generateForce(...) method, but the two are complete independent.
> 
> Side note: due to the fixed MUL loading issue from #4213 and using similar functionality, did additional checking specifically for fixed MUL affects.
> 
> All these changes allow roles (EDIT: these are the force generator roles, from .../docs/RAT Stuff/rat-generator.txt) to be passed down from the 'generated force' level to the individual random unit selection. This has two immediate effects: artillery, and infantry in hostile environments.
> 
> Previously, artillery was read from a flag in the scenario settings and passed down through multiple levels using a specific boolean method parameter. That parameter has been removed, and the scenario flag is now used to add the ARTILLERY role to the force selection.
> 
> Infantry for hostile environments is an entirely new feature.  When certain hostile conditions are present, either the XCT or MARINE roles are automatically added to the force selection for infantry, limiting what types of infantry are generated.  Because there are so few official units that fit these parameters, a fall-back position has been included - if no explicit hostile environment infantry is available, normal infantry is generated and the armor kit is swapped out for hostile environment suits.  As a bonus, under less hostile conditions that are still deadly to unprotected infantry, normal infantry will have their infantry kit swapped out for something appropriate e.g. snow/heatsuits for temperature, light environmental suits for low atmosphere pressure.  The role selection and armor kit changes are applied to base scenario forces and scenario modifier forces, including infantry that is loaded into transport bays of APCs.
> 
> This PR adds in hooks to to allow force generator role filters to be added to the scenario template and scenario modifier XMLs.  Each force can have an optional roleChoices element with one or more forceRole elements, using the same role formatting as the force generator XMLs.  A single forceRole element will always be used while multiple instances will result in random selection.  As an example, the HostileFacility.XML scenario template might look like this:
> 
> ```
> <scenarioForces>
>    ...
>    <entry>
>             <key>Primary Opfor</key>
>             <value>
>                ...
>                <roleChoices>
>                    <forceRole></forceRole>
>                    <forceRole>urban</forceRole>
>                    <forceRole>urban</forceRole>
>                </roleChoices>
>                ...
>             </value> 
>    </entry>
>    ...
> </scenarioForces>
> ```
> 
> Two thirds of the time the Primary OpFor force will generate using the URBAN role, while the rest will be normal generation without roles (current behavior).
> 
> Similarly, the AlliedReinforcments.XML might look like this:
> 
> ```
> <forceDefinition>
>     ...
>     <forceName>Allied Ground Reinforcements</forceName>
>     ...
>     <roleChoices>
>         <forceRole></forceRole>
>         <forceRole>fire_support</forceRole>
>     </roleChoices>
>     ...
> </forceDefinition>
> ```
> Each time the modifier is used, it randomly chooses between standard generation and fire support.
> 
> Using an empty forceRole entry will select with no roles, as does not including the <roleChoices> block (which is current behavior).
> 
> Multiple roles can combined using commas and no spaces e.g. "support,cargo".  The more roles that are combined the smaller the number to choose from, so use with caution.  Roles may have restrictions that are incompatible with other force properties e.g. the RECON role should not be used on a force that may generate heavy or assault weight units.  Naturally these role selection elements have no affect on fixed MUL unit selection, but should still be omitted for clarity.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
@@ -94,7 +94,7 @@ public class AtBDynamicScenarioFactory {
      * @param template The template to use when populating the new scenario.
      * @param contract The contract in which the scenario is to occur.
      * @param campaign The current campaign.
-     * @return
+     * @return A new Scenario object with the provided settings
      */
     public static AtBDynamicScenario initializeScenarioFromTemplate(ScenarioTemplate template, AtBContract contract, Campaign campaign) {
         AtBDynamicScenario scenario = new AtBDynamicScenario();
@@ -214,8 +214,8 @@ public class AtBDynamicScenarioFactory {
      * @param scenario    Scenario for which we're generating forces
      * @param contract    The contract on which we're currently working. Used for skill/quality/planetary info parameters
      * @param campaign    The current campaign
-     * @param weightClass The maximum weight class of the units to generate (ignored )
-     * @return How many "lances" or other individual units were generated.
+     * @param weightClass The average weight class across all forces
+     * @return How many "lances" or other individual units were generated?
      */
     private static int generateForces(AtBDynamicScenario scenario, AtBContract contract, Campaign campaign, int weightClass) {
         int generatedLanceCount = 0;
@@ -308,7 +308,7 @@ public class AtBDynamicScenarioFactory {
      * @param campaign           The current campaign
      * @param effectiveBV        The effective battle value, up to this point, of player and allied units
      * @param effectiveUnitCount The effective unit count, up to this point, of player and allied units
-     * @param weightClass        The maximum weight class of the units to generate (ignored )
+     * @param weightClass        The average weight class to generate this force at
      * @param forceTemplate      The force template to use to generate the force
      * @param isScenarioModifier true if the source of generateForce() was a scenario modifier
      * @return How many "lances" or other individual units were generated.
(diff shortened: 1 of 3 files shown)


SuperStucco:
Thanks for this.  A quick scan shows the high points are all there, so it gets a stamp of approval here.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -492,15 +659,13 @@ public static int generateForce(AtBDynamicScenario scenario, AtBContract contrac
...
+                    ArrayList<Entity> arrayGeneratedLance = new ArrayList<>(generatedLance);
                     // bin fill ratio will be adjusted by the loadout generator based on piracy and quality
                     ReconfigurationParameters rp = TeamLoadoutGenerator.generateParameters(
                             cGame,
                             cGame.getOptions(),
                             arrayGeneratedLance,
-                            factionCode,
-                            new ArrayList<Entity>(),
-                            new ArrayList<String>(),
+                            factionCode, new ArrayList<>(), new ArrayList<>(),
```
Sleet01, line 668: Minor stylistic nit: if arguments are expanded into a vertical list, each should be on a separate line so that the eye does not skip over them (as I did here, briefly thinking the `new ArrayList<>()` args had gone missing).

Sleet01 (approved):
LGTM!

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
IllianiCBT: Ah, that's a 'My Bad'. I'll fix it in a sec :)

PR #4441, opened 2024-07-19
Implemented Incremental Loyalty Changes

IllianiCBT:
This PR attempts to make Loyalty more dynamic by introducing event-based Loyalty changes. Now, as personnel 'live' their lives events may occur which result in Loyalty being tested. Sometimes it will improve, other times it will be reduced.

Triggering events include...

- The death of a spouse (especially if the spouse was pregnant)
- The death of a parent (if individual is a child)
- The death of both parents
- The death of a child
- Marriage (affects both parents)
- The birth of a child (affects both parents)
- Graduating from an academy
- The campaign misses a paycheck
- The execution or jettisoning of prisoners

Both _Turnover & Retention_ and _Education_ documentation have been updated to reflect these changes.

**Important**
This PR merges in changes made in #4439 and #4440 and _must_ be merged last to avoid conflicts.


HammerGS (approved):
Tested and I think it works.

PR #4446, opened 2024-07-20
Added Random Personality Functionality

IllianiCBT:
This PR adds the ability to grant personnel randomized personality traits & quirks. These are currently for flavor only, however they will later be utilized by the random events module to personalize events based on the individual's personality.

### Examples
image
image
image


IllianiCBT:
Flipping this to live. Description has been updated accordingly.

PR #4456, opened 2024-07-22
Added Nag for Inability to Afford Next Jump

### LINKED ISSUE #4327: RFE: Advance day warning if unable to afford jump
Given the presence of the other advance day warnings that don't let you advance without confirmation it's somewhat odd that there isn't one for when you're too poor to make a jump. As the cost of the next jump isn't visible, it can catch people out.

IllianiBird:
A new nag warning has been implemented when the campaign cannot afford the next jump. This includes adding a new checkbox option in the MHQOptionsDialog, creating a new UnableToAffordJumpNagDialog class, and appropriate constant and GUI property updates for the new nag option.

This will assist players in being aware of their campaign fund status related to jump affordability. Previously, days would just keep advancing until the user noticed the report error. This was particularly frustrating when using advance multiple days, as there is no way to cancel out of the day advance to address the issue.

Closes #4327


Sleet01 (approved):
Looks good!  Also, and I mean this sincerely, _thank you_.

PR #4463, opened 2024-07-22
Decoupled Dependents from AtB, Reworked Dependents, Reworked New Personnel Ages, Added Documentation

### LINKED ISSUE #3570: RFE: cap additional adult dependents based on current admin levels
Every January 1st my campaign gets 80-100+ adult dependents added (current campaign is somewhere north of 800, they have a small town living in their backyard).  Kind of nice as a renewing resource for when extra personnel are needed but quickly blows out any semblance of keeping pace with required numbers of administrators.

Rather than blindly adding more adult dependents every year, it would be nice to have some kind of upper limits of what the unit can support.  This may be raw numbers of administrator personnel, and/or numbers of admin/HR members.  Either way, when January 1st rolls around dependents are only added up to the capped number.

This would not include child dependents of current members, staff, or dependents.

IllianiBird:
This PR primarily reworks the random addition and removal of Dependents from a campaign. In doing so it decouples dependents from AtB and moves to a monthly addition/removal model, rather than annual. This creates a more dynamic feel to addition and removal.

The number of random Dependents is now soft-capped at 5% of the total active, adult, non-dependents in the campaign. This helps reduce the issue of larger campaigns ending up with small towns traveling with them.

While I was reworking dependents I also incorporated the information on how they work into the prisoners & defection documentation. Neither random dependents nor prisoners are sufficiently detailed to warrant their own documentation, so it made sense to merge them.

Finally, I slightly adjusted the ages new personnel spawn at. In particular, dependents now have much larger range of possible ages and Veteran and Elite personnel will spawn at higher ages.

Closes #3570


IllianiCBT:
Updated to remove conflicts

IllianiCBT:
Flipping this to draft so I can't merge it by accident

IllianiCBT:
Flipping this to live. Assuming the tests pass, it should be good for review.

PR #4466, opened 2024-07-22
Added Option to Grant Random Toughness when Using Toughness

### LINKED ISSUE #3939: RFE:  MHQ Toughness
We discussed this recently in another PR. 

Right now, Toughness is unusuable unless manually assigned.

As far as I can tell, Characters cannot spawn with it, and it cannot be purchased with XP.  Sometimes characters will spawn with "Toughness 0" but never with any actual points in it.

I propose it that it become a "birth" stat that they can be generated with, and that they can spawn with 0-2 points.

Purchasing via XP would be ok too, but since I view this as a "natural" stat, I think it would be better if it wasnt purchaseable.  Some people are just tough, and some are not.  But either way would be fine.

I think its control should belong here, on this campaign options tab:

image

IllianiCBT:
This PR adds a campaign option to provide personnel with a random Toughness bonus or malus. This is based on a 2d6 roll, with a 2 granting -1 Toughness; and 12 granting +1. This is accompanied by a biography entry that describes why the individual has this modifier.

This only affects personnel if Toughness is enabled. Furthermore, only new personnel will be affected.

This PR should be merged after #4446 as that PR has been rebased onto this point to avoid conflicts.

Closes #3939


Thom293:
Very nice. Can't wait to check out the new backgrounds. Thank you.

HammerGS (approved):
Tested and seems to work.

PR #4468, opened 2024-07-23
Implemented Monthly Personnel Data Cleanup

IllianiBird:
Added a feature to delete personnel data on the last day of each month. The cleanup applies to individuals who have departed the unit over a month ago.

Campaign options have been added to exempt dead or retired personnel.

The campaign option to track departure date has been removed, the functionality is now always enabled. This functionality is needed for this feature to work fully. And, really, these doesn't seem to be a good reason not to track that data.

This addition should help reduce save bloat. As we don't do anything which these individuals, there is no need to keep them in the save. I opted to have the cleanup occur on the last day of each month, as most of our major personnel updates occur on the 1st; so we can get rid of junk data ahead of the main workhorse functions. I opted to only remove personnel who have departed the campaign over a month ago, as this gives the user plenty of time to undo any unwanted departures.


HammerGS:
Has merge conflicts.

HammerGS (approved):
Tested and seems to work.

IllianiCBT:
There will probably be more conflicts as stuff gets merged over, but they're fixed for now

IllianiCBT:
Noting that this was modified after approval. I changed the handling of legacy campaigns so that personnel who have departed the campaign, but lack a departure date, have their departure date set as the current campaign date. Previously they were just marked for removal.

PR #4480, opened 2024-07-24
Updated CamOps Reputation Calculations

IllianiCBT:
This PR adds a number of improvements to how we handle the CamOps unit rating. The below list isn't definitive, but covers the main points:

- Update a number of unit rating calls so that they use available boolean methods, rather than matches. i.e. `c.getCampaignOptions().getUnitRatingMethod().equals(UnitRatingMethod.CAMPAIGN_OPS)` becomes `c.getCampaignOptions().getUnitRatingMethod().isCampaignOperations()`
- Renamed 'Unit Rating' to 'Reputation' to match current CamOps terminology
- Added the tracking of non-combatant personnel transport requirements
- Added the ability for combat personnel to pull 'double-duty' to reduce Admin personnel needs, as per CamOps
- Removed Medics from counting towards Admin Rating, as current printings do not make a reference to Medics
- Periods of inactivity now affect unit rating, crime is still not tracked, but will be in a future update
- Vessels that are not fully crewed no longer provide benefits such as bays and docking collars.
- Incorporated personality characteristics into unit rating (if personality characteristics are enabled). Later updates are going to add Campaign Options to toggle this.

Future plans include adding the tracking of crime and figuring out how to handle AtB unit rating related modifiers.


IllianiCBT:
After discussing with Hammer, our plans regarding this have changed. Closing this PR while I work to implement our plans.

PR #4489, opened 2024-07-26
Refactored Personality Characteristic Generation Logic

IllianiCBT:
Simplified the personality characteristic generation by removing redundant methods and reducing the complexity of trait selection.

Adjusted probability checks to prevent trait bloat and ensure uniqueness, and added a recursive check to guarantee at least one characteristic is generated per person.

Relocated characteristic descriptions into their own resource bundle.

Updated documentation.


HammerGS (approved):
Tested and seems to work.

PR #4492, opened 2024-07-27
Added Promotion-Based autoAwards Processing

### LINKED ISSUE #4336: RFE: ADD THE RESERVED WORD "IMPLICIT " TO RANKS
I know what I want too say but not sure how too put into words. Hope this is ok

Add the Reserved word "implicit" for the <range></range> of the< item>RANKS</item>

with "implicit",the trigger would only fire when the exact rank  is selected. IMO, this would be much more usable than Inclusive or exclusive. Well, It would be for me. 

KG_BRANDENBURG

IllianiCBT:
This has been a feature much requested since I first started work on autoAwards. Via this PR I have added a new 'promotion' sub-category of Rank Awards. These awards are triggered whenever someone changes rank, but rather than being tied to specific thresholds, each award is assigned a specific rank. It will only be issued to personnel with that specific rank and, when issued, all other promotion type awards will be removed.

The use case here is for users who want to use autoAwards to distribute and manage rank insignia. To that end, I have also added functionality so that users can optionally limit promotion awards to only be valid for specific Rank Systems. For example, you might have one set of rank insignia for the SLDF and another for the MAF.

Documentation has not been updated for this PR and will be included in a follow-up PR.

Closes #4336

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/PersonAwardController.java
+++ b/MekHQ/src/mekhq/campaign/personnel/PersonAwardController.java
@@ -188,6 +188,27 @@ public class PersonAwardController {
     }
 
     /**
+     * Removes an award given to this person (without logging the removal) based on:
+     * @param setName is the name of the set of the award
+     * @param awardName is the name of the award
+     * @param awardedDate is the date it was awarded, or null if it is to be bulk removed
+     */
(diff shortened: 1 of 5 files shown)


HammerGS (approved):
Tested and seems to work

PR #4493, opened 2024-07-27
Added Ability to Tell autoAwards to Ignore Individual Custom Award Sets

IllianiBird:
This PR introduces another long requested feature: the ability to filter award sets so that autoAwards won't use specific custom award sets. This is as simple as entering the names of the award sets you don't want autoAwards to use into a free text box in campaign options. As it runs autoAwards will check whether an award set appears in that list, and if it does it will ignore all awards from that set.


HammerGS (approved):
Tested and seems to work

PR #4494, opened 2024-07-27
Improved Award Tooltips with XP & Edge Benefits

IllianiCBT:
The tooltips for awards will now show what XP and Edge awards have been granted by those awards. This respects campaign options, so will not show Edge if Edge benefits from awards have been disabled.


HammerGS (approved):
Tested and seems to work

PR #4498, opened 2024-07-27
Added AToW Mission Completion XP Awards, Added Global XP Cost Multiplier Option

### LINKED ISSUE #4484: RFE: Enable XP Multiplier for the XP table/SPAs
The addition of an XP multiplier for the XP table would allow players to use a preset to set their table, whilst using the multiplier to effect the speed of their progression.

EXAMPLE

If the cost to increase a personnel's piloting was 4, and the player set the multiplier to 5, then the final XP cost would be 20. 

Having this expanded upon to even include the SPAs would allow further simplicity in the ability for players to modify their XP systems.

IllianiCBT:
Implemented functionality to award XP based on mission outcomes, including failure (1 XP), success (3 XP), and outstanding success (5 XP). All values configurable.

XP is awarded to all active personnel.

Outstanding Success is limited to StratCon only and is awarded for completing a contract with 3+ VP. This works nicely with the bonus exchange program as now users have a reason to not exchange all but one of their VP into Bonus Parts.

Furthermore, I added an XP Cost Multiplier option to campaign options. This allows users to modify the global costs of Skills, SPAs, and Edge purchases without needing to edit individual costs. This is useful for things like our new player preset, where we can globally reduce the cost of improvements without needing to change the ratio of costs between skill levels (for example). Closes #4484.

This PR is required for the Preset Reset project.


HammerGS (approved):
Tested and seems to work

PR #4501, opened 2024-07-27
Added Hospital Beds to Facility Report, Added Maximum Patients Per Doctor Campaign Option

IllianiCBT:
This PR adds hospital beds to the unit information facility report. This information is not displayed if Advanced Medical is disabled, as currently only Advanced Medical uses this information.

image

Furthermore, a new campaign option was added allowing users to specify how many patients can be serviced per doctor (default 25 to match CamOps).


MekHQ/src/mekhq/campaign/CampaignSummary.java
```diff
@@ -332,11 +332,33 @@ public String getAdministrativeCapacityReport(Campaign campaign) {
...
+        }
+
+        if (campaign.getCampaignOptions().isUseAdvancedMedical()) {
+            int patients = (int) campaign.getPatients().stream()
+                    .filter(patient -> patient.getDoctorId() != null)
+                    .count();
+
+            int doctorCapacity = campaign.getActivePersonnel().stream()
+                    .filter(person -> (person.getPrimaryRole().isDoctor()) || (person.getSecondaryRole().isDoctor()))
+                    .mapToInt(person -> 35)
```
NickAragua, line 353: Not sure what this magic number is.

NickAragua (changes requested):
Looks good, just one magic number.

MekHQ/src/mekhq/campaign/CampaignSummary.java
IllianiCBT: Ah, I should have commented here. This is the number of personnel each doctor can service. It's defined by Advanced Medical, however it never occurred to me to check to see whether AM has that information available to a getter. I'll take a look shortly, thanks for pointing this out.

IllianiCBT: So, it turns out that it wasn't defined in advanced medical because (at least insofar as I could see) doesn't have its own class. All its components (barring the actual injuries) _appear_ to be nestled in `Campaign.java.

Instead, maximum number of patients was just a magic number defined in `InfirmaryTab.java` and used to check whether the doctor was over their limit before assigning a new patient. I opted to create a campaign option so users could define how many patients each doctor can have. I've set the default value to `25` as this matches CamOps.

Once we implement CamOps Medical we can re-use that campaign option and have it apply to both CamOps Medical and Advanced Medical.

NickAragua (approved):

PR #4504, opened 2024-07-27
Clear Assigned Forces when Removing StratCon Scenario

IllianiCBT:
Previously, when a scenario was removed, its assigned forces were not cleared, which could lead to inconsistencies. This commit ensures that the assigned forces of a scenario are properly cleared.

This fix is experimental and needs further testing.


IllianiCBT:
Developmental deadens

PR #4507, opened 2024-07-28
Removed Vestigial 'Uneventful' Graduation Event

IllianiCBT:
Removed the 'uneventful' graduation event, as the relevant resource bundle had been removed at some past date and the reference was causing missing resource errors.

I also added a shuffle operation to randomize the order of graduation events in the `EducationController` because I'm fancy like that.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
+++ b/MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
@@ -897,7 +897,7 @@ public class EducationController {
         }
 
         if (graduationRoll >= 99) {
-            if (Compute.d6(1) > 5) {
+            if (Compute.d6(1) >= 5) {
                 if (academy.isHomeSchool()) {
                     String reportMessage = "<span color='" + MekHQ.getMHQOptions().getFontColorPositiveHexColor() + "'>"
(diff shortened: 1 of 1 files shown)


NickAragua (approved):
Fine, although I'm curious why it was removed.

IllianiCBT:
The actual event 'uneventful' got removed a while ago. Basically, there is a chance of a random flavor event occurring when someone graduates. Originally if no event was rolled a message was printed to the daily report that said something along the lines of `Commander Barry has attended their graduation, but it was uneventful`.  The 'but it was uneventful' string was the way to tell that no event occurred. Otherwise that string would show the event text instead.

However that ended up with daily reports looking like this:

`Sergeant Smith attended their graduation ceremony, but it was uneventful
Corporal Bob attended their graduation ceremony, but it was uneventful
Private Porkins attended their graduation ceremony, but it was uneventful`

I realized at some point that I could just use an empty string and remove the repetition. However, apparently, I forgot to remove the call to the old event so very rarely (something like a 0.139% chance per graduating character) we'd try to call the (now missing) resource, resulting in a missing resource error.

If I had thought about it when I removed the old 'uneventful' event I would have just made it another event that could randomly occur, but that didn't occur to me back then and I chose not to restore the resource string when writing this PR because that also did not occur to me :D

PR #4523, opened 2024-07-31
Prevented Automated Removal of Genealogically Relevant Characters

IllianiCBT:
Added a check to ensure characters related to a genealogy with at least one active member are not removed from the campaign by automated personnel cleanup. This helps maintain family relationships, as removal severs existing genealogical relationships.


HammerGS (approved):
Tested and seems to work.

Issue #2925, opened 2021-10-27
[RFE] Order multiple from Repair bay> Parts Acquisition box


FelixGold:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
MekHQ 0.49.3
Windows
Java 11

### Description
I think it would be useful to add an "Order X" button when acquiring items you are missing equipment on the Parts Acquisition box, so you can order multiple tons of ammo, or backup weapons that are in demand rather than ordering a single part, then navigating to another part of the program to order more.
![](https://media.discordapp.net/attachments/458706545379639296/903032416019562526/unknown.png)

repligator:
Closing as completed.

PR #4553, opened 2024-08-05
Add "Buy in Bulk" Button to Parts Acquisition in Repair Bay

Algebro7:
This PR adds a "Buy in Bulk" button to the Parts Acquisition dialog in the Repair Bay, closing #2925.

I re-used the "Buy in Bulk" dialog from the CampaignGUI so very little new logic had to be implemented. The user clicks the "Buy in Bulk" button, a prompt opens up with a text box asking how many, and then the parts are added to the procurement list like normal:

buyinbulk1

View from the main Campaign GUI after ordering multiple parts:

buyinbulk2

Note that I also simplified/removed some redundant code when determining the text to display on the order button.

I did not touch the properties files in resources, although I think this is necessary for translations to work. Just let me know what I need to do there and I'll be happy to take care of it.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/AcquisitionsDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/AcquisitionsDialog.java
@@ -487,32 +487,42 @@ public class AcquisitionsDialog extends JDialog {
             btnUseBonus.addActionListener(ev -> useBonusPart());
             actionButtons.add(btnUseBonus, gbcActions);
             gbcActions.gridy++;
-
+            JButton btnOrderOne;
+            JButton btnOrderInBulk;
             if (partCountInfo.isCanBeAcquired()) {
-                JButton btnOrderOne = new JButton("Order One");
-                btnOrderOne.setToolTipText("Order one item");
-                btnOrderOne.setName("btnOrderOne");
-                btnOrderOne.addActionListener(ev -> {
-                    campaignGUI.getCampaign().getShoppingList().addShoppingItem(part.getAcquisitionWork(),
-                            1, campaignGUI.getCampaign());
-                    refresh();
-                });
-                actionButtons.add(btnOrderOne, gbcActions);
-                gbcActions.gridy++;
+                btnOrderOne = new JButton("Order One");
+                btnOrderInBulk = new JButton("Order in Bulk");
+            } else {
+                btnOrderOne = new JButton("Order One (TN: Impossible)");
+                btnOrderInBulk = new JButton("Order in Bulk (TN: Impossible)");
             }
+            btnOrderOne.setToolTipText("Order one item");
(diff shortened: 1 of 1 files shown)


IllianiCBT:
While the code looks fine, I'm curious about the use case compared to the user just clicking 'order' multiple times and whether that use case justifies the additional button.

Algebro7:
Sure, that works too, and is what I've done in the past. It does get cumbersome when I want to put in 20+ orders of armor on a regular basis though.

I assumed since the RFE was still open with multiple tags added by a maintainer that it was acknowledged as a valid enhancement.

Thom293:
I cant test right now, but I think pushing a button multiple times wont add multiple to the order list.  Or maybe its just for armor.  Dont remember. But I know I wished for such a button more than once...  Though I cannot explain why.

IllianiCBT:
If pressing a button multiple times isn't adding multiple items then that's a bug we need to investigate (although, I've not noticed it working incorrectly), adding a new button because the existing ones don't work properly isn't a good solution.

> I assumed since the RFE was still open with multiple tags added by a maintainer, including "Good First Issue", that it was acknowledged as a valid enhancement.
Me opening a discussion about whether this is the best solution doesn't mean it isn't a valid enhancement. It means I think there needs to be a discussion about whether this is the best solution.

HammerGS:
I don't see an issue with just entering a number myself. This has been a long requested RFE, but I seem to recall we have some other Bulk ordering ones that would link to and around this.

Algebro7:
> If pressing a button multiple times isn't adding multiple items then that's a bug we need to investigate (although, I've not noticed it working incorrectly), adding a new button because the existing ones don't work properly isn't a good solution

Repeatedly pressing "order one" is working fine for me both in 49.20 and current master, and with both normal components and armor.

> This has been a long requested RFE, but I seem to recall we have some other Bulk ordering ones that would link to and around this.

Thanks, just let me know.

IllianiCBT (approved):
Seeing that I'm the only one thinking this is unnecessary and the code is fine, I'm going to go ahead and mark it as approved. :)

Algebro7:
> Seeing that I'm the only one thinking this is unnecessary and the code is fine, I'm going to go ahead and mark it as approved. :)

Do I need to update the properties files in resources to make the localization work correctly?

IllianiCBT:
> > Seeing that I'm the only one thinking this is unnecessary and the code is fine, I'm going to go ahead and mark it as approved. :)
> 
> Do I need to update the properties files in resources to make the localization work correctly?

_To my knowledge_ it should default to English and only pull from the localized ResourceBundles if the client is set to a non-English language and an alternative language entry exists for that string. I'll be honest, our localization is spotty at best due to how few non-English speakers we have on the team.

PR #4558, opened 2024-08-06
Added DropShip Bounty Settings

### LINKED ISSUE #3577: RFE:  Campaign Option to convert salvaged Dropships to Bounty
This has been discussed a number of times on discord over the years, but I dont see an RFE for it. 

Dropship salvage breaks the economy in StratCon.  Getting one early quickly takes away all financial pressures.  And when the financial pressures are gone, and you can buy and hire what and who you want when you want, you essentially become invincible.

So RFE is this.  Have checkbox Campaign option to set salvaged dropships into a Cbill bounty on the salvage screen.  Basic example would be a % of value I think?  It could be configurable % or fixed.  

If fixed, Id suggest 5-10% of value.  Perhaps +2% for B rated units and +5% for A rating.  Something more elaborate could be done, but that is a basic way.

I think many people do this manually already.  But seems like something that could be automated and would keep us from being tempted. :)

EDIT:  As JohnyBurner pointed out below, even if you dont salvage them, they break the salvage ratio, so you still get extremely rich even if you dont salvage them.  I like his idea better.

Thank you.

IllianiCBT:
## Requires https://github.com/MegaMek/megamek/pull/5862

Introduced a new campaign option to award a percentage bonus for destroyed DropShips. This bonus can be configured in the campaign settings and (if enabled) will replace the salvaging of DropShips killed on ground maps with a set percentage of the unit's sale value.

In universe, this represents the difficulty of salvaging large vessels from the field, the salvage site instead being sold to an interested third party.

This was introduced to offer an alternative to the sudden finance spike experienced when killing DropShips. Or, if the employer is allowed to salvage the unit, the ability to suddenly be able to salvage everything until the end of the contract.

Personally, I find a setting of 5% works best, providing a nice boost to finances without derailing the economic/equipment growth of the campaign.

## Closes #3577


Thom293:
most excellent!

HammerGS (approved):
Tested and seems to work

PR #4561, opened 2024-08-07
Improved Handling of Child Dependents When Attending Academies

### LINKED ISSUE #4522: [RFE] Education: Children born to parents at school should stay with the parent until they return
### Description
Per request in Discord: Currently babies just magically teleport back to your company if their mother is at school when they give birth.   It would be nice if children stayed with their parent until they return from school/had the option of being sent straight to boarding school or something.

IllianiCBT:
Added tracking of child dependents when a character attends an out-of-unit education.

Now, when characters attend an outside-unit academy they will take any child dependents with them. This is true even if there is another parent remaining with the unit. The logic there is that academies are perceived as safer than an active military unit.

Any children tagging along in this way will have their status set to `On Leave` for the duration of the academic education, even if they cease being considered a child during that time. Babies born while the mother is at an academy are automatically factored into this system.

When the education concludes and the parent successfully returns to the unit all 'tag alongs' have their status set to `Active`.

### Closes #4522

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -212,6 +212,7 @@ public class Person {
     private int eduJourneyTime;
     private int eduEducationTime;
     private int eduDaysOfTravel;
+    private List<UUID> eduTagAlongs;
     //endregion Education
 
     //region Personality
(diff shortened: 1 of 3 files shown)


HammerGS (approved):
Tested and seems to work

PR #4567, opened 2024-08-07
Updated Text for Against the Bot Campaign Option

### LINKED ISSUE #4550: 50.00 "Use StratCon Rules" appears twice on that Campaign Options tab.
Also the top one seems confusingly named.  

image

IllianiCBT:
Clarified that StratCon requires Against the Bot rules in the checkbox text. This makes it explicit that users need both sets of rules for StratCon.

### Closes #4550


HammerGS (approved):
Not sure if we already do this, but could you make it that when you click Stratcon, it activates the AtB option.

IllianiCBT:
It's not possible to enable StratCon without enabling AtB first, as deselecting AtB disables the entire tab.

PR #4570, opened 2024-08-07
Separated Handling of Prisoners & Prisoner-Defectors in Resolve Mission Dialogs

### LINKED ISSUE #4505: [RFE] - End of Contract Prisoner Nag and Defectors
This very helpful End of Contract Nag reminds you to ransom or free them.  

But I believe it should exclude people willing to defect (the *Prisoners).  

Or perhaps some other handling?  A pop up before the Nag "the following people are willing to defect, would you like to hire them?"  Simiar to the Awards screen. 
 But that's gui work and might be a pain.  

But at the least excluding them from the ransom/free dialog would be helpful. Thank you.

IllianiBird:
Updated the BriefingTab to differentiate between regular prisoners and defectors during the ransom process. Implemented a new prompt to handle defectors, aligning the flow with existing prisoner ransom logic. Added a new localization string for the defector ransom prompt. Now, users will be prompted with two ransom dialogs: one for prisoners and one for prisoner-defectors.

I opted not to not include a dialog to recruit all prisoner-defectors, as recruiting all prisoner-defectors is not universally a good idea. Especially when using personalities, loyalty, and later _especially_ when using random events (once implemented).

I also added a space between the Prisoner/Defector title extension and a characters title (rank). So no more PrisonerSergeant Smiths. I could have sworn there was an RFE for that, but I couldn't find it.

### Closes #4505

### DIFF
--- a/MekHQ/resources/mekhq/resources/GUI.properties
+++ b/MekHQ/resources/mekhq/resources/GUI.properties
@@ -71,6 +71,7 @@ confirmExecute.format=Do you really want to execute %s?
 jettisonQ.text=Jettison?
 confirmJettison.format=Do you really want to push %s out an airlock?
 ransomQ.format=Ransom %d prisoners for %s?
+ransomDefectorsQ.format=You have %d prisoners willing to defect, ransom them for %s?
 ransomFriendlyQ.format=Ransom friendly %d prisoners for %s?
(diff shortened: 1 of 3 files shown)


HammerGS (approved):
Tested and seems to work

Issue #4473, opened 2024-07-23
[0.49.20] Bug: BA c-bill cost calculation errors/mismatches


Gribbly1:
Bundled Megamek with MekHQ 0.49.20
Windows 10, Java Version: 11.0.13

Tech Manual p281 BATTLE ARMOR STRUCTURAL COSTS AND AVAILABILITY, p285 for final unit cost.

BA financial values are inconsistent across the purchase unit menu, procurement info pane, Hangar and financials tab.

[BAtest.cpnx.gz](https://github.com/user-attachments/files/16348710/BAtest.cpnx.gz)

Note: parts value set to 1 for clarity.

![longinus](https://github.com/user-attachments/assets/d21720bd-046d-4830-8d25-24b965816d08)

![long2](https://github.com/user-attachments/assets/0bb7dd5d-7686-4452-99da-f61496f378c4)

![long3](https://github.com/user-attachments/assets/6c4807f1-e77d-4bc4-b2ef-254f9e87f965)

![long4](https://github.com/user-attachments/assets/b1372058-d784-4dd3-b00d-1466edf1d582)

**MML dry calc for comparison:**

![image](https://github.com/user-attachments/assets/5c7b6094-114f-4181-9aa2-dae68d650d0c)

repligator:
Math time:

Cost With Ammo=2548320
Dry Cost=2440320
Mystery Cost=1948320

2548320-1948320=600000

600000= the cost to train 1 Longinus Battle Armor Pilot multiplied by 4

MHQ is charging you 600,000 C-Bills less because you are only buying the suits, you have to supply your own pilots.

Gribbly1:
Math time:
1,948,320 does not equal 6,195,280

repligator:
> Math time: 1,948,320 does not equal 6,195,280

Apologies if I sounded glib earlier, that was not my intention.
After further investigation, it appears MHQ is using the `INF_JUMP` `priceMultiplier` (2.6) on these BA's. 
2382800(the base cost of the parts according to `partsValue`)x2.6=6195280

This can be confirmed by setting breakpoints on lines 963 and 964 of `megamek/common/Infantry.java`

Gribbly1:
No worries.

Is there a way to see exactly what partsValue is made up from? It doesn't seem to correspond to the dry cost or purchase cost.

Also the ammo cost should be negligible for one shot SRM2's, not the full 27k*4 for 45 shots per suit.

repligator:
[...]

Issue #2552, opened 2021-04-29
Infantry Salvage Values are absurdly high.


Dzlan:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
0.48.0
Windows 10
openjdk 11 2018-09-25
OpenJDK Runtime Environment 18.9 (build 11+28)
OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)

### Description
![Capture](https://user-images.githubusercontent.com/22495256/116495869-e6f2ea80-a8f7-11eb-9f47-59d0c2c79f6b.PNG)

I notice infantry have absurd values in salvage screen. You can get this full amount via ransom.

### Files
[megameklog.txt](https://github.com/MegaMek/mekhq/files/6395971/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/mekhq/files/6395978/mekhqlog.txt)

Thom293:
Came here to post same thing.  My screen is from Dev Snapshot.  Note field gunners and Mechanized bridgelayer.

![Screenshot 2021-04-28 221202](https://user-images.githubusercontent.com/23645569/116498389-3fca7080-a86f-11eb-831f-c1f4833a2b5c.png)

HammerGS:
These are RAW see Tech Manual pg 276

_Darrell’s conventional infantry platoon is a motorized formation of 28 troopers. These troops possess standard laser rifles as a Primary weapon, and carry no Secondary weapons. After consulting the Conventional Infantry Weapons costs table (pp. 298-301), Darrell looks up the cost of the laser rifle (1,250 C-bills), and finds that the base trooper cost for the platoon will be 70,710.678 C-bills (2,000
x √[1,250] = 70,710.678). As these troops are motorized, this base per-trooper cost becomes 113,137.085 C-bills (70,710.678 x 1.6 for motorized = 113,137.085). The entire 28-man platoon costs 3,167,838 C-bills (113,137.085 x 28 = 3,167,838.38, rounded down to 3,167,838). This is considered
the platoon’s final value (unless Darrell chooses to spring for anti-’Mech training)._

If made Anti-Mek then multiply that by 5.

Thom293:
Wow.  Well, not sure if its possible to just errata it for MHQ or give them a salvage value of zero or something.  

[...]

PR #4572, opened 2024-08-08
Adjusted Sale Value for Infantry and Battle Armor Units

### LINKED ISSUE #1795: Hanger Value of Battle Armor much greater than Purchase Price
Searched for issues with word "Battlearmor", did not find issue

Using MekHQ 0.47.6 dev with Windows 10
java (build 1.8.0_241-b07)

Battle armor in hanger is priced (and sold for) as much as 10 times it costs (and is paid for) in the Purchase Units screen.

Battle Armor Cost Bug
Battle Armor Cost Bug2

[Tercio Saltillo30930228.zip](https://github.com/MegaMek/mekhq/files/4730959/Tercio.Saltillo30930228.zip)

[factions.zip](https://github.com/MegaMek/mekhq/files/4730968/factions.zip)

IllianiCBT:
Adjusted the sale value calculation for infantry (and BA) units by using their alternate cost method. This ensures that the sale of infantry is computed correctly and appropriately matches their purchase value.

Whether this final value is entirely correct is a matter for bigger brains than mine to debate. However, this at least matches the equipment value without factoring in personnel training. As you're selling the equipment, not the personnel.

This change will affect any instance where we use the sale value of an infantry or battle armor unit. Salvage immediately comes to mind, but likewise so does the campaign option to use unit sale value when calculating contract pay.

### Closes #1795 & #2552 & #4473

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -71,6 +71,7 @@ import java.io.PrintWriter;
 import java.math.BigInteger;
 import java.util.List;
 import java.util.*;
+import java.util.Map.Entry;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


IllianiCBT:
While I was in `Unit.java` I did a little light refactoring, but I did spot a lot of unparameterized uses of `Mounted` alongside a number of unused methods that seem to have been supplanted by `Entity.java` at some point in the distant past. I opted to leave these alone as I didn't want to remove an unused method, incase this was a false-positive; and I left the `Mounted` instances alone as I don't know enough about that class to make a reasonable judgement for how it should be used.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -808,37 +803,25 @@ public String getDescHTML() {
...
-                return new TargetRoll(0, "transport bay");
-            case SITE_FACILITY:
-                return new TargetRoll(-2, "maintenance facility");
-            case SITE_FACTORY:
-                return new TargetRoll(-4, "factory");
-            default:
-                return new TargetRoll(0, "unknown location");
-        }
+        return switch (site) {
+            case SITE_FIELD -> new TargetRoll(2, "in the field");
```
HammerGS, line 807: PG 193 of CO has the new names for these. It would be nice to move to these with a seperate PR. 

image

HammerGS (approved):

MekHQ/src/mekhq/campaign/unit/Unit.java
IllianiCBT: I'll address that in a separate PR, but gimme a couple of minutes and I'll see it done

PR #4573, opened 2024-08-08
Fixed Formation Level Calculations

IllianiCBT:
Added a missing recursion element to ensure the TO&E is read correctly.


HammerGS (approved):
Tested and seems to work

PR #4591, opened 2024-08-09
Added CamOps Planetary Condition Modifiers to Maintenance Checks

### LINKED ISSUE #4575: RFE: Implement Planet Modifers to Repairs
Campaign Ops has rules for planet types impacting repairs. Pg 193 

image

With all the planets having details on this would be a fun option for hardcore players.

IllianiCBT:
Integrated various planetary conditions (such as gravity, atmosphere, pressure, and temperature) into the campaign's targeting calculations. Improved the Atmosphere class by adding methods to check specific atmospheres and refactoring the parseAtmosphere method using a switch expression.

### Closes #4575


HammerGS (approved):
Tested and seems to work. But should this be an option.

IllianiCBT:
> Tested and seems to work. But should this be an option.

I'm glad you mentioned this, because I know we spoke about that but apparently I forgot to implement it. Consider it done.

PR #4596, opened 2024-08-09
Added Personality Characteristics and Loyalty to Personnel Tab

### LINKED ISSUE #4543: RFE: Personnel tab display mode "Personality"
I would like to see an additional display mode in the personnel tab that shows characters´ personality traits, their intelligence and their loyalty; right now it is impossible to sort for any of these, or get an overview without looking at each person individually. With intelligence and loyalty in particular, there are times when it would be useful to sort for them, such as to find good education candidates.

IllianiCBT:
Enhanced the `PersonnelTabView` to include new columns for personality characteristics such as aggression, ambition, greed, social, and intelligence. Added getters for these traits and updated sorting and display logic to support the new attributes. Also included a test for personality traits in PersonnelTabView.

### Closes #4543


IllianiCBT:
I should mention that this PR removes 'portrait path' as something that can be viewed via the personnel table. This information is dubiously useful, and even if it was useful the string is too long to be visible on the table anyway. That we don't have any open issue reports for that tells me it's not information users take advantage of, so I replaced it with Loyalty.

HammerGS (approved):
Tested and seems to work.

Issue #3610, opened 2023-02-19
Vehicle crewmembers throwing off vehicle crew experience rating.


pheonixstorm:
Issue appears to be as old as the vehicle crewmember role

Support vehicles have three types of crew, driver, gunner, and crewmember. When calculating the experience rating for a vehicle crew it is not ignoring crewmembers who use the mechanic skill rather than the two vehicle skills. For a vehicle with a driver, no gunner, and several crewmembers this can result in an experience rating of green or worse for the gunnery rating of the unit. This in turn causes the overall unit rating to be worse than it should be, and that in turn affects the rating score for setting up contracts

IllianiCBT:
I have closed off the Reputation side of this bug in #4605, but the part where we transfer data to mm will need to be handled separately

IllianiCBT:
Investigating how we calculate composite Gunnery and Piloting skills shows we're only taking into account characters whose possess the relevant Gunnery or Piloting skills. At least insofar as I can see, it does appear to being calculated correctly on our side.

Would you be able to provide an example of where this is an issue? That will help me narrow it down.

pheonixstorm:
Looking at this again it seems the unit rating no longer is affected but the experience rating in the Unit/Graphic tab and right TOE pane are showing the incorrect experience rating. Left pane shows correctly. Before, all three locations were showing incorrect values

![TOE View](https://github.com/user-attachments/assets/a2f966a2-272f-49d6-b194-5b8a1d3fe1bd)

pheonixstorm:
Campaign file for testing
[My Campaign30670101.cpnx.gz](https://github.com/user-attachments/files/16577736/My.Campaign30670101.cpnx.gz)

IllianiCBT:
Can you chuck up a save demonstrating the issue, I can then take a look. I think it’ll be a lot easier for me to appreciate the issue if I can see it in action :)

IllianiCBT:
Nevermind, you’re one step ahead of me!

IllianiCBT:
So this is actually (kinda) working as intended. The reason why it's showing those crew members is because those are the unit commanders. If you give a rank to any other character in the unit, they will become the unit commander and that's who you'll see as the 'face' of the unit.

Which raises the question: why is it picking crew members instead of other, more experienced, personnel as the commander.

Closing this issue as 'working as intended' and I'll investigate how we pick unit commanders separately.

pheonixstorm:
[...]

PR #4605, opened 2024-08-11
Overhauled CamOps Unit Reputation Calculations and Report

### LINKED ISSUE #3475: 0.49.10 Unit rating off due to imaginary tech shortage caused by having large craft.
Another odd bug, this one affects the unit rating when dealing with large craft and normal unit techs. In the case of the image below, mech techs. This appears to affect both FMMr and Cam Ops ratings. The image shows only 3 techs listed in the unit rating panel, but four are listed under the personnel tab. A Leopard dropship is also part of the unit. If the Leopard is mothballed or removed, the unit rating tech count returns to normal. The missing tech count happens with all tech types, mech, aero, or mechanic.

No errors in the log.
tech_count

[_transportbug2.zip](https://github.com/MegaMek/mekhq/files/9714834/_transportbug2.zip)

IllianiCBT:
As requested by Hammer, this PR goes through and rewrites how we handle CamOps Reputation (previously known as Unit Rating). This incorporates current errata and streamlines a lot of the calculations.

Furthermore, CamOps reputation was detached from `IUnitRating`, as this gave me a lot more freedom in how I could implement the new system.

The second reason was that I wanted to isolate FM:Mr's Unit Rating calculations entirely. While FM:Mr is considered deprecated at this point, by isolating it it's my hope that we can reduce the likelihood of some future change breaking FM:Mr's Unit Rating, which would then potentially result in its' removal. While CamOps rules are current and what we will be supporting, moving forward, I recognize that some users prefer to use FM:Mr. I wanted to make one last gesture of support before 50.0 ships and FM:Mr becomes visible as 'deprecated' to the wider userbase.

I also added Crime Tracking, because it amused me to do so. Right now it only tracks whether the campaign is a Pirate faction, but I intend to have some fun with this at a later date.

### Closes #3475


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -6022,33 +6129,49 @@
...
+     * If the unit rating method is neither FMMR nor Campaign Operations, "N/A" is returned.
+     *
+     * @return The text representation of the unit rating
+     */
     public String getUnitRatingText() {
-        return getUnitRating().getUnitRating();
+        UnitRatingMethod unitRatingMethod = campaignOptions.getUnitRatingMethod();
+
+        if (unitRatingMethod.isFMMR()) {
+            return getUnitRating().getUnitRating();
```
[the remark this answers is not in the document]
IllianiCBT: I deprecated it, go away

MekHQ/src/mekhq/campaign/rating/CamOpsReputation/TransportationRating.java
```diff
@@ -0,0 +1,355 @@
...
+            // this is a binary bonus, so we only need to flip the value once
+            if ((hasJumpShipOrWarShip == 0) && (entity.isJumpShip() || entity.isWarShip())) {
+                hasJumpShipOrWarShip = 1;
+            }
+
+            dockingCollars += entity.getDockingCollars().size();
+
+            // Iterate through each bay in entity
+            for (Bay bay : entity.getTransportBays()) {
+                if (bay instanceof SmallCraftBay) {
```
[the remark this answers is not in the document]
IllianiCBT: While it's not the most elegant of executions, I opted to go with the chain of type tests as it made for easy to follow code.

pheonixstorm:
@IllianiCBT if you haven't already you may also want to check out #3610 as it also deals with the Cam Ops rating.

repligator:
My unit Rep Report shows (+-10) for 1 failed mission. Is that meant to be read as 'plus negative ten', and if so could it be re-written as '-10' the same way No Dropship is -5?

repligator:
NPE on first day of the month
```
11:10:55,807 ERROR [mekhq.MekHQ] {AWT-EventQueue-0}
mekhq.MekHQ.lambda$main$1(MekHQ.java:294) - Uncaught Exception Detected
java.lang.NullPointerException: temporal
	at java.base/java.util.Objects.requireNonNull(Objects.java:235)
	at java.base/java.time.LocalDate.from(LocalDate.java:394)
	at java.base/java.time.LocalDate.until(LocalDate.java:1641)
	at java.base/java.time.temporal.ChronoUnit.between(ChronoUnit.java:272)
	at mekhq.campaign.Campaign.processUnitReputation(Campaign.java:3805)
```
[mekhq.log](https://github.com/user-attachments/files/16588780/mekhq.log)
[Omega Galaxy31540201.cpnx.gz](https://github.com/user-attachments/files/16588783/Omega.Galaxy31540201.cpnx.gz)

HammerGS:
I think we go with the CodeQL has the sanity check and merge this. Then it's easier for QA to test in the nightlies up to release.

IllianiCBT:
Updated the mission string format to display the correct multiplier and added a null check for dateOfLastCrime in the `Campaign` class.

IllianiCBT:
This is the updated Action: https://github.com/MegaMek/mekhq/actions/runs/10357521893

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3770,6 +3794,34 @@ public boolean newDay() {
         return true;
     }
 
+    private void processUnitReputation() {
```
Sleet01, line 3797: This function seems to deal primarily with crime score and pirates; is this function name accurate?

IllianiCBT: When I wrote the function I anticipated it being needed for more than just crime. Good catch, I'll get it updated shortly :)

MekHQ/src/mekhq/campaign/rating/CamOpsReputation/ReputationController.java
```diff
@@ -0,0 +1,618 @@
...
+
+    // total
+    private int reputationRating = 0;
+
+    public ReputationController(Campaign campaign) {
+        initializeReputation(campaign);
+    }
+
+    //region Getters and Setters
+    @SuppressWarnings(value = "unused")
```
Sleet01, line 78: Is it truly necessary to have this decorator on every method?  That makes me dubious about the warnings themselves...

But if so, supposedly you can decorate the entire class to suppress the warnings for everything in it, which would certainly reduce code bloat:

    @SuppressWarnings(value = "unused")
    public class ReputationController {

(Note, however, that I have not tried this myself)

MekHQ/unittests/mekhq/campaign/rating/CampaignOpsReputationTest.java
Sleet01, line 1: I'm leery of deleting an entire unit test file without at least attempting to replicate its coverage.  Is there a plan to implement unit tests of the new code at some point in the near future?

Sleet01 (approved):
A couple questions, but overall looks good!

MekHQ/src/mekhq/campaign/rating/CamOpsReputation/ReputationController.java
rjhancock: I'm against suppressing the warnings. If the parameter isn't used in the code it should be removed. If a child class needs it, they can implement the method with the param.

IllianiCBT: So I was under the impression, from feedback on a very early PR, that I should be creating Getters and Setters for all variables that might need to be called remotely. Is that not the case? Because if not, well that will save me a lot of time and tedium.

rjhancock: Personally I wouldn't create any methods you don't need to. There is also the new Java Record type (I think) which automatically creates all that are needed. Might look into that.

MekHQ/src/mekhq/campaign/rating/CamOpsReputation/ReputationController.java
```diff
@@ -0,0 +1,602 @@
...
+        NodeList subNodeList = wn2.getChildNodes();
+
+        for (int i = 0; i < subNodeList.getLength(); i++) {
+            Node node = subNodeList.item(i);
+
+            if (node.getNodeType() == Node.ELEMENT_NODE) {
+                if (isTechnicianRequirements) {
+                    String[] numbers = node.getTextContent().substring(1, node.getTextContent().length() - 1).split(",\\s*");
+                    List<Integer> list = Arrays.stream(numbers)
+                            .map(Integer::parseInt)
```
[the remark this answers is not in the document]
IllianiCBT: Resolved: added try-catch

MekHQ/src/mekhq/campaign/rating/CamOpsReputation/ReputationController.java
```diff
@@ -0,0 +1,602 @@
...
+                            .collect(Collectors.toList());
+                    technicianRequirements.put(node.getNodeName(), list);
+                } else {
+                    map.put(node.getNodeName(), Integer.parseInt(node.getTextContent()));
```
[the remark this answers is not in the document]
IllianiCBT: Resolved: added try-catch

PR #4609, opened 2024-08-11
Integrated Academy Information into Interstellar Map Tab

IllianiCBT:
This PR is another Hammer request: the academies present in a system are now visible in the system information, including their description.

Furthermore, I went ahead and implemented an 'academies' filter, so users can visually spy which systems have non-local, non-homeschool academies.

These changes are also designed to handle custom academy sets, so if users have one or more custom academy sets, those academies will appear in the filter and system description.

As always, I did some light refactoring in the files I edited. Nothing too exciting, just clean up where appropriate.


HammerGS (approved):
tested and seems to work

PR #4618, opened 2024-08-13
WIP: sell parts from parts in use dialog

### LINKED ISSUE #1838: RFE: Selling inventory from Parts in Use report
Request a method for selling parts directly from the Parts in Use report.  The only way I've been able to use it is by sorting the "In Use" column and jumping back to the Warehouse tab to sell items.  This would simplify selling salvaged inventory after a contract.

Algebro7:
This PR adds "Sell" and "Sell in Bulk" options to the "Parts in Use" screen. I had to add the part ID to the `PartInUse` model to be able to locate the part in the warehouse, as before this was only set up for new part acquisitions. Note there is a "magic" value for the `PartInUse.id`, because there are two constructors that don't take a `Part` and thus don't have a defined `id`. Those constructors are never called so I'm not sure what the use-case is, but maybe they could be refactored out later. 

I'm opening this as a draft for now because there seems to be a bug where it will let you continue selling parts even when you don't have any more to sell. Basically, the following condition in `Quartermaster.java` is not being hit for some reason:

```
        quantity = Math.min(quantity, part.getQuantity());
        if (quantity <= 0) {
            return;
        }
```

I'm not sure if this is a bug in `Quartermaster` or something I'm doing incorrectly, so if anyone has any suggestions that'd be great. It seems like the quartermaster is _supposed_ to be correctly detecting whether you have spare parts to sell and isn't, but I didn't see any other bug reports related to that.

Closes #1838


Sleet01 (approved):
LGTM.

Sleet01:
> ```
>         // Do not sell more than we have
>         quantity = Math.min(quantity, part.getQuantity());
>         if (quantity <= 0) {
>             return;
>         }
> ```
> 
> I'm not sure if this is a bug in `Quartermaster` or something I'm doing incorrectly, so if anyone has any suggestions that'd be great. It seems like the quartermaster is _supposed_ to be correctly detecting whether you have spare parts to sell and isn't, but I didn't see any other bug reports related to that.

I'd suggest creating a new unit test (copied from one of the existing tests) and using debug mode with breakpoints to trace the execution.  I'd guess that either the code execution is taking a different path, or that we are never getting a 0-or-less value.

Algebro7:
Marking this ready for review. It was a little more involved than I thought at first due to the fact that the Parts In Use Report dialog was "lossy" and didn't have a good way to map back to spare parts in the warehouse, but I figured out how to do it relatively cleanly IMO.

I did need to add 1 field to the `PartInUse` class to make this possible; a name representing the part name so that I could search the available parts for it.

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
```diff
@@ -127,6 +130,43 @@ public void actionPerformed(ActionEvent e) {
                 refreshOverviewSpecificPart(row, piu, partToBuy);
             }
         };
+
+        Action sell = new AbstractAction() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                int row = Integer.parseInt(e.getActionCommand());
+                PartInUse piu = overviewPartsModel.getPartInUse(row);
```
IllianiCBT, line 138: I'd prefer if this were `partsInUse`, this is just personal preference, but I've had to dig through enough old code to know that no matter how obvious an abbreviated variable name is when you write in; in ten years time nobody is going to know at a glance what it is.

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
```diff
@@ -127,6 +130,43 @@ public void actionPerformed(ActionEvent e) {
...
+            }
+        };
+
+        Action sellInBulk = new AbstractAction() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                int row = Integer.parseInt(e.getActionCommand());
+                PartInUse piu = overviewPartsModel.getPartInUse(row);
+                campaign.getWarehouse().getSpareParts().stream().filter(p ->
+                                Objects.equals(p.getName(), piu.getName()))
```
IllianiCBT, line 153: I see you're comparing Strings here, is it not possible to compare `part` instead? That way we wouldn't need to pass it into a String.

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
```diff
@@ -127,6 +130,43 @@ public void actionPerformed(ActionEvent e) {
...
+                        .findFirst()
+                        .ifPresent(p -> {
+                            int quantity = 1;
+                            PopupValueChoiceDialog pcd = new PopupValueChoiceDialog(gui.getFrame(), true,
+                                    "Sell how many " + p.getName(), quantity, 1, CampaignGUI.MAX_QUANTITY_SPINNER);
+                            pcd.setVisible(true);
```
IllianiCBT, line 159: As an illustration of my earlier point, I've reached this and I've no idea what `pcd` stands for. Eventually I found it on line 157, but if it were called `popupChoiceDialog` I would have immediately known what the variable was for.

IllianiCBT (changes requested):
Our players are going to _love_ this change. Warehouse management has long been a major thorn in everyone's side. I do have a couple of minor points of feedback, these predominately come down to points of preference.

I've lost hours of my life trying to hunt down the meaning of one or two letter variables, so you'll get used to me having a personal vendetta against them :D

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
```diff
@@ -127,6 +129,43 @@
                 refreshOverviewSpecificPart(row, piu, partToBuy);
             }
         };
+
+        Action sell = new AbstractAction() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                int row = Integer.parseInt(e.getActionCommand());
```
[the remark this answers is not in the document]
IllianiCBT: Can I verify what we're 100% certain the value of `e.getActionCommand()` is going to be an Integer; or does this need to be placed in an if-catch?

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
```diff
@@ -127,6 +130,43 @@
...
```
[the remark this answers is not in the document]
IllianiCBT: I'm assuming from the code snippet this is a duplicate of the earlier warning?

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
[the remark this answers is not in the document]
Algebro7: Yeah, [PartsInUseTableModel.java line 323](https://github.com/MegaMek/mekhq/blob/master/MekHQ/src/mekhq/gui/model/PartsInUseTableModel.java#L308) constructs the event containing the row number when a button is pressed. It's not possible for it to be something else

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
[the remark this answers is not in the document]
Algebro7: Yep, same thing, it can be ignored as it's impossible for the value not to be an integer (inside of a string)

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
Algebro7: No worries, done in cdc6ddb :)

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
Algebro7: Done in cdc6ddb

MekHQ/src/mekhq/gui/dialog/PartsReportDialog.java
Algebro7: So I thought I tried using IDs earlier and it didn't work, but it turns out I was wrong (or I understand the parts API much better than I did this afternoon). I've refactored this to use the existing `checkForExistingSparePart()` method and IDs instead of string names and it's much cleaner. Thanks for the suggestion!

IllianiCBT (approved):
LGTM

PR #4633, opened 2024-08-14
Refactored Force Commander Logic & TOE Information Display

### LINKED ISSUE #4607: RFE: Display Commanding Officer of Formation when Formation is selected for Formation Larger than Lance
### Environment
Java 17
Linux
Master (latest PR https://github.com/MegaMek/mekhq/commit/a16a4fbdef2ed9c113f8eb1f3c8955af8a4bdb96)

### Description
Currently, when a Lance (or equivalent) is selected on the TOE, the commanding officer is displayed in the upper right. In this case, Major General Brita Mellem.
Screenshot_20240811_101836

However, when I travel up a level on the TOE, the Company commander is not shown. The Lance Commanders are show below, but not the Company Commander.
Screenshot_20240811_101910

When I travel up another level in the TOE, the Battalion Commander is not shown, _but_ the Company Commanders are shown below, with Major General Brita Mellem being accurately depicted as the Company Commander.
Screenshot_20240811_101942

When I travel up another level in the TOE, the Regimental Commander is shown (because they use the commander flag), and the Battalion Commanders are shown below, with Major General Brita Mellem being accurately depicted as the Battalion Commander.
Screenshot_20240811_102020

It would be nice if the formation commander would be displayed when the formation they are commanding was selected, instead of needing to travel up a level. In this case, when selecting the Company or Battalion, it would display Commander: Major General Brita Mellem, just like the Lance does. I believe this would also finish the already partially completed RFE #3873

### Files
[My HUGE Campaign31510104.cpnx.gz](https://github.com/user-attachments/files/16574017/My.HUGE.Campaign31510104.cpnx.gz)

IllianiCBT:
Improved the logic for updating force commanders by refactoring the method to streamline the selection process. Used a list of eligible commanders based on units and sub-forces, and enhanced the selection by incorporating a tiebreaker mechanism.

Further, we now track force commander beyond the lance level, ensuring a proper chain of command is displayed throughout the TO&E. To that end force commander is now visible for nodes beyond lance and origin.

I also discovered we were re-calculating force commanders when viewing sub-forces in the side-bar. This meant that changes to force commander were not reflected when viewing sub-forces. Now sub-forces correctly display the current force commander.

Finally, I adjusted font sizes in a couple of places to improve readability.

### Closes #4607

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -860,6 +860,8 @@ public class Campaign implements ITechManager {
                 lances.put(id, new Lance(force.getId(), this));
             }
         }
+
+        force.updateCommander(this);
     }
 
     public void moveForce(Force force, Force superForce) {
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -75,6 +75,7 @@ public class Force {
     private final Vector<UUID> units;
     private int scenarioId;
     private UUID forceCommanderID;
+    private UUID overrideForceCommanderID;
     protected UUID techId;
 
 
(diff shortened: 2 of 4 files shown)


IllianiCBT:
That is a good shout, so this change _does_ respect manually selected Commanders, however the system we use isn't great:
- it doesn't stick if you move new units into the force
- it isn't respected when viewing forces from one step up

Both of these are bugs that predate this PR

Thom293:
It was too sticky before and would not clear so this is a compromise I believe.  Just didn't know if you knew about it and if it would mess up awards or something else so I mentioned it.

IllianiCBT:
I've solved the first bullet point, but the second is proving difficult to track down

HammerGS (approved):
Tested and seems to work with test campaign.

PR #4636, opened 2024-08-14
Rename strat ops personnel market to campaign ops to reflect rulebook reorganization

Algebro7:
This PR simply renames the Strat Ops personnel market method to "Campaign Ops" which reflects the fact that these rules moved to a different book. I added a check to the XML loader to automatically use Campaign Ops when saves contained the old Strat Ops string so backward compatibility works. 

As a sidenote, I doublechecked the personnel generation code and it matches what is in camops, although it doesn't seem to use the same skill generation method...or at least I couldn't find it. I will keep looking for that and push a separate PR once that's done.

### DIFF
--- a/MekHQ/mmconf/campaignPresets/Campaign Operations (StratCon).xml
+++ b/MekHQ/mmconf/campaignPresets/Campaign Operations (StratCon).xml
@@ -461,7 +461,7 @@
 		<sharesForAll>false</sharesForAll>
 		<useTaxes>false</useTaxes>
 		<taxesPercentage>30</taxesPercentage>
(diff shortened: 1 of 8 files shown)


IllianiCBT (changes requested):
Can you also take a look at the campaign presets stored in `mmconf/campaignPresets` to ensure they are correctly pointing towards CamOps Recruitment, as they were previously set to point towards StratOps.

The line you're looking to change is `<personnelMarketName>Strat Ops</personnelMarketName>`

Algebro7:
Thanks! I think this is resolved by cef4c6c but let me know if I missed anything.

IllianiCBT (approved):

PR #4651, opened 2024-08-16
Fixed Support Rating Administration Requirements Calculations

IllianiCBT:
Refactored the calculation of personnel count by merging the division operation into a single step. Corrected the divisor values to match CamOps.


NickAragua (approved):
Pretty sure CamOps doesn't require administrators to be administrated (page 21), so why aren't we subtracting the # of admins from the total personnel count?

NickAragua (changes requested):
Meant to "Comment" or "Request Changes" rather than "Approve"

IllianiCBT:
> Pretty sure CamOps doesn't require administrators to be administrated (page 21), so why aren't we subtracting the # of admins from the total personnel count?

`getTotalPersonnelCount()` fetches personnel count based on the requirements listed in CamOps, which doesn't include Administrators. So removing Admin personnel here was unnecessary.

IllianiCBT:
For ease of reading, here is `getTotalPersonnelCount()`

```
/**
     * Calculates the total personnel count required for the campaign, taking into account technician requirements.
     *
     * @param campaign The campaign for which to calculate the total personnel count.
     * @param technicianRequirements The number of technicians required for the campaign.
     * @return The total personnel count required for the campaign.
     */
    private static int getTotalPersonnelCount(Campaign campaign, int technicianRequirements) {
        int totalPersonnelCount = technicianRequirements;

        // Count personnel
        for (Unit unit : campaign.getActiveUnits()) {
            if (unit.isMothballed()) {
                continue;
            }

            Entity entity = unit.getEntity();

            if (entity.isMek() || entity.isAerospaceFighter() || entity.isConventionalFighter()) {
                totalPersonnelCount++;
            } else if (entity.isVehicle()) {
                totalPersonnelCount += (int) Math.ceil(entity.getWeight() / VEHICLE_WEIGHT_DIVIDER);
            } else {
                totalPersonnelCount += unit.getFullCrewSize();
            }
        }

        return totalPersonnelCount;
    }
```

NickAragua (approved):
Ok, good to go then.

PR #4655, opened 2024-08-16
Refactored Random Company Name Generator to Use CSV

IllianiBird:
This PR switches the random company name generator from using enum-based values to using a handful of csv files, similar to how we handle Bloodnames and Callsigns. This means users can edit and customize the available pool of name segments, with full support for the user folder.


NickAragua (approved):

IllianiCBT:
Flipped this to draft, so it can't be accidentally merged into 50.0

IllianiCBT:

PR #4659, opened 2024-08-16
Add initial support for Hiring Hall levels and add new hiring halls

Algebro7:
This refactors some internals in `AtBConfiguration` to provide support for different levels of hiring halls, which is necessary to implement support for CamOps. The "public" API hasn't changed except for a new `getHiringHallLevel()` method, and old save files are of course supported. This information will also be useful for displaying the hiring hall level in the UI, such as on the planetary map.

This also adds a handful of hiring halls that previously weren't in the game based on CamOps and Sarna. I added an approximation of the level based on Sarna in cases where an explicit official hiring hall level wasn't given.


Algebro7:
I've cherry picked these commits into my larger camops contract market branch, and given that they aren't needed for other features ATM I'm thinking it makes sense to close this and just merge these changes with PR #4644 . Closing for now but let me know if we want to re-open later for some reason.

PR #4665, opened 2024-08-17
Revert change to find parts by name when selling from parts in use

### LINKED ISSUE #4663: [v0.50.0-SNAPSHOT] Unable to sell most parts through the Parts In Use Report.
### Environment
04:50:26,741 INFO  [mekhq.MekHQ] {main}
mekhq.MekHQ.initializeLogging(MekHQ.java:315) - Starting MekHQ v0.50.0-SNAPSHOT
    Build Date: 2024-08-17T01:16:16.522116870
    Today: 2024-08-17
    Origin Project: MekHQ
    Java Vendor: Eclipse Adoptium
    Java Version: 21.0.3
    System Locale: en_GB
    MM Code Revision: 5d74b455d1bfe3c55797e4e3daac3bd0168ca828
    MML Code Revision: fc321e451b068ba6292e43ecc1f6ce80345d79cb
    MHQ Code Revision: b3e0e4457a766489e5545398116ef0d341862003

### Description
The Parts in Use Report (found in the Command Center tab) now has buttons that are meant to allow selling parts directly from this report. This only seems to function sporadically, and will often fail to sell any parts. 

### Files
[customs.zip](https://github.com/user-attachments/files/16643208/customs.zip)
[1st Rimward Hussars.zip](https://github.com/user-attachments/files/16643209/1st.Rimward.Hussars.zip)
[mekhq.log](https://github.com/user-attachments/files/16643210/mekhq.log)

Algebro7:
Commit a4cc0f9 introduced a bug where several types of parts could no longer be sold. This PR simply reverts it and goes back to looking up the target part by name rather than ID.

I agree with the original comment that it's not ideal to search by strings, but it appears there isn't really a better way to do it with how this dialog is constructed. I think there are some opportunities for refactoring it but I think those should be done in the future rather than now. Let me know what you think.

Closes #4663

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/PartInUse.java
+++ b/MekHQ/src/mekhq/campaign/parts/PartInUse.java
@@ -9,7 +9,7 @@ import mekhq.campaign.work.IAcquisitionWork;
 
 public class PartInUse {
     private String description;
-    private int id;
+    private String name;
     private IAcquisitionWork partToBuy;
     private int useCount;
     private int storeCount;
@@ -35,13 +35,14 @@ public class PartInUse {
             appendDetails(sb, part);
         }
         part.setUnit(u);
-        this.id = part.getId();
+        this.name = part.getName();
         this.description = sb.toString();
         this.partToBuy = part.getAcquisitionWork();
         this.tonnagePerItem = part.getTonnage();
         // AmmoBin are special: They aren't buyable (yet?), but instead buy you the ammo inside
         // We redo the description based on that
         if (partToBuy instanceof AmmoStorage) {
+            this.name = ((AmmoStorage) partToBuy).getName();
             sb.setLength(0);
             sb.append(((AmmoStorage) partToBuy).getName());
             appendDetails(sb, (Part) ((AmmoStorage) partToBuy).getAcquisitionWork());
@@ -76,8 +77,8 @@ public class PartInUse {
         return description;
     }
 
-    public int getId() {
-        return id;
+    public String getName() {
+        return name;
     }
 
     public IAcquisitionWork getPartToBuy() {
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/parts/PartInUse.java
```diff
@@ -35,13 +35,14 @@ public PartInUse(Part part) {
             appendDetails(sb, part);
         }
         part.setUnit(u);
-        this.id = part.getId();
+        this.name = part.getName();
```
NickAragua, line 38: How does this play with loading files from the previous version?

NickAragua (commented):
Flagging for some extra testing due to backwards compatibility requirements.

MekHQ/src/mekhq/campaign/parts/PartInUse.java
Algebro7: So the PartInUse is not actually stored in save files, it's just a (lossy) model constructed for this dialog that is deleted when the dialog is closed. There should be no impact on save files/backwards compatibility.

NickAragua: Ok. If PartInUse isn't being saved, then we can consider this good to go.

Algebro7: Alright yeah, I just double checked and it's not saved anywhere. The only part that is persistent is the `IAcquisitionWork partToBuy` field, which is pulled from the `Part` this is being constructed from and saved to the shopping list when buying items, but neither PR touched that object. So we should be good.

IllianiCBT (approved):

PR #4673, opened 2024-08-19
Added Conditional Check Before Showing 0 Personnel Eligible autoAwards Dialog

IllianiBird:
Previously, the dialog showing no award eligibility appeared unconditionally. Now, it only shows if `isManualPrompt` is true, preventing unnecessary dialogs when not prompted manually.


Sleet01 (approved):
LGTM

PR #4674, opened 2024-08-19
Fixed NPE in CamOps Mission Date Calculation

IllianiCBT:
Previously, mission date calculation could throw a NullPointerException if there were no missions. Added a null check to ensure `oldestMissionDate` defaults to the current date if no missions exist. This prevents the exception.


Sleet01 (approved):
LGTM

PR #4679, opened 2024-08-19
Refactored DropShip Bonus Calculation Logic

IllianiCBT:
Simplified the conditionals for DropShip bonus calculation by merging nested checks. Updated tooltip text to clarify DropShip salvage rules.


Sleet01 (approved):
LGTM

PR #4686, opened 2024-08-21
Refresh camops personnel market daily

Algebro7:
Per CamOps rules, forces can attempt to recruit new personnel every maintenance/repair cycle, which is defined as 1 day. This PR makes the CamOps personnel market refresh daily. The serialization methods have been removed as it's no longer necessary to store the `daysSinceRolled` value in the XML, and previous and current saves load just fine.

Future refactoring could remove the `removePersonnelForDay()` override since it doesn't actually do anything except return the list that was passed in, but I don't think it's a good idea to mess with the interface at the moment.

@IllianiCBT  Feel free to hold off until 0.50 if you'd rather not introduce this now. Thanks!

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/PersonnelMarketCampaignOps.java
+++ b/MekHQ/src/mekhq/campaign/market/PersonnelMarketCampaignOps.java
@@ -18,16 +18,13 @@
  */
 package mekhq.campaign.market;
 
-import java.io.PrintWriter;
 import java.time.LocalDate;
 import java.util.Collections;
 import java.util.List;
 
 import mekhq.campaign.personnel.enums.PersonnelRole;
-import org.w3c.dom.Node;
 
 import megamek.common.Compute;
-import mekhq.utilities.MHQXMLUtility;
 import mekhq.campaign.Campaign;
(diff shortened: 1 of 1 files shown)


IllianiCBT:
How do these changes interact with the other market methods?

Algebro7:
I don't think they should because there can only be one active method at a time right?

IllianiCBT:
I think the area of concern is removing the tracking of days since last roll. Do the other methods not use that variable?

I have guests all week, so having to read your code on mobile means I can't see the bigger picture. :)

Algebro7:
> I have guests all week, so having to read your code on mobile means I can't see the bigger picture. :)

All good :)
 
> Do the other methods not use that variable?

They do not--it's private to this particular personnel market implementation, so the callers don't know about it/don't care. It was only being stored for PersonnelMarketCampaignOps to know when to actually generate new personnel, because the caller just calls it each day without knowing whether it actually generated anything, if that makes sense.

IllianiCBT (approved):
We should merge this before 50.0, as otherwise the CamOps personnel market (the new default) will feel like a massive nerf.

PR #4690, opened 2024-08-22
Corrected Experience Generation Logic and Refined Campaign Options GUI

IllianiCBT:
Fixed bug in experience level generation that caused rolls of 2 to result in Ultra-Green personnel.

Replaced direct value checks with a clamped switch case for experience level generation.

Streamlined the campaign options pane by removing unused skill levels and updating labels for clarity.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -19,30 +19,10 @@
  */
 package mekhq;
 
-import java.awt.Graphics2D;
-import java.awt.Image;
-import java.awt.image.BufferedImage;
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM

Thom293:
Sorry - just downloaded the nightly to test.  Does this make substantive changes or are these cosmetic?  Looks like lot of campaign options stuff.  If it needs testing I can wait for the artifact build and redownload...

IllianiCBT:
The only non-cosmetic change is that a roll of a 2 on the table to determine starting Experience Level will generate Green personnel, not Ultra-Green to match TW.

PR #4691, opened 2024-08-22
Revised CamOps Reputation Derived AtB Modifier, Renamed `getUnitRatingMod` Method

IllianiCBT:
Consolidated ELITE, HEROIC, and LEGENDARY levels into a single rating value. This more closely matches the values expected by AtB and will prevent unexpected behavior.

This PR also renames the `getUnitRatingMod` to `getAtBUnitRatingMod` to more closely match its' use. This touches a lot of Classes, making this PR look a lot scarier than it actually is.


IllianiCBT:
This PR has been updated to include name changes.

Issue #4695, opened 2024-08-23
[rfe] Add current edge total to the tech selection screen in the repair bay/warehouse


azhais:
While doing a high cost repair task, I wanted to use a tech with edge to add some cover for a "destroyed part" failure. It would be helpful if you could see how much edge a tech has from the repair bay rather than having to scroll over the in the personnel tab, find someone with edge, then go back to the bay and find the tech by name.

(I will be implementing this myself, I just don't have permissions to set the assignee)

IllianiCBT:
Closing as resolved

PR #4696, opened 2024-08-23
Add current edge total to the tech information

azhais:
Resolves RFE #4695

Show current edge as long as edge for support personnel is enabled in the options and the specific tech has the repair failure edge trigger enabled.


IllianiCBT:
While I was amused to see you make that RFE, I want to be clear that I only asked you to make an RFE because I didn’t realize you were offering to implement this yourself xD

azhais:
Oh, I'd have made it either way.   Long habit from work that all commits need a ticket.

IllianiCBT (approved):

PR #4699, opened 2024-08-25
Refactor Initial Education Level Assignment

IllianiCBT:
Implemented a centralized method to determine initial education levels based on age and experience. Replaced multiple instances of manual education level setting with the new method to ensure consistency and reduce code redundancy.


IllianiCBT:

PR #4714, opened 2024-08-26
Fixed Maintenance Interval Calculations

### LINKED ISSUE #4709: Maintenance checks should occur a quarter as often in garrison conditions
Page 191 of the fourth printing of _Campaign Operations_ has the following:

> Maintenance checks are run every week in the field, or once every four weeks when in garrison conditions (not in the field or in combat). Garrison conditions includes when being transported between systems for ground units and small aerospace units.

MekHQ currently only has a single option for controlling how often maintenance checks are performed, and it simply starts each day by checking if a unit hasn't had a maintenance check recently enough. That behavior means that changing it back and forth can lead to every single unit needing checks at the same time, which is especially problematic as this tends to occur the instant you actually need them in the field.

I propose the following change: each unit should have a floating-point value tracking the number of days elapsed since the last maintenance check. This value should increase by 0.25 under garrison conditions and 1.00 in the field. If that value exceeds the period specified by the relevant campaign option (or double/triple that depending on the Rugged quirk), a maintenance check is performed. That value should be made visible to the player.

When a maintenance check is performed, the timer resets to zero.

IllianiCBT:
Updated the `incrementDaysSinceMaintenance` method to account for time scaling based on active missions, and changed maintenance time tracking from `integer` to `double`. This ensures maintenance frequency matches CamOps.

### Closes #4709

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -47,6 +47,8 @@ import mekhq.campaign.event.UnitArrivedEvent;
 import mekhq.campaign.finances.Money;
 import mekhq.campaign.force.Force;
 import mekhq.campaign.log.ServiceLogger;
+import mekhq.campaign.mission.AtBContract;
+import mekhq.campaign.mission.Mission;
 import mekhq.campaign.mission.Scenario;
 import mekhq.campaign.parts.*;
 import mekhq.campaign.parts.equipment.*;
@@ -135,9 +137,9 @@ public class Unit implements ITechnology {
     private int mothballTime;
     private boolean mothballed;
 
-    private int daysSinceMaintenance;
-    private int daysActivelyMaintained;
-    private int astechDaysMaintained;
+    private double daysSinceMaintenance;
+    private double daysActivelyMaintained;
+    private double astechDaysMaintained;
     private int maintenanceMultiplier;
 
     private Campaign campaign;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4989,11 +4991,24 @@ public int getMaintenanceTime() {
...
+        double timeIncrease = 0.25;
+
+        for (Mission mission: activeMissions) {
+            if (mission instanceof AtBContract) {
+                if (((AtBContract) mission).getContractType().isGarrisonDuty()) {
+                    continue;
+                }
+            }
+
+            timeIncrease = 1;
```
NickAragua, line 5005: If we're setting this to a fixed value, we can break out of the loop after the first time we set it.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -5013,15 +5028,15 @@ public int getDaysSinceMaintenance() {
...
+        return (daysActivelyMaintained / daysSinceMaintenance);
     }
 
     public boolean isFullyMaintained() {
         return daysActivelyMaintained == daysSinceMaintenance;
     }
 
     public int getAstechsMaintained() {
-        return (int) Math.floor((1.0 * astechDaysMaintained) / daysSinceMaintenance);
+        return (int) Math.floor((astechDaysMaintained) / daysSinceMaintenance);
```
NickAragua, line 5039: I don't think we need parens around a single variable

NickAragua (changes requested):

PR #4718, opened 2024-08-26
Added StratCon Introduction Promo on Campaign Start

IllianiBird:
Implemented a feature to display a promotional introduction for StratCon when the Campaign Options dialog is closed, and it is the first day of the campaign, _and_ StratCon is enabled.

This has been implemented to try and reduce confusion among users transferring from vanilla AtB to StratCon now that StratCon is our dominate single player mode.

image


HammerGS (approved):
Tested and seems to work

PR #4719, opened 2024-08-26
Adjusted Target BV Percentage Logic for BVScaled OpFor Generation.

IllianiBird:
## NOT FOR INCLUSION IN 50.0

Replaced static minimum BV percentage with dynamic target percentage calculation during force generation. Ensured scenarios more accurately adhere to BV budget constraints and improved general generation behavior.

OpFor generation, insofar as BV is concerned now undergoes the following steps:
- mhq rolls a 8-sided die, minus 4 multiplied by 5. This gives us a number in the range of +/- 5-20. This added to 100 gives the target percentage of the OpFor's BV Allowance. Essentially this means the OpFor BV allowance will be equal to the player force, modified by difficulty, and then further modified by the die roll. This creates a variance of OpFor sizes without being under or overwhelming. Via this system users should see a greater variety of battle difficulties, some easy, some hard, but without battles being impossible - outside of modifiers, which players have a lot of control over.
- mhq generates lances (or regional variants) until it exceeds the percentage of the BV allowance determined above. At that point it will randomly remove units until it falls back within allowance. This was already done for Unit Count OpFor generation, but not for BV based generation.

What this does is it creates a more predictable OpFor size, but without being static. Players can still be faced with difficult encounters, especially if they're also getting modifiers, but it should no longer feel unfair or oppressive.


IllianiCBT:
Here is an example of this system in play:

Commander Bill fields a 5k BV lance, with difficulty set to Veteran (120%). This gives a base OpFor BV allowance of 6k BV.

**Old System**
Commander Bill's campaign has a A Unit Rating, that means at minimum the OpFor BV allowance will be 95% of the base BV. So, at minimum 5.7k BV. However there are a lot of dice rolls happening behind the scene, so it could be anywhere from 95-195% of the base BV allowance. mhq will keep adding lances (or regional variants) until this bar is passed, at which point generation stops.

No excess BV is trimmed (unlike balancing by Unit Count), so in a _very extreme_ example mhq could target 150% of the base BV allowance, roll a handful of lances, be 1 BV below the target, at which point she will then roll an _entire_ extra lance. This is why we sometimes see players encountering extraordinarily unbalanced OpFors. Some people like this, but really it's bad design. In this example that could see the OpFor fielding 12.5k BV against Commander Bill's 5k (150% of the allowance, plus an extra 5k lance). If not more, if mhq rolls a particularly strong lance.

**New System**
mhq rolls a d8-4, rolling a 6. Some math later, this corresponds to +10%. That means the base OpFor BV allowance is increased by 10%, 6.6k BV. Potentially it could be as little as -20% (4k) or as much as +20% (7k). Neither extreme is effortlessly easy, or oppressively difficult.

mhq generates lances (or regional variant) until it exceeds the adjusted BV allowance, as with the old system.

Finally, mhq will remove random units from the OpFor until it falls within the adjusted BV allowance. This means, if the OpFor is 1 BV below the target BV it won't suddenly add an entire extra lance. This might mean players occasionally fight under-strength lances, but that can be attributed to battlefield attrition and is entirely within the realms of canon.

IllianiCBT:
Flipping this to live

Thom293:
Not directly related to this but maybe worth mentioning.  Do asf generate a "Base 2" unit (flight - the traditional starter BT ASF unit) or "base 4" unit (traditional ground lance size). 

It seems like it's 4 instead of 2, and causes ASF generations to be even more lopsided than normal if the player stick to a base 2 unit size (and is the increment of how most DS carry ASF).  

Again may not belong here. But it's a related subject that was brought up recently.

IllianiCBT:
So this would actually render that pretty irrelevant, as even if it generates base-4, it will just trim down if it’s over BV’d. And if it’s base-2 and under BV’d it will generate another base 2 force (trimming down, if necessary). Effectively reaching the same result.

IllianiCBT:

PR #4720, opened 2024-08-27
Replaced TaharqaSkillGenerator with StratConSkillGenerator in Force Generation Calculations

IllianiBird:
### Requires https://github.com/MegaMek/megamek/pull/5948

Switched the skill generator implementation from `TaharqaSkillGenerator` to the new `StratConSkillGenerator`.

Previously, when OpFors generated there was a lot of variance in the Skill Levels across the force. This led to weird scenarios where you would find yourself against 1/2 Pilots among a supposedly Green OpFor force. While not outside the realms of possibility, this is something that I've seen players request we change off and on for months.

Now, when OpFors generate they will generate at a Skill Level equal to, or one step above/below the rating of the OpFor force. So a Regular force will consist of Green, Regular, and Veteran crews. While an Elite force will consist of Veteran, Elite, and Heroic crews, and so on.


NickAragua (approved):

IllianiCBT:
Flipping this to live

PR #4722, opened 2024-08-27
Refactored Difficulty Multiplier Calculation

IllianiBird:
### This should not be merged into 50.0

Revised the `getDifficultyMultiplier` method to use a switch expression based on skill levels. I also went ahead and adjusted the multipliers based on Discord feedback this means the difficulty campaign option will have a far larger impact on OpFor generation, especially when combined with my other two OpFor PRs.

Below, I've included the percentage of the player's BV that will be used as base BV allowance when generating OpFors. The old values are in parentheses. As you can see, the impact of each difficulty is substantially larger than the old values. This should give players far better control over their experience and will mean that newer players won't get the pants kicked out of them as often, while veteran players will be able to enjoy the challenge they desire.

Ultra-Green: 50% (80%)
Green: 75% (90%)
Regular: 100% (100%)
Veteran: 125% (110%)
Elite: 150% (120%)
Heroic: 175% (130%)
Legendary: 200% (140%)


IllianiCBT:
Flipping this to live.

@Thom293 when this gets merged you'll need to update your StratCon guide

IllianiCBT:

PR #4727, opened 2024-08-27
Fixed Recursive Node Reader in Force.java

### LINKED ISSUE #4713: MekHQ-0.50.0-SNAPSHOT: Formation Type not Auto Set for Lance Level Formations
### Environment
Version: MekHQ-0.50.0-SNAPSHOT
OS: Win11
Java: 17

### Description
When creating a detailed TO&E for a campaign, formation levels auto-assign at Regiment, Battalion, and Company level.

However, lance level formations do not auto-assign.

This may want to be a separate RFE, but the default top-level has an override, whilst lower levels don't. Personally I'd find it super useful to be able to override the formation level for any TO&E force.

image
image - I'd like this feature for all forces. Let me know if it needs a separate RFE.
image
image
image

### Files
[The Ramshackle Renegades30620207.cpnx.gz](https://github.com/user-attachments/files/16742008/The.Ramshackle.Renegades30620207.cpnx.gz)

@IllianiCBT

IllianiCBT:
Simplified recursive depth calculation in `getMaximumDepth` and improved the method `populateFormationLevelsFromOrigin` by combining related logic. Also adjusted the `populateOriginNode` function to return an integer.

Also removed a couple of 'tabs' that somehow snuck in at some point over the past couple of years.

### Closes #4713

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -130,7 +130,7 @@ public class Force {
 
     public Camouflage getCamouflageOrElse(final Camouflage camouflage) {
         return getCamouflage().hasDefaultCategory()
-                ? ((getParentForce() == null ) ? camouflage : getParentForce().getCamouflageOrElse(camouflage))
+                ? ((getParentForce() == null) ? camouflage : getParentForce().getCamouflageOrElse(camouflage))
                 : getCamouflage();
     }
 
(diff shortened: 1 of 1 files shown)


NickAragua (approved):
Going forward, if you have an excessive number of line break changes, can you either a) roll those back or b) call out where the actual changes are? Sorting through the line breaks and spacing changes makes it difficult to identify where the actual code changes are.

PR #4739, opened 2024-08-28
Fixed NPE in `getEligibleCommanders`

### LINKED ISSUE #4738: MekHQ-0.50.0-SNAPSHOT: Commander loading failure with existing save - error message mentions academies
### Environment
Version: MekHQ-0.50.0-SNAPSHOT - CI1597 Nightly - Megamek launched via MekHQ
OS: Win11
Java: 17

### Description
image

When loading a save created on the 0.50.0 CI-1597 nightly, the above error message appears. This is a blocker and prevents MHQ from loading the save at all.

Some possible conditions?:

- A unit commander flag has been set
- All meks are loaded onto dropships in the TO&E

Split from: https://github.com/MegaMek/mekhq/issues/4728

### Files
[The Nova Dragoons30000826.cpnx.gz](https://github.com/user-attachments/files/16787898/The.Nova.Dragoons30000826.cpnx.gz)

IllianiCBT:
Replaced stream processing with a loop to check for null commander units, preventing potential `NullPointerExceptions` during data loading (if, for example, we attempt to build the TO&E before the Unit has fully initialized).

### Closes #4738

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -434,11 +434,16 @@ public class Force {
      * @return a list of UUIDs representing the eligible commanders
      */
     public List<UUID> getEligibleCommanders(Campaign campaign) {
-        List<UUID> eligibleCommanders = getUnits().stream()
-                .map(unitId -> campaign.getUnit(unitId).getCommander() != null ?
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM!

Issue #4742, opened 2024-08-29
[0.50.0 Nightly CI #1598] Rapid unplanned disassembly of entire force when loading save files created in #1598


LadyAdia:
### Environment
What version of MekHQ does your issue apply to? 0.50.0 Nightly CI 1598
What operating system are you using?  MacOS 14.6.1
What java version are you using?  17.0.2

### Description
Issue: When loading save files created in the current nightly build, nearly all units in the force are disassembled.
First three screenshots shows the status of things when loading a save created in Nightly CI 1596.

<img width="1440" alt="Command Center Before" src="https://github.com/user-attachments/assets/5a3c352b-d18d-4f16-b5cc-c70607a3fc5a">

<img width="1440" alt="Hangar Before" src="https://github.com/user-attachments/assets/68143938-d8a1-47c3-8159-df115a7fd138">

<img width="1440" alt="Warehouse Before" src="https://github.com/user-attachments/assets/22c3bd74-9d79-4902-99ac-dd97ef14efa4">

Upon loading and taking these screenshots, I immediately saved the game and reloaded from the new save. The following shows the result

<img width="1440" alt="Command Center After" src="https://github.com/user-attachments/assets/7c32987b-c9a4-4edd-8064-ea319bfd40d2">

<img width="1440" alt="Hangar After" src="https://github.com/user-attachments/assets/3011f269-4f34-4b7a-b1a5-3060405094af">

<img width="1440" alt="Warehouse After" src="https://github.com/user-attachments/assets/283bfa90-c658-4d8d-b689-821ee7b28e06">

The .zip file attached below contains both the before and after save files, the above screenshots, the log file, and all my userdata

### Files
[MekHQ Save Error 24-08-28 V2.zip](https://github.com/user-attachments/files/16791210/MekHQ.Save.Error.24-08-28.V2.zip)

LadyAdia:
Quick note: the save names "House KaruCRASH" is the before file. I forgot to rename the file before compressing it

UlyssesSockdrawer:
I can replicate this with my Nova Dragoons campaign that migrated from nightly 1597 to 1598.

As an interesting point, my mothballed unit is NOT disassembled, so still gets read properly when loading the save.

![image](https://github.com/user-attachments/assets/99fcbf24-490c-462c-90cc-60e84240333c)

![image](https://github.com/user-attachments/assets/8dd2e8ef-fb77-4c09-a043-bab5694c54b1)

[...]

PR #4743, opened 2024-08-29
Prevent Rapid unplanned disassembly (band-aid)

repligator:
Prevents Rapid unplanned disassembly of units by removing one line. Shameless (band-aid solution, as the real cause of the problem is that the number of days since last maintenance check is being stored with a decimal. I'm 99% sure that we don't have any need to measures how many fractions of a day since last maintenance, since it can only ever occur once per day at max.

Ran MHQ unit tests
Saved and loaded several times

Fixes https://github.com/MegaMek/mekhq/issues/4742

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -2034,7 +2034,6 @@ public class Unit implements ITechnology {
             }
         } catch (Exception ex) {
             LogManager.getLogger().error("Could not parse unit {}", idNode.getTextContent().trim(), ex);
-            return null;
         }
 
         if (version.isLowerThan("0.49.3")) {


Sleet01:
I'm hesitant to take this PR; it seems like you're very close to finding the root cause.
Could you do a bisect of the MHQ repo to find the first commit that introduced the save issue (since I agree with the assessment that this is a save-side error)?
Then it should be relatively simple to revert the change that introduced this bug in the first place.

Thom293:
Yes.  May not be the cause but that's the one I was thinking of.

repligator:
https://github.com/MegaMek/mekhq/pull/4714 does indeed seem to be the root cause. Doubles have decimal places, ints don't. However, simply changing the `daysSinceMaintenance` and `getDaysSinceMaintenance` double back to int breaks maintenance in a way I don't understand.

IllianiCBT:
The issue was caused by me forgetting to switch from int to double when we unmarshal `daysSinceMaintenance` and is fixed by #4746

IllianiCBT:
Issue has since be resolved. Closing as resolve.

PR #4777, opened 2024-09-03
Include more information about available personnel in Camops Personnel Market Refresh Report

Algebro7:
This adds a small amount of information about new personnel in the market refresh report when the CamOps market is being used. I decided to only do this for CamOps as it generates a single unit per day and thus won't flood the report with messages for "higher yield" methods.

This change was in response to player feedback in Discord that it was tedious to check the personnel market every day to see what was available. The intention here is to show the player at a glance the important information (exp level and role, mainly) so they can skip checking the market if they aren't interested.

Here's what the report looks like:

camops-market-report


IllianiCBT:
This looks like it has a failing test; you likely need to update the branch to current in mhq, mm, and mml. Just clicking the 'update branch' button will only update the branch to master on mhq.

Algebro7:
Thanks, looked like some weird intermittent error. Rebased and tests passed this time.

IllianiCBT (approved):

PR #4792, opened 2024-09-05
Second attempt: Sell Parts from Parts in Use Dialog

### LINKED ISSUE #4723: [0.50.0-SNAPSHOT] Selling items from Parts In Use Menu with "Name (Tonnage)" name format causes wrong items to sell
### Environment
00:49:38,142 INFO  [mekhq.MekHQ] {main}
mekhq.MekHQ.initializeLogging(MekHQ.java:315) - Starting MekHQ v0.50.0-SNAPSHOT
    Build Date: 2024-08-25T01:25:18.350793281
    Today: 2024-08-26
    Origin Project: MekHQ
    Java Vendor: Eclipse Adoptium
    Java Version: 21.0.3
    System Locale: en_GB
    MM Code Revision: 38c5c63334f27ef2c0e2cf39b9b634883a948f14
    MML Code Revision: b1b0bee27fee84bfe1af4375aba9a5423258cea7
    MHQ Code Revision: d9a07a3c8f79c902c2c188a1590b53ee786de3bd

### Description
Trying to sell an item that has Name (tonnage) format from the Parts In Use Menu results in a different, similar item being sold instead. This follows specific behaviour outlined below:

Example: selling 20t foot actuators will sell a 65t foot actuator instead. It will continue selling all 65t foot actuators, before moving on to a different weight of foot actuator again, selling all of these, etc.  In some instances, the item fails to sell at all. 

This behaviour appears reliably with items that follow this naming convention, e.g.: actuators, limb parts, Mech sensors, Jump Jets, and so on.

The sell function also does not distinguish between: 
LB 10-X AC
or
LB 10-X AC (Omni Pod)
In this example, trying to sell the former will result in it selling parts of the latter.

Algebro7:
This PR re-implements selling parts from the parts in use dialog in a (hopefully) much more robust way. Rather than storing strings representing the name and looking up the parts in the warehouse, which was incredibly unreliable due to inconsistent naming under the hood, references to the legit spare part objects are stored in the PartInUse row when building out the table. These objects are rebuilt everytime the table refreshes or is re-opened so there should be no concern about the extra persistent state.

I also added a `getSellableQuantity()` method to the `Part` class, primarily to be overridden by `AmmoStorage` and `Armor`, which track their quantities different than all other parts. This allows the caller to not have to explicitly check the type of the Part being sold.

Closes #4723


MekHQ/src/mekhq/campaign/parts/Part.java
```diff
@@ -1275,6 +1275,10 @@ public int getQuantity() {
         return quantity;
     }
 
+    public int getSellableQuantity() {
+        return getQuantity();
```
IllianiCBT, line 1279: If we’re just returning the value of getQuantity() would it make more sense to just call getQuantity() directly instead of indirectly via a new method?

IllianiCBT (approved):
Marking this as approved, despite a minor bit of feedback. Overall the code is clean and looks functional. I would definitely want this in the hands of QA so they can vigorously test it, so once it’s merged please flag it in the QA channel and ask them to break it. With this being such a high demand function, I wanna make sure it hits release in a 10/10 state, if that makes sense.

MekHQ/src/mekhq/campaign/parts/Part.java
Algebro7: Thanks for the feedback. The problem is getQuantity() is what you want for for selling most types of parts, but not others. Without this method, the code calling getQuantity() has to check what type of part they're dealing with, which breaks polymorphism and ends up being more brittle. With `getSellableQuantity()` in the base class, the calling code doesn't have to care what type it is--the responsibility of figuring out what the legal sellable quantity is is delegated to the part instead. 

If the caller doesn't care about selling parts at all it can continue using `getQuantity()` directly instead. Hope that helps answer your question!

IllianiCBT: It flew straight over my head, but you’ve got a solid reason which is good enough for me - even if I don’t entirely understand it :D

IllianiCBT:
This has caught some conflicts following the recent name changes.

PR #4810, opened 2024-09-09
Mechwarrior class rename

rjhancock:
Mech -> Mek

Lots of updates. I'm sorry.


rjhancock:
Still working through it all. It's a very painful process to say the least.

I'm doing a final pass now for Mech to Mek over all 3 and hitting all folders/files

Issue #4764, opened 2024-09-02
Non mechwarriors always get the natural aptitude piloting SPA


ThomasMarik:
### Environment
What version of MekHQ does your issue apply to? 0.50
What operating system are you using?  Windows, Mac, or Linux? Windows
What java version are you using?  17

### Description
When generating a company all non mechwarriors who are selected to get an SPA always gets assigned natural aptitude piloting. While nice that some people might have hidden talents it is very strange that it is always Natural aptitude piloting and not something specific to their jobs.

IllianiCBT:
Based on the description this may be a return of a previously resolved issue caused by these SPAs not having prerequisites.

Another potential cause is that the affected personnel all qualify for the SPAs via Small Arms. Potentially, the SPA pool for affected personnel might be so small that this is the only SPA they qualify for.

Ultimately, this whole issue is caused by how we determine what random SPAs a character receives.

IllianiCBT:
Just confirmed, Natural Aptitude SPAs lost their prerequisites again. So that's the cause.

IllianiCBT:
Ok, I can confirm this was entirely due to incompetence on my part. It turns out I forgot to add the missing requirements for the Natural SPAs when I made the PR that was specifically meant to address that issue.

IllianiCBT:
Closing as resolved

PR #4817, opened 2024-09-12
Updated Campaign Presets

### LINKED ISSUE #4417: Oblique artilleryman SPA should have a pre requisite of at least regular in artillery?
as per title, currently has no pre requisites, which is weird for people who don;t artillery, and could lead to pointless SPA gain upon reg>vet experience gain

IllianiCBT:
This PR factors in a number of pieces of feedback and oversights which came to light following the release of 50.0

### Closes #4417 & #4764


IllianiCBT:
Updated this PR to bring it in line with the recent naming overhaul and to fix a couple of minor issues that were preventing preset load.

IllianiCBT:
I also removed an unnecessary compatibility handler for Clan phenotype that was causing duplicate data to be displayed in campaign options. As we store this information as an Integer not a String, the handler wasn't necessary here, so it was simple enough to just remove.

PR #4846, opened 2024-09-16
Restored Missing Portrait Folders

IllianiCBT:
A number of portrait folders went missing due to the changes made by the name change project and their presence in Git Ignore. This commit restores them


IllianiCBT:
Merging this unreviewed as it only includes portraits and merging now means I can restore gitignore's filtering of these folders.

PR #4848, opened 2024-09-17
Added Tooltip for Crew Requirements in Hangar Tab Display

IllianiBird:
Added tooltips to crew requirements in the Hangar tab. See the below image, for example. This should it much easier for users to identify what specific crew role they're missing.

image

Added comprehensive unit tests to validate the behavior of the new `getCrewTooltip` and `appendReport` methods.

Also, made minor adjustments to column display titles and sizes.


MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -97,7 +97,7 @@ public String getColumnName(int column) {
             case COL_FORCE -> "Force";
             case COL_CREW -> "Crew";
             case COL_TECH_CRW -> "Tech Crew";
-            case COL_MAINTAIN -> "Maintenance Costs";
+            case COL_MAINTAIN -> "Maintenance";
```
IllianiCBT, line 100: This helps reduce the amount of space required to fully display this column's information.

MekHQ/unittests/mekhq/gui/model/UnitTableModelTest.java
```diff
@@ -0,0 +1,180 @@
...
+     // In the following tests the getCrewTooltip() method is called, and its response is checked
+     // against expected behavior
+
+     @Test
+     public void noGunner() {
+         initializeGunners(0,0);
+         initializeEntityClass(true, null);
+         initializeCrew(0,0);
+
+         String expected = "<html><br><b>Drivers: </b>1/1</html>";
```
SJuliez, line 137: I don't do much in the way of unit tests, so please take it with a grain of salt: These tests check GUI output. It seems much more likely that they'll need an update for someone improving the tooltip than they would find an error.

MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -125,16 +126,80 @@ public int getAlignment(int col) {
...
         };
     }
 
+    /**
+     * Returns the tooltip for the crew status of a given unit.
+     *
+     * @param unit the unit for which to get the crew tooltip
+     * @return the crew tooltip as an HTML string
+     */
+    static String getCrewTooltip(Unit unit) {
```
SJuliez, line 154: I suppose this could be private?

MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -125,16 +126,80 @@ public int getAlignment(int col) {
...
-            case COL_QUIRKS -> u.getQuirksList();
+    /**
+     * Returns the tooltip for the specified row and column.
+     *
+     * @param rowIndex    the index of the row
+     * @param columnIndex the index of the column
+     * @return the tooltip for the specified row and column, or {@code null} if no tooltip is available
+     */
+    public @Nullable String getTooltip(int rowIndex, int columnIndex) {
+        Unit unit = getUnit(rowIndex);
```
SJuliez, line 137: With tables... might want to give this a null guard :) (meaning, return null if unit is null)

MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -125,16 +126,80 @@ public int getAlignment(int col) {
...
+
+    /**
+     * Appends a crew report line to the provided StringBuilder.
+     *
+     * @param report the {@link StringBuilder} to append to
+     * @param title the title of the crew role (e.g., "Driver", "Gunner")
+     * @param assigned the number of crew members assigned to the role
+     * @param needed the number of crew members needed for the role
+     */
+    static void appendReport(StringBuilder report, String title, int assigned, int needed) {
```
SJuliez, line 199: also private

SJuliez (approved):
Have some comments. Will leave it to you if you want to make changes.

MekHQ/src/mekhq/gui/model/UnitTableModel.java
IllianiCBT: Unless I'm thinking about things wrongly, this needs to be package-private so it's accessible to the unit test

MekHQ/src/mekhq/gui/model/UnitTableModel.java
IllianiCBT: As with the other instance, I made this package-private so I can unit test it. If that assumption is wrong, please lemme know as otherwise I'm gonna be making this mistake a lot :D

MekHQ/unittests/mekhq/gui/model/UnitTableModelTest.java
IllianiCBT: Any suggestions on how I could improve them? I only learned how to do Unit Tests over the weekend, so I'd greatly appreciate any suggestions :)

MekHQ/src/mekhq/gui/model/UnitTableModel.java
SJuliez: Welllll tbh I have done only little reading w.r.t. unit tests. I will read the pragmatic guide neo suggested. I really recommend J. Blochs Effective Java, it barely touches unit tests but has loads of useful insight from a guy who designed Java. Personally I follow his rule: never upgrade visibility beyond what's strictly necessary, and unit tests are not strictly necessary. Another perspective is this: Don't unit test a class' inner workings. They are likely to change often and in fine ways and then you have to change the unit test. Also, when you change the inner workings, you know you'll have to get it right within that class but you don't need to care about anything outside of it. Rather test the interface of classes, i.e. the methods that other classes can use. When you change the behavior of such a method, other classes may or may not suddenly break and you can't very well check all the code of those.

IllianiCBT: So, based on that, a good rule of thumb would be: if the method is public or used by multiple methods disparate, then I should unit test it; if it's only used by the one method and within a single class then there is no need to Unit Test?

SJuliez: I guess. On that I'm no authority. And dont forget you're doing this on your sweet spare time.

IllianiCBT: I've opted to include a unit test for `getCrewTooltip()` due to its' complexity and the need to ensure it's reporting correctly. However, I have reworked the unit tests based on your feedback. They should be a bit more robust now and easier to maintain.

PR #4865, opened 2024-09-20
Implemented Force Generation 3, Including Clan Bidding & Batchall System

IllianiBird:
### Requires [#6021](https://github.com/MegaMek/megamek/pull/6021)

This PR aims to retool our OpFor generation system for StratCon to provide a better play experience for our users.

First of all, this PR adjusts the AtB force generation tables used to generate formations of a certain weight. The old AtB system identified what weight force it wanted to generate and then rolled a 'die'. Depending on the roll the final force would be at weight, or lighter. What this meant is that a target 'heavy' lance could be aimed for, but three light lances would be spawned instead. Now, if the OpFor is meant to get a 'heavy' lance, they will always get a 'heavy' lance, though the exact make-up of that lance will vary.

Speaking of which, the lance compositions (what weight unit goes into what weight of lance) have been reworked to match the tables on 265 of TW.

Regarding what weight force should be generated. Previously, this would be based on the mean weight of the player force. Now it is randomized, based on the random force weight table on p265 of TW. Furthermore, this is rolled on a per-bot basis. Previously, one roll would be made and applied to all bot forces. Now each force will make a separate roll. That means bot A might bring an 'assault' force, while bot b might favor a 'medium' force.

Difficulty modifiers were being applied twice. Now they are not. Whoops!

The main beneficiaries of these changes will be Aerospace players and players with very large BV forces (25K+ per lance). However, with the other changes made previously, I think the overall feel and experience of fighting MekHQ OpFors will be greatly improved.

The following images show a regular OpFor with C-rated equipment vs. the following lance. Note that all pilots are 'legendary' (0/1).
image

Old System
image

New System
image

Oh and I added a dynamic Batchall system for the clans, nothing big ;)

image


IllianiCBT:
I've loosened up unit culling (when a force is over budget). Now, instead of just checking 'am I over budget' it now only starts culling if the resulting force is >25% over budget. The original system was too consistently producing under budget forces.

IllianiCBT:
Added an option to balance based on Generic BV instead of True BV (enabled by default)

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -786,67 +724,380 @@
...
             }
         }
 
+        // Generate the force
         BotForce generatedForce = new BotForce();
         generatedForce.setFixedEntityList(generatedEntities);
         setBotForceParameters(generatedForce, forceTemplate, forceAlignment, contract);
         scenario.addBotForce(generatedForce, forceTemplate, campaign);
 
+        boolean batchallAccepted = true;
```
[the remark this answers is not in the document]
IllianiCBT: It's read 5 lines later, silly bot

IllianiCBT:
Added Clan bidding and batchall mechanics.

Flipping this to live. It's now feature complete and I'm going back to bug fixes.

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -49,6 +38,12 @@
 import mekhq.campaign.rating.UnitRatingMethod;
 import mekhq.service.mrms.MRMSOption;
 import mekhq.utilities.MHQXMLUtility;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import java.io.PrintWriter;
+import java.util.*;
+import java.util.Map.Entry;
```
rjhancock, line 47: This is going to be a nitpick related to the style guide and the recently added EditorConfig MIGHT trigger for your IDEs.

Imports alphabetized, statics at top, and the wildcard filter set to 10 before triggering.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -410,8 +407,7 @@ public static int generateForce(AtBDynamicScenario scenario, AtBContract contrac
                 }
                 break;
             default:
-                logger.warn(
-                        String.format("Invalid force alignment %d", forceTemplate.getForceAlignment()));
+                logger.warn(String.format("Invalid force alignment %d", forceTemplate.getForceAlignment()));
```
rjhancock, line 410: I'll have to double check but I THINK warn accepts variable parameters.

MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -786,67 +724,380 @@ public static int generateForce(AtBDynamicScenario scenario, AtBContract contrac
...
         if (forceTemplate.getGenerationMethod() == ForceGenerationMethod.BVScaled.ordinal()) {
-            while (((forceBV / forceBVBudget * 100) > targetPercentage) && (generatedEntities.size() > 1)) {
-                int targetEntity = Compute.randomInt(generatedEntities.size());
-                forceBV -= generatedEntities.get(targetEntity).calculateBattleValue();
-                generatedEntities.remove(targetEntity);
+            String balancingType = "";
+            if (campaign.getCampaignOptions().isUseGenericBattleValue()) {
+                balancingType = " Generic";
             }
+            logger.info("Generated a force with " + forceBV + '/' + forceBVBudget + ' ' + balancingType + " BV");
```
rjhancock, line 738: I know the Info method does support variable args and would be better to use the string formatting stuff.

IllianiCBT: It never occurred to me to use String.format, thanks for the suggestion

rjhancock: Yea, it's a vague security issue that ensures the data you're trying to pass into the string is actually what you intended to pass into it.

IE: Not trying to pass an object (for some reason) to a string when it is expecting a number.

IllianiCBT: Sounds good. Won't take me long to get that resolved tomorrow. It'll be a useful thing for future too.

repligator:
I'm getting multiple bachall transmissions for each scenario. One from each of the enemy commanders. IE one from the primary OPFOR, then one from the infantry, then one from the irregulars. BA is also generating as escorts for offboard artillery.

[CBS302530250117.cpnx.gz](https://github.com/user-attachments/files/17099255/CBS302530250117.cpnx.gz)
[CBS302530250110.cpnx.gz](https://github.com/user-attachments/files/17099256/CBS302530250110.cpnx.gz)
[mekhq.log](https://github.com/user-attachments/files/17099264/mekhq.log)
[megamek.log](https://github.com/user-attachments/files/17099268/megamek.log)

IllianiCBT:
So multiple Batchalls, one per enemy bot force, is intentional.

There was a discussion about having it trigger once per contract. I'm torn as to which is the best option.

Off-board units and aerospace certainly shouldn't get random BA tagging along, though. I didn't think to exclude them.

MekHQ/src/mekhq/campaign/CampaignOptions.java
IllianiCBT: Where can I configure that? Or will I need to be doing that by hand?

rjhancock: If using IntelliJ, https://www.jetbrains.com/help/idea/editorconfig.html should help enable it IF you have master merged in with the config file.

The config file probably still needs some tweaks to get it to match the style where it can and to work with the various IDEs but it should be a good base for now.

IllianiCBT:
Flipping this once again from Draft to Live.

At this point any remaining issues can be resolved in a future PR, I think we need to get this merged and tested so further changes can be made using 'real' data.

IllianiCBT:
Ok, now I'm finished for real. The only change impending is replacing the last couple of placeholders in one of the resource bundles, but that can be added post-merge.

HammerGS (approved):
I've been testing this during development and haven't seen issues. But needs to go to the test pilots for balancing.

PR #4869, opened 2024-09-20
Compliance Mech/Tech Fix.

rjhancock:
Fixed names for recently re-added portraits due to .gitignore issue.
Adjusted .gitignore to NOT ignore portraits.

Should we move the other Tech's to Tek's?


IllianiCBT:
Ahhh, sorry Tex, I missed that the portrait file names included prohibited terms. My bad.

rjhancock:
Nah, all good. It happens. The .gitignore is how they got missed initially as I did do the name change but it stopped the adding which is why I removed it in this PR and re-added/fixed them.

PR #4888, opened 2024-09-25
Added 'Part Quality Report' Dialog

### LINKED ISSUE #4773: RFE: Maintenance/Supply tab in Hangar, and maybe ammo too.
Given the recent update, there's a lot more the average player needs to know about keeping their mechs in fighting shape. Currently you can see an average "Quality" rating for a unit in the Hangar, but there's not really a quick way to see an itemized list of it's parts and their respective quality ratings. You can move to the next maintenance cycle and look at the previous report, but this is a text heavy window that is clunky to read.

So my suggestion. In the Hangar, on the right where there is typically a Technical Readout for the currently selected machine... add an extra tab to that window that could be labeled Maintenance/Supply (or something else more fitting). 

Near the top of this tab, could have a button "Buy Parts", this would open a Parts In Use window listing parts of the currently selected unit, including currently equipped ammos, as well as other ammos that could be equipped to the current unit. 

Next a few lines dictate the unit's current Monthly Supply Costs. Then a list of all parts installed in the unit that the game tracks quality for. Splash of red/orange on A/B rated parts that are liable to break, black background white text on parts that are destroyed maybe.

Finally down the bottom (or maybe a 3rd tab) could put the currently equipped ammo, with dropdowns for selecting other ammos a-la swapping them in the MegaMek lobby.

IllianiBird:
Implemented a dialog interface to generate and display a 'part quality report' for a single unit. This includes adding a new menu option, defining the necessary command, and creating the `PartQualityReportDialog` class for handling the report generation and display.

image

I imagine someone else can probably come in to make the report pretty, but I cannot stand GUI work so that person won't be me :D

### Partially Closes #4773

### DIFF
--- a/MekHQ/resources/mekhq/resources/GUI.properties
+++ b/MekHQ/resources/mekhq/resources/GUI.properties
@@ -376,6 +376,10 @@ HangarReportDialog.title=Hangar Report
 MaintenanceReportDialog.title=Maintenance Report
 MaintenanceReportDialog.Unit.title=Maintenance Report for %s
 
+### PartsReportDialog Class
+PartQualityReportDialog.title=Part Quality Report
+PartQualityReportDialog.Unit.title=Parts Quality Report for %s
+
 ### MonthlyUnitCostReportDialog Class
 MonthlyUnitCostReportDialog.title=Monthly Cost Report
 MonthlyUnitCostReportDialog.Unit.title=Monthly Cost Report for %s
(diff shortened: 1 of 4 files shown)


HammerGS (approved):
Tested and seems to work. Does need to look improvements

IllianiCBT:
As much as I don't like GUI work and as much as I specifically said someone will need to do it... 

... how would you like it appearing? If you have a specific design in mind I can see what we can do.

PR #4903, opened 2024-09-26
Implement dynamic hiring halls

Algebro7:
This switches the Hiring Hall system from a static list provided in `atbconfig.xml` to a dynamic one based on various planetary system factors, including tech rating, HPG rating, government, and population. Hiring Hall ratings are shown in the system info pane to the right of the map:

info

And a filter has been added to show hiring halls on the map:

map

Right now I've assigned pretty simple methods with arbitrary target values to determine the quality of the hiring hall, but it should be very easy to adjust based on feedback from players.


IllianiCBT (approved):
LGTM

PR #4911, opened 2024-09-26
Simplified Tech Personnel Checks

IllianiCBT:
Removed experience level checks from tech personnel checks to streamline the determination of technician roles. This removes the really weird situation where ultra-green techs can't be assigned to units (or, in some cases, repair tasks). Whether they _should_ be assigned is something best left to the player. This was particularly odd, as this is the only role that had the 'must be Green or better' restriction.


IllianiCBT:
For the curious, this was why sometimes you'd end up with techs not assigned to units, when using the company generator.

HammerGS (approved):

PR #4943, opened 2024-09-29
Added Contract Difficulty Estimate, Including GUI Support

IllianiBird:
### Requires [#6046](https://github.com/MegaMek/megamek/pull/6046) (tests will fail until that PR is merged)

Refactored the `AtBContract` class to include methods for calculating and displaying contract difficulty. Added UI elements in `ContractMarketDialog` and `ContractSummaryPanel` to show difficulty estimates.

Added dialog calls for when the player is picking unusually easy or overly challenging contracts.

Enhanced code maintainability by removing a long switch statement for faction logos and replaced it with a function call. Expanded the switch to include all factions we have logos for, and not just the Clan factions.

image
image
image
image
image


HammerGS (approved):
Tested and seems to work.

PR #4945, opened 2024-09-29
Added Era Buttons to Date Chooser and Modernized GUI

### LINKED ISSUE #4662: RFE: Include era-bookmarks with the calendar when starting a new campaign.
When starting a new campaign, there is a step when the user specifies the campaign starting date.

This can be enhanced by providing era-specific bookmarks that will automatically set the date to the specific era, which will allow new and returning users to quickly play in the period of their choosing. These eras are already tracked by date in MekHQ, usually displayed at the top of the window along with the date. 

These bookmarks should only set the date, not automatically confirm it (to allow users to browse). Optionally, a textsnippet can provide a contextual overview of the era selected. 

@IllianiCBT - Tagged on request.

IllianiBird:
Refactored the DateChooser class to support internationalization using ResourceBundle for dynamic text handling. Introduced new era selection buttons for different eras in the UI.

image

### DIFF
--- a/MekHQ/resources/mekhq/resources/DateChooser.properties
+++ b/MekHQ/resources/mekhq/resources/DateChooser.properties
@@ -1,0 +1,104 @@
+# monthNames
+monthNameJanuary.text=January
+monthNameFebruary.text=February
+monthNameMarch.text=March
+monthNameApril.text=April
+monthNameMay.text=May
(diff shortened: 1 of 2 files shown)


HammerGS (approved):
Tested and seems to work.

PR #4981, opened 2024-10-04
Implemented Campaign Options IIC Preset Picker

IllianiBird:
Replaced the existing preset picker class with `SelectPresetDialog` - the preset picker that was developed for the Campaign Options IIC project.

I figured compartmentalizing would make reviewing easier. Especially as CO IIC is going to be a hefty merge.

image
image

**Confirm** is a new dialog option added, this completely bypasses the campaign options dialog (users can still pick their date, of course). Once CO IIC is out, we'll have the contents of the General Tab pop up after the date chooser, but for now it will just apply the relevant campaign options and be done.


IllianiCBT:
I forgot to leave a note about the naming of the resource bundle. This is a temporary name in-use while CO IIC is in-dev. It’ll be given a more appropriate name once we no longer have elements from both current CO and IIC

HammerGS (approved):
Tested and seems to work.

PR #4992, opened 2024-10-05
Add Customization Option to Refit Dialog

WeaverThree:
Using a refit kit is easier than and has different rules from customizing a mek, and until now the only way to use the customization rules was to use the meklab manually.

This setup allows you to use the customization rules to change mek to a any available variant without using the meklab tab, even when refit-to-canon-units-only is on. This means you can do more without using the meklab tab, and that customizing units in MegaMekLab externally is more viable in more scenarios.

Utilities.getAllVariants was changed but it's a single-client function.

As a bonus, the refit dialog now resizes a little better, with the mek view panels now enlarging with the window size.

This screenshot shows how only customize is available when the refit-to-canon-units-only option is turned on.


IllianiCBT:
Thank you for this contribution and welcome to the project!

HammerGS (approved):
Tested and seems to work.

PR #5002, opened 2024-10-08
correct issues with fixed map generation

kuronekochomusuke:
- correct filter to be ```!getTerrainType().equals("Space")``` was proventing fixed map generation
- allow fixed map when doing a map reroll

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBScenario.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBScenario.java
@@ -26,6 +26,9 @@ import java.text.ParseException;
 import java.time.LocalDate;
 import java.util.*;
 
+import megamek.common.*;
+import megamek.common.util.fileUtils.MegaMekFile;
+import megamek.utilities.BoardClassifier;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/mission/AtBScenario.java
```diff
@@ -492,6 +487,38 @@ public void setMapFile() {
         setMapFile(getTerrainType());
     }
 
+    /**
+     * If there are maps of the appropriate size available and we roll higher than
+     * the given threshold, replace the scenario's generated map with a fixed map
+     * from data/boards
+     */
+    public void setScenarioMap(int mapChance) {
```
IllianiCBT, line 495: So something I've been thinking about: when we check for maps of `x` size is there value in adding some flexibility? For example: `<=x*1.1 && >=x*0.9`, this would increase the chances of hitting a valid map.

IllianiCBT (approved):
Looks great. Only one comment, which is more of a discussion point rather than anything that should stop this getting merged.

MekHQ/src/mekhq/campaign/mission/AtBScenario.java
kuronekochomusuke: xVariance and yVariance increases the range of board sizes searched from the current size.   making the variance a campaign option would be nice to have.
```
    public List<String> getMatchingBoards(int width, int height, int xVariance, int yVariance,
            List<Tags> tags) {
```

IllianiCBT: Ah ha! Completely missed that. I agree it would be nice to have as a campaign option.

Issue #3168, opened 2022-03-13
[0.49.7] Preserve percent of force fails if original lance is undeployed and new lance deployed


mjbroekman:
### Environment
	Release Date: 2022-02-18
	Today: 2022-03-13
	Origin Project: MekHQ
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.13
	Platform: Windows 10 10.0 (amd64)
	System Locale: en_US
	Total memory available to MegaMek: 8 GB

### Description
When a scenario generates and "allocates" a lance for the scenario, the "Preserve % of listed forces" objective will fail if you undeploy the allocated lance and deploy a different lance in its place.
![Screenshot 2022-03-13 14 23 07](https://user-images.githubusercontent.com/5141448/158073764-23615084-0873-47fd-a2d1-4109fd54dd42.png)

In the above scenario, my 'Fredasa' lance was too damaged from a previous combat, so I deployed my 'Kinnol' lance after undeploying 'Fredasa'. All units survived the scenario and my BV was ~89% of total at the end of combat.

As a side note, with the success of the recon raid, it said I gained 1 AtB roll, but no roll was logged post-combat:
![Screenshot 2022-03-13 14 31 14](https://user-images.githubusercontent.com/5141448/158073987-ef516b42-7d1f-4e0c-8862-eeb244bc7ba4.png)

Daily log: Combat occurred after the addition of ammo to the procurement list and the only things added were prisoners, my recovered mechs, and salvaged opfor units.
![Screenshot 2022-03-13 14 33 54](https://user-images.githubusercontent.com/5141448/158074161-a7878f0e-5bb9-41e2-8ba5-8daf9ee2d802.png)

### Files
[customs.zip](https://github.com/MegaMek/mekhq/files/8240356/customs.zip)
I think this is the pre-combat autosave: [Autosave-1-Star League Reborn-30140530.cpnx.gz](https://github.com/MegaMek/mekhq/files/8240359/Autosave-1-Star.League.Reborn-30140530.cpnx.gz)
If not, this is the prior day autosave: [Autosave-2-Star League Reborn-30140529.cpnx.gz](https://github.com/MegaMek/mekhq/files/8240360/Autosave-2-Star.League.Reborn-30140529.cpnx.gz)
[entitystatus.txt](https://github.com/MegaMek/mekhq/files/8240364/entitystatus.txt)
[megamek.log](https://github.com/MegaMek/mekhq/files/8240365/megamek.log)
[...]

Issue #2749, opened 2021-07-16
Problems with non-combat formations in StratCon


SuperStucco:
In 49.2, formations flagged as non-combat are being selected as primary forces for scenarios.  In a few cases this is appropriate e.g. as part of a convoy in a Convoy Escort scenario, or in a Base Defense scenario.  However they are also being assigned as primary forces for other scenarios, as in the attached campaign.

[Jansenn's Horde30420214.cpnx.gz](https://github.com/MegaMek/mekhq/files/6831692/Jansenn.s.Horde30420214.cpnx.gz)
[Customs.zip](https://github.com/MegaMek/mekhq/files/6831694/Customs.zip)

There are other problems as a result of this.  Because the formation is flagged as non-combat they are not always assigned to the scenario i.e. the scenario is generated and no forces are assigned; in this specific example, because the primary formation is non-combat the map size is 0x0 and no defending forces are generated.

Sometimes a scenario has no primary force assigned but the victory conditions still list the non-combat formation.  While this formation could be left out and a combat formation assigned, in the scenario resolution after playing it out in MegaMek the units do not have check boxes in the victory conditions window leaving it as an automatic fail.  If the AtB Campaign State "Manage Scenario" dialog is used to assign the non-combat formation to the scenario, because it is flagged as non-combat it is not deployed to MegaMek and the units are still not in the scenario resolution window.  The only way for this to function is to read the victory conditions thoroughly, find the listed formations, ensure they are flagged as combat, and manually deploy them from the TO&E tab.

Short version: non-combat formations should be handled better in StratCon.  There needs to be a point to the combat/non-combat distinction.  They should not be assigned as primary forces.  They could be assigned as secondary forces in scenarios such as Convoy Escort or Base Defense.  When scenario victory conditions include preserving part of a formation, those formations (combat or non-combat) should be automatically added to the scenario and also sent to MegaMek.

UlyssesSockdrawer:
Should be fixed by https://github.com/MegaMek/mekhq/pull/5010 - can be closed once merged.

IllianiCBT:
Closing as resolved

PR #5010, opened 2024-10-08
Hacky Fix: Makes StratCon Scenario Generation Respect Non-Combat Flag in TO&E

UlyssesSockdrawer:
This is a very simple hacky fix to StratconRulesManager.java's scenario generation.

Please be aware I don't know how to write or add unit tests, so haven't done any of that here.

I've added a line that, when trying to generate a new scenario on a Monday, as well as Nick's original checks of whether a unit is already deployed and is assigned to a track, also checks if the force is flagged as combat.

I've seen it step over a force that was flagged as non-combat in a very simple test campaign using the debugger so I think this works, but this needs careful testing to make sure.

I've attached a save file to this PR that people can use to test by generating some contracts and then seeing if you ever get 'lance-truck' picked as the seed lance.

 It has one combat force and one non-combat force - lance-mek (combat) and lance-truck (non-combat).

[NonCombatTestUnit-Base.cpnx.gz](https://github.com/user-attachments/files/17298574/NonCombatTestUnit-Base.cpnx.gz)

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -1087,6 +1087,7 @@ public class StratconRulesManager {
                 .mapToInt(key -> key)
                 .mapToObj(campaign::getForce).filter(force -> (force != null)
                         && !force.isDeployed()
+                        && force.isCombatForce()
                         && !forcesInTracks.contains(force.getId()))
                 .map(Force::getId)
                 .collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


IllianiCBT (approved):

HammerGS:
I know we have issues that should be identfied and closed with this. Can someone reference those for closure.

UlyssesSockdrawer:
@HammerGS looks like:

https://github.com/MegaMek/mekhq/issues/2749 - can definitely be closed by this

https://github.com/MegaMek/mekhq/issues/3168 - only the 'non-combat' part of this, so might not be fully closed here but might be closed by the work @IllianiCBT did around ForceGen3?

IllianiCBT:
I'm going to leave #3168 open as this is part of a bigger issue where legacy ATB functionality causes issues with StratCon. For StratCon users shouldn't be deploying from the TO&E and shouldn't be manually un-deploying forces as that will (as illustrated in the cited report) cause issues.

PR #5017, opened 2024-10-10
Display Advanced Medical Injuries in UI

WeaverThree:
The numerical hit counter in the force display in the TO&E and the injuries column in the Personnel tab were both only showing "hits" from when advanced medical is turned off. This fixes them so they work with both systems. Are there any other places I need to hit with this?


IllianiCBT (approved):

HammerGS:
This has picked up conflicts.

IllianiCBT:
Just a thought while you fix these conflicts, it might be useful to have fatigue displayed in the TO&E, too.

You'll want something that looks like this:

```
if ((campaign.getCampaignOptions().isUseFatigue()) && (person.getEffectiveFatigue(campaign) > 0)) {
        int effectiveFatigue = person.getEffectiveFatigue(campaign);
        ++ colored text code ++
}
```

WeaverThree:
Not sure why that was a conflict, it was in a separate part of the file.



It doesn't look great unless you expand the panel, but then, neither do their names.

IllianiCBT:
Could we perhaps arrange it so that its...

> <Name>
> <Veterancy and Role>
> <Injuries/Hits and Fatigue>

...I don't know if that would look better.

HammerGS:
> Now it's like this:
> 
> 

I prefer this.

IllianiCBT:
LGTM

PR #5025, opened 2024-10-11
Pin/sticky force view tab on TO&E panel

WeaverThree:
Another tiny one but it will save me many many clicks. It makes the force view tab sticky, so if you're viewing a unit and select a different TO&E entry, it stays on unit. Or goes back there when its relevant.

There's probably a better way to do this like rearchitecturing the pane to not be recreated all the time and to just have two subpanels that get populated as needed, but this is the least invasive change I could come up with.


IllianiCBT (approved):
Everything looks fine, but this is something I'd want tested before it gets merged.

HammerGS:
@IllianiCBT I'll let you decide when you want to merge this.

PR #5029, opened 2024-10-11
Better visual distinction for GM mode / overtime toggles

SJuliez:
Using MMToggleButton

image


IllianiCBT (approved):
Thank you for doing this

PR #5030, opened 2024-10-11
Replaced autoAwards Award Ceremony Placeholder Image

IllianiCBT:
Updated the image file format for the award ceremony from PNG to JPG and adjusted its scaled height from one-seventh to one-fourth of the screen height.

Replaced placeholder image with final.


IllianiCBT:
Merging this due to the very small amount of changes made and wanting to restore the gitignore change I needed to make for this to be viable.

PR #5031, opened 2024-10-11
Cleanup related to new naming scheme and refactoring, fixes refits to blank model names.

Setsul:
Removed old kludges obsoleted by getFullChassis()
Refactoring of retrieveOriginalUnit() to enable
refactoring of populateRefits() that properly throws exceptions.

This fixes the bug that ever since `(Standard)` was purged from model names, those models with now blank names stopped showing up as targets for refits in the GUI.
And if something similar happens again we will get an EntityLoadingException now.

Screenshots from 49.7 (last milestone before the bug) vs 50.0 vs this commit for illustration purposes.
MekHQ refit bug3
MekHQ refit bug1
MekHQ refit bug5

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -1331,23 +1331,21 @@ public class Utilities {
      * @throws EntityLoadingException
      */
     public static MekSummary retrieveOriginalUnit(Entity newE) throws EntityLoadingException {
-        MekSummaryCache cacheInstance = MekSummaryCache.getInstance();
-        cacheInstance.loadMekData();
-
+        // this might be better with refreshUnitData() and/or elsewhere
+        MekSummaryCache.getInstance().loadMekData();
+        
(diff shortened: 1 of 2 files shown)


IllianiCBT:
Welcome to the project :)

Setsul:
Weaver did almost instantly point me at the line that was causing the problem, so credit where credit is due.

They did look over it already and chose this version over a literal 1 line fix that left the accumulated cruft in place, for obvious reasons, so I expect we'll get their official blessing shortly.

WeaverThree:
Looks good to me. I checked out `retrieveOriginalUnit()` and found that it's a single-client function used in `Refit.saveCustomization()`, which i checked and it still works fine :>. I also tested that it fixed the issue, and it did, it let me refit a Warhammer IIC 2 to a plain Warhammer IIC no problem.

IllianiCBT:
Perfect, thanks for taking a look :)

IllianiCBT (approved):

PR #5035, opened 2024-10-11
Fix #5028 and allow random camo to go into deeper directories

### LINKED ISSUE #5028: [0.50.01 2024-10-10] AtBContract.pickRandomCamoflage can select directory instead of file

### Severity *

### Brief Description *

From my save file. Here's a working ally/enemy camo decleration

```
<camouflage>
    <category>Federated Suns/Avalon Hussars</category>
    <filename>34th Avalon Hussars (Ardan Sortek&apos;s Battalion).jpg</filename>
</camouflage>
```

Here's a non-working one

```
<camouflage>
    <category>Capellan Confederation</category>
    <filename>Chesterton Reserves</filename>
</camouflage>
```

It appears to have picked a directory instead of a file.

This appears to be caused by [AtBContract.pickRandomCamoflage](https://github.com/WeaverThree/mekhq/blob/8961ffeea53fd8fa721aff301bd1e3eebb55e370/MekHQ/src/mekhq/campaign/mission/AtBContract.java#L236)

It looks like the collection adding on line 252 or possibly the loop below that need to be filtering on `isFile` or similar, but I want someone who knows Java better to have a look at it.

Thanks to @atmafox for seeing the problem.

### Steps to Reproduce

Play game until it randomly picks bad camo. Not much else to do for it.

### Operating System *

Linux

### Java Version *

17.0.12

### MekHQ Suite Version *

Free Text (type manually)

### Custom MekHQ Version

0.50.01 HEAD 2024-10-10

### Attach Files

mekhq.MekHQ.initializeLogging(MekHQ.java:306) - Starting MekHQ v0.50.1-SNAPSHOT
    Build Date: 2024-10-10T19:36:25.665243020
    Today: 2024-10-10
    Origin Project: MekHQ
    Java Vendor: Eclipse Adoptium
    Java Version: 17.0.12
    Platform: Linux 6.11.2-203.fsync.fc40.x86_64 (amd64)
    System Locale: en_US
    MM Code Revision: e7d09b5391c560f7f47ff03222ef17c83ffdc038
    MML Code Revision: 446bd8d9c7e28bfdf49d35f312d838c44e16941a
    MHQ Code Revision: 8961ffeea53fd8fa721aff301bd1e3eebb55e370

There was nothing in the logs about bad camo.

Here's my save in case it helps
[0.50.01-2024-10-10 Camo.zip](https://github.com/user-attachments/files/17337857/0.50.01-2024-10-10.Camo.zip)

### Final Checklist

WeaverThree:
The random camouflage picker was sometimes picking directories instead of files to use as camo for a group in a generated contract. It also wasn't going into all of the subdirectories, only the first level under the faction, so some files were getting left out. This fixes both of those problems. 

It seems like Camoflauge or AbstractIcon should have some verification that the files you pass them exist, but...

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -260,7 +260,8 @@ public class AtBContract extends Contract {
             Path randomPath = allPaths.get(new Random().nextInt(allPaths.size()));
 
             String fileName = randomPath.getFileName().toString();
-            String fileCategory = randomPath.getParent().toString(); // randomFile.getParent().replaceAll("\\\\", "/");
+            String fileCategory = randomPath.getParent().toString()
+                .replaceAll("\\\\", "/"); // Is this necessary on windows machines?
             fileCategory = fileCategory.replaceAll(ROOT_DIRECTORY, "");
 
             return new Camouflage(fileCategory, fileName);
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -64,12 +64,17 @@ import org.w3c.dom.NodeList;
 import javax.swing.*;
 import java.awt.*;
 import java.io.File;
+import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.ParseException;
 import java.time.DayOfWeek;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static java.lang.Math.round;
 import static megamek.client.ratgenerator.ModelRecord.NETWORK_NONE;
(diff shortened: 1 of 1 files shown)


IllianiCBT:
On my mobile so can’t review, but thank you for fixing this

MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -238,41 +243,32 @@ public static Camouflage pickRandomCamouflage(int currentYear, String factionCod
...
-        if (!allFiles.isEmpty()) {
-            File randomFile = allFiles.get(new Random().nextInt(allFiles.size()));
+        if ((null != allPaths) && (!allPaths.isEmpty())) {
+            Path randomPath = allPaths.get(new Random().nextInt(allPaths.size()));
 
-            String fileName = randomFile.getName();
-            String fileCategory = randomFile.getParent().replaceAll("\\\\", "/");
+            String fileName = randomPath.getFileName().toString();
+            String fileCategory = randomPath.getParent().toString()
+                .replaceAll("\\\\", "/"); // Is this necessary on windows machines?
```
IllianiCBT, line 264: Yes, to answer your question. Basically, windows gets angry if you're using mixed slashes in a directory string. So we convert to use all forward slashes, which it is happy to accept - because all the slashes are in the same direction. Windows is specifically the reason we do this.

There is almost certainly a better way to handle it.

IllianiCBT (approved):

MekHQ/src/mekhq/campaign/mission/AtBContract.java
WeaverThree: If I was still on windows I'd throw a log on there and confirm if the file functions are actually giving us backslashes because MOST development systems will just give forward slashes on all systems, but I don't know about java... :|

IllianiCBT: When we initially launched this we ran into issues when using mixed slashes, because `getParent` returns `parent/child` on good OSs and `parent\child` on windows. So you end up in a situation where you're seeing `parent\child/myLittleCamo.png` which was confusing the heck out of windows.

IllianiCBT: Or something to that effect

Issue #4978, opened 2024-10-04
[RFE] Save MekHQ MekLab customs to user dir if user dir is set


WeaverThree:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the RFE to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

When you customize a mek in the meklab INSIDE mekhq, it does not give you an option as to where to save the .mtf (or presumably .blk for other unit types), they just go to 

`<program dir>/data/mekfiles/customs/<campaign name>/`

I propose that if the user directory option is set, they instead go to

`<user dir>/<campaign name>/`

Or possibly

`<user dir>/campaigns/<campaign name>/`

This would make it easier to deal with different megamek versions and such.

You can't really give the user the choice of where to save the file, which would also resolve this, because MekHQ needs to have that file be immediatly accessable for the campaign to continue, so this is the best option I can think of.

### MekHQ Suite Version *

v0.50.1-SNAPSHOT: 2024-0-012

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

Issue #4554, opened 2024-08-05
[0.50.0-SNAPSHOT] Refitting through MekLab within MekHQ freezes application. 


Tzahr:
### Environment
17:25:38,461 INFO  [mekhq.MekHQ] {main}
mekhq.MekHQ.initializeLogging(MekHQ.java:303) - Starting MekHQ v0.50.0-SNAPSHOT
    Build Date: 2024-08-04T01:23:25.882857488
    Today: 2024-08-05
    Origin Project: MekHQ
    Java Vendor: Eclipse Adoptium
    Java Version: 21.0.3
    Platform: Windows 11 10.0 (amd64)
    System Locale: en_GB
    Total memory available to MekHQ: 8 GB
    MM Code Revision: 0f0d10d35f6bbc190c42be24563140865de228c5
    MML Code Revision: 35aa5ce66ca59272c0d1e05c3d174cd4c0335b3c
    MHQ Code Revision: 76e9cea125956ab9a9d8af389ee60c2bb2eca81d

### Description
Loading a mech into the MekLab in MekHQ, creating a custom refit, and applying it causes the application to freeze up. 
Curiously, after force-closing the application and reloading it, the custom variant does appear as a refit kit, meaning it does save the mechfile without issue. 

### Files
[mekhq.log](https://github.com/user-attachments/files/16502491/mekhq.log)

repligator:
See also 
#4072

I've experienced this exact issue (freeze, but the refit kit is still available) multiple times in 49.19 and 49.20, but I haven't encountered it yet in 50.

Tzahr:
It happens pretty much every tiue I do a refit even in the Nightly, and it's been happening to my fried also, unfortunately.

Sleet01:
@Tzahr Could you post a campaign file and zipped custom mechs folder so we can try to repro the issue with your setup?

Tzahr:
I've updated Nightly a few times since the initial report; I have not been refitting a whole lot, so I tried it again on the same save to check if this was still relevant, and lo and behold, it now appears to function just fine the few times I tested. 

I'll tinker a bit, and should it happen again, I'll report back with the specifics. Until then, though, this appears to have resolved itself, either on my end, through a knock-on effect.

LadyAdia:
More data points for the issue. Nightly CI1600. I'm including my files to see if it'll help reproduce the issue. Of note: if I customize shortly after starting a play session, things work fine. It's only after playing for some time that the freeze will occur.

[...]

PR #5043, opened 2024-10-13
New names cleanup and potential bugfixes

### LINKED ISSUE #4791: units.cache gaining extra duplicates of refits and sometimes the entire unit database
### Environment
Windows 10 / Java 17 / 0.50.00 release

### Description
units.cache is receiving duplicates of refits when they are made. If units.cache is deleted to attempt resolution, units.cache recieves a duplicat of all BattleTech units (not just customs), doubling the file size the first time and increasing by the same amount each successive time, resulting in multiple duplicates of everything.
image
image
image
The 'Firebird' was a refit made before moving to 0.50.00, the (Food Truck) was made after, both are duplicated. Removing the units.cache and the mechfiles and even replacing the units.cache with one from another verified 'clean' installation seems ineffective.
### Files
[FOODIES.zip](https://github.com/user-attachments/files/16885492/FOODIES.zip)
[customs.zip](https://github.com/user-attachments/files/16885494/customs.zip)
[units.zip](https://github.com/user-attachments/files/16885498/units.zip)
The Units.cache file attached has duplicates of custom refits.

Setsul:
That should be the last of MekHQ updated to follow the new "IS name (clan name) model" naming scheme. MM and MML still have some bits.
Please excuse the funny list of commits, only [bd0e259](https://github.com/MegaMek/mekhq/pull/5043/commits/bd0e2599c8c294e150304c29ea57d8c1af0eac9c) is actually new.

This will prevent custom refits of clan units from reusing already taken model names. I don't think that bug was reported though.
The change from `loadMekData()` to `refreshUnitData()` might help with #4554 and potentially fix #4791 but I haven't tested either and the proper way for the former would still be to enable the (un)loading of a single specified file.

The rest is some incidental cleanup and refactoring to prepare for implementing #4978


IllianiCBT (approved):
How much testing has the cache portion of this had? The code looks fine, I’m just wary of exacerbating the existing refit issues.

IllianiCBT:
Also, this is approved pending the JavaDoc error being fixed

Setsul:
> How much testing has the cache portion of this had? The code looks fine, I’m just wary of exacerbating the existing refit issues.
"It works on my machine"
Not very extensive, but the few refits I've tried did show up and it didn't lag or freeze.
`refreshUnitData()` is used elsewhere so if it's broken we need to fix it anyway.

IllianiCBT:
That was my assumption, but I’ve learned not to assume. When this gets merged let’s flag it with the QA folks so they know to test it, speaking of which I should probably give you to contributor role on Discord…

PR #5053, opened 2024-10-16
Colorise Skill Levels

WeaverThree:
This started because I wanted a way to easily distinguish skill levels in Tech List of the Repair Panel:

This followed on to the repair section of the Warehouse:

And the refit selector:

And then on to non-techs, we have the Doctor List, which probably needs to be sorted by skill as well but who has that many doctors?

And then it was on to the rest of the places skill shows up in the program:

Of course, in the Personnel List, things can become a bit overboard with row highlights:

But in a week of playing with it, I didn't have a list nearly that bad and I don't think it works out too bad even if you do.

Of course, the colors are customizable. I believe these are reasonable defaults, though:

Along the way, I also cleaned up a fair bit of code to use StringBuilders, better HTML, better case statements, and `ReportingUtilities.messageSurroundedBySpanWithColor()` and such.


IllianiCBT:
So one point of feedback, before I review this properly: as a rule we use green to denote 'good' and red to denote 'bad'. While I see why you chose this scale, I think flipping the colors would be better. Purple for Elite is 10/10 though, let's keep that.

MekHQ/src/mekhq/campaign/market/PersonnelMarket.java
```diff
@@ -157,7 +157,7 @@ public void generatePersonnelForDay(Campaign campaign) {
                     stringBuilder.append("<br>A ");
                 }
                 stringBuilder.append("<b> ")
-                    .append(expLevel)
+                    .append(SkillType.getColoredExperienceLevelName(person.getExperienceLevel(campaign, false)))
```
IllianiCBT, line 160: This might be something for another PR, but I think we should overload this method with a version that defaults secondary to false. Given the sheer frequency of `person.getExperienceLevel(campaign, false)` in our codebase, I think that would have value.

MekHQ/src/mekhq/campaign/personnel/SkillType.java
```diff
@@ -153,6 +138,82 @@ public static String getExperienceLevelName(int level) {
...
+     * @return String hex code for a font tag
+     */
+    public static String getExperienceLevelColor(int level) {
+        return switch (level) {
+            case EXP_ULTRA_GREEN -> MekHQ.getMHQOptions().getFontColorSkillUltraGreenHexColor();
+            case EXP_GREEN -> MekHQ.getMHQOptions().getFontColorSkillGreenHexColor();
+            case EXP_REGULAR -> MekHQ.getMHQOptions().getFontColorSkillRegularHexColor();
+            case EXP_VETERAN -> MekHQ.getMHQOptions().getFontColorSkillVeteranHexColor();
+            case EXP_ELITE -> MekHQ.getMHQOptions().getFontColorSkillEliteHexColor();
+            case -1 -> "";
```
IllianiCBT, line 168: This will default to the default color based on the theme, correct?

IllianiCBT (approved):
A couple of questions, but nothing that'll stop this from being merged. :)

MekHQ/src/mekhq/campaign/market/PersonnelMarket.java
WeaverThree: That sounds like a good idea for another PR, yeah. I'll look at doing that when I clear my current stuff :>.

WeaverThree: Made a note of it even.

MekHQ/src/mekhq/campaign/personnel/SkillType.java
WeaverThree: It should, because the functions that use it don't even generate a tag when the result is "".

WeaverThree:
For the colors, does this look good?

IllianiCBT:
Perfect! :)

PR #5054, opened 2024-10-16
Implement CamOps Contract Negotiation

Algebro7:
This implements the CamOps RAW Negotiation mechanics for the CamOps Contract Market. Players are allowed to reroll each term once, and the term is recalculated based on the MOS/MOF of an opposed negotiation check. For each 2 points of MOS/MOF, the roll on the supplemental terms table is adjusted up or down by 1.

I ended up adding some additional Camops-specific state in AtBContract. Specifically, the initial rolls for the contract so that they can later be adjusted if the terms are negotiated.

~~Flipping this to Draft since while writing I realized that we would need to serialize the initial rolls as well.~~


IllianiCBT (approved):
Everything looks good. Let's get this in the hands of the testers.

PR #5083, opened 2024-10-20
Fix StratCon vs Normal deployment status of units

WeaverThree:
Right now, there's this problem:

Fire Support I was deployed via the right click menu in TO&E and Scout I was deployed via the StratCon Map.

Of note is that if you save and load the game, the units would get the deployment flag then anyway.

I consolodated the code for updating unit deployments into Force.SetScenarioID.

Note that this does not add or remove issues regarding TO&E drag and drop that I will document in another issue. (#5084)

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -896,7 +896,7 @@ public class Campaign implements ITechManager {
         int id = lastForceId + 1;
         force.setId(id);
         superForce.addSubForce(force, true);
-        force.setScenarioId(superForce.getScenarioId());
+        force.setScenarioId(superForce.getScenarioId(), this);
         forceIds.put(id, force);
         lastForceId = id;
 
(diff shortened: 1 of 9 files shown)


IllianiCBT:
Can you clarify what this PR does? :)

WeaverThree:
Sorry for late reply!

So right now, there are some bits of code that do force deployment like this:

```java
force.setScenarioID(myScenario)
for every unit in force {
  unit.setScenarioID(myScenario)
}
```

And there are other bits of code that do force deployment like this:

```java
force.setScenrioID(myScenario)
// and then nothing
```

The latter method causes the force to be considered deployed in the UI but none of the units in the force to be. (Unless you save and load the game). This means that with the latter method, the UI won't show the units and people as deployed in the TO&E, hangar, and personnel list but the former method will.

Notably, TO&E right click deployment uses the former method and Stratcon Map-based deployment uses the latter method.

This PR aims to make all deployments work the same by moving the

```java
for every unit in force {
  unit.setScenarioID(myScenario)
}
```

INTO `force.setScenarioID()` (and removing it from other places).

This would enforce consistency across deployment code.

IllianiCBT (approved):

PR #5088, opened 2024-10-20
Added 100+ New Quirks, Refactored Personality Traits & Quirk Classes

IllianiCBT:
This PR adds around 100 new personality quirks, more than doubling the pool. It also adjusts saving and loading of traits to use ordinal values rather than hardcoded. This will make adding new quirks much easier and is just cleaner code. I also removed a bunch of unused boolean comparators.


IllianiCBT:
The deprecated warnings can be ignored: I was the one who deprecated those methods.

HammerGS (approved):
Tested and seems to work .

PR #5103, opened 2024-10-21
Globally fix slow mousewheeling issues, with scaling!

WeaverThree:
`JScrollPane`. They're everywhere. Sometimes the mousewheel scrolling works fine with them, sometimes it doesn't. There doesn't seem to be any pattern to when it works.

MegaMek and MegaMekLab already control their `JScrollPane`s with `get*ScrollBar().setUnitIncrement(int)`, seemingly doing it for each one separately. HQ only does it for the StratCon map.

Something like two-thirds of HQ's are anonymous, which makes setting properties on them a bother.

This implements a drop-in replacement for `JScrollPane`, `JScrollPaneWithSpeed`, which sets its own increment modified by the GUI scale factor, and replaces all uses of `JScrollPane` with it. (Except that Stratcon Map one.) This provides a single point for tuning and would allow for customization down the line.


MekHQ/src/mekhq/gui/utilities/JScrollPaneWithSpeed.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2019-2024 - The MegaMek Team. All Rights Reserved.
```
IllianiCBT, line 2: New files only need the current year

MekHQ/src/mekhq/gui/utilities/JScrollPaneWithSpeed.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * It's a JScrollPane that manages its scrollspeed based on the UI scale
+ */
+public class JScrollPaneWithSpeed extends JScrollPane {
+    final static int BASE_INCREMENT = 16;
+
+    /**
+     * @see JPanel#JPanel()
+     */
+    public JScrollPaneWithSpeed() {
```
IllianiCBT, line 36: Would it make sense to use overrides in this class?

WeaverThree: You mean `@Override`?  I didn't know that worked on constructors

IllianiCBT: Sorry, that was just poor placement on my part. I had to stick the comment somewhere, but it was more a general comment on the class as a whole.

IllianiCBT: Conversation can be ignored.

IllianiCBT (approved):
LGTM

PR #5105, opened 2024-10-22
ResolveScenarioWizardDialog - Some UX Improvements

WeaverThree:
This increases the size of the Resolve Scenario wizard a bit so you can see all the components horizontally, and makes it stop changing size as much. It also increases the size of the popups for viewing personnel to something useful.

There are still some issues with window placement and size wiggling that I intend to look into but this should increase the UX of these dialogs a little bit at least.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
@@ -26,6 +26,7 @@ import megamek.client.ui.dialogs.EntityReadoutDialog;
 import megamek.client.ui.preferences.JWindowPreference;
 import megamek.client.ui.preferences.PreferencesNode;
 import megamek.client.ui.swing.UnitEditorDialog;
+import megamek.client.ui.swing.util.UIUtil;
 import megamek.common.GunEmplacement;
 import megamek.common.options.OptionsConstants;
 import megamek.logging.MMLogger;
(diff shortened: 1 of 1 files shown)


HammerGS (approved):

MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
```diff
@@ -1769,8 +1737,8 @@ private void showPerson(PersonStatus status, boolean isPrisoner) {
         gridBagConstraints.fill = GridBagConstraints.NONE;
         dialog.getContentPane().add(btn, gridBagConstraints);
 
-        dialog.setMinimumSize(new Dimension(600, 300));
-        dialog.setPreferredSize(new Dimension(600, 300));
+        dialog.setMinimumSize(new Dimension(450, 700));
```
IllianiCBT, line 1740: Can we replace these fixed values with `dialog.setMinimumSize(UIUtil.scaleForGUI(450, 700)` and `dialog.setPreferredSize(UIUtil.scaleForGUI(450, 700)`. That will ensure the dialog correctly scales with the users scaling settings.

IllianiCBT (changes requested):

MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
WeaverThree: Done in all 3 places

IllianiCBT (approved):

PR #5107, opened 2024-10-22
Updated Interstellar Map View Defaults & Optimized 'ISW' view

IllianiCBT:
This PR updates the interstellar map to make the old ISW the default. It makes some substantial optimizations under the hood to ISW to make it less of a resource hog. As well as some general visual changes to that view and renamed it to 'Territory'.

This has a few added benefits. In addition to looking closer to CGL's official maps, the territory won't disappear when the user changes map filters. So you avoid situations like the image below:
image

Instead you get...
image


HammerGS (approved):
Tested and seems to be working.

PR #5111, opened 2024-10-22
Better multi-day repair and refit reports.

WeaverThree:
Changes the multi-day repair message to specify days remaining instead of always saying "tomorrow". Does similar for the refit messages.


HammerGS:
This is something long needed

IllianiCBT (approved):

PR #5133, opened 2024-10-26
Refit - Reformat, document, annotate, comment, clarify

WeaverThree:
This is me spending a day or two to go through and understand Refit.java, and clean it up as much as I can without (gods willing) changing the logic of it at all. I've added doc comments to nearly everything, I've added a lot of fixmes and some todos marking stuff that needs to be changed or considered. I've renamed a bunch of variables to more descriptive names.  I've pulled boolean variables out of some of the worse if statements. I've made a minor change to the way error strings work and added a few ReportingUtilities functions in there.

I'm submitting this as its own PR because having these changes done separately means that the actual changes to function can be evaluated without all the mess of reformatting and cleanup going on around them at the same time.


HammerGS (approved):
Tested and seems to work

PR #5134, opened 2024-10-26
Refit - Fix some minor and easy things

WeaverThree:
**Requires #5133**

Make some quick fixes before I buckle down for more exstensive reftting of the refit system...

Fix some instance checks
Fix checking engine side crits
Fix default refurbish time to something sane
Resolve a few fixme/todos that I now understand
Set time and skill multipliers to CamOps values
Fix an accidental wrong newEntity reference and make a function static that should be.


HammerGS (approved):
Tested and seems to work

PR #5135, opened 2024-10-26
A few text joining functions for ReportingUtilities

WeaverThree:
Figured I'd break these out of my Refit rework.

For example, `surroundIf("(", maybeText, ")")` will return `"(<maybeText>)"` or `""` depending on the status of `maybeText`

`separateIf(text1, ", ", text2)` will return one of `"<text1>, <text2>"`, `"<text1>"`, `"<text2>"`, or `""` based on the status of the first and last parameters.

They're pretty simple, but can declutter a lot of the simpler cases of things where using a StringBuilder or StringJoiner seems like overkill.

The names are inspired by WinAmp title formatting marcos ^^;.


WeaverThree:
If you're reading this on the Discord and it looks confusing, Discord ate half of my exapmle strings >.>

HammerGS (approved):

PR #5156, opened 2024-10-31
Refactor Supply Depot SP Modifiers to be Monthly

IllianiCBT:
Changed the support point modifiers gained from Supply Depot facilities from weekly to monthly.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -658,10 +658,10 @@ public class StratconRulesManager {
      * Applies time-sensitive facility effects.
      */
     private static void processFacilityEffects(StratconTrackState track,
-            StratconCampaignState campaignState, boolean isMonday) {
+            StratconCampaignState campaignState, boolean isStartOfMonth) {
         for (StratconFacility facility : track.getFacilities().values()) {
-            if (isMonday) {
-                campaignState.addSupportPoints(facility.getWeeklySPModifier());
+            if (isStartOfMonth) {
+                campaignState.addSupportPoints(facility.getMonthlySPModifier());
             }
         }
     }
@@ -1903,6 +1903,7 @@ public class StratconRulesManager {
             return;
         }
         boolean isMonday = ev.getCampaign().getLocalDate().getDayOfWeek() == DayOfWeek.MONDAY;
+        boolean isStartOfMonth = ev.getCampaign().getLocalDate().getDayOfMonth() == 1;
 
         // run scenario generation routine for every track attached to an active
         // contract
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -658,10 +658,10 @@ private static void setupFacilityScenario(StratconScenario scenario, StratconFac
      * Applies time-sensitive facility effects.
      */
     private static void processFacilityEffects(StratconTrackState track,
-            StratconCampaignState campaignState, boolean isMonday) {
+            StratconCampaignState campaignState, boolean isStartOfMonth) {
```
Scoppio, line 661: I would prefer if this function received a date instead of a boolean, and the date itself had a function call that answers if it is the start of month.
This could enable some neat stuff, like scheduled events on facilities, etc.
Also, it would reduce the number of late changes, if we want to move from monthly back to weekly, or biweekly, it has to be changed in this funciton, and in the function that call it.

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -1903,6 +1903,7 @@ public void handleNewDay(NewDayEvent ev) {
             return;
         }
         boolean isMonday = ev.getCampaign().getLocalDate().getDayOfWeek() == DayOfWeek.MONDAY;
+        boolean isStartOfMonth = ev.getCampaign().getLocalDate().getDayOfMonth() == 1;
```
Scoppio, line 1906: If the functions were to receive the localdate and be made to take up whatever decision they can make, we would eliminate those magical booleans in the arguments of a few function calls, and would also not need to create variables like "isMonday" and "isStartOfMonth", because the functions that care about what day it is would be the ones to deal with that kind of trouble.
Since we are not doing those things ate "FPS" scale this should not be such a performance hit for it to worry us.

Scoppio (commented):
The code itself looks fine, just a couple of comments

Scoppio (approved):
LGTM, tested locally and is working fine, no apparent problems

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
IllianiCBT: That would be beyond the scope of this PR

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
IllianiCBT: If you want to make a function that does that, it would definitely see use, however checks like this are pretty standard in our codebase and major refactoring is beyond the scope of this PR.

PR #5157, opened 2024-10-31
Fixed StratCon Reinforcement Deployment Check

IllianiCBT:
Removed redundant check for victory points in deployment eligibility. Previously we removed the automatic exchange of CVP to SP when reinforcing a non-Fight Stance'd force. However, this check was missed so would display non-Fight forces even if the player has no SP. Now, those forces will only be visible if the player has SP adding visible value to the Fight stance.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -1527,8 +1527,7 @@ public class StratconRulesManager {
         }
 
         // otherwise, the force requires support points / vps to deploy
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -1527,8 +1527,7 @@ public static ReinforcementEligibilityType getReinforcementType(int forceID, Str
         }
 
         // otherwise, the force requires support points / vps to deploy
```
Scoppio, line 1529: Change in code make the commentary above it wrong, should be changed to match the code

HammerGS (approved):

PR #5158, opened 2024-11-01
Implemented Resupply Module

IllianiCBT:
This PR implements the new Resupply Module. For more information please see [here](https://megamek.org/development/stratcon/2024/10/31/StratCon-and-Logistics-The-Supply-Chain-Ahead.html)

This is a fairly complicated module and documentation will be provided prior to the launch of 50.02. However, for the benefit of QA I'm going to provide a brief summary.

Each month, while on contract, the player will be provided a resupply of items tailored to their TO&E. The size of the convoy's delivery is based on the average weight of combat units in the TO&E. Players can provide their own units in exchange for a bonus. For independent contracts players must supply their own units.

Convoys can be intercepted, with the player able to intercede should they wish it. Interception chance is based on force tonnage, with larger convoys easier for the enemy to detect; and enemy morale, with higher morale increasing the chance of an interception. Convoys can weigh up to 200t before suffering a detection malus, while lighter convoys will enjoy a bonus. Abandoning a player convoy will result in the convoy's automatic defeat. All units will be lost and all crew will be captured or killed.

Ground convoys can run into role-play events. I would like to extend these events to aerial convoys, but that will take a bunch of work and time is short.

On guerrilla contracts players cannot provide their own convoys, but instead are approached by smugglers. These resupplies cost a value based on the contents of the convoy with a chance the player will be scammed.

To define a force as a convoy the force must be selected in the TO&E and assigned as a supply convoy via the right-click menu. While any forces can be assigned as a convoy there are some limitations. All forces descended from the selected force will be considered a part of the convoy. It is not possible to tell MekHQ to ignore specific forces. It's an all or nothing. A force that has an ancestor that is classified as a 'convoy' cannot itself be made as a convoy. When you assign a convoy all ancestors and descendants will have their convoy status revoked.


repligator:
Error popped up. I'm guessing it's because I had my supply convoy assigned elsewhere at the end of the month.
[mekhq.log](https://github.com/user-attachments/files/17602519/mekhq.log)
[IS Logistics30680701.cpnx.gz](https://github.com/user-attachments/files/17602521/IS.Logistics30680701.cpnx.gz)
[Autosave-1-IS Logistics-30680628.cpnx.gz](https://github.com/user-attachments/files/17602522/Autosave-1-IS.Logistics-30680628.cpnx.gz)

IllianiCBT:
> Error popped up. I'm guessing it's because I had my supply convoy assigned elsewhere at the end of the month. [mekhq.log](https://github.com/user-attachments/files/17602519/mekhq.log) [IS Logistics30680701.cpnx.gz](https://github.com/user-attachments/files/17602521/IS.Logistics30680701.cpnx.gz) [Autosave-1-IS Logistics-30680628.cpnx.gz](https://github.com/user-attachments/files/17602522/Autosave-1-IS.Logistics-30680628.cpnx.gz)

Fixed. Thanks for the report :)

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -0,0 +1,1929 @@
...
+     * {@code HashMap}, thus ensuring that the pool remains independent of future changes to the
+     * potential parts {@code HashMap}.
+     */
+    private List<Part> buildPool(Map<String, PartDetails> potentialParts) {
+        List<Part> partsPool = new ArrayList<>();
+
+        for (PartDetails potentialPart : potentialParts.values()) {
+            // We need to use ceil here, otherwise well-stocked campaigns will end up with very
+            // mediocre Resupplies.
+            for (int entry = 0; entry < Math.ceil(potentialPart.getWeight()); entry++) {
```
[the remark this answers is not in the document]
IllianiCBT: Fixed

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3730,6 +3739,45 @@ && getLocation().getJumpPath().getLastSystem().getId().equals(contract.getSystem
...
+                                // assuming anyone who isn't dead has been captured.
+                                if (Compute.d6(2) > 7) {
+                                    status = PersonnelStatus.POW;
+                                }
+                                crewMember.changeStatus(this, currentDay, status);
+                            }
+                        }
+                        removeUnit(unitID);
+                    }
+                }
```
Scoppio, line 3776: This is a suggestion on how to use modern java with streams for data manipulation and reduce the number of branching operation in the CPU.

The following code was not tested and possibly has bug (like the simultaneos removal of units while iterate over its list, probably need to keep new ArrayList in there (commented out, and probably slightly wrong.

```java
    /**
     * Processes an abandoned convoy. The player is presented with a defeat dialog,
     * and if they have independent command rights, it checks each player template force in the scenario
     * for being a convoy force. If it is a convoy force, its units are treated as abandoned units.
     * Each crew member of these units is set as either KIA or POW based on a die roll.
     * It finally removes each unit from the campaign.
     *
     * @param contract The current {@link AtBContract}.
     * @param scenario The relevant {@link AtBDynamicScenario}.
     */
    private void processAbandonedConvoy(AtBContract contract, AtBDynamicScenario scenario) {
        convoyFinalMessageDialog(this, contract.getEmployerFaction());

        if (!contract.getCommandRights().isIndependent()) {
            return;
        }

        scenario.getPlayerTemplateForceIDs().stream()
            .map(this::getForce).filter(Objects::nonNull)
            .filter(Force::isConvoyForce)
            .map(Force::getUnits)
            // .flatMap(new::ArrayList) ??? maybe this is necessary here? I cant test this case now
            .map(this::decideCrewFate)
            .forEach(this::removeUnit);
    }
    
    /**
     * decideCrewFate
     * <p>
     *     Decide the fate of the crew of a unit that was abandoned in a time of need
     * </p>
     * @param unitID
     * @return
     *
     */
    private UUID decideCrewFate(UUID unitID) {
        Optional.of(getUnit(unitID)).ifPresent(unit -> unit.getCrew().forEach(this::decideCrewmemberFate));
        return unitID;
    }

    /**
     * decidePersonFate
     * <p>
     *      We're using the CamOps rules for infantry survival here and
     *      assuming anyone who isn't dead has been captured.
     * </p>
     * @param person The person to decide the fate of.
     */
    private void decideCrewmemberFate(Person person) {
        PersonnelStatus status = KIA;
        if (Compute.d6(2) > 7) {
            status = PersonnelStatus.POW;
        }
        person.changeStatus(this, currentDay, status);
    }
```

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3798,6 +3858,24 @@ private void processNewDayATB() {
...
+     * This method checks if the contract type is not Guerrilla Warfare or if a
+     * d6 roll is greater than 4. If any of these conditions is met, it calculates the maximum
+     * resupply size based on the contract's required lances, creates an instance of the
+     * {@link Resupply} class, and initiates a resupply action.
+     *
+     * @param contract The relevant {@link AtBContract}
+     */
+    private void processResupply(AtBContract contract) {
+        if (!contract.getContractType().isGuerrillaWarfare() || Compute.d6(1) > 4) {
+            int maximumResupplySize = (int) Math.max(1, Math.floor((double) contract.getRequiredLances() / 3));
```
Scoppio, line 3873: I think this line is a good candidate to be its own function, so we have a place that is easy to find and change the max resuply size.

MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -163,6 +165,31 @@ public void setCombatForce(boolean combatForce, boolean setForSubForces) {
...
+     * @param setForSubForces {@code true} to recursively apply convoyForce status to all subforces,
+     *                         {@code false} to only set for this force.
+     */
+    public void setConvoyForce(boolean convoyForce, boolean setForSubForces) {
+        this.convoyForce = convoyForce;
+        if (setForSubForces) {
+            for (Force force : subForces) {
+                force.setConvoyForce(convoyForce, true);
+            }
+        }
```
Scoppio, line 190:
```suggestion
        this.convoyForce = convoyForce;
        if (setForSubForces) {
            subForces.forEach(force -> force.setConvoyForce(convoyForce, true));
        }
```

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3670,6 +3673,12 @@ && getLocation().getJumpPath().getLastSystem().getId().equals(contract.getSystem
                         if (stub) {
                             scenario.convertToStub(this, ScenarioStatus.DEFEAT);
                             addReport("Failure to deploy for " + scenario.getName() + " resulted in defeat.");
+
+                            // I really don't like checking against a String here, but I couldn't find a way to
+                            // fetch the scenario's original template
+                            if (Objects.equals(scenario.getName(), "Emergency Convoy Defense")) {
```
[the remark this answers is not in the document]
IllianiCBT: This was a great idea, btw

IllianiCBT:
Flipping this to live. Description has been updated accordingly.

MekHQ/src/mekhq/campaign/market/procurement/Procurement.java
```diff
@@ -0,0 +1,366 @@
...
+
+        if (techEra < ERA_CLAN) {
+            return RATING_X;
+        }
+
+        availability++;
+
+        if (availability > RATING_F) {
+            return Compute.d6() > 3 ? RATING_F : RATING_X;
+        }
```
Scoppio, line 310: same as before about availability randomly changing

MekHQ/src/mekhq/campaign/market/procurement/Procurement.java
```diff
@@ -0,0 +1,366 @@
...
+            } else {
+                int extinctionYear = part.getExtinctionDate(true);
+
+                if ((techEra == ERA_SW)
+                    && (gameYear >= extinctionYear)) {
+                    return Compute.d6() > 3 ? RATING_F : RATING_X;
+                } else {
+                    return availability;
+                }
+            }
```
Scoppio, line 335: same

MekHQ/src/mekhq/campaign/market/procurement/Procurement.java
```diff
@@ -0,0 +1,366 @@
...
+            if (part.getBaseAvailability(techEra) >= RATING_E) {
+                if (availability == RATING_FSTAR || availability == RATING_X) {
+                    int extinctionYear = part.getExtinctionDate();
+
+                    if ((techEra == ERA_SW)
+                        && (gameYear >= extinctionYear)) {
+                        return Compute.d6() > 3 ? RATING_F : RATING_X;
+                    } else {
+                        return availability;
+                    }
```
Scoppio, line 359: same

MekHQ/src/mekhq/campaign/mission/Scenario.java
```diff
@@ -736,16 +736,12 @@ public boolean isTraitor(Entity en, Campaign c) {
         }
         // also make sure that the crew's external id does not match a traitor in
         // case of ejected pilots
-        if ((null != en.getCrew()) && !"-1".equals(en.getCrew().getExternalIdAsString()) &&
-                isTraitor(UUID.fromString(en.getCrew().getExternalIdAsString()))) {
-            return true;
-        }
-        return false;
+        return (null != en.getCrew()) && !"-1".equals(en.getCrew().getExternalIdAsString()) && isTraitor(UUID.fromString(en.getCrew().getExternalIdAsString()));
```
Scoppio, line 739: split this into individual variables, test each one independently and then do the final check which says if they are ... well, whatever this is supposed to say.

MekHQ/src/mekhq/campaign/mission/Scenario.java
```diff
@@ -1023,6 +1011,8 @@ public static Scenario generateInstanceFromXML(Node wn, Campaign c, Version vers
 
                 if (wn2.getNodeName().equalsIgnoreCase("name")) {
                     retVal.setName(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("scenarioType")) {
+                    retVal.setStratConScenarioType(ScenarioType.fromOrdinal(Integer.parseInt(wn2.getTextContent())));
```
Scoppio, line 1015: purely stylistic choice, I prefer to use the enum name instead of the ordinal value, as there is no meaningful change here and allow us to add more values in betwen enums, and is human readable.

MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
```diff
@@ -201,22 +204,15 @@ public void updateObjectiveEntityState(Entity entity, boolean forceEntityEscape,
...
-            }
+            return switch (objective.getObjectiveCriterion()) {
+                case Destroy -> entityIsDestroyed(entity, opponentHasBattlefieldControl);
+                case ForceWithdraw -> entityIsForcedWithdrawal(entity);
+                case Capture -> entityIsCaptured(entity, !opponentHasBattlefieldControl);
+                case PreventReachMapEdge -> !entityHasReachedDestinationEdge(entity, objective);
+                case Preserve -> !entityIsDestroyed(entity, opponentHasBattlefieldControl);
+                case ReachMapEdge -> entityHasReachedDestinationEdge(entity, objective);
+                default -> false;
+            };
```
Scoppio, line 215: 👍🏼

MekHQ/src/mekhq/campaign/mission/ScenarioObjectiveProcessor.java
```diff
@@ -341,12 +337,21 @@ public String processObjective(ScenarioObjective objective, int qualifyingUnitCo
     }
 
     /**
-     * Processes an individual objective effect.
-     * @param effect
-     * @param scaleFactor If it's scaled, how much to scale it by
-     * @param tracker
+     * This method processes individual objective effects based on their type.
```
Scoppio, line 340: what does it means "processes" in this context? It transform? return another value? derives some information?

MekHQ/src/mekhq/campaign/mission/enums/ScenarioType.java
```diff
@@ -0,0 +1,54 @@
...
+            if (scenarioType.ordinal() == ordinal) {
+                return scenarioType;
+            }
+        }
+
+        final MMLogger logger = MMLogger.create(ScenarioType.class);
+        logger.error(String.format("Unknown Scenario Type ordinal: %s - returning NONE.", ordinal));
+
+        return NONE;
+    }
```
Scoppio, line 53: make it throw an InvalidParameter or something like that instead, and have the function that is handling it implement an error handling that will force the NONE instead.
This way you avoid adding unnecessary things inside the enum (MMLogger) and you know exactly what it will do.

Or keep the NONE return as default but keep the knowledge of what it means to receive a NONE for scenario type in the function that is handling it. 
I prefer the exception tough, because this means the person HAS to implement something, also allows us to not have the NONE enum value and the user will have to either accept a null value in whatever they are implementing or they will have to add some new state in the enum.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -0,0 +1,2371 @@
...
+import static mekhq.campaign.mission.enums.AtBMoraleLevel.STALEMATE;
+import static mekhq.campaign.stratcon.StratconContractInitializer.getUnoccupiedCoords;
+import static mekhq.campaign.stratcon.StratconRulesManager.generateExternalScenario;
+import static mekhq.campaign.stratcon.StratconRulesManager.getRandomTrack;
+import static mekhq.campaign.unit.Unit.getRandomUnitQuality;
+import static mekhq.campaign.universe.Factions.getFactionLogo;
+import static mekhq.utilities.ReportingUtilities.CLOSING_SPAN_TAG;
+import static mekhq.utilities.ReportingUtilities.spanOpeningWithCustomColor;
+
+public class Resupply {
```
Scoppio, line 79: If possible, break most if not all the Panels/Jmenus/Etcs into their own classes.
Keep the data into a data object, lets call it "Ressuply", and have all the GUI be a class called RessuplyGUI.
The Ressuply GUI will interact with the Ressuply data through its interface, updating, increasing, reducing, etc. However, if I ever need to work with the "resupply" programatically I dont have to worry about a GUI poping out of nowhere.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -0,0 +1,2371 @@
...
+
+        if (negotiator != null) {
+            Skill skill = negotiator.getSkill(SkillType.S_NEG);
+
+            if (skill != null) {
+                int skillLevel = skill.getFinalSkillValue();
+                negotiatorSkill = skill.getType().getExperienceLevel(skillLevel);
+            }
+        }
+    }
```
Scoppio, line 265: Most of those tests can be moved to small functions that just return true or false, so this function becomes less indented.

Also, try to use Optionals more frequently in those situations, we already use null alot in the system, we may keep passing null forward, its fine, but optionals of things that can be null would more easily help work with some of the evaluations we have to do like in functions such as this.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -0,0 +1,2371 @@
...
+                    if (cargoCapacitySubTotal > 0) {
+                        totalPlayerCargoCapacity += cargoCapacitySubTotal;
+                        playerConvoys.put(force, cargoCapacitySubTotal);
+                    }
+                }
+            }
+        }
+    }
+
+    public void createPlayerConvoyOptionalDialog() {
```
Scoppio, line 2073: Really, this need to be split into many small classes, one for each individual part of it. The data has to be split from it too.
This is better done now than later, because later it will be too much of a pain to do properly, and its only going to grow more and more. Lets keep the new stuff neat and not repeat the errors of the legacy code.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -0,0 +1,2371 @@
...
+        int width = UIUtil.scaleForGUI(500);
+        return String.format("<html><i><div style='width: %s; text-align:center;'>%s</div></i></html>",
+            width, convoyMessage);
+    }
+}
+
+/**
+ * Class representing details about a part, including the part itself and its drop weight.
+ */
+class PartDetails {
```
Scoppio, line 2330: Maybe this could be a record... except by the setWeight, but we could create new PartDetails from old partDetails to have new part weights.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,643 @@
...
+    private Entity getEntity(MekSummary unitData) {
+        try {
+            return new MekFileParser(unitData.getSourceFile(), unitData.getEntryName()).getEntity();
+        } catch (Exception ex) {
+            logger.error("Unable to load entity: {}: {}",
+                unitData.getSourceFile(),
+                unitData.getEntryName(), ex);
+            return null;
+        }
+    }
```
Scoppio, line 245: We need to stop reusing the exact same code everywhere and just create an "EntityUtils" 🤣

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,643 @@
...
+        // Dialog dimensions and representative
+        final int DIALOG_WIDTH = 400;
+        final int DIALOG_HEIGHT = 200;
+
+        // Creates and sets up the dialog
+        JDialog dialog = new JDialog();
+        dialog.setTitle(resources.getString("dialog.title"));
+        dialog.setLayout(new BorderLayout());
+        dialog.setSize(UIUtil.scaleForGUI(DIALOG_WIDTH, DIALOG_HEIGHT));
+        dialog.setLocationRelativeTo(null);
```
Scoppio, line 408: If the GUIs and dialogs are just to allow users to interact with a very small a mount of things that can be interacted also through the API of the class I am more prone to be ok with it, but I still would prefer it to be a separated class which this one can instantiate and show to the user.

MekHQ/src/mekhq/campaign/stratcon/StratconCampaignState.java
```diff
@@ -176,6 +176,20 @@ public void useSupportPoint() {
...
     }
 
+    /**
+     * Decreases the number of support points by the specified increment.
+     *
+     * @param increment The number of support points to use/decrease.
+     */
+    public void useSupportPoints(int increment) {
+        supportPoints -= increment;
+    }
```
Scoppio, line 186: I guess thats a decrement

MekHQ/src/mekhq/campaign/stratcon/StratconCampaignState.java
```diff
@@ -176,6 +176,20 @@ public void useSupportPoint() {
...
+
+    public void convertVictoryToSupportPoint() {
+        victoryPoints--;
+        supportPoints++;
+    }
```
Scoppio, line 191: lacking check if VP are greater than zero.

MekHQ/src/mekhq/campaign/stratcon/StratconContractInitializer.java
```diff
@@ -224,6 +224,26 @@ public static void initializeCampaignState(AtBContract contract, Campaign campai
...
+//                    if (hiddenCache != null) {
+//                        logger.info(String.format("A secret cache has been spawned for contract %s",
+//                            contract.getName()));
+//                    }
+//                } else {
+//                    logger.error("'Chasing a Rumor' scenario failed to deserialize");
+//                }
+//            }
+        }
+
```
Scoppio, line 246: just to mark

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -2191,6 +2193,35 @@ public static void processScenarioCompletion(ResolveScenarioTracker tracker) {
...
+                                CacheType.TRASH_CACHE.ordinal());
+
+                            // The rumor is a dud
+//                            if (false) { // TODO replace placeholder value
+//                                cache.createDudDialog(track, scenario);
+//                            } else {
+//                                if (Objects.equals(cache.getFaction().getShortName(), "SL")) {
+//                                    cache.createProposalDialog();
+//                                }
+//                            }
```
Scoppio, line 2222: Just to mark

Scoppio (changes requested):
The quality is as always great. But there are some parts that require special attention, and at least two classes need a large refactor, alongside a few functions need to be extracted from where they are now.

IllianiCBT:
While I see this has been approved, it should not be merged until I've had the opportunity to review Luana's suggested changes.

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
IllianiCBT: You might have noticed a couple of these - they're related to the LosTech portion of this module that will be launched separately. We decided to split implementation in two mid-production, as this allows us to work on LosTech Caches while addressing bugs and feedback with the logistics portion of the module. I should have mentioned this in the PR description. mea culpa.

MekHQ/src/mekhq/campaign/stratcon/StratconCampaignState.java
IllianiCBT: I'm glad you spotted this, because that whole method should have been removed when merging in master.

MekHQ/src/mekhq/campaign/stratcon/StratconCampaignState.java
IllianiCBT: I do believe you would be correct, yes :D

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3847,6 +3854,57 @@ && getLocation().getJumpPath().getLastSystem().getId().equals(contract.getSystem
...
+                        }
+
+                        removeUnit(unitID);
+                    }
+                }
+            } catch (Exception ex) {
+                logger.warn(ex.getMessage(), ex);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
IllianiCBT: Wholeheartedly agree. The issue we have right now is that campaign.java is so all encompassing and touching so many different aspects, the likelihood that a new function will be related to something in campaign.java (and therefore logically placed _in_ that class) is fairly high.

Something I think we'd really benefit from is someone ripping campaign.java apart and democratizing a lot of the methods housed there.

MekHQ/src/mekhq/campaign/market/procurement/Procurement.java
```diff
@@ -0,0 +1,366 @@
...
+        int targetNumber =  switch (availability) {
+            case RATING_A -> 2;
+            case RATING_B -> 3;
+            case RATING_C -> 4;
+            case RATING_D -> 6;
+            case RATING_E -> 8;
+            case RATING_F -> 10;
+            // This value is deliberately impossible on 2d6
+            default -> 13; // X or F*
+        };
```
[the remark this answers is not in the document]
IllianiCBT: Thankfully no. In those cases we just change what is shown to the user (i.e. A is shown in the UI as F, or whatever); what's handled internally remains consistent.

MekHQ/src/mekhq/campaign/market/procurement/Procurement.java
```diff
@@ -0,0 +1,366 @@
...
+        if (part.isExtinct(gameYear, originFaction.isClan(), factionTechCode)) {
+            int extinctionYear = part.getExtinctionDate(originFaction.isClan(), factionTechCode);
+
+            if ((gameYear > extinctionYear) && useHardExtinction) {
+                return RATING_X;
+            }
+
+            if (gameYear < (extinctionYear + 10)) {
+                return Compute.d6() > 3 ? availability : RATING_X;
+            }
```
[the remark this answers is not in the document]
IllianiCBT: This is a part of the rules document I was provided, with a basis in RAW. Changing availability shouldn't cause any issues. The idea, as it was described in the document and during the meeting I had prior to creating this class, is that it's meant to reflect the fading availability of items that have become extinct. During that decade window specific parts that are listed as extinct may still be available, but not consistently and not easily.

MekHQ/src/mekhq/campaign/mission/Scenario.java
IllianiCBT: I'll need to double check this, but I think it might actually be pre-existing code or potentially a bad merge. I can't remember why I made this change, making me think it might not have been intentional. I'll review and correct accordingly. :)

MekHQ/src/mekhq/campaign/mission/Scenario.java
IllianiCBT: Yeah, I think that would be best, too. Originally I didn't use enum name because I couldn't work out a way to handle that without some really clunky code. Reading through your review popped the answer in my head, so I'll get that done.

MekHQ/src/mekhq/campaign/mission/enums/ScenarioType.java
IllianiCBT: I'm not sure I follow. Would you be able to expand on this point? :)

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
IllianiCBT: I concur, because this is obnoxious.

MekHQ/src/mekhq/campaign/Campaign.java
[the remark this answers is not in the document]
Scoppio: Campaign became one of two "façade" objects for the project, MHQ being the other one. It is fine, and it happens, however this also means that our efforts should be focused on every time we "touch" the campaing, instead of adding code to it, we take the code that we want to interact with and move to a new place. It is better if campaign become something like a "context", as if we had to communicate with a database and campaign were that database.

MekHQ/src/mekhq/campaign/mission/Scenario.java
Scoppio: this looks like just an aggregation/simplification of a condition check, the original was also very bad.

the "best" here would be to do something like this:

```java

var crewIsPresent = en.getCrew() != null
if (!crewIsPresent) {
  return false;
}
var crewExists = en.getCrew.getExternalIdAsString() != "-1";
var crewIsTraitor = isTraitor(UUID.fromString(en.getCrew().getExternalId()));
return crewExists && !crewIsTraitor;
```

PS.: I dont know if those names are correct or if thats the function of the method, this is just an example.

MekHQ/src/mekhq/campaign/mission/Scenario.java
Scoppio: Enum.valueOf is a default in Java and transform the `name` into the Enum.

MekHQ/src/mekhq/campaign/mission/enums/ScenarioType.java
Scoppio:
```java
    
    int ordinal = 12;
    ScenarioType ScenarioType = null;
    try {
        scenarioType = ScenarioType.values()[ordinal]
    }  catch (IndexOutOfBoundsException e) {
        scenarioType = ScenarioType.FOOBAR
        logger.warn("Unknown scenario type {}, valid scenario types are", ordinal, ScenarioType.values())
    }        
```

or you can do

```java
        // returns a default value if the ordinal is wrong.
        public static ScenarioType fromOrdinal(int ordinal, ScenarioType defaultValue) {
        for (ScenarioType scenarioType : values()) {
            if (scenarioType.ordinal() == ordinal) {
                return scenarioType;
            }
        }
        return defaultValue;
    }      
```

Scoppio: I prefer the first, because the function that needs that value should know what are the possible values it can have or handle, so it shuold also be the place where errors are dealt with, this is important mainly when we are dealing with backward compatibility, like trying to load a new type of scenario in an old version of the code, or an old file in a new code. The loader process has to be as lenient as possible with what it reads, and it should handle its own mistakes and errors.

IllianiCBT:
Flipping this to live.

The code backing this PR has just finished undergoing a massive refactor. In particular a lot of the functionality has be democratized into separate, smaller classes. A big push was made to move out the GUI stuff into its' own package, rather than having everything dumbed in the one class. Furthermore, in most cases, actual processes are no longer being handled inside of dialogs. This should make things easier to maintain, and edit as necessary.

Anyone reviewing this may spot some vestigial code relating to Star League Caches. Ignore that entirely, that's a 50.03 thing. Originally Caches and Resupplies were going to be launched together, but fate had other plans, so that portion has been disabled.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+                dialogDismissActionListener.actionPerformed(null);
+            }
+        });
+
+        // Prepares and adds the icon of the representative as a label
+        JLabel iconLabel = new JLabel();
+        iconLabel.setHorizontalAlignment(JLabel.CENTER);
+
+        ImageIcon speakerIcon = getSpeakerIcon(false);
+//        speakerIcon = scaleImageIconToWidth(speakerIcon, UIUtil.scaleForGUI(100));
```
Sleet01, line 423: If you have the spoons, please delete the commented-out lines.

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+        iconLabel.setIcon(speakerIcon);
+        dialog.add(iconLabel, BorderLayout.NORTH);
+
+        // Prepares and adds the description
+//        JLabel description = new JLabel(
```
Sleet01, line 428: Ditto

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+//            String.format("<html><div style='width: %s; text-align:center;'>%s</div></html>",
+//                UIUtil.scaleForGUI(DIALOG_WIDTH), getDudDialogText(track, stratconCoords)));
+//        description.setHorizontalAlignment(JLabel.CENTER);
+
+        JPanel descriptionPanel = new JPanel();
+        descriptionPanel.setBorder(BorderFactory.createTitledBorder(
+            String.format(resources.getString("dialogBorderTitle.text"), "PLACEHOLDER")));
+//        descriptionPanel.add(description);
```
Sleet01, line 436: Ditto

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+                dialogDismissActionListener.actionPerformed(null);
+            }
+        });
+
+        // Prepares and adds the icon of the representative as a label
+        JLabel iconLabel = new JLabel();
+        iconLabel.setHorizontalAlignment(JLabel.CENTER);
+
+        ImageIcon speakerIcon = getSpeakerIcon(true);
+//        speakerIcon = scaleImageIconToWidth(speakerIcon, UIUtil.scaleForGUI(100));
```
Sleet01, line 485: Ditto

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+        iconLabel.setIcon(speakerIcon);
+        dialog.add(iconLabel, BorderLayout.NORTH);
+
+        // Prepares and adds the description
```
Sleet01, line 489: Ditto

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+//        JLabel description = new JLabel(
+//            String.format("<html><div style='width: %s; text-align:center;'>%s</div></html>",
+//                UIUtil.scaleForGUI(DIALOG_WIDTH), getProposalText(proposal)));
+//        description.setHorizontalAlignment(JLabel.CENTER);
+
+        JPanel descriptionPanel = new JPanel();
+        descriptionPanel.setBorder(BorderFactory.createTitledBorder(
+            String.format(resources.getString("dialogBorderTitle.text"),
+                resources.getString("senderUnknown.text"))));
+//        descriptionPanel.add(description);
```
Sleet01, line 499: Ditto

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
```diff
@@ -0,0 +1,640 @@
...
+                dialogDismissActionListener.actionPerformed(null);
+            }
+        });
+
+        // Prepares and adds the icon of the representative as a label
+        JLabel iconLabel = new JLabel();
+        iconLabel.setHorizontalAlignment(JLabel.CENTER);
+
+        ImageIcon speakerIcon = getSpeakerIcon(true);
+//        speakerIcon = scaleImageIconToWidth(speakerIcon, UIUtil.scaleForGUI(100));
```
Sleet01, line 558: Ditto

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
Sleet01: Is this still the operating plan?

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/StarLeagueCache.java
IllianiCBT: So this class can be ignored in its entirety. It's initial work for a 50.03 feature. Resupplies and Caches got split up late in development, and it was easier to comment out the stuff that would cause problems than delete it and have to rewrite it later down the line

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
IllianiCBT: Yeah, absolutely. It’s for the star league caches I mentioned yesterday

PR #5159, opened 2024-11-01
CampaignGUI Tab reordering

Scoppio:
Screenshot 2024-11-01 at 01 11 17

Changes order of the tabs in CampaignGUI, interstellar map is now right beside the command center mostly due to "vibes".

I think a command center should have easy access to a map. Then we have the TO&E, Briefing Room and StratCon, this is the "war room" basically, then we have personnel and hangar. Warehouse moved after the repair bay due to ergonomics, its not so frequent that we "visit" the warehouse, and it has more to do with book keeping most of the time, checking how many tons we have left, if there is anything to sell, etc. Infirmary, Finances, and in the last tab MekLab.
I think that MekLab in the last position is reasonable since it is mostly an empty screen unless you are activelly changing something, and as soo
n as you accept a design it disappears from the mek lab again, so it should be fine staying out of the way when navigating the tabs.

This is, of course, open for comments and suggestions as always.
Lets just try to avoid bikeshed arguments and just come up to "a solution", it does not need to be the best one, but it needs to be one.

### DIFF
--- a/MekHQ/src/mekhq/gui/enums/MHQTabType.java
+++ b/MekHQ/src/mekhq/gui/enums/MHQTabType.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2022 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2017-2024 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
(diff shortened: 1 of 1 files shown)


IllianiCBT (approved):
LGTM

PR #5210, opened 2024-11-18
Update Rank Systems to Version 0.50.2-SNAPSHOT

IllianiCBT:
Incremented the rank systems version in ranks.xml from 0.50.1-SNAPSHOT to 0.50.2-SNAPSHOT. This stops IDEA from pestering us to update it any time we make a change in mhq.

### DIFF
--- a/MekHQ/userdata/data/universe/ranks.xml
+++ b/MekHQ/userdata/data/universe/ranks.xml
@@ -1,3 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<rankSystems version="0.50.2-SNAPSHOT">
+<rankSystems version="0.50.02-SNAPSHOT">
 </rankSystems>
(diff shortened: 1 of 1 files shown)


MekHQ/userdata/data/universe/ranks.xml
```diff
@@ -1,3 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<rankSystems version="0.50.1-SNAPSHOT">
+<rankSystems version="0.50.2-SNAPSHOT">
```
HammerGS, line 2: Shouldn't this be 50.02

IllianiCBT: So, yes, but not according to our versioning. I think Tex might need to take a look at that.

HammerGS: I just fixed it :)
image

IllianiCBT: Perfect! I'll get this PR updated when I'm home :)

PR #5223, opened 2024-11-19
Adjusted Command Rights Scenario Modifiers

IllianiCBT:
Adjusted the command rights modifiers based on player feedback.

Originally the intent was to roll these modifiers back to their 50.0 values, however that presented a couple of issues. As those modifiers use fixed unit counts we end up in a situation where we have to use magic numbers to guesstimate the power of those units when preparing the contract difficulty estimate. It also fails to scale with the player force, leading to an undesirable scenario where the player might have a tiny force (say, a couple of Warrior VTOL) and the allied princess drops two 100t assault 'meks.

Instead, I've opted to keep things scaled by BV, but adjusted the BV multiplier to reduce the overall size of engagements and to reduce the players' reliance on the performance of their allied Princess (the two chief causes of concern).

Independent: No Allies -> _no change_
Liaison: 1x Heavy or Assault 'Mek -> _no change_
Integrated: 200% Allies -> 50% Allies
House: 100% Allies -> 25% Allies

This means a 7,500 GBV player force can expect 3,750 GBV of allies on Integrated contracts, or 1,875 of allies on House contracts. This should amount to a couple of extra units, similar to the fixed values found in <50.01. Note that with the current set up it is possible for the allied force to exceed this target value by around 25%. We have a solution in the works, but it's a really big project so will come later.

### DIFF
--- a/MekHQ/data/scenariomodifiers/HouseOfficerAir.xml
+++ b/MekHQ/data/scenariomodifiers/HouseOfficerAir.xml
@@ -1,10 +1,10 @@
 <AtBScenarioModifier>
-    <additionalBriefingText>Your employer is dispatching a force to assist with this objective.</additionalBriefingText>
+    <additionalBriefingText>Your force will be deployed under the command of an allied commander.</additionalBriefingText>
     <benefitsPlayer>true</benefitsPlayer>
     <eventTiming>PreForceGeneration</eventTiming>
     <forceDefinition>
         <actualDeploymentZone>-1</actualDeploymentZone>
-        <allowAeroBombs>true</allowAeroBombs>
+        <allowAeroBombs>false</allowAeroBombs>
         <allowedUnitType>9</allowedUnitType>
         <arrivalTurn>0</arrivalTurn>
         <canReinforceLinked>false</canReinforceLinked>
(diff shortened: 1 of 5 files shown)


Setsul:
The ally weights in the contract difficulty estimate will need to be adjusted to match this.

IllianiCBT:
Ack! I made a mental note about that yesterday and then completely forgot it by this morning. Thanks for the catch. Any recommended values?

IllianiCBT:
Actually, I know exactly what I'll do

IllianiCBT:
The scope of this PR has changed. The title & description has been updated accordingly.

Setsul:
I see what you're going for.
You might still want to set the estimate ratios slightly higher than the guaranteed allied forces', since there are random scenario modifier that we still want to factor into the difficulty estimate a bit.
Liaison also probably also has to be adjusted downwards a bit, since we don't want that to have more of an effect on the difficulty estimate than House/Integrated.
Though to be honest, Liaison is highly dependent on the player's average number of unit and GBV per lance compared to a single unit within a certain weight range, so it's all guesswork without making that part of the method much more complex.

IllianiCBT:
Liaison is a big thorn in our side, one that I'm not entirely sure how to resolve cheaply.

My current thinking is that the difficulty estimate is 'good enough'. If you want to take another look at the values, though, I definitely wouldn't turn down the help. Especially as you seem to have a better grasp of maths than me :D

Setsul:
We're actually doing most of the work to do it properly already, it'll just be awkward to get the data since it's done in a different method. We can fairly easily figure out the average BV per player lance. What would need to be added is the logic get the average BV of an allied heavy/assault mek, but that's still easy enough.

I think the end goal should be to overhaul the difficulty estimate and make it factor in vehicles as well, but either way that's beyond the scope here.

Again, I'd call it good enough for now and look into again after scenario modifiers are settled a bit more. For a better estimate I'll also have to look into the chances of random modifiers that add allied forces.

Setsul (changes requested):
I think you forgot to adjust the actual ratios and only changed the comments.

IllianiCBT:
Ugh, no, I'm just clearly one coffee short of a braincell this morning. Fixing

HammerGS (approved):

Issue #5181, opened 2024-11-08
Manually Generated New Contract Throws Arithmetic Error


jschmetzer:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MekHQ documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

When creating a new contract using the Add Mission button , and all fields are populated, creating the contract throws an arithmetic error. 

### Steps to Reproduce

1. Download and unpack latest Nightly
2. Start a new Campaign with the complete Stratcon preset
3. Create a new contract mission; fill in any not-generated fields
4. Press Ok

### Operating System *

Mac

### Java Version *

17.0.12

### MekHQ Suite Version *

Free Text (type manually)

### Custom MekHQ Version

50.01

### Attach Files

[mekhq.log.zip](https://github.com/user-attachments/files/17683353/mekhq.log.zip)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

repligator:
```
14:34:19,175 ERROR [mekhq.MekHQ] {AWT-EventQueue-0}
mekhq.MekHQ.lambda$main$1(MekHQ.java:291) - Uncaught java.lang.ArithmeticException detected. Please open up an issue containing all logs, the game save file, and customs at https://github.com/MegaMek/megamek/issues . If Sentry is enabled, the error has also been logged.
java.lang.ArithmeticException: / by zero
	at mekhq.campaign.stratcon.StratconContractInitializer.trackObjectDistribution(StratconContractInitializer.java:247)
	at mekhq.campaign.stratcon.StratconContractInitializer.initializeCampaignState(StratconContractInitializer.java:107)
	at mekhq.gui.dialog.NewAtBContractDialog.btnOKActionPerformed(NewAtBContractDialog.java:549)
```

[...]

PR #5232, opened 2024-11-21
Generate required lances for manual AtB contracts for StratCon init

jslater89:
StratCon initialization requires a nonzero requiredLances to generate a nonzero number of tracks, and zero tracks causes a division by zero further along in initialization.

Solves issue #5181.


IllianiCBT (approved):
I thought we fixed this? Clearly not :D

Issue #5200, opened 2024-11-15
Stratcon allows deployment but not use of non-combat forces


Tempest-56:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MekHQ documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

In Stratcon, when you deploy forces it gives you the option to deploy forces that are marked as non-combat on your TOE. Should the enter a combat scenario (such as an encounter when revealing an unreconned hex), the scenario will generate with the force as the seed and assigned to it. However as a non-combat force you cannot actually start the scenario as it doesn't count them as deployed. Similarly, you cannot assign a different unit to it without consuming a support point for them as reinforcements, nor can you advance the day as the incomplete scenario nag pops up.

### Operating System *

Windows 11

### Java Version *

17

### MekHQ Suite Version *

v0.50.0

### Custom MekHQ Version

Snapshot as of 11/1

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

IllianiCBT:
For anyone picking this up: we should be filtering out non-combat forces from this dialog.

Tempest-56:
Oops. Apologies, I thought I _had_ attached a save. Corrected and attached.
[Alley Cats30161229.cpnx.gz](https://github.com/user-attachments/files/17807609/Alley.Cats30161229.cpnx.gz)

IllianiCBT:
Closing as resolved

PR #5250, opened 2024-11-25
Renamed 'Lances' to 'Strategic Formations', Expanded Functionality

IllianiCBT:
**Key Note:** `Lances` are what MHQ called any combat `Force`, i.e. any combat node in your TO&E that contains at least one ground `Unit`. When reading the below any instance of 'lance' is referring to the internal class and _not_ the IS Formation.

To reduce confusion `Lance.java` has been renamed to `StrategicFormation.java`. The relevant methods and classes have also been renamed.

A StratFor is considered to be any combat force in the TO&E that contains at least one ground unit and has no ancestor or descendant nodes which are also StratFors. Players can manually define individual forces to always be considered StratFors, or never considered StratFors. These overrides can be easily removed. Assigning or removing an override is accessed via the TO&E right-click menu.

StratFors are easily recognized by their bold nameplate. StratFors that have had their state overridden are also underlined.

A force that has had its StratFor status set to `true` will consider all units in that force, and all descendant forces, to be a part of that StratFor. This allows for deployment of entire companies at will (or greater).

The Briefing Room has been updated to consider units in descendant forces, when determining force weight. Weight is derived from a mean of all individual forces in the StratFor, ignoring empty forces.

StratCon and AtB will consider any StratFor fair game when picking seed forces, so players should be advised to try and keep their forces roughly equal, where possible.

StratCon will no longer consider non-combat forces as fair game for assignment to a track (manual or via integrated command rights), or scenario seeding.

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -18,46 +18,32 @@
  */
 package mekhq;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.UUID;
-
-import javax.swing.JOptionPane;
-
 import io.sentry.Sentry;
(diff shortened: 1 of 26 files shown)


IllianiCBT:
### Closes #5200

IllianiCBT:
Why do I keep doing that: reopening.

HammerGS (approved):
Limited testing and GitHub Co-Pilot didn't identify any issues.

PR #5253, opened 2024-11-26
allows to select a previous engineer or new crew member to activate large vessels

### LINKED ISSUE #4529: Can't Activate Dropship
### Environment
Java 17
Linux
Master (latest PR https://github.com/MegaMek/megamek/pull/5830)

### Description
I mothballed a dropship. Several months later, I tried to activate it. It shows as 'Activating (3840m)', but as soon as I advance day, it goes right back to mothballed. GM activate does work as intended.

### Files
[mekhq.log](https://github.com/user-attachments/files/16445069/mekhq.log)
[dropship-Fronc Marshals30800610.cpnx.gz](https://github.com/user-attachments/files/16445073/dropship-Fronc.Marshals30800610.cpnx.gz)

Scoppio:
Because we have the concept of "engineer" inside MM to handle the self-healing properties of large vessels, funcitonalities like "mothball" and "activate" are compromised due to them requiring a tech.

This fix does two things:

- fix #4529, it does that by adding the previous "engineer" to the crew of the mothballed vessel.
- Adds `vessel techs` to the selection of techs in the "select a techie" drop down window when you want to activate a unit.

The later was necessary to deal with a rare but possible situation. If the engineer that originally mothballed the large vessel (present in the mothballInfo object) is not active, or were removed/sacked by any reason then they would not be available to activate the ship, this means that the ship would be locked in a situation that could only be circunvented by GM activating.

PS.: I hope the second change does not cause any trouble

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -3753,8 +3753,13 @@ public class Person {
     }
 
     public boolean canTech(final Entity entity) {
+        if (entity == null) {
+            return false;
+        }
         if ((entity instanceof Mek) || (entity instanceof ProtoMek)) {
             return hasSkill(SkillType.S_TECH_MEK);
+        } else if (entity instanceof Dropship || entity instanceof Jumpship) {
+            return hasSkill(SkillType.S_TECH_VESSEL);
         } else if (entity instanceof Aero) {
             return hasSkill(SkillType.S_TECH_AERO);
         } else if (entity instanceof BattleArmor) {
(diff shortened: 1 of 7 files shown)


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -3901,6 +3916,19 @@ public boolean isTech() {
         return isTechMek() || isTechAero() || isTechMechanic() || isTechBA();
     }
 
+    /**
+     * Checks if the person is a tech, includes mektek, mechanic, aerotek, BAtek and the non-cannon "large vessel tek"
+     * @return true if the person is a tech
+     */
+    public boolean isTechExpanded() {
```
psikomonkie, line 3923: Are there other functions than maintaining a mothballed thing, or a more descriptive name for this group that could be used for a method name?

Scoppio: its exclusivelly used by the function `pickTechForMothballOrActivation`

I can come up with a better name, but `expanded` does cover pretty well the intent, which brings `techs` and an expansion on the variety.

HammerGS (approved):

PR #5257, opened 2024-11-28
Add Persistent Initiative Bonus to campaign

### LINKED ISSUE #4600: [RFE] CamOps Initiative Modifier
CamOps includes a lot of modifiers applied to initiative rolls made during scenarios. These can result from using a WMD or poor morale, to use the two examples that jump immediately to mind.

Due to the interactions needed with MegaMek, this falls outside my immediate skillset. So I'm putting out this RFE in the hope someone else is willing to take a shot.

### Requirements
- An initiative modifier `int` saved in `campaign` with setter and getter, and a third method that allows you to increment/decrement the modifier. This will need to be marshaled and unmarshaled to xml.
- The ability to pass that initiative modifier into a scenario and have it apply to initiative rolls made for the duration of that scenario

Implementing the various CamOps systems that would take advantage of this functionality I will be doing myself, so no need to worry about that side of things.

DM0000:
Added initiative bonus to campaign and passes it to megaMek's constantInitBonus.

Currently doesn't have bonus/negative ceiling, but can implement something along those lines if needed.  
The initiativeBonusIncrement function takes a boolean-->  true increases by 1, false decrease by one.

closes #4600

In future could tie bonus to fatigue/morale module

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -132,6 +132,7 @@ public class AtBGameThread extends GameThread {
 
                 client.getLocalPlayer().setCamouflage(app.getCampaign().getCamouflage().clone());
                 client.getLocalPlayer().setColour(app.getCampaign().getColour());
+                client.getLocalPlayer().setConstantInitBonus(campaign.getInitiativeBonus());
 
                 if (started) {
                     client.getGame().getOptions().loadOptions();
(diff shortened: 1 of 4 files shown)


IllianiCBT:
You’re a star, thank you!

mill get this reviewed later tonight, if nobody else does in the meanwhile.

HammerGS (approved):

IllianiCBT:
Saklad raised a question, here: https://github.com/MegaMek/mekhq/pull/5250

PR #5263, opened 2024-11-30
Refactored Strategic Formation Weight Categories

IllianiCBT:
Refactored the manner in which we calculate Strategic Formation weights (Heavy, Light, etc). Previously we inherited the old Legacy AtB weight categories which considered any `Force` with a weight greater than 390 to be a Super Heavy force. This older method also failed to properly account for non-IS campaigns.

Now weight categories are correctly tied to unit weight categories, with faction exceptions applied where appropriate.

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/StrategicFormation.java
+++ b/MekHQ/src/mekhq/campaign/force/StrategicFormation.java
@@ -2,7 +2,7 @@
  * Lance.java
  *
  * Copyright (c) 2011 - Carl Spain. All rights reserved.
- * Copyright (c) 2020 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2020-2024 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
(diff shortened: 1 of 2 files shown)


RaozSpaz:
For players set as Mercenary, will there be something they can set to define lance size or what faction base method they are using?

IllianiCBT:
No plans at this time - that behavior has not been changed from the prior implementation.

MekHQ/src/mekhq/campaign/force/StrategicFormation.java
```diff
@@ -268,8 +283,8 @@ public boolean isEligible(Campaign campaign) {
          */
         if (campaign.getCampaignOptions().isLimitLanceNumUnits()) {
             int size = getSize(campaign);
-            if (size < getStdLanceSize(campaign.getFaction()) - 1 ||
-                    size > getStdLanceSize(campaign.getFaction()) + 2) {
+            if (size < getStandardForceSize(campaign.getFaction()) - 1 ||
+                    size > getStandardForceSize(campaign.getFaction()) + 2) {
```
Scoppio, line 287: Some magical number that would be nice to get rid of in the future.

MekHQ/src/mekhq/campaign/force/StrategicFormation.java
```diff
@@ -566,35 +581,37 @@ public static StrategicFormation generateInstanceFromXML(Node wn) {
...
      * @return Total force weight
      */
     public static double calculateTotalWeight(Campaign campaign, int forceId) {
         double weight = 0.0;
 
         for (UUID id : campaign.getForce(forceId).getUnits()) {
-            Unit unit = campaign.getUnit(id);
-            if (null != unit) {
+            try {
+                Unit unit = campaign.getUnit(id);
```
Scoppio, line 592: Shorter and neater.

I usually prefer to have this kind of try-catch-continue to use be changed by a `getOptional`, so you let the place that is closes to the error deal with the error (in this case a getUnitOptional inside campaign would throw the error, capture it, log it, and return an optional empty)

But its just preference.

Scoppio (approved):

HammerGS (approved):

PR #5271, opened 2024-12-02
ACS-Like autoresolve [WIP]

Scoppio:
ACS-like autoresolve
---

This autoresolve implements a mechanism that is inspired in the Abstract Combat System to play out the result of combats instantaneously.

## How it works

It uses alot of the work created by @SJuliez for SBF, like the action handlers, phase managers, mostly "dumb" units (so only handlers and services can do things), it also implements a fairly generic memory to units, so they can remember actions they took in their turn, attacks they suffered, etc.

It first redistributes all the forces into ACS Valid force formations, then it creates each formation and links up the AlphaStrike representation of each Entity with its original entity counterpart.

It also reloads every entity as new references, and recreates all crews as new references, otherwise they will stay attached to the original units on the Campaign, which will cause damage and other changes to propagate to the units even if you cancel the resolution.

After all formations are set, a game starts, it uses the initiative, deployment, movement, firing and end phases. Movement phase is used for Engagement and Control which currently is kind of a "thermometer", it mostly change the balance of the game betwen rounds so you dont have a single formation steamrolling everything always.

During firing phase the units hit random units inside enemy formations, as in the quick/simple ACS rules for damage designation.

During the endphase units have their morale tested as per SBF rules (maybe I should change for ACS rules here), they are also checked for withdraw (there is a low chance for a formation to withdraw). 
All units that leave the game take some amount of damage.

## How to use it

There is a toggle implemented in the campaign options, where you can select if you want your autoresolve to be played by bots (selected as default) or not. With the option unchecked it will use the ACS-like autoresolve instead of opening MM with bots. Now just click in the Autoresolve


IllianiCBT:
I am eager to take a look at this, but it might not be until the end of the week as I really need to maximize my CO IIC time once thanksgiving guests leave, tomorrow.

Is there anything specific you want me to take a look at?

Scoppio:
> I am eager to take a look at this, but it might not be until the end of the week as I really need to maximize my CO IIC time once thanksgiving guests leave, tomorrow.
> 
> 
> 
> Is there anything specific you want me to take a look at?

Yes, the force creation/setup process, if there is anything that I should take into consideration but isnt, then the post game resole too. The Damage handlers so you can see if it lacks anything you wanted.

Scoppio:
CodeQL is pointing false positives for errors that font exist.

MekHQ/mmconf/defaultKeyBinds.xml
```diff
@@ -14,139 +14,139 @@ shift, etc).
...
-    <isRepeatable>false</isRepeatable>
-  </KeyBind>
-
-  <KeyBind>
-    <command>centerOnSelected</command>
-    <keyCode>32</keyCode> <!-- Spacebar -->
-    <modifier>0</modifier>
-    <isRepeatable>0</isRepeatable>
-  </KeyBind>
+             xsi:noNamespaceSchemaLocation="keyBindingSchema.xsl">
```
IllianiCBT, line 17: I know xml can sometimes get grumpy when we change whitespaces. Just wanting to verify the changes here won't cause any issues?

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -516,6 +527,90 @@ public void gamePhaseChange(GamePhaseChangeEvent e) {
         // Why Empty?
     }
 
+    // TODO: LUANA
```
IllianiCBT, line 530: I know this is still WIP, so just highlighting this TODO in case it got missed.

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -516,6 +527,90 @@ public void gamePhaseChange(GamePhaseChangeEvent e) {
...
+            resolveDialog.setVisible(true);
+            if (resolveDialog.wasAborted()) {
+                for (UUID personId : tracker.getPeopleStatus().keySet()) {
+                    Person person = getCampaign().getPerson(personId);
+                    // person.setStatus(PersonnelStatus.ACTIVE);
+                    person.setHits(person.getHitsPrior());
+                }
+
+                // should I do something here? Maybe do the trigger?
+                // MekHQ.triggerEvent(new ScenarioResolvedEvent(currentScenario));
```
IllianiCBT, line 555: I _think_ we want the trigger here, so it can be picked up by subscribers - though I don't know any that listen for this event, it's unlikely to hurt anything.

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -516,6 +527,90 @@ public void gamePhaseChange(GamePhaseChangeEvent e) {
...
+                return;
+            }
+
```
IllianiCBT, line 558: There seems to be a lot of duplicated code here, is there value in grabbing the original code and partitioning them into discrete methods and calling those, instead?

MekHQ/src/mekhq/campaign/ai/utility/Memory.java
```diff
@@ -0,0 +1,67 @@
+package mekhq.campaign.ai.utility;
+
+import java.util.*;
+import java.util.function.Function;
+
+public class Memory {
```
IllianiCBT, line 6: We definitely need a JavaDoc on this class as it's not immediately obvious what it does.

MekHQ/src/mekhq/campaign/ai/utility/HasMemory.java
```diff
@@ -0,0 +1,7 @@
+package mekhq.campaign.ai.utility;
```
IllianiCBT, line 1: Don't forget the legal notice on your new classes. You're missing them in a couple of places. I know this is still WIP, so you might not have got around to them yet, but figured it was worth mentioning just in case.

MekHQ/src/mekhq/campaign/autoResolve/AutoResolveGame.java
```diff
@@ -0,0 +1,710 @@
...
+     *
+     * @param currentFormationID the start point of the formation search. Need not
+     *                           match an actual formation
+     * @param phase              the phase to check
+     * @param direction          the selection to seek the next or previous
+     *                           formation
+     * @return the formation that precedes or follows the given formation ID, if one
+     *         can be found
+     */
+    private Optional<AcsFormation> getEligibleFormationImpl(int currentFormationID, GamePhase phase,
```
IllianiCBT, line 602: Seeing QL picking up a couple of 'useless params' - false positive, or something we can look at trimming?

MekHQ/src/mekhq/campaign/autoResolve/damageHandler/CrewMustSurvive.java
```diff
@@ -0,0 +1,5 @@
+package mekhq.campaign.autoResolve.damageHandler;
```
IllianiCBT, line 1: Does this need to be a class, or can we attach it as an enum to one of your other classes?

MekHQ/src/mekhq/campaign/autoResolve/damageHandler/EntityMustSurvive.java
```diff
@@ -0,0 +1,6 @@
+package mekhq.campaign.autoResolve.damageHandler;
```
IllianiCBT, line 1: Similar to my last comment, does this need to be a separate class? Can we merge the two enums, seeing that they have the same entries with identical ords

MekHQ/src/mekhq/campaign/autoResolve/helper/RandomUtils.java
```diff
@@ -0,0 +1,187 @@
+package mekhq.campaign.autoResolve.helper;
+
+import org.apache.commons.lang3.stream.Streams;
+
+import java.util.*;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class RandomUtils {
```
IllianiCBT, line 10: We could do with some JavaDocs for this class, and/or methods. It might also be worth taking a look at `ObjectUtility` to see whether there is value in merging the two classes. I would have mentioned this back when you first suggested a random workhorse method, but I only discovered it existed earlier today.

MekHQ/src/mekhq/campaign/autoResolve/helper/SetupTeams.java
```diff
@@ -0,0 +1,144 @@
...
+        ).toList());
+    }
+
+    static private AutoResolveForce playerForce(int playerID, String forceName, List<Unit> units) {
+        return new AutoResolveForce(1, forceName, playerID, units.stream().map(SetupTeams::breakEntityRef).filter(Objects::nonNull).toList());
+    }
+
+    @Nullable
+    public static Entity breakEntityRef(Unit unit) {
+        var entity = ASConverter.getUndamagedEntity(unit.getEntity());
```
IllianiCBT, line 50: While it's not marked as such `getEntity()` can return `null`. Something to be aware of as I _believe_ if that occurs here it will result in an NPE.

MekHQ/src/mekhq/campaign/autoResolve/helper/UnitStrengthHelper.java
```diff
@@ -0,0 +1,23 @@
...
+
+import megamek.common.*;
+import mekhq.campaign.autoResolve.scenarioResolver.components.UnitStrength;
+
+public class UnitStrengthHelper {
+
+    static public UnitStrength getUnitStrength(Entity entity) {
+        // The values here are completely arbitrary and should be adjusted as needed, probably should be in a config file.
+        if (entity instanceof Mek) {
+            return new UnitStrength(entity.calculateBattleValue(), +1, entity);
```
IllianiCBT, line 11: It might be worth taking a look at `entity.getGenericBattleValue()` - GBV is static per unit type & weight, so you might be able to do some math to derive inter-unit type power modifier from that. I don't think the arbitrary values are inherently wrong, mind, just wanted to give you options. :)

MekHQ/src/mekhq/campaign/autoResolve/scenarioResolver/abstractCombatSystem/AcsSimpleScenarioResolver.java
```diff
@@ -0,0 +1,90 @@
+/*
```
IllianiCBT, line 1: It's 02:30 local, so I'm going to call it here. Leaving this message so I know where I was up to.

IllianiCBT (commented):
Everything I've seen has me very excited to see this implemented. This is only a partial review as this is hardily a small PR and I need sleep. :D

PR #5273, opened 2024-12-02
Implemented Advanced StratCon Reinforcements

### LINKED ISSUE #4733: RFE; Admin and Reinforcements, conversation archived for @Illiani
Thom293 — 08/26/2024 9:21 PM
@Illiani meant to reply to this earlier.  Perhaps an option to start a contract with a Single SP (not VP) might be a worthwhile idea?  The start of contract is the most deadly time.  having an optional single SP to starts (some?) contracts for emergency reinforcements might help some.

Algebro — 08/26/2024 9:23 PM
I've noticed those first couple scenarios without SP are savage

Algebro — 08/26/2024 9:27 PM
Maybe starting sp based on something the player does, like strategy skill level or something

Illiani — 08/26/2024 9:27 PM
I was thinking the same

Algebro — 08/26/2024 9:28 PM
Would be good to have command admins have a use in stratcon outside of negotiating contracts

Gryphon — 08/26/2024 9:29 PM
could possibly have it use the strategy skill OF admin personel to give them more things to spend XP on

Illiani — 08/26/2024 9:29 PM
There is definitely mileage here

Algebro — 08/26/2024 9:30 PM
Transport admins could help with reducing deployment times or something (like tracks where deploying to hexes takes 3 turns or whatever)
I think it's command and transport that don't do anything

Illiani — 08/26/2024 9:31 PM
I believe lance leader tactics influences reinforcement times atm
That or Strategy
Illiani — 08/26/2024 9:31 PM
But I don’t see being able to double-up on reinforcement reduction to be a wholly bad thing
I would really like to see Command Admin get more use

Algebro — 08/26/2024 9:32 PM
That's reinforcing
Some tracks take multiple days to scout hexes and come back or whatever

Illiani — 08/26/2024 9:33 PM
Yes, I see what you mean now
Admin/Transport would be great for that

Illiani — 08/26/2024 9:33 PM
So Admin/Command for starting SP; Transport for deployment/return times
Logistics I would like to have it so that only logistics admins get to roll for acquisition (or Techs, if that’s enabled). That’s a change I’ve had planned for a while

Monky — 08/26/2024 9:37 PM
nice
HR for recruitment of prisoners/thwarting contract breach by employer?

Illiani — 08/26/2024 9:38 PM
I have secret plans for HR

Monky — 08/26/2024 9:38 PM
or dealing with 'disloyalty' in a professional productive way

Gryphon — 08/26/2024 9:39 PM
yeah, I use scrounge and put all my people with scrounge into admin, logistics, but that's just me making the system make more sense to me.  It would be good for logistics to handle aquisitions officially

Illiani — 08/26/2024 9:40 PM
Scrounge might not be long for this world. I want to do a sweep of our skills and bring them up to current rules. There are a couple in there from the old MechWarrior rpg from the 80s that have been since replaced

RaozSpaz — 08/26/2024 10:33 PM
Replying to the reinforcement topic of starting with SP. Sounds like a not terrible idea. Typically you can always reinforce, but the chance factors attached to reinforcing without SP are just really bad [not to mention make 0 sense].

RaozSpaz — 08/26/2024 10:45 PM
Yeah. I will never understand the random chance for more bad stuff when you do a non SP reinforce. I understand if it was a chance to fail [they arent ready in time or whatever] but causing the player more problems. How? Why?

Algebro — 08/26/2024 10:54 PM
I like how it forces you to decide whether to risk making things worse but it doesn't make a lot of sense logically 
If reinforcing without sp just had a chance to not work then you wouldn't really have an incentive to not just reinforce constantly

Algebro — 08/26/2024 10:58 PM
I don't think it necessarily needs to stay how it is but I do think there should be risk or some other cost, at least that has made for some really interesting decisions in my contracts 
I normally used to never make my scout lances try and win without reinforcements but now ive had to a couple times

Thom293 — 08/26/2024 11:00 PM
Best feeling  in the world to win an outmatched battle with just a scout lance.

RaozSpaz — 08/26/2024 11:32 PM
I mean there are plenty of risk factors that aren't "Suddenly the opfor got bigger somehow"
Intercepted - Generate another battle using the reinforcing lance
Sidetracked - Greatly delay the reinforcing time for the unit
Mishap - Your reinforcement group deploys onto the region like normal, but fails to properly reinforce [will return after the deployment window like normal]
Bait and Switch - Your OpFor has flipped the engagement onto your reinforcements. The contract resets with your reinforcing lance being the primary lance, and all other assigned units now reinforcing it instead.
Logistical Failure - Your reinforcements fail and all other reinforcements are removed from the engagement.

Obviously some of these would be ridiculous to code, but some examples of 'non larger opfor' penalties.

Gryphon — 08/26/2024 11:44 PM
Re: intercepted.  Isn't there a battle type that amounts to "get across this battlefield with minimal losses within X timeframe"?

RaozSpaz — 08/26/2024 11:45 PM
Yeah. Breakthrough
If we had the ability to do lead up missions would be a great option
But we dont have that quite yet

Illiani — 08/26/2024 11:48 PM
The random extra modifiers are meant to represent the OpFor detecting your reinforcements and sending reinforcements of their own

Illiani — 08/26/2024 11:49 PM
I do like the idea of reinforcements getting intercepted on a failed roll, though. I think that would be really cool
I do need to learn how to initiate scenarios from within mhq…

RaozSpaz — 08/26/2024 11:50 PM
The prequel to learning how to do chain scenarios 😄

Illiani — 08/26/2024 11:50 PM
Mutinies, actually

Illiani — Yesterday at 12:00 AM
If I were to implement it, I think there would be four possible outcomes to reinforcing without SP:

reinforcement successful
delayed: successful, but with delayed arrival time
unsuccessful: the force is unavailable as if they had participated to the battle, but don’t arrive in time to do any actual fighting
interception: a new battle is spawned with the reinforcement force assigned to it
The easiest way to implement ‘unsuccessful’ would be to just have it identical to delayed, but increase the arrival time to something like 30 turns. So potentially they could still arrive just not any time soon
Honestly, the only difficulty here would be interception and that’s not impossible, I just don’t personally know how to do it yet

IllianiCBT:
### Introduction
The previous reinforcement implementation created some balancing concerns. The gameplay loop would be this: scenario spawns, player deploys a lance, the player then reinforces that scenario using Forces in Fight Stance. This is easily done by converting CVP to SP, if necessary. The player will usually out-BV Princess at this point, allowing for a much easier victory. This generates a CVP, making future scenarios easier. So long as the player can maintain CVP above 1 they can infinitely reinforce. The only limitation is the number of Forces available to them. As they are taking less casualties they end up in a victory spiral where past victories ensure future victories. There is a bit more to it than that, but that's the general gist.

Following a handful of discussions on Discord it was determined that a more advanced implementation was desirable.

### Overview
- All reinforcement attempts cost 1 SP, no matter what role the force is in.
- All reinforcements require a reinforcement check, no matter what role the force is in. Forces in the Fight role re-roll this check using the best result.

### Reinforcement Checks
1. We start by identifying the most senior character in the Admin/Command role. This character is handling communication between the Primary Force, the would-be reinforcements, and any local allied forces. Their Administration skill is used to determine the base Target Number for the reinforcement role.
- If the campaign has no personnel in the Admin/Command role, the check is automatically failed with no Support Points spent.
- If the most senior character in the Admin/Command role does not have the Administration skill (somehow), the check is automatically failed however a Support Point is still spent.

2. Next we identify any _Facility Modifier_. Currently a -1 is applied to the Target Number for each allied facility in the relevant Sector, while a +1 malus is applied for each enemy facility.
3. Then we identify the _Skill Modifier_. This is equal to the skill level of your enemies, minus the skill level of your employer. If using CamOps Reputation and on an Independent contract, your Experience Rating is used instead of your employer's skill.
4. If the contract's command rights are _Liaison_ the Target Number is reduced by 1. If the contract's command rights are _House_ or _Integrated_ the Target number is reduced by 2.
5. Finally, a 2d6 roll is made. If this equals or beats the Target Number the reinforcement check has been passed.
- Forces in the Fight role make two attempts, picking the best result.
- If the 2d6 roll is less than the Target Number or a natural, unmodified 2, then the check has been failed. The Support Point is lost and no reinforcement occurs. Otherwise the check is successful, however we now need to determine whether the enemy has intercepted the reinforcements.

### Interception Check
- Interception checks only occur if the reinforcement was successful.

1. First we roll a d100 and compare the result to the scenario chance for the relevant track. If the roll beats the scenario chance no interception occurs.
- Scenario chance is directly linked to enemy Morale, so the players will need to judge for themselves whether they can risk interception when reinforcing.

2. If an interception occurs, we need to determine whether the leader of the reinforcements was able to evade the interception. The evasion target number is equal to 9 - the force leader's _Tactics_ skill (including Skill Bonus, if present).
- 'Force Leader' is the character marked as the leader of the reinforcing force in your TO&E
- If the force leader does not have the _Tactics_ skill evasion is automatically failed.

3. If an interception occurs, but is evaded, then the reinforcements will arrive but will have been delayed. This means their arrival time is doubled or tripled (based on a d2+1 die roll).
- Evading an interception counts as a scenario, for the purposes of Fatigue.
4. If an interception occurs, and evasion is failed, then the reinforcements have been intercepted. A special interception scenario is spawned with the reinforcements automatically assigned to that scenario.
- Players can choose to reinforce interception scenarios, as normal.
- The date of the interception scenario is equal to the current date, so depending on the date of the original scenario it may be possible for the interception force to return and then assist with the original scenario.
- The scenario can occur in any unoccupied hex in the track. Currently StratCon doesn't handle having multiple scenarios in the same hex very well, so players will just have to use their imagination.

### Fixed Exploits
- It is no longer possible to bypass reinforcement rolls by assigning multiple units as the primary force.
- It is no longer possible to bypass the Support Point cost of reinforcement rolls by picking multiple reinforcement forces at the same time.

### Other
- It should be noted that a campaign's Admin/Transport personnel generate Support Points on a monthly basis. This was added by a prior PR, but is being mentioned here for those users unaware of that change.
- StratCon contracts will now initialize with Support Points based on the skill of the campaign's Admin/Transport personnel.

### Closes #4733

### DIFF
--- a/MekHQ/data/scenariotemplates/Low-Atmosphere Reinforcements Intercepted.xml
+++ b/MekHQ/data/scenariotemplates/Low-Atmosphere Reinforcements Intercepted.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<ScenarioTemplate>
+    <name>Low-Atmosphere Reinforcements Intercepted</name>
+    <shortBriefing>Destroy hostile air forces.</shortBriefing>
+    <detailedBriefing>Hostile air units have intercepted your reinforcements. Destroy at least 50% of the enemy force while preserving 50% of your own units.</detailedBriefing>
+    <mapParameters>
+        <allowedTerrainTypes />
+        <allowRotation>false</allowRotation>
+        <baseHeight>50</baseHeight>
+        <baseWidth>50</baseWidth>
+        <heightScalingIncrement>5</heightScalingIncrement>
+        <mapLocation>LowAtmosphere</mapLocation>
+        <useStandardAtBSizing>true</useStandardAtBSizing>
+        <widthScalingIncrement>5</widthScalingIncrement>
+    </mapParameters>
+    <scenarioForces>
+        <entry>
+            <key>Player</key>
+            <value>
+                <actualDeploymentZone>-1</actualDeploymentZone>
+                <allowAeroBombs>false</allowAeroBombs>
+                <allowedUnitType>-3</allowedUnitType>
+                <arrivalTurn>0</arrivalTurn>
+                <canReinforceLinked>true</canReinforceLinked>
+                <contributesToBV>true</contributesToBV>
+                <contributesToMapSize>true</contributesToMapSize>
+                <contributesToUnitCount>true</contributesToUnitCount>
+                <deployOffboard>false</deployOffboard>
+                <deploymentZones>
+                    <deploymentZone>0</deploymentZone>
+                    <deploymentZone>1</deploymentZone>
+                    <deploymentZone>2</deploymentZone>
+                    <deploymentZone>3</deploymentZone>
+                    <deploymentZone>4</deploymentZone>
+                    <deploymentZone>5</deploymentZone>
+                    <deploymentZone>6</deploymentZone>
+                    <deploymentZone>7</deploymentZone>
+                    <deploymentZone>8</deploymentZone>
+                    <deploymentZone>9</deploymentZone>
+                    <deploymentZone>10</deploymentZone>
+                </deploymentZones>
+                <destinationZone>5</destinationZone>
+                <fixedUnitCount>0</fixedUnitCount>
+                <forceAlignment>0</forceAlignment>
+                <forceMultiplier>1.0</forceMultiplier>
+                <forceName>Player</forceName>
+                <generationMethod>0</generationMethod>
+                <generationOrder>1</generationOrder>
+                <maxWeightClass>4</maxWeightClass>
+                <minWeightClass>0</minWeightClass>
+                <retreatThreshold>50</retreatThreshold>
+                <startingAltitude>0</startingAltitude>
+                <syncDeploymentType>None</syncDeploymentType>
+                <useArtillery>false</useArtillery>
+            </value>
+        </entry>
+        <entry>
+            <key>OpFor</key>
+            <value>
+                <actualDeploymentZone>-1</actualDeploymentZone>
+                <allowAeroBombs>false</allowAeroBombs>
+                <allowedUnitType>-3</allowedUnitType>
+                <arrivalTurn>0</arrivalTurn>
+                <canReinforceLinked>true</canReinforceLinked>
+                <contributesToBV>false</contributesToBV>
+                <contributesToMapSize>true</contributesToMapSize>
+                <contributesToUnitCount>false</contributesToUnitCount>
+                <deployOffboard>false</deployOffboard>
+                <deploymentZones />
+                <destinationZone>5</destinationZone>
+                <fixedUnitCount>0</fixedUnitCount>
+                <forceAlignment>2</forceAlignment>
+                <forceMultiplier>1.0</forceMultiplier>
+                <forceName>OpFor</forceName>
+                <generationMethod>1</generationMethod>
+                <generationOrder>5</generationOrder>
+                <maxWeightClass>4</maxWeightClass>
+                <minWeightClass>0</minWeightClass>
+                <retreatThreshold>50</retreatThreshold>
+                <startingAltitude>5</startingAltitude>
+                <syncDeploymentType>OppositeEdge</syncDeploymentType>
+                <syncedForceName>Player</syncedForceName>
+                <useArtillery>false</useArtillery>
+            </value>
+        </entry>
+    </scenarioForces>
+    <scenarioObjectives>
+        <scenarioObjective>
+            <associatedForceNames>
+                <associatedForceName>OpFor</associatedForceName>
+            </associatedForceNames>
+            <associatedUnitIDs />
+            <successEffects>
+                <successEffect>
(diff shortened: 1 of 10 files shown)


IllianiCBT:
- Base target number for reinforcement is now 9, down from 12.
- Interception checks are only made if the reinforcement roll is failed, but is not a critical failure.
- Fixed a bug with skill modifiers on Independent contracts
- Reduced reinforcement deployment times by roughly 33% - this will vary based on the movement of the force in question, but generally this will mean that reinforcements actually arrive in time for the battle and having those reinforcements get delayed isn't as crippling. To use the examples in the documentation:
`a group of Atlases (3/5) should arrive at turn 7 (20 / 3)` down from 10
`a group of jump-capable Griffins (5/8/5) should arrive on turn 3 (20 / 6, rounded down)` down from 5 (Jump increases walk speed by 1 for the purpose of arrival speed - note this has not been changed since 2018)
`a group of Ostscouts (8/12/8) should arrive on turn 2 (20 / 9, rounded down)` down from 3

HammerGS (approved):
Co-Pilot sees no issues.

PR #5280, opened 2024-12-03
Centralized Formation Size Logic & Adjusted Contract Required Lance Count

IllianiCBT:
Unified the various methods for calculating standard force sizes based on factions, enabling consistent and flexible force size determination. Formation sizes are now housed in `StrategicFormation.java`.

Adjusted the logic for determining how many forces a contract requires. Previously, we did some math, looking at the number of combat units in the TO&E and then doing some math. Often this would give the user a spare lance or two, but from what I can tell this was a byproduct of some math and not something that was actually enforced. Furthermore, there was no support for factions that didn't use base 4 i.e. Clans, or ComStar.

Now the base number of strategic forces required for a contract is based on faction. How this works is we scan the TO&E for combat units and then count the number of 'points' each unit contributes (based on faction preferences). We divide this number by the standard lance-level formation for the campaign faction. This part of the code is identical to the previous implementation, the only thing that we've changed is using faction lance-level size, rather than always assuming an IS base-4.

For example, if you are a Clan campaign with 10 'Meks (1 point), 5 Vees (0.5 point each), and 5 Aerospace Fighters (0.5 point each) you will be considered to have 15 points total. This will mean your base required forces is 3. I'm not going to go into the full workings here, as that code is all unchanged.

Once we have determined base force count we drop 1-in-3 forces. That means in the above example, the Clan player would only be required to provide 2 Stars, not 3. We do this so that the player always has reserve forces unless they choose to over-extend (see below).

Finally we roll 2d6 with the result adjusting the number of required forces:

     2: 25%
     3: 50%
     4: 75%
     5-9: 100%
     10: 125%
     11: 150%
     12: 175%

The number of required forces will never go below 1, or above the maximum number of deployable forces available to the campaign (again, remember we are calculating this based on the number of combat units in the TO&E not the number of explicit forces).

Contract pay is adjusted accordingly. This means the player can choose to take a less intense contract, where they know they will have plenty of reserve forces, or they can reduce the number of available reserves in exchange for a bigger payout.


HammerGS (approved):
Co-Pilot sees no issues.

PR #5281, opened 2024-12-03
Refactored Leadership Unit Selection Logic

IllianiCBT:
This PR revises Leadership Unit selection to no longer use the lowest BV in the previously assigned forces. Furthermore, players are not granted `x` number of auxiliary units per level of _Leadership_.

Instead, players are granted a 500 BV2 budget, multiplied by the number of levels of _Leadership_ obtained by the force commander. A character with 3 _Leadership_, for example, can supplement their deployment with 1,500 BV2 of supplemental units. There is no limit on the number of units the player can deploy in this way, so long as the total BV2 does not exceed the budget. Leadership units are still required to be a different type to the primary unit for the parent force.

Only units that are present in the TO&E are eligible to be chosen as Leadership Units. This allows the player to better control what units are listed.


IllianiCBT:
- Increased the leadership unit budget from 500/level to 1000/level (capped at 5,000). This may need further adjustments, but should be a good starting point*
- Leadership Unit costs are based on crewless BV2*
- Leadership Units are no longer restricted to types other than the primary force's primary type. This means 'Meks can reinforcement 'Mek forces, and Vees can reinforce Vee forces*
- Updated the Leadership unit picker GUI to include more information

*These changes were suggested by our user base

IllianiCBT:
- Leadership unit picker now sorts entries based on formation names, not unit names. This is done using a reversed force hierarchy. So we sort at battalion level, then company level, then lance level (and so on). This ensures units are grouped intuitively, which should make finding the right units easier for players with large campaigns.

HammerGS (approved):
Co-Pilot sees no issues.

HammerGS:
@IllianiCBT Merge Conflicts

PR #5282, opened 2024-12-03
Refactored CVP Conversion to Removal of CVP.

IllianiCBT:
Removed the conversion of victory points to support points and replaced it with a GM function to remove victory points.

At will SP is very powerful, and while recent changes have reduced the effectiveness of SP, it was decided that we should remove at will SP sources.

Players are awarded with extra XP if they get over a certain number of CVP, meaning excess CVP are not wasted.

The primary source of SP is Admin/Transport personnel who generate monthly SP.


HammerGS (approved):
Co-Pilot sees no issues.

PR #5283, opened 2024-12-03
Adjusted Support Points Negotiation Logic

IllianiCBT:
Adjusted the logic for calculating negotiated support points.

Each month personnel in the Admin/Transport role use their administration skill to generate Support Points. Previously the number of Support Points that could be generated was capped at the number of Sectors assigned to the contract.

This PR changes this so that the maximum number of Support Points that can be generated, per month, is equal to the number of combat forces required by the contract.

Furthermore, for every 4 points a character exceeds their Admin skill check, that character will generate an extra Support Point. This should allow Elite personnel to carry less skilled characters.


HammerGS (approved):
Co-Pilot sees no issues.

HammerGS (approved):

PR #5286, opened 2024-12-03
Removed Predictable Scenario Generation Schedule

IllianiCBT:
Previously scenarios would always be spawned on a Monday, with the player having advanced notice for all scenarios that would be encountered that week.  I've received a lot of feedback over the past months from the dev team and streamers that this creates a very predictable timetable of affairs which, while very civilized, removes a lot of tension from deployments.

Now, every Monday, we make scenario rolls as normal. For every scenario scheduled we secretly store a date. On that date we will spawn the scenario. The player will still have some forewarning, based on the deployment time for the chosen sector, however this makes engagements a lot less predictable.


IllianiCBT:
Added more intelligent force picking for Integrated contracts.

Previously it would just pick a random force from a list of available combat forces. Now it creates a pool with intact forces having a higher chance of being picked over less operational forces.

HammerGS (approved):
Co-Pilot sees no issues.

HammerGS:
@IllianiCBT Merge Conflicts

PR #5287, opened 2024-12-03
Added Sector Column to `ScenarioTableModel`

IllianiCBT:
Implemented a new sector column in the `ScenarioTableModel` to display relevant sector information for StratCon scenarios. This changes improves the user interface by ensuring users know which sector they should be looking in, when trying to track down a scenario for deployment purposes.


HammerGS (approved):
Co-Pilot sees no issues.

PR #5292, opened 2024-12-04
Updated Sector Display & Expanded Scenario Information

IllianiCBT:
This PR updates the StratCon sector display to display a list of sectors, rather than using a drop-down. This makes it easier for users to spot how many sectors they are assigned to, for any given contract.

I also went ahead and expanded the scenario information to include ally (including player) and hostile (including 3rd parties) BV. This should make it easier for users to determine what forces to assign to what scenario without needing to bounce between the AO and Briefing tabs quite as much.

Players currently on Integrated contracts, when updating, will notice a minor visual bug where things don't use the best alignment. It's very minor and doesn't break anything. This bug will not exist for campaigns that pick Integrated contracts _after_ updating.


IllianiCBT:
Also, I hate GUI work.

PR #5339, opened 2024-12-11
Refactored Cargo Capacity Calculations and Reporting.

### LINKED ISSUE #5324: Implemented Resupply Module #4518 - MekHQ not recognizing Cargo Mek's cargo capacity, relevant for Logistics Overhaul/Upgrade

### Severity *

### Brief Description *

Using two salvaged Daedalus 'meks, I was trying to get a convoy working, and got it set up and assigned. The issue presents in MekHQ - it does not recognize the meks as having any cargo capacity. It picks up 4.35 tons from an insulated cargo space on a Coolant Truck (Hover) and truncates to 4 tons, but does not pick up the combined 30 tons in the two stock Daedalus GTX2A 'Stevedore' meks. This is relevant for satisfying logistical requirements for the resupply feature.
image
image
image
image

### Steps to Reproduce

_No response_

### Operating System *

Windows 10

### Java Version *

17

### MekHQ Suite Version *

None

### Custom MekHQ Version

 Implemented Resupply Module #4518 

### Attach Files

_No response_

### Final Checklist

IllianiCBT:
Campaign cargo capacity was not correctly reporting cargo units stored in non-bays. For example, many industrial 'Meks have parts labeled 'cargo'. Mechanically, these are not cargo containers so do not get picked up as available cargo. I believe the intent is that at TW level these units are always presumed to be carrying cargo of some description. Rather than this being available cargo space.

I went ahead and modified how we fetch total cargo capacity from a unit, so now these amorphous 'cargo' parts are considered cargo capacity. This uses string comparison which is not my favorite method, but after a while searching through our codebase it doesn't seem we have a better option as these 'cargo' parts are not treated as special equipment so we don't have another way to confirm whether the part we're looking at _is_ 'cargo'.

I also adjusted cargo capacity so that units that are not fully crewed do not contribute cargo capacity. No more filling the hangar with empty cargo trucks.

Finally, I adjusted the cargo report displayed on the command center so that it's rounded to 1 decimal place, rather than nearest whole number. This will prevent users mistakenly bug reporting 0.5t capacity displaying as 1t.

Closes #5324

### DIFF
--- a/MekHQ/src/mekhq/campaign/CampaignSummary.java
+++ b/MekHQ/src/mekhq/campaign/CampaignSummary.java
@@ -32,6 +32,8 @@ import mekhq.campaign.unit.HangarStatistics;
 import mekhq.campaign.unit.Unit;
 import org.apache.commons.lang3.StringUtils;
 
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


gsparks3:
> I believe the intent is that at TW level these units are always presumed to be carrying cargo of some description. Rather than this being available cargo space.

Going by TechManual page 239, the "Cargo" crits on a mek explicitly are for Cargo Bays, which can be emptied or filled as desired (this can even be done at combat-relevant speeds if Lift Hoists are also present). The primary difference between the mek version and the vehicle version is that a bay requires 1 critical slot per ton on the former, but only 1 slot total on the latter (as vehicles track item slots differently). This can also involve distinguishing between e.g. 5 tons worth of "Cargo" split into five 1-ton bays, or a single 5-ton bay that occupies 5 crits in a contiguous, linked block (with the primary difference _there_ being that a critical hit to _any_ slot of the 5-ton, 5-crit bay destroys _all_ cargo in said bay). 

This is provided primarily for informational reasons, as it does not change that this is the easiest fix at the moment.

HammerGS (approved):

PR #5355, opened 2024-12-13
Disable Advance Day Button While Advance Day is in Progress

IllianiCBT:
Added logic to disable the Advance Day button to prevent multiple activations while the advance day event is still being resolved. Re-enabled the button after the process completes, ensuring proper functionality and avoiding unintended behavior.

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -1635,8 +1635,21 @@ public class CampaignGUI extends JPanel {
                 if (getCampaign().isWorkingOnRefit(tech) || tech.isEngineer()) {
                     continue;
                 }
-                String nameBuilder = "<html>" + tech.getFullName() + ", <b>" + SkillType.getColoredExperienceLevelName(tech.getSkillLevel(getCampaign(), false)) + "</b> " + tech.getPrimaryRoleDesc() + " (" + getCampaign().getTargetFor(r, tech).getValueAsString() + "+), " + tech.getMinutesLeft() + '/' + tech.getDailyAvailableTechTime() + " minutes</html>";
-                name = nameBuilder;
+                StringBuilder nameBuilder = new StringBuilder(128);
+                nameBuilder.append("<html>")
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1623,21 +1635,8 @@ public void refitUnit(Refit r, boolean selectModelName) {
...
-                    .append(tech.getPrimaryRoleDesc())
-                    .append(" (")
-                    .append(getCampaign().getTargetFor(r, tech).getValueAsString())
-                    .append("+), ")
-                    .append(tech.getMinutesLeft())
-                    .append('/')
-                    .append(tech.getDailyAvailableTechTime())
-                    .append(" minutes</html>");
-                name = nameBuilder.toString();
+                String nameBuilder = "<html>" + tech.getFullName() + ", <b>" + SkillType.getColoredExperienceLevelName(tech.getSkillLevel(getCampaign(), false)) + "</b> " + tech.getPrimaryRoleDesc() + " (" + getCampaign().getTargetFor(r, tech).getValueAsString() + "+), " + tech.getMinutesLeft() + '/' + tech.getDailyAvailableTechTime() + " minutes</html>";
```
Sleet01, line 1638: This seems less clean, plus if we're not using a StringBuilder you can just concatenate everything into `name` and skip the second line.

Was this an IDE-suggested change?  Apparently most will turn this form into the StringBuilder form under the hood, so I'm not sure this change gets us anything.

Sleet01 (changes requested):
I'm not sure about the nameBuilder change; do you mind giving the rationale for that?

Otherwise the fix looks good.

IllianiCBT:
That was IDEA 'helping'

IllianiCBT:
> I'm not sure about the nameBuilder change; do you mind giving the rationale for that?

Should be good now. Glad you caught that. IDEA launched a new version and their file clean-up in previous versions was less aggressive. Disabled it moving forward.

Sleet01 (approved):
LGTM!

Issue #4767, opened 2024-09-02
Paid recruitment rolls being debited multiple times per instance


OrbMonky:
### Environment
Windows 10, Java 17, 0.50.00 release

### Description
One week's paid recruitment recursed 7 times after the original, all on the same date.
![image](https://github.com/user-attachments/assets/00627a49-b2d6-400a-9869-55d752151c6b)
Using ATB Market if that matters. Is it possibly rolling for each potential recruiting admin?
It did seem to give results commensurate with such an expenditure (or I just got really lucky) but definitely an unintuitive 'where did all my money go' thing.
### Files
[mekhq.log](https://github.com/user-attachments/files/16840804/mekhq.log)
[customs.zip](https://github.com/user-attachments/files/16840822/customs.zip)
[RECRUITMENT.zip](https://github.com/user-attachments/files/16840823/RECRUITMENT.zip)

OrbMonky:
Went back and ran it again and got 9 instead of 8 so it looks like a possible 'roll to failure' effect that is happening?

Algebro7:
So I'm not sure what's going on, but when I loaded your save and had a play around with it, I was only only getting 1 debit each week afterwards as expected. I was also unable to replicate with other saves. Are you still seeing this behavior if you restart MekHQ and continue advancing days to the following Mondays?

My _guess_ having looked at the code is that there was some kind of bad state, possibly due to switching between Campaign options. There is [no loop involved](https://github.com/MegaMek/mekhq/blob/master/MekHQ/src/mekhq/module/atb/AtBEventProcessor.java#L76) , so the only way I can see that this could fire multiple times on the same day is if there are multiple `AtbEventProcessor` instances registered for the `NewDayEvent`. This shouldn't normally happen and I've been unable to induce it again, but it's something we should keep an eye out for and fix if we can figure it out.

OrbMonky:
I have had it on every roll except for one period after the latest contract, when it started back up again after a few weeks of just making one roll. It seems random in number, not pertaining to any number of staff or other discernible quality of the campaign. I can repro on my end, but I am not sure what to look for in terms of traceable behavior.

IllianiCBT:
We're having a really hard time replicating this issue. My best guess is that you managed to find yourself in a weirdly specific game state that we've been update to recreate.

[...]

PR #5357, opened 2024-12-13
Fixed Multiple-Nag Bug by Refactoring in-app New Campaign Handling

IllianiCBT:
For a good long while there has been a weird bug where sometimes advancing day would cause nag dialogs to trigger multiple times. This became more noticeable with the introduction of the Turnover module, as that nags users on a monthly basis increasing the likelihood users will interact with this bug.

It's taken me literal months to track down and identify the cause of this bug. It turns out that when a user creates a new campaign, from within an existing campaign, event listeners are not unregistered causing all events to be triggered multiple times. For example, it would mean that any New Day event would be triggered twice. Technically, it would trigger once for each new in-app campaign created. Create 20 new campaigns, one after the other, without exiting mhq? Your new day events would trigger 20 times.

However, if we unregister these event listeners and the user happened to cancel out of the in-app new campaign dialog then _no_ events would trigger until the campaign was reloaded or mhq exited.

Previously we've papered over in-app new campaign issues by having a dialog pop-up advising the user not to cancel once they pass a certain point. That worked well enough for the old 'preset' bug, but that wouldn't work here. At least, not without a really poor play experience.

Instead, what I've done is passed some logic down the chain so that a few things happen when a player triggers an in-app new campaign:
- We ask the user to confirm they want to start a new campaign. If they fail to confirm, we stop here.
- We ask the user if they would like to save their existing campaign. If they cancel the save prompt, or a save fails, we stop here.
- The point of no return is reached. We unregister all event listeners and trigger the new campaign preset picker dialog.
- If the preset picker dialog is canceled, we exit MekHQ entirely.
- If the customize preset dialog is canceled, we exit MekHQ entirely.


IllianiCBT:
A couple of addendums:
- I could have sworn there were a couple of bug reports related to this, but I couldn't find them.
- The duplicate events issue isn't just restricted to New Day events, but effectively any gui-based event. I anticipate this fix having some performance improvements for users who start a lot of campaigns back-to-back.

psikomonkie:
Do you think this would fix #4767 ?

IllianiCBT:
Potentially. iirc we were never able to track down the root cause, but multiple event triggers _could_ cause that, if paid recruitment is a part of the functionality triggered by a new day event.

PR #5362, opened 2024-12-14
Scouts, Scout: Refactored StratCon Scenario Deployment to Support Scout Role Behavior

IllianiCBT:
It has long been a point of frustration for players that when scouting a StratCon strategic map, with a force assigned to the scout role, if a random scenario spawns it will spawn on top of the scouts forcing them into often uneven fights. While there is a degree of realism there, it doesn't offer the best play experience.

Now, when scouting, if a random scenario is spawned it will spawn in an empty adjacent empty hex (if possible). Only if all adjacent are occupied (highly unlikely) will the scenario spawn on top of the scout force.

While I recognize there may be some desire for scouts to occasionally fail to evade enemy forces, or to have scout suitable scenarios, I opted for this system as we do not currently categorize scenario templates based on force role suitability. This would be a non-zero effort implementation.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -18,7 +18,6 @@
  */
 package mekhq.campaign.stratcon;
 
-import megamek.codeUtilities.ObjectUtility;
 import megamek.common.Minefield;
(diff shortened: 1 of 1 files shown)


HammerGS:
I think this is a starting point, but we should look at having scouting fails that lead them into combat at some point.

Issue #5269, opened 2024-12-01
[v0.50.01] Incorrect lances being pulled into scenarios in subcontract


LadyAdia:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MekHQ documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

 I'm in a garrison contract and I took up a short relief subcontract with integrated command that only needed one lance. I assigned one of my assault lances to it and left the rest on the garrison duty, but it seems to be really common for lances not assigned to the subcontract to be deployed on scenarios despite how I have my assignments set.
Example: Lance assigned to Contract A is pulled into a scenario in Subcontract B. 

Included in the attached .zip file is my user files, custom builds, MekHQ log, current save file with the notification of the deployment, and my previous autosaves from the days leading up to the point the scenario generated.

[MHQ Contract Bug 2024-12-01.zip](https://github.com/user-attachments/files/17970633/MHQ.Contract.Bug.2024-12-01.zip)

### Steps to Reproduce

1. Accept garrison contract
2. Accept subcontract with integrated command
3. allow a scenario to generate. Deployed lance bat be one not assigned to the subcontract.

### Operating System *

Mac

### Java Version *

17.0.12

### MekHQ Suite Version *

Free Text (type manually)

### Custom MekHQ Version

0.50.01

### Attach Files

[MHQ Contract Bug 2024-12-01.zip](https://github.com/user-attachments/files/17970636/MHQ.Contract.Bug.2024-12-01.zip)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

IllianiCBT:
Closing as resolved

PR #5365, opened 2024-12-14
Refactored Logic for Retrieving Deployable Combat Forces

IllianiCBT:
Simplified and clarified the method to determine which combat teams are available for deployment. Ensured combat teams set to the Unassigned (aka In Reserve) role are not considered eligible for scenarios.

It's important to note that the modified method is only utilized during Integrated contracts. The goal here is to give users a way to declare that forces should _not_ be assigned to scenarios, while under Integrated command.

I wanted to avoid users needing to muck around with Combat Team overrides, non-combat flags, and that sort of thing - which the currently the only solution, short of ripping forces out of the TO&E.

Now, if a force is too battered to fight the user can put them 'In Reserve' and mhq will ignore them when it comes to assigning forces.

The reason we do it this way, is putting a force in the In Reserve/Unassigned role doesn't count it towards the required lances for a contract. That means that there is a limit on just how many forces the player can 'hide' in this way.


IllianiCBT:
Updated this PR since it's initial posting so that Integrated contracts will only consider combat teams that have been assigned to that contract as eligible for deployment. This one simple change actually allowed us to remove a bunch of excess code and no longer required loops.

Closes #5269

PR #5387, opened 2024-12-16
Adjusted Marshaling and Unmarshaling of New AutoLogistics Module

IllianiCBT:
Adjusted how the new autoLogistics module, added by @Lapras, marshals and unmarshals to/from xml.

Previously, it was converting the PartQuality to string (adjusted based on whether the user had reverse quality names enabled). This is a bit cumbersome and I'm not confident it would properly handle users enabling/disabling that option mid-campaign.

Instead, I've changed it so that it's using the enum value, as this remains consistent no matter what options the user has enabled. QUALITY_A is always QUALITY_A, the only thing that changes is how that's presented to the user.


IllianiCBT:
PR approved by @Lapras via Discord

PR #5389, opened 2024-12-16
Abstract Combat Auto Resolve

Scoppio:
# Abstract Combat Auto Resolve - ACAR

This PR implements ACAR into MHQ. It allows the system to simulate multiple combats very quickly.

It adds configuration for auto resolve, the autoresolve itself, the player may also select which kind of autoresolve to use (Princess or ACAR), etc.

It is also fully implemented with the scenario resolution tracker. Has a test for quick evaluation, it also writes a gamelog into a file named simulation.html


MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -1132,10 +1135,6 @@ public CampaignOptions() {
         phenotypeProbabilities[Phenotype.VEHICLE.ordinal()] = 0;
         phenotypeProbabilities[Phenotype.PROTOMEK.ordinal()] = 95;
         phenotypeProbabilities[Phenotype.NAVAL.ordinal()] = 25;
-
-        // Remove Milestone after 0.49.19
-        phenotypeProbabilities[Phenotype.MEKWARRIOR.ordinal()] = 95;
-        phenotypeProbabilities[Phenotype.PROTOMEK.ordinal()] = 95;
```
Scoppio, line 1138: this is duplicated for absolutelly no reason

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -654,6 +667,158 @@ public IconPackage getIconPackage() {
...
+        List<Future<?>> futures = new ArrayList<>();
+        for (int i = 0; i < numberOfGames; i++) {
+            futures.add(executor.submit(() -> {
+                new Resolver(getCampaign(), units, scenario, new SimulationOptions(getCampaign().getGameOptions()), simulationScore::addResult)
+                    .resolveSimulation();
+            }));
+        }
+
+        // Wait for all tasks to complete
+        for (Future<?> future : futures) {
```
IllianiCBT, line 810: Do we need to use a wildcard here, or can we use a specific type?

MekHQ/src/mekhq/campaign/autoresolve/acar/SimulationContext.java
```diff
@@ -0,0 +1,784 @@
...
+    private final List<String> forceMustBePreserved = new ArrayList<>();
+    private final List<Action> pendingActions = new ArrayList<>();
+
+    /**
+     * Game Phase and rules
+     */
+    private GamePhase phase = GamePhase.UNKNOWN;
+    private GamePhase lastPhase = GamePhase.UNKNOWN;
+
+    private final Map<Integer, SkillLevel> playerSkillLevels = new HashMap<>();
```
IllianiCBT, line 73: I think CodeQL might be right in this instance. I'm not spotting where this is used.

MekHQ/src/mekhq/campaign/autoresolve/acar/SimulationContext.java
```diff
@@ -0,0 +1,784 @@
...
+import org.apache.commons.lang3.NotImplementedException;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+/**
+ * @author Luana Coppio
+ */
+public class SimulationContext implements IGame {
```
IllianiCBT, line 53: This is a mm sourced class, so my input is going to be very limited.

@SJuliez I saw you reviewed the mm half of this. I don't suppose you'd mind taking a look at this specific class?

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
```diff
@@ -0,0 +1,68 @@
...
+    private static void processFormationModifiers(ManeuverToHitData toHit, Formation formation) {
+        if (formation.getEngagementControl() == EngagementControl.FORCED_ENGAGEMENT) {
+            toHit.addModifier(1, "Forced Engagement");
+        }
+        if (formation.isAerospace()) {
+            toHit.addModifier(2, "Aerospace Formation");
+        }
+    }
+
+    private static void processCombatUnit(ManeuverToHitData toHit, Formation formation) {
```
IllianiCBT, line 48: Would the SkillLevel enums be a better fit here, rather than Strings? 

e.g. `SkillLevel.ELITE.toString()` or `SkillLevel.ELITE.name()`

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
```diff
@@ -0,0 +1,68 @@
...
+    public static ManeuverToHitData compileToHit(Formation formation) {
+        var toHit = new ManeuverToHitData(formation.getTactics(), "Tactics");
+        processFormationModifiers(toHit, formation);
+        processCombatUnit(toHit, formation);
+        return toHit;
+    }
+
+    private static void processFormationModifiers(ManeuverToHitData toHit, Formation formation) {
+        if (formation.getEngagementControl() == EngagementControl.FORCED_ENGAGEMENT) {
+            toHit.addModifier(1, "Forced Engagement");
```
IllianiCBT, line 41: If this is a user-facing String we need it to be in a resource bundle, as per the style guide. I usually make an exception for older code that predates that requirement, unless we're making a lot of String-based changes, but for new files we wanna stick to resource bundles.

With it being in a TargetRoll extension I'm not making an assumption as to use, as they can be both hidden and user facing. So feel free to ignore this comment if it's not relevant.

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
```diff
@@ -0,0 +1,68 @@
...
+            case 3 -> toHit.addModifier(0, "Veteran");
+            case 2 -> toHit.addModifier(-1, "Elite");
+            case 1 -> toHit.addModifier(-2, "Heroic");
+            case 0 -> toHit.addModifier(-3, "Legendary");
+            default -> toHit.addModifier(TargetRoll.IMPOSSIBLE, "Invalid skill");
+        }
+
+        switch (formation.moraleStatus()) {
+            case SHAKEN -> toHit.addModifier(+0, "shaken");
+            case UNSTEADY -> toHit.addModifier(+1, "unsteady");
```
IllianiCBT, line 63: Do these enums have a string name we can grab, instead of defining it here? Would mean we only need to update a centralized location if we ever want to change stuff up.

MekHQ/src/mekhq/campaign/autoresolve/acar/action/RecoveringNerveActionToHitData.java
```diff
@@ -0,0 +1,56 @@
...
+        //noinspection OptionalGetWithoutIsPresent
+        var formation = game.getFormation(recoveringNerveAction.getEntityId()).get();
+        RecoveringNerveActionToHitData toHit = new RecoveringNerveActionToHitData(formation.getMorale(), "Morale");
+        processSkill(toHit, formation);
+        return toHit;
+    }
+
+    private static void processSkill(RecoveringNerveActionToHitData toHit, Formation formation) {
+        switch (formation.getSkill()) {
+            case 7 -> toHit.addModifier(+2, "Wet behind the ears");
```
IllianiCBT, line 45: This comes back to my earlier comment about possibly using variants of `SkillLevel.ELITE.toString()`

MekHQ/src/mekhq/campaign/autoresolve/acar/action/WithdrawToHitData.java
```diff
@@ -0,0 +1,59 @@
...
+            toHit.addModifier(2, "Formation is infantry only");
+        }
+        if (formationIsVehicleOnly) {
+            toHit.addModifier(1, "Formation is vehicle only");
+        }
+    }
+
+    private static void processMorale(WithdrawToHitData toHit, Formation formation) {
+        switch (formation.moraleStatus()) {
+            case SHAKEN -> toHit.addModifier(+1, "Shaken morale");
```
IllianiCBT, line 53: As per earlier, do we have enum names we can pull here, rather than defining them multiple times separately?

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+            damage += 1;
+        }
+        damage += attackingUnit.getCurrentDamage().getDamage(attack.getRange()).damage;
+
+        damage = Math.max(1, processDamageByEngagementControl(attacker, target, damage));
+        return damage;
+    }
+
+    private void handleCrits(Formation target, SBFUnit targetUnit, int criticalRollResult, SBFUnit attackingUnit) {
+        if (criticalRollResult <= 4) {
```
IllianiCBT, line 132: Could we remove this chain of if-else, replacing it with a switch?

If you want the results to fit between x and y, MathUtility.clamp() has you covered.

Not a major issue, purely stylistic, but I find switches a lot easier to read at a glance when doing results tables.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+            // Destroyed
+            countKill(attackingUnit, targetUnit);
+            targetUnit.setCurrentArmor(0);
+            target.setHighStressEpisode();
+            reporter.reportUnitDestroyed();
+        }
+    }
+
+
+    private void countKill(SBFUnit attackingUnit, SBFUnit targetUnit) {
```
IllianiCBT, line 159: CodeQL seems to have a vendetta against this method. I'm leaving it up to you as to whether these are legitimate or false positives.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+        var engagement = engagementControlMemories
+            .stream()
+            .filter(f -> f.getOrDefault("targetFormationId", Entity.NONE).equals(target.getId()))
+            .findFirst(); // there should be only one engagement control memory for this target
+
+        if (engagement.isEmpty()) {
+            return damage;
+        }
+
+        var isAttacker = (boolean) engagement.get().getOrDefault("attacker", false);
```
IllianiCBT, line 181: I've noticed you use `var` a lot, is there value in narrowing down the type to specific uses? `var` becoming `boolean`, in this case?

I know `var` lets the compiler infer type, based on what comes after, but it feels like it's opening us up to type cast errors if we, for example, put a boolean in `isAttacker` and then later try and treat it as an incompatible class.

If this isn't a concern, I'd appreciate it if you could explain why. Purely for my own edification :)

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+        var wonEngagement = (boolean) engagement.get().getOrDefault("wonEngagementControl", false);
+        var engagementControl = (EngagementControl) engagement.get().getOrDefault("engagementControl", EngagementControl.NONE);
+        if (wonEngagement) {
+            return processDamageEngagementControlVictory(engagementControl, damage);
+        }
+        return processDamageEngagementControlDefeat(engagementControl, damage);
+    }
+
+    private int processDamageEngagementControlDefeat(EngagementControl engagementControl, double damage) {
+        switch(engagementControl) {
```
IllianiCBT, line 196: We probably want to be using enhanced switches here, unless you want specific fall through behavior, which it doesn't look like you do.

Enhanced switches only became available to us when we updated to j17, so you'll only see it in newer code.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+            case NONE:
+                // nothing happens
+                break;
+            case OVERRUN:
+                break;
+            case STANDARD:
+                break;
+            case EVADE:
+                damage = damage * 0.5;
```
IllianiCBT, line 205: Actually, on second thoughts, seeing that you only have one case where something happens, would it not be better as an if-statement?

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
```diff
@@ -0,0 +1,231 @@
...
+        return (int) damage;
+    }
+
+    private int processDamageEngagementControlVictory(EngagementControl engagementControl, double damage) {
+        switch(engagementControl) {
+            case NONE:
+                // nothing happens
+                break;
+            case OVERRUN:
+                damage = damage * 0.25;
```
IllianiCBT, line 219: Either an enhanced switch here, or a pair of if-statements.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/WithdrawActionHandler.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * This is not up to rules as written, the intention was to create a play experience that is more challenging and engaging.
+     * The rules as written allow for a very simple withdraw mechanic that in this situation is very easy to exploit and would
+     * create too many games which result in no losses.
+     */
+    @Override
+    public void execute() {
+        var withdraw = (WithdrawAction) getAction();
```
IllianiCBT, line 52: This is coming back to my earlier comments of using `var` v. specific type. I am really nervous with just how much you're using `var` throughout this module.

MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhaseEndManager.java
```diff
@@ -0,0 +1,87 @@
...
+package mekhq.campaign.autoresolve.acar.manager;
+
+import megamek.common.enums.GamePhase;
+import mekhq.campaign.autoresolve.acar.SimulationManager;
+import mekhq.campaign.autoresolve.acar.report.PublicReportEntry;
+
+public record PhaseEndManager(SimulationManager simulationManager) implements SimulationManagerHelper {
+
+    public void managePhase() {
+        switch (simulationManager.getGame().getPhase()) {
```
IllianiCBT, line 29: If you used an enhanced switch here you wouldn't need to include `break` after each case, removing a lot of risk from writing switches. Generally, post 50.0, we should only be using normal switches in cases where we want specific fall through behavior - as enhanced switches don't allow for fall through.

MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java
```diff
@@ -0,0 +1,60 @@
...
+
+package mekhq.campaign.autoresolve.acar.manager;
+
+import mekhq.campaign.autoresolve.acar.SimulationManager;
+
+public record PhasePreparationManager(SimulationManager simulationManager) implements SimulationManagerHelper {
+
+    public void managePhase() {
+        clearActions();
+        switch (game().getPhase()) {
```
IllianiCBT, line 28: This would be a lot more succinct as an enhanced switch:

```
switch (game().getPhase()) {
            case DEPLOYMENT, SBF_DETECTION, MOVEMENT, FIRING -> {
                resetEntityPhase();
                simulationManager.initiativeHelper.determineTurnOrder(game().getPhase());
            }
            case INITIATIVE -> clearActions();
            case END, VICTORY, default -> {
                clearReports();
                clearActions();
            }
        }
```

MekHQ/src/mekhq/campaign/autoresolve/acar/phase/FiringPhase.java
```diff
@@ -0,0 +1,222 @@
...
+            var actingFormationMos = Compute.rollD6(2).getMarginOfSuccess(actingFormationToHit);
+            var targetFormationMos = Compute.rollD6(2).getMarginOfSuccess(targetFormationToHit);
+
+            if (actingFormationMos > targetFormationMos) {
+                if (EngagementControl.OVERRUN.equals(actingFormation.getEngagementControl())
+                    && !actingFormation.isEngagementControlFailed()) {
+                    range = ASRange.SHORT;
+                } else {
+                    range = WeightedDoubleMap.of(
+                        ASRange.LONG, actingFormation.getStdDamage().L.damage,
```
IllianiCBT, line 93: Can we confirm whether the below CodeQLs are false-positive or not, it's hard to tell when reviewing through GitHub

MekHQ/src/mekhq/campaign/autoresolve/acar/report/EndPhaseReporter.java
```diff
@@ -0,0 +1,73 @@
...
+import megamek.common.IGame;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class EndPhaseReporter {
+
+    private final Consumer<PublicReportEntry> reportConsumer;
+    private static final Map<Integer, Integer> unitDestroyedMessageMap = Map.of(
+        IEntityRemovalConditions.REMOVE_DEVASTATED, 3337,
```
IllianiCBT, line 33: Using integers like this is really confusing, at least looking at it as a reviewer. If these integers parse to a specific meaning, would there be value in converting them into enums, instead?

MekHQ/src/mekhq/campaign/autoresolve/acar/report/MoraleReporter.java
```diff
@@ -0,0 +1,70 @@
...
+    private final IGame game;
+    private final Consumer<PublicReportEntry> reportConsumer;
+
+    public MoraleReporter(IGame game, Consumer<PublicReportEntry> reportConsumer) {
+        this.game = game;
+        this.reportConsumer = reportConsumer;
+    }
+
+    public void reportMoraleCheckStart(Formation formation, int toHitValue) {
+        // 4500: Start of morale check
```
IllianiCBT, line 39: What are these integers in reference to? This is the type of thing that probably makes absolute sense to you, as the writer, but means nothing to anyone just looking at your comment :D

IllianiCBT (commented):
Needing to take a break before I move onto the last handful of classes. Also, because having so much code and comments open is causing GitHub to die around me! :D

My initial thoughts are...
- This is every bit as cool as I hoped
- We probably don't want to be using normal switches any more, instead let's use enhanced switches where appropriate
- I'm _really_ nervous about how reliant this code is on the use of `var` instead of specific types. I've outlined my concerns in my review comments, but I'd like to add another: when I look at a `var` I do not immediately know what type of value it's going to hold. Compare to `int`, I immediately know exactly what kind of information that variable is going to hold and therefore what I can and cannot do to it.

MekHQ/src/mekhq/campaign/handler/PostScenarioDialogHandler.java
```diff
@@ -0,0 +1,113 @@
...
+import mekhq.gui.dialog.RetirementDefectionDialog;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * @author Luana Coppio
+ */
+public class PostScenarioDialogHandler {
```
IllianiCBT, line 41: I'm a little concerned about this class. With this implementation we have, effectively, three separate ways to conclude a scenario with slightly different implementations: concluding a mm scenario and going back to mhq; resolving manually via the Briefing Tab; and now this.

My concern is that without any centralized system if someone wants to add a new action to occur during scenario resolution they now need to update three different locations, without anything in the code inherently pointing them to do so. It's already bad enough that we have two different resolution systems.

I'm not sure what we want to do to resolve this, so this is more just to highlight a potential issue than requesting specific change.

MekHQ/src/mekhq/campaign/unit/damage/AeroDamageApplier.java
```diff
@@ -0,0 +1,75 @@
...
+import megamek.common.HitData;
+import megamek.common.IEntityRemovalConditions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static megamek.common.Compute.randomInt;
+
+/**
+ * Applies damage to an Aero entity.
```
IllianiCBT, line 32: Does this apply to larger aero units, like DropShips, WarShips, or JumpShips? If not, how difficult would it be for you to add those handlers too?

MekHQ/src/mekhq/utilities/RandomUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+package mekhq.utilities;
+
+import megamek.common.Compute;
+
+import java.util.List;
+import java.util.Optional;
+
+;
+public class RandomUtils {
```
IllianiCBT, line 28: What benefit does this give me over `ObjectUtility.getRandomItem()`? And, is there value in moving this into that class, instead of having two separate classes that, on the surface, both seem to be doing the same thing.

IllianiCBT (commented):
Finished my review. For the mm-source stuff I'm about as much use as a chocolate firepoker, so I can't really offer any useful insight there.

For the rest, I've outlined my concerns. I suspect you'll have some comments of your own, so I'm not going to mark this as 'request changes' as I'm more interested in hearing what you have to say than forcing my preferences on you.

I'm also not going to mark it as approved just yet, at least not until you've had a chance to go through my feedback and reply/action as you see fit. :)

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -654,6 +667,158 @@ public IconPackage getIconPackage() {
...
+            var simulatedVictories = calculateNumberOfVictories(numberOfSimulations, scenario, units);
+            stopWatch.stop();
+            if (simulatedVictories.getRuns() == 0) {
+                message = "Commander, we were unable to simulate any combat scenarios. Do you want to proceed?";
+            } else {
+                message = "Commander, we ran " + simulatedVictories.getRuns() + " simulated combat scenarios and our forces came out victorious"
+                    + " in " + simulatedVictories.getVictories() + ", lost " + simulatedVictories.getLosses() + ", and drew "
+                    + simulatedVictories.getDraws() + " times. This gives us a "
+                    + (simulatedVictories.getVictories() * 100 / simulatedVictories.getRuns()) + "% chance of victory."
+                    + " Do you want to proceed?\n\t" + stopWatch + " elapsed time.\n\t"
```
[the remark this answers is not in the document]
Scoppio: Its an interesting information, would be better in a debug environment tbh (and added in the log)

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -654,6 +667,158 @@ public IconPackage getIconPackage() {
...
+            tracker.setEvent(autoResolveConcludedEvent);
+            tracker.processGame();
+
+            ResolveScenarioWizardDialog resolveDialog =
+                new ResolveScenarioWizardDialog(campaignGUI.getFrame(),
+                    true, tracker);
+            resolveDialog.setVisible(true);
+            if (resolveDialog.wasAborted()) {
+                // I think I can get rid of this, but I'm not sure atm, testing
+//                for (UUID personId : tracker.getPeopleStatus().keySet()) {
```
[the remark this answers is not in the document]
Scoppio: Turns out it is useless indeed. The thing is that the persons do not receive damage in the simulations because I am applying the damage to entities without reference to the persons, this means that if you cancel the reasolution wizard nothing happens to your personel or meks.

MekHQ/src/mekhq/MekHQ.java
```diff
@@ -654,6 +667,158 @@ public IconPackage getIconPackage() {
...
+                victories.incrementAndGet();
+            } else if (game.getVictoryTeam() == 2) {
+                losses.incrementAndGet();
+            } else {
+                draws.incrementAndGet();
+            }
+            gamesRun.incrementAndGet();
+        }
+
+        public int getVictories() {
```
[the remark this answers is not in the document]
Scoppio: Not in this case, since it is counting losses, victories, draws, this is very short and simple, and the object itself is also very direct on its responsibiliites `SimulationScore`.

MekHQ/src/mekhq/MekHQ.java
Scoppio: It has to be a wildcard here unfortunatelly, but I think I can change one or another thing around and solve this.

MekHQ/src/mekhq/campaign/autoresolve/acar/SimulationContext.java
Scoppio: I fixed that , there is a PR with this being used now.

MekHQ/src/mekhq/campaign/autoresolve/acar/SimulationContext.java
Scoppio: I did multiple changes based on Juliez feedback and it improved it very much

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
Scoppio: Not really, its not a 1:1 match with SkillLevel

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
Scoppio: Fair

MekHQ/src/mekhq/campaign/autoresolve/acar/action/ManeuverToHitData.java
Scoppio: I will look into it.

MekHQ/src/mekhq/campaign/autoresolve/acar/action/RecoveringNerveActionToHitData.java
Scoppio: I will look into it.

MekHQ/src/mekhq/campaign/autoresolve/acar/action/WithdrawToHitData.java
Scoppio: It is inline with every single "ToHit" present in the game.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
Scoppio: The amount of false positives CodeQL raises makes me wonder how it can be used by anyone.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
Scoppio: This isnt a concern, var is not like writting `?` or `Object` and then letting the compiler discover, its instead just a shorthand for the type of the variable, so here I could use `boolean isAttacker` or write down `var isAttacker` without problem because in both cases the IDE and the compiler knows I am using a boolean here.

The casting to (boolean) I am doing in this case is specifically because the `getOrDefault` is dealing with a variable that is a `Map<String, Object>`, so I need the typecasting here otherwise it will resolve the variable type to `Object`.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
Scoppio: fixing

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
Scoppio: fixing

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/WithdrawActionHandler.java
Scoppio: Screenshot 2024-12-16 at 19 44 56
Screenshot 2024-12-16 at 19 45 17
Screenshot 2024-12-16 at 19 45 33
Screenshot 2024-12-16 at 19 45 48

There is no need for using anything other than var inside a function to be honest with you.

MekHQ/src/mekhq/campaign/autoresolve/acar/handler/StandardUnitAttackHandler.java
IllianiCBT: Every day is a school day. Thanks for the explanation :)

MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhaseEndManager.java
Scoppio: I am banking on fallthrough behavior in this switch statement

```java
switch (game().getPhase()) {
            case DEPLOYMENT:
            case SBF_DETECTION:
            case MOVEMENT:
            case FIRING:
                resetEntityPhase();
                simulationManager.initiativeHelper.determineTurnOrder(game().getPhase());
                case INITIATIVE:
                clearActions();
                break;
                
                
            case END:
            case VICTORY:
            default:
                clearReports();
                clearActions();
                break;

File: MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java (hunk 18 above)
Comment (Scoppio) [reply MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java:28]: but would also be wrong, as initiative also run `resetEntityPhase` and the determineTurnOrder.

File: MekHQ/src/mekhq/campaign/autoresolve/acar/phase/FiringPhase.java (hunk 19 above)
Comment (Scoppio) [reply MekHQ/src/mekhq/campaign/autoresolve/acar/phase/FiringPhase.java:93]: thats a false positive, It Is incapable of understanding the flow of this function.

File: MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java (hunk 18 above)
Comment (IllianiCBT) [reply MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java:28]: Ahhhh, I hate reviewing through GitHub, you're quite right

File: MekHQ/src/mekhq/campaign/autoresolve/acar/report/EndPhaseReporter.java (hunk 20 above)
Comment (Scoppio) [reply MekHQ/src/mekhq/campaign/autoresolve/acar/report/EndPhaseReporter.java:33]: Those are messageIDs for the gamelogger. It is just a reference map saying "this number refers to that other number" so it knows which report to load in the acs-report-message-LOCALE.properties, this was the best I could do since I rely on the IEntityRemovalConditions due to the DamageApplier class.

File: MekHQ/src/mekhq/campaign/autoresolve/acar/report/MoraleReporter.java (hunk 21 above)
Comment (Scoppio) [reply MekHQ/src/mekhq/campaign/autoresolve/acar/report/MoraleReporter.java:39]: Screenshot 2024-12-16 at 19 56 27

I was using those to guide me, so I could go back in the code and note all the message numbers I had to write them down in the properties file when I finished.

File: MekHQ/resources/mekhq/resources/acs-report-messages_en.properties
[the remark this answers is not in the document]
Comment (Scoppio) [reply MekHQ/resources/mekhq/resources/acs-report-messages_en.properties:1]: Yes, it is a placeholder for the english locale, if there is any entry missing here it will fallback to the "default" file which should contain the report entry there.

This also serve the function of proving the i18n is working, I can just rewrite one of the many entries from the default file in here and see the change in the game when it is loaded up.

File: MekHQ/src/mekhq/campaign/autoresolve/converter/SetupForces.java [hunk 28]
```diff
@@ -0,0 +1,659 @@
...
+            Map<Integer, Container> balancedForces = new HashMap<>();
+
+            for (int team : entitiesByTeam.keySet()) {
+                createTopLevelForTeam(balancedForces, team, new ArrayList<>(entitiesByTeam.get(team)), topCount);
+            }
+
+            return new ArrayList<>(balancedForces.values());
+        }
+
+        private static void createTopLevelForTeam(Map<Integer, Container> cmap, int team, List<Integer> allEnt, int topCount) {
```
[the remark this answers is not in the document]
Scoppio: I didn even realize I forgot the cmap here, thanks

MekHQ/src/mekhq/campaign/handler/PostScenarioDialogHandler.java
Scoppio: Oh, I knew I was missing something, this class also substitute `BriefingTab#resolveScenario`. let me look back into it.

MekHQ/src/mekhq/campaign/unit/damage/AeroDamageApplier.java
Scoppio: I've not tried, but it shuold work fine in those. I need to setup a test class to eval it.

MekHQ/src/mekhq/utilities/RandomUtils.java
Scoppio: I wasant aware of that function. But the benefit is the Optional, which does allow for some safer object manipulation as the null test is a bit harder to get around.

MekHQ/src/mekhq/campaign/autoresolve/acar/manager/PhasePreparationManager.java
Scoppio: you can use split or unified view, both have some ups and downs for revieweing and make it a bit easier depending on the type of difficulty you are having. I particularly like using github for reviewing code.

IllianiCBT (approved):
Looking forward to seeing them merged and tested :)

PR #5443, opened 2024-12-19
Updated Naming of AtBLanceRole Class, Added Auxiliaries Combat Role

### LINKED ISSUE #3688: Allow units not in the TO&E to be undeployed when used as auxiliary forces
Using either Leadership or Tactical skill bonuses, it's possible to deploy auxiliary forces such as infantry, support and combat vehicles, and aircraft through StratCon.  I keep most of mine out of the TO&E because infantry-only and support vehicle formations just get overrun, vehicle-only formations struggle just a little too much, and I'm not currently interested in air/space combat scenarios (this gets down to there being no workable "Non-combat"/"Logistics"/etc. role so they don't get picked for scenarios, but that's another issue).

Once selected as auxiliary forces for a scenario there is no way to correct a mistake - the only "Undeploy" option is through the TO&E tab.  There should be a way to remove units that are not in the TO&E which are assigned in error.

### LINKED ISSUE #3745: RFE: Expanded formation roles
On the Briefing Room tab, formations can be assigned a role which influences what scenarios they are assigned to and what benefits they get from commander skills of Leadership, Tactics, and Strategy.  Currently the selections are Fight, Scout, Defend, Training, and Unassigned.  The last doesn't seem to do much other than not provide any benefits, and the formations with this role are still routinely assigned as primary forces for scenarios.  I have to keep both combat and non-combat support outside of formations so they aren't inadvertently assigned to scenarios they should never be present in.  This means that support formations, such as an artillery battery or non-combat units, cannot be organized in the TO&E.

I would like to see an expansion of roles to help organize an entire force, combat, combat support, and non-combat, in the TO&E.  These would also enable pulling the non(?) functional Non-Combat Formation flag from the TO&E tab.

Some examples:

Support: this formation will never be deployed as the primary force, except in the case of a Base Defense scenario or if they are "Stay Deployed" in a StracCon location.  They would be available to deploy as reinforcements via a Support point (or possibly 2 points).  This would be used for artillery batteries, company/battalion sized infantry, Elemental stars/binaries/trinaries, and similar forces which are intended to back up more heavily armed and armored primary formations.  They will be considered in counting formations for determining the number of tracks and objectives in a contract/mission.

Logistics: this formation will never be deployed as the primary force.  Instead, they would be added to Convoy Escort, Defend Allied Facility, and Base Defense scenarios the same way a players DropShip may be used in a Defend Grounded DropShip scenario.  They may be deployed on the StratCon map but only to allied facilities.  This would allow transport and other non-combat formations to be included in the TO&E for tracking command structure and other organizational purposes.  They will be excluded for determining the number of tracks in a contract/mission.

Resting: again, will not be deployed as a primary formation and cannot be used as reinforcements except via a similar check to the reinforcement check for Fight and Scout roles.  This also ties into strategic fatigue, allowing formations to reset their fatigue levels after a preset period of time.

Intercept and Ground Attack: for conventional/aerospace fighter formations as #2732 

Unassigned: because there is an optional setting to limit the number of formations with roles, this may still be necessary although it might be replaced with the Support role instead.  If retained these forces should not be chosen as primary forces for scenarios unless all other formations with the primary combat roles (Fight, Scout, Defend, Training) are already assigned to scenarios.  They should be available as reinforcements in the same manner as those with the Support role.

### LINKED ISSUE #4190: [RFE] - Add HQ Deployment Role for StratCon lances
Right now there is no way to put forces in your TOE without them being pulled as a primary force into battle.  

Not sure of the best way to do it, but an HQ Role for support forces that might occasionally be drawn into battle, but it would have a greatly reduced chance of being pulled as a combat force, but not be immune.  E.g. Traning Lances in AtB defaulted to a 10% encouter chance.  StratCon works differently, so I am not sure how it can work, but this is a common complaint.

So it would be in addition to a non-combat flag (which currently does not work) that would presumably be used for organizational purposes.  REMFs that never get picked for battle.  Like your recovery vehicles, or mobile kitchen lance, etc.  Some good ideas detailed here:  https://github.com/MegaMek/mekhq/issues/2749

And then you could have the HQ Role that can be picked for battle but at a greatly reduced chance.  Like traning lances, Arty Lances, etc.

Thank you.

### LINKED ISSUE #5192: [RFE] Auxiliary formations

- [X] I'm reporting the RFE to the correct repository:

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

The implementation of combat versus non-combat forces in the TOE is good but could be refined further. I propose a third designation: auxiliary units. 

The intended use is for forces that are combat facing but not intended to be primary units such as an artillery lance. Right now such lances are given the same weight as any other lance for the purposes of scenario generation. This can create weird, unfun, or suboptimal scenarios. Under AtB you could affect that by assigning different roles (or leaving them Unassigned) but the move to Stratcon has gotten rid of that.

An auxiliary unit would have a significantly reduced weight when generating scenarios, allowing for the occasional 'your artillery unit got ambushed' while making that a rare surprise while making it so the majority of your combat scenarios use actual combat-oriented forces. 

### MekHQ Suite Version *

v0.50.01

### Attach Files

_No response_

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

IllianiCBT:
This PR renames the `AtBLanceRole` class to `CombatRole` to better reflect its current use-cases.

It also introduces the new 'Auxiliary' Combat Role. Forces in this role are not counted towards the number of forces deployed to the contract, but can be deployed to the StratCon AO as a Primary Force or Reinforcements.

If deploying as Reinforcements, Auxiliaries still pay 1 SP but roll their reinforcement check twice using only the best result. This benefit is shared with forces in the 'Fight' role.

Auxiliaries will never be automatically assigned to scenarios created while under 'Integrated' command rights.

Fixes #5192
Fixes #4190
Fixes #3745
Fixes #3688

Addendum: for Legacy AtB the 'Auxiliary' role is identical to the 'In Reserve' role.


Sleet01 (approved):
Looks good!

PR #5444, opened 2024-12-19
Removed Remaining Instance of Undeploy Functionality when StratCon is Enabled

IllianiCBT:
Previously, the "Undeploy Unit" hangar option was always enabled, regardless of the campaign's StratCon setting. Added logic to disable the menu item when the StratCon option is enabled. This is to remove the likelihood the user will run into a couple of squirrelly bugs related to deploy-undeploy behavior in StratCon.


Sleet01 (approved):
LGTM

PR #5447, opened 2024-12-19
Refactored Conventional Fighter Handling in StratCon

IllianiCBT:
This PR updates our handling of Conventional Fighters in StratCon based on feedback from QA.
 
It was found that, despite our best prior efforts, aircraft OpFors were still generating with an overabundance of Conventional Fighters. This was due to a handful of causes, not least because of how ridiculously cheap CF units are, vs. the usefulness of an excess of action economy.

To resolve this we added in some additional steps:
- Conventional Fighters will not normally spawn as part of an NPC force (OpFor, or Allied).
- Instead, if the owning faction also owns the planet the force is being generated for, an additional flight of Conventional Fighters will be added to each aircraft force based on the number of Airbase facilities present in the relevant track. No airbases, no conventional fighters.

While there are some exceptions, a 'flight' is _usually_ 2 units, though this does vary some degree based on faction. ComStar-based forces will spawn 6 and Capellan-based forces will spawn 2-3.


Sleet01 (approved):
LGTM

PR #5449, opened 2024-12-19
Fixed Force Removal Bug in TO&E, Fixed Convoy-Support Force Option Bug

IllianiCBT:
Added removal of child forces when a parent force is deleted, ensuring proper cleanup of associated sub-forces.

Disabled certain menu actions for convoy forces to prevent invalid operations.


Sleet01 (approved):
LGTM

PR #5450, opened 2024-12-19
Fixed Combat Team Eligibility Check by Rolling Back Use of `getAllUnits()`

IllianiCBT:
Replaced `getAllUnits() with `getUnits()` to prevent parent nodes from incorrectly being marked as Combat Teams. Added a comment for clarification of the logic.


Sleet01 (approved):
Looks good.

PR #5472, opened 2024-12-21
Renamed Supply Depots to Space Ports, Buffed Support Point Gain

IllianiCBT:
In #5471 we went ahead and added a mechanical benefit to Allied Industrial Centers. As this effect is identical to Supply Depots I wanted to go ahead and improve Supply Depots, however while doing so I decided to rename Supply Depots to Space Ports. This was to warrant the additional SP gained through these facilities while also to cover a facility type that was missing from <50.02


HammerGS (approved):

HammerGS:
@IllianiCBT Conflicts

PR #5477, opened 2024-12-21
Rewrote & Reworked Training Combat Role

IllianiCBT:
While the original implementation of Training Combat Teams was very serviceable, I've been wanting to rework it for several months. As I was writing the documentation for Combat Teams, I realized this was likely the best opportunity I would have. I discussed it with QA and Monky came up with the seed for a very cool system.

I won't go into the full details now, as that will be covered in the upcoming documentation, however, a brief overview:

Personnel assigned to a Combat Team with the _Training_ combat role will automatically improve their combat skills. The skills improved will be based on the combat skills held by the Combat Team's commander, and in turn based on the commander's combat profession/s. A tank commander is not going to be able to teach MekWarriors how to Mek, after all.

The time to improve a skill is _roughly_ 28 days per experience level due to be obtained. So going from Ultra-Green to Green would take 28 days, while going from Green to Regular would take 56 days. The experience multiplier set in campaign options and the characters Intelligence (if this option is enabled) both influence the speed in which skills are learned.

Characters cannot improve their skills above Regular, or the skill experience level held by the commander (minus 1). So a commander with Regular in Gunnery/Mek can only teach people up to Green, while they would need Veteran to get trainees to Regular.

Commanders are not judged on their rank, so enlisted characters can train personnel.


IllianiCBT:
This PR also incorporates a number of Combat Team fixes and adjustments that would have been a nightmare of conflicts had they been pushed separately.

HammerGS (approved):

HammerGS:
@IllianiCBT Conflicts

HammerGS:
@IllianiCBT  Co-Pilot says 

The error in the logs indicates that the method isInReserve() is not found for the variable combatRole of type CombatRole in the following files:

MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java at line 2008
MekHQ/src/mekhq/campaign/Campaign.java at line 3794

To fix this error, ensure that the CombatRole class has the isInReserve() method defined. If it is missing, you need to add this method to the CombatRole class. If the method exists in a different class or location, make sure you are importing the correct class and using the method appropriately.

PR #5489, opened 2024-12-21
Fix NPE when searching for primary force in Stratcon scenario generation

jslater89:
In its original form, the modified loop causes the following exception when the scenario is on any track but the first:

```
java.lang.NullPointerException: Cannot invoke "mekhq.campaign.stratcon.StratconScenario.getPrimaryForceIDs()" because "stratconScenario" is null
    at mekhq.campaign.mission.AtBDynamicScenarioFactory.setPlayerDeploymentTurns(AtBDynamicScenarioFactory.java:3652)
    at mekhq.gui.BriefingTab.startScenario(BriefingTab.java:908)
```

I'm not sure if warning is the correct action if a primary force isn't found, or if it needs to be an exception.


IllianiCBT (approved):
This isn't a necessary fix, as this will only affect campaigns where the user updates mid-contract, which we deliberately don't support. However, seeing that you've already put the work in, I don't see any harm in merging this.

IllianiCBT:
Actually, I'm going to take back my earlier comment: this _is_ a necessary fix as otherwise it breaks out ability to test with <50.02 campaigns.

PR #5501, opened 2024-12-22
Updated Resupply Mechanics, Added Documentation

IllianiCBT:
Added early-exit logic to skip interception checks when enemies are routed, ensuring successful deliveries.

Added documentation.

Please see comments for post-posting changes.


IllianiCBT:
- Set a minimum weight when adding items to a convoy, as some items have a weight of 0 which didn't entirely break things, but was weird.
- Added Small Craft to the unit exclusion list.
- Removed Deployed units from the unit exclusion list.
- Adjusted parts ticket calculations, so that having more similar parts in the warehouse slowly degrades ticket count, rather than being a binary all/nothing.
- Adjusted scenario resolution loot rules so that Resupply scenarios distribute loot based on the cargo capacity of the surviving units.
- Improved post-scenario notifications when loot is acquired.
- Removed some post-scenario campaign log entries as these were no longer necessary.

HammerGS (approved):

PR #5510, opened 2024-12-22
Updated Force Refresh Logic in `TOEMouseAdapter`

IllianiCBT:
Added event triggering for all forces after modifying their status to ensure consistent updates. Adjusted the `refreshForceView` call position for better logical flow and UI updates.

This fixes a legacy bug that meant non-combat (support) forces were not being correctly removed from the list of Combat Teams if they were eligible to be a Combat Team _before_ receiving the Support designation.


IllianiCBT:
This was not the PR I intended to merge...

PR #5513, opened 2024-12-23
ACAR: More damage for ejected units

Scoppio:
Ejected units were not ending up with "enough" damage. Improved on that a bit.


IllianiCBT (approved):
LGTM

HammerGS (approved):

PR #5541, opened 2024-12-27
Further Adjustments to Starting Ages

IllianiCBT:
Fixed some edge cases and better refined Clan starting ages.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -922,7 +922,7 @@ public class Utilities {
             age += roll;
         }
 
-        // Clam age, if necessary
+        // Clamp age, if necessary
         if (experienceLevel != EXP_NONE) {
             age = max(18, age);
         }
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/Utilities.java
```diff
@@ -345,10 +348,10 @@ public static int generateExpLevel(int bonus) {
 
         return switch (roll) {
             case 1 -> SkillType.EXP_ULTRA_GREEN;
```
Sleet01, line 350: Can we remove `SkillType.` here as well?

IllianiCBT: Yes we certainly can - I don't know why that didn't get nixed automatically

MekHQ/src/mekhq/Utilities.java
```diff
@@ -891,13 +915,18 @@ public static int getAgeByExpLevel(int experienceLevel, boolean isClan) {
...
 
-            if (isClan) {
+            if (isClan && (experienceLevel != EXP_NONE)) {
                 roll = (int) Math.ceil(roll / 2.0);
             }
 
             age += roll;
         }
 
+        // Clam age, if necessary
```
Sleet01, line 925: :clamp: rather than :oyster: I think.

Sleet01 (changes requested):
Just a couple touch-ups, then good to go.

Sleet01 (approved):

PR #5551, opened 2024-12-28
Refactored Initial Education Logic

IllianiCBT:
Following a Discord report that personnel were not spawning with education higher than `High School` I took the opportunity to review how we initialize education when spawning characters. I replaced duplicated logic with a centralized method and incorporated Intelligence (for players using random personalities). Now characters will more reliably spawn with education levels that better reflect their experience level.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
+++ b/MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
@@ -28,15 +28,19 @@ import mekhq.campaign.finances.Money;
 import mekhq.campaign.finances.enums.TransactionType;
 import mekhq.campaign.log.ServiceLogger;
 import mekhq.campaign.personnel.Person;
-import mekhq.campaign.personnel.SkillType;
 import mekhq.campaign.personnel.enums.PersonnelStatus;
 import mekhq.campaign.personnel.enums.education.EducationLevel;
 import mekhq.campaign.personnel.enums.education.EducationStage;
+import mekhq.campaign.personnel.randomEvents.enums.personalities.Intelligence;
 import mekhq.utilities.ReportingUtilities;
 
 import java.time.DayOfWeek;
 import java.time.LocalDate;
 import java.util.*;
+
+import static megamek.common.Compute.d6;
+import static mekhq.campaign.personnel.SkillType.EXP_REGULAR;
+import static mekhq.campaign.personnel.SkillType.EXP_VETERAN;
 
 /**
  * The EducationController class is responsible for managing the education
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
```diff
@@ -1919,4 +1883,140 @@ private static List<String> graduationEventTable() {
...
+     * a random pass/fail system, and their primary role.
+     * Additionally, assigns a pre-nominal ("Dr") for individuals who achieve a doctorate-level education.
+     *
+     * <p>The method applies the following rules:</p>
+     * <ul>
+     *   <li>Persons younger than 16 are automatically assigned the "Early Childhood" education level.</li>
+     *   <li>For individuals aged 16 or older, a pass/fail rate (based on intelligence and default thresholds)
+     *   determines whether they flunk or succeed at further education.</li>
+     *   <li>The assigned education level depends on their experience level and whether they have a combat or non-combat role.</li>
+     *   <li>If the person's education reaches the doctorate level, they are granted the pre-nominal "Dr".</li>
```
HammerGS, line 1898: Random question - If the player is hiring a Doctor is all this just automatically set them to Doctor?

HammerGS (approved):

MekHQ/src/mekhq/campaign/personnel/education/EducationController.java
IllianiCBT: I'm not sure about the question, so this might not answer it, but...

Doctor, as in the role, does not guarantee the character has a doctorate; however, if the character is in a civilian role and is Elite (or better) then there is a chance they will spawn with a doctorate. At which point we give them the 'Dr' prenominal.

HammerGS: Was thinking specifically of the MD. I hire an MD and they should come with the Dr. bit. Aka...make sure we avoid. High School education but is a Doctor.

IllianiCBT: I can add a special handler to that, sure. However it means we're going to end up with some odd results. For example, characters do not require a doctorate to take the Doctor role. However, all market characters would appear _with_ doctorates. Despite not having a skill level that would match what a character _with_ a doctorate would have if they went through the Education module.

IllianiCBT: I think it's a very valid point. I _think_ the best approach is to leave it as is, then some time around 50.4 (or late 50.03) I'll put in the work to implement Random Backgrounds which would remove this issue entirely. I think that's the play.

PR #5559, opened 2024-12-29
Refactored Combat Team Handling in Scenarios for Better `null` Safety

IllianiCBT:
Revised combat team handling in several AtB scenario classes to improve null safety and replace hardcoded IDs with constants. Adjusted enemy force generation logic to account for scenarios where combat teams are not assigned.

This PR was a byproduct of me trying to fix an issue with @OrbMonky's save that had been broken by a bug earlier in the dev cycle. While I was unable to fix their saves, this PR should help reduce the risk of something like this happening again.


NickAragua:
Looks like there's some merge conflicts here that we'll want to resolve before moving forward with this.

NickAragua (commented):
Looks good other than the outstanding merges.

PR #5566, opened 2024-12-30
Refactored Nag Dialog Checks to Improve Performance

IllianiCBT:
We ran into an issue where nag dialogs were tanking new day perform. This was caused by a few factors, but predominantly due to the number of JDialogs we were creating on new day. This PR updates the nags and nag controller so that we only initialize a nag JDialog if we intend to use it.


Sleet01 (approved):
From what I recall of the earlier Nag work, this should be good.
These nags get disposed when closed, right?

IllianiCBT:
> From what I recall of the earlier Nag work, this should be good. These nags get disposed when closed, right?

Yeah, their default close action is set to dispose on close. And all buttons include dispose()

PR #5577, opened 2024-12-31
Leadership Units & Support Point Scarcity Hot Fixes

IllianiCBT:
The player was not correctly being shown the dialog to pick Leadership Units, or Frontline bonus units/minefields. Now they do.

Furthermore, it was found that Support Points were much scarcer than intended. Your Admin/Command personnel will now roll three times each, at the start of the contract; and once every week (instead of every month).


IllianiCBT:
Oh and I also fixed a very rare out-of-bounds error that could occur if NPC pilots were generated with impossible skills.

IllianiCBT:
- Reduced scenario intensity substantially.
- Adjusted Turning Points on House command right contracts to be in-line with other command rights (i.e. 1in4 chance of a scenario being a Turning Point).
- Removed auto-fail condition when reinforcements are attempting to evade an interception and their commander has no Tactics skill.
- Updated reinforcement documentation to reflect this change.
- Further improved initial Support Points (those determined at contract start).

PR #5602, opened 2025-01-02
Adjusted AtB Bonus Events

IllianiCBT:
AtB Bonus Events could sometimes trigger a resupply, which created a lot of confusion as users would end a month; get their normal resupply, then get a second seemingly out of nowhere.

Now only post-scenario bonus events can trigger a resupply. Normal, monthly, events give 1 SP instead.

Also expended a few other options so that if users had certain settings enabled they would still get a bonus.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -764,29 +764,28 @@ public class AtBContract extends Contract {
     }
 
     /**
-     * Performs a bonus roll with varying outcomes based on the roll results.
+     * Performs a bonus roll to determine and execute a random campaign bonus.
+     * The roll is simulated using 1d6, and the outcome triggers different bonus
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -766,30 +764,30 @@ public int getContractScoreArbitraryModifier() {
...
      */
-    public boolean doBonusRoll(Campaign campaign) {
+    public boolean doBonusRoll(Campaign campaign, boolean isPostScenario) {
+        final CampaignOptions campaignOptions = campaign.getCampaignOptions();
+
         int number;
         int roll = d6();
 
-        switch (roll) {
+        return switch (roll) {
```
Sleet01, line 788: This is a very interesting construction; nice work!

Sleet01 (approved):
LGTM

PR #5616, opened 2025-01-02
feat: adds some helpful tooltips for contracts

firefly2442:
Could someone please double-check the "Command Rights" text?

Helps with user-experience (UX) and usability.

### DIFF
--- a/MekHQ/data/scenariomodifiers/LiaisonGround.xml
+++ b/MekHQ/data/scenariomodifiers/LiaisonGround.xml
@@ -1,5 +1,5 @@
 <AtBScenarioModifier>
-    <additionalBriefingText>Your employer has deployer an observer to monitor your performance.</additionalBriefingText>
+    <additionalBriefingText>Your employer has deployed an observer to monitor your performance.</additionalBriefingText>
     <allowedMapLocations>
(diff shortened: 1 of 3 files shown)


IllianiCBT:
I don't believe those '0-4' are correct. Take a look at the scenario modifiers for those command rights. iirc they use a % of the player's xBV and not fixed counts.

HammerGS (changes requested):
Per Illiani Comments

firefly2442:
OK thank you.  I believe this is correct now?  Given the XML Scenario Modifiers and looking at the `ScenarioForceTemplate.java` it has a `forceMultiplier` applied.  The comments denote it adjusts it based on the Battle Value (BV) which I assume is correct and *shouldn't* be the General Battle Value (G)BV?

IllianiCBT:
If you just say 'Battle Value' that'll cover both, as there is no guarantee the user has GBV enabled and it applies to both.

IllianiCBT:
You'll have to forgive me, but I can't figure out how to review on the app. However, 'contract score' isn't used any more so probably shouldn't be mentioned as that's likely to cause confusion.

PR #5618, opened 2025-01-02
Enable Auto-Resolve Button Conditionally for AtB Scenarios

IllianiCBT:
Previously, the auto-resolve button was always enabled based on the `canStartGame` flag. This change ensures it is only enabled if the scenario is an `AtBScenario`. Running ACAR on a non-AtB Scenario results in Class Cast Errors.

Posting this PR as a suggested edit, assuming Luana hasn't got plans to expand ACAR to non-AtB Scenarios. If she does, then this can act as a temporary solution until her edits go live.


HammerGS (dismissed):
Not merging till @Scoppio provides feedback

Scoppio (approved):
Good catch

Scoppio:
I am going to enable non-atb scenarios in the near future for it, but some work is needed first

PR #5625, opened 2025-01-03
Updated StratCon Data Center to Increase Scan Range Instead of Sector Reveal

IllianiCBT:
Replaced the "track reveal" mechanic with a new "scan range" system, allowing facilities and units to extend the area they can reveal. Updated all relevant classes, methods, and XML definitions to reflect this change.

This change was made to reduce the effectiveness of Allied Data Centers. In prior versions the presence of a single Data Center completely removed the scouting stage from a contract for an entire Sector.

Now for each Allied Data Center present in the Sector the player's scan range will be increased by 1. This means, with a single Data Center the sight range of Patrol Forces will be increased from 1 to 2. While other combat roles will have their sight range increased from 0 to 1. The effectiveness of Allied Data centers stack. So a Patrol force, supported by two Data Centers, would have a sight range of 3.

This does not devalue Patrol forces, as Patrol forces will never have scenarios spawn directly on top of them, when scouting. Making deploying Patrols still the optimum strategy for exploring the sector.

In addition, I have adjusted how scouting works. Previously, if you revealed a fixed scenario when scouting, that scenario would immediately 'activate'. This meant that you never wanted to scout unless you had a second force on hand to immediately respond if you stumble across an objective scenario. We've had a lot of feedback from the players about this, and while I get the logic, it was creating a lot of poor experiences. Now, when a fixed scenario is uncovered it remains dormant until interacted with by a player force - identical to the behavior seen when the player GM Reveals a Sector.

### DIFF
--- a/MekHQ/data/stratconfacilities/AlliedDataCenter.xml
+++ b/MekHQ/data/stratconfacilities/AlliedDataCenter.xml
@@ -3,7 +3,7 @@
 	<displayableName>Data Center</displayableName>
 	<facilityType>DataCenter</facilityType>
 	<owner>Allied</owner>
-	<revealTrack>true</revealTrack>
-	<userDescription>This facility reveals all information about this track as long as it is active.</userDescription>
+	<increaseScanRange>true</increaseScanRange>
+	<userDescription>Increases the scan range of all friendly forces by 1.</userDescription>
 	<visible>true</visible>
 </StratconFacility>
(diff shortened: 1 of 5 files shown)


HammerGS:
General comment, wouldn't it make sense to keep a reveal track in the code so allow for a future option. Versus eliminating it and needing to add it back in later.

IllianiCBT:
I see the value, yeah. I’ll try and find time to make that change tomorrow :)

HammerGS (approved):

HammerGS:
@IllianiCBT Merging this and will leave to you to remember the comment.

IllianiCBT:
Added a fresh PR with the functionality restored. I didn't trust myself to remember tomorrow.

PR #5634, opened 2025-01-04
DRAFT: Issue 3650 - RFE Add Ability to Link Units in MekHQ

### LINKED ISSUE #3650: RFE: Add ability to link units together for loading/linking
There's a variety of composite units in BattleTech, with inconsistent support in the MegaMek suite right now.

---

To start with, personnel are assigned to units: this is fairly well-implemented with the ability to attach the former to the latter, which I feel is a reasonable design baseline for the rest, at least in terms of user interface.

Next, loading units (mainly, but not exclusively, infantry) onto carriers is currently supported only in MegaMek, which leads to a lot of tedious setup when starting a fight from MekHQ. This doesn't stop them from being distinct units, but it does necessitate that they exist on the same force. I'd say that's a reasonable implementation step, and it provides groundwork for the remainder.

Then there's the most urgent reason I opened this: [trailers are not meant to count as separate units at all, at least while attached to a tractor](https://bg.battletech.com/forums/techmanual/tractortrailer-combinations-and-%27number-of-units%27/). Treating them as separate units makes them unusable in MekHQ due to things like crew requirements. To implement that correctly, however, would likely require the creation of a new "component" unit type that can merge with other units and henceforth be treated as part of it.

Finally, implementing "composite" units would allow handheld weapons, which are currently unimplemented everywhere, to be added with relative ease.

### LINKED ISSUE #4006: [0.49.19 Nightly] Battle Armor unable to be loaded into troop transports in TO&E
### Environment
megamek.MegaMek.initializeLogging(MegaMek.java:118) - Starting MegaMek v0.49.19-SNAPSHOT
	Build Date: 2024-04-17T01:08:45.274342600
	Today: 2024-04-17
	Origin Project: MegaMekLab
	Java Vendor: Eclipse Adoptium
	Java Version: 17.0.6
	System Locale: en_US
	Total memory available to MegaMek: 8 GB

### Description
Battle Armor is unable to be assigned to infantry bay troop transports in TO&E, as seen in the screenshot below. All 3 Cardinal transports shown have 5.0t infantry bays, but the only option for loading the BA is onto the dropship that has a BA bay.

I do NOT have the "BA weights" optional rule enabled, and even if it was, the Elemental II squad shown below is only 5.0t as it is a 5-man medium squad.

### Files
image

image

[BA Troop Transport.zip](https://github.com/MegaMek/mekhq/files/15013732/BA.Troop.Transport.zip)

### LINKED ISSUE #4712: MekHQ-0.50.0-SNAPSHOT: Assigning a TO&E Force to Transports Does not Assign All Units in Force to Available Bays
### Environment
Version: MekHQ-0.50.0-SNAPSHOT
OS: Win11
Java: 17

### Description
When right clicking a force in the TO&E, an option exists to assign the force to a transport.

When this option is clicked, only the top unit in the force is assigned to the transport. Other units are not. Expected behaviour should be that all units in the force are assigned to available bays in the transport.

Assigning individual units to transports works as expected.

Step 1: 
image

Step 2:
image

Step 3 - where the bug happens:
image

Additional:

image
Assigning unit individually works.
image

### Files
[The Ramshackle Renegades30620207.cpnx.gz](https://github.com/user-attachments/files/16741993/The.Ramshackle.Renegades30620207.cpnx.gz)

@IllianiCBT

### LINKED ISSUE #5593: [50.02] TO&E reporting incorrect available space in Transport Ship after save and reload

### Severity *

### Brief Description *

Screenshot 2025-01-01 135936
In the TO&E, when you're assigning Aerospace to a Transport ship, it shows how much space is left in it. However, after a save and reload, it shows the space remaining as though it were empty and not transporting the TO&E already assigned to it. 

In the screenshot, The 'Fury TT' should be full and not showing 3 spots of availability for Aerospace Fighters as it only has space for 3. That availability went away when I filled it initially, but after a save and reload its showing space available again.

### Steps to Reproduce

1. Navigate to TO&E
2. Assign an aerospace unit to a transport ship. Availability reduced by 1.
3. Save, reload.
4. Availability now reads as completely empty, even though the aerospace unit is still assigned to the transport ship.

### Operating System *

Windows 11

### Java Version *

17.0.12

### MekHQ Suite Version *

Free Text (type manually)

### Custom MekHQ Version

0.50.02

### Attach Files

[logs.zip](https://github.com/user-attachments/files/18286792/logs.zip)
[Valkyrie30420303.cpnx.zip](https://github.com/user-attachments/files/18286794/Valkyrie30420303.cpnx.zip)
[mekfiles.zip](https://github.com/user-attachments/files/18286795/mekfiles.zip)

### Final Checklist

psikomonkie:
Fixes #3650 
Fixes #4006 
Fixes #4712
Fixes #5593 

Dependent on [MegaMek#6348](https://github.com/MegaMek/megamek/pull/6348).

image
image

# Linking Units in MekHQ

## Features
- Fixed "Ship Transport"-related bugs. 
- "Tactical Transport": allows the player in MekHQ to assign units to a transport, including infantry in APCs and Battle Armor on Meks. 
  - When both are deployed to MegaMek together via AtB the player will be prompted to load the units into their transports. This will also update the transported unit's deployment turn to be the same as the transport. 
  - If a unit, its tactical transport, and its ship transport are all deployed, the unit will prefer its ship transport. Reminder: players can still edit assigned transports & deployment turns in the lobby.

### To Do
- [ ] Encapsulate Clear Transports logic
- [ ] New/Fix unit tests
- [ ] Correct + Improve Javadocs
- [ ] Message Keys
- [ ] Remove vestigial code & unnecessary comments
- [ ] General style cleanup

### Out of Scope
- Towing: Towing isn't supported in MegaMek lobbies yet, so no support here. If we want to add it just in MekHQ for logistics it should be possible to do it through this.
- Cargo: Cargo is just another thing that be loaded into a Transporter. While it's not implemented in MM, it should JustWork. This'd be useful for, say, that Quickdraw that carries weapons on its back in a lift hoist, and allowing the player to manage what it's loaded with in MekHQ.


psikomonkie:
Functionally, this is ready to be looked at. The assorted bugs related to transport ship loading have been squashed. The tactical transporter is working how I want. I still little things to do (See To Do), but the overall structure shouldn't change much more.

psikomonkie:
This is ready for review. 

I'm gonna do another pass or two later today over my changes resolve the Git Bot errors (though I'm pretty sure most of them are false-positives from the old code I was touching) and to see if I can find anymore style errors, unused code, missing JavaDocs etc. That said, this would be okay if it was merged as-is.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1301,6 +1302,25 @@
                     menuItem.setEnabled(true);
                     popup.add(menuItem);
                 }
+                Unit[] unitsArr = units.toArray(new Unit[0]);
+                //AssignForceToTacticalTransportMenu assignForceToTransportMenu = new AssignForceToTacticalTransportMenu(gui.getCampaign(), unitsArr);
+                JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToTacticalTransportMenu(gui.getCampaign(), unitsArr));
```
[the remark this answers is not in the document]
psikomonkie: It's on a menu built with a deprecated menu type; I've migrated more of the menu to use the new JScrollableMenu than before. I think this can be ignored.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1657,6 +1677,26 @@
                 popup.add(menuItem);
             }
 
+            Unit[] unitsArr = units.toArray(new Unit[0]);
+            JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToTacticalTransportMenu(gui.getCampaign(), unitsArr));
```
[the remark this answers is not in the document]
psikomonkie: It's on a menu built with a deprecated menu type; I've migrated more of the menu to use the new JScrollableMenu than before. I think this can be ignored.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1193,111 +1196,40 @@
...
-                // Only display the Assign to Ship command if your command has at least 1 valid
-                // transport
-                // and if your selection does not include a transport
-                if (!gui.getCampaign().getTransportShips().isEmpty()) {
-                    for (Unit ship : gui.getCampaign().getTransportShips()) {
-                        if (ship.isSalvage() || (ship.getCommander() == null)) {
-                            continue;
+                Unit[] unitsArr = units.toArray(new Unit[0]);
+                //AssignForceToTacticalTransportMenu assignForceToTransportMenu = new AssignForceToTacticalTransportMenu(gui.getCampaign(), unitsArr);
+                JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToShipTransportMenu(gui.getCampaign(), unitsArr));
```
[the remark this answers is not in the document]
psikomonkie: It's on a menu built with a deprecated menu type; I've migrated more of the menu to use the new JScrollableMenu than before. I think this can be ignored.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1649,10 +1466,40 @@
                 popup.add(menuItem);
             }
 
-            if (StaticChecks.areAllUnitsTransported(units) && !StaticChecks.areAnyUnitsDeployed(units)) {
+            Unit[] unitsArr = units.toArray(new Unit[0]);
+            JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToShipTransportMenu(gui.getCampaign(), unitsArr));
```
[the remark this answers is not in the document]
psikomonkie: It's on a menu built with a deprecated menu type; I've migrated more of the menu to use the new JScrollableMenu than before. I think this can be ignored.

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,332 @@
...
+     * Recalculates transport capacity
+     * @param transporters What transporters are we tracking the details of?
+     */
+    @Override
+    public void initializeTransportCapacity(Vector<Transporter> transporters) {
+        transportCapacity.clear();
+        Set<Entity> oldTransportedEntities = clearTransportedEntities();
+        loadTransportedEntities();
+        if (transporters != null && !transporters.isEmpty()) {
+            for (Transporter transporter : transporters) {
```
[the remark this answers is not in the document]
psikomonkie: Yes Git Bot, that's why I have the null guard. Is there something I'm supposed to fix for this?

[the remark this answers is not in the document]
IllianiCBT: The bot is really dumb with null values some times. So long as you have the guard in place you should be good.

MekHQ/src/mekhq/gui/menus/AssignForceToTransportMenu.java
```diff
@@ -0,0 +1,105 @@
...
+            transportMenu.addActionListener(evt -> { transportMenuAction(evt, transporterType, transport, units); });
+
+            transportMenus.add(transportMenu);
+        }
+        return transportMenus;
+    }
+
+    protected abstract Set<Class<? extends Transporter>> filterTransporterTypeMenus(final Unit... units);
+
+    protected abstract void transportMenuAction(ActionEvent evt, Class<? extends Transporter> transporterType, Unit transport, Unit... units);
```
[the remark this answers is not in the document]
psikomonkie: Am I supposed to do something with this?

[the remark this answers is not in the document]
IllianiCBT: As it's an abstract there is every chance the event might be utilized by an implementation of the method, so it's best to leave it there. Even if the abstract method itself isn't doing anything with it.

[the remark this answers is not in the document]
psikomonkie: Sounds good, thanks Illiani!

MekHQ/src/mekhq/campaign/unit/ShipTransportedUnitsSummary.java
```diff
@@ -0,0 +1,169 @@
...
+        for (Unit transportedUnit : transportedUnits) {
+            Unit oldTransport = loadTransport(transporterType, transportedUnit);
+            if (oldTransport != null) {
+                oldTransports.add(oldTransport);
+            }
+        }
+
+        //Let's get matching transporters and then recalculate our transport capacity for this transporter type
+        Vector<Transporter> transporters = new Vector<>(transport.getEntity().getTransports().stream()
+            .filter(transporter -> transporter.getClass() == transporterType).toList());
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/campaign/unit/ShipTransportedUnitsSummary.java
```diff
@@ -0,0 +1,169 @@
...
+     * both the target bay and the transport ship. Once in the MM lobby,
+     * this data
+     * will be used to actually load the unit into a bay on the transport.
+     *
+     * @param transportedUnits units being loaded
+     * @param transporterType type (Class) of bay or Transporter
+     * @return old transports; what were  the units' previous transport, if they had one?
+     */
+    public Set<Unit> loadTransportShip(Vector<Unit> transportedUnits, Class<? extends Transporter> transporterType) {
+        Set<Unit> oldTransports = new HashSet<>();
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/campaign/unit/ShipTransportedUnitsSummary.java
```diff
@@ -0,0 +1,169 @@
...
+     * and set
+     * both the target bay and the transport ship. Once in the MM lobby,
+     * this data
+     * will be used to actually load the unit into a bay on the transport.
+     *
+     * @param transportedUnits units being loaded
+     * @param transporterType type (Class) of bay or Transporter
+     * @return old transports; what were  the units' previous transport, if they had one?
+     */
+    public Set<Unit> loadTransportShip(Vector<Unit> transportedUnits, Class<? extends Transporter> transporterType) {
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1315,39 +1319,100 @@ public CurrentLocation getLocation() {
...
+        shipTransporters.addTransporter(unit);
+    }
+
+    /**
+     * Adds an entry to the list of tactical transporters . We'll use this
+     * to assign units later
+     * @see CampaignTransporterMap
+     * @param unit - The unit we want to add to this Map
+     */
+    public void addTacticalTransporter(Unit unit) {
```
psikomonkie, line 1366: Generic vs specific methods in Campaign & Unit: There are many methods in Campaign.java & Unit.java that have a "Ship" version, a "Tactical" version, and a "Generic" version. This is an example of that:
- public void addShipTransporter(Unit unit)
- public void addTacticalTransporter(Unit unit)
- public void addCampaignTransport(CampaignTransportType campaignTransportType, Unit unit)

I've preferred to use the generic version when possible, as that often let me handle both/any transport cases with a single method outside of Unit/Campaign. 

Should we keep the "specific" versions of these methods:
- Specific methods that aren't used yet - addTacticalTransporter is never used, the generic addCampaignTransport is always used. Should I keep addTacticalTransporter so someone else could use it in the future, or remove it because it's unused?
- Specific methods that are used - I can also remove all of the specific methods and replace them with their generic version everywhere. When it is a specific one we want to use, pass in the CampaignTransportType as a specific enum you want (SHIP_TRANSPORT or TACTICAL_TRANSPORT currently) to get the exact same functionaltiy as the "specific" one. Should I refactor and remove the specific ones when possible, or leave these usages as-is?

Thanks,
Psi

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * TransportShipAssignment.java
+ *
+ * Copyright (c) 2020 The Megamek Team. All rights reserved.
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
```
psikomonkie, line 8: I copied some files during this, which brought over this copyright/license. I also created some new files, which didn't include this. What's the deal with adding/updating these -  Should every new file I create have this w/ the updated year, is it unnecessary to add, or is it handled by someone else?

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,235 @@
...
+     * Recalculates transport capacity - make sure you pass in all the transporters
+     * of a given type (class), or just all the transporters an entity has.
+     * @param transporters What transporters are we recalculating?
+     * @see Entity#getTransports()
+     */
+    @Override
+    public void recalculateTransportCapacity(@Nullable Vector<Transporter> transporters) {
+        if (transporters != null && !transporters.isEmpty()) {
+            // First let's clear the transport capacity for each transport type in transporters
+            for (Transporter transporter : transporters) {
```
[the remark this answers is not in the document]
psikomonkie: Yes Git Bot, that's why there's the null guard. Silly bot.

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
Scoppio: You have to handle the copyright by yourself.

I use this for megamek (GPL2)

```
Copyright (c) $originalComment.match("Copyright \(c\) (\d+)", 1, "-", "$today.year")$today.year - The MegaMek Team. All Rights Reserved.

This program is free software; you can redistribute it and/or modify it
under the terms of the GNU General Public License as published by the Free
Software Foundation; either version 2 of the License, or (at your option)
any later version.

This program is distributed in the hope that it will be useful, but
WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
for more details.

```

and this for mekhq (GPL3)

```
Copyright (c) $originalComment.match("Copyright \(c\) (\d+)", 1, "-", "$today.year")$today.year - The MegaMek Team. All Rights Reserved.

 This file is part of MekHQ.

 MekHQ is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.

 MekHQ is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 GNU General Public License for more details.

 You should have received a copy of the GNU General Public License
 along with MekHQ. If not, see <http://www.gnu.org/licenses/>.
 ```
 
 Those are formatted for intelliJ profile templating

File: MekHQ/src/mekhq/AtBGameThread.java [hunk 13]
```diff
@@ -235,6 +239,10 @@ public void run() {
                     }
                 }
 
+                Map<Class, Map<UUID, List<UUID>>> potentialTransports = new HashMap<>();
+                potentialTransports.put(SHIP_TRANSPORT.getTransportAssignmentType(), new HashMap<>());
+                potentialTransports.put(TACTICAL_TRANSPORT.getTransportAssignmentType(), new HashMap<>());
```
Scoppio, line 244: Avoid using such complex setup and instead move this to a class and make the access of each thing explicit.
You can make that class extend Container<?> so you can still use all the different facilities like for-each, iterators, contains, add/append, etc.

example:

[Orders.java](https://github.com/MegaMek/megamek/blob/1fee799f589375d8f2130585cc64b8836b8a65b2/megamek/src/megamek/common/autoresolve/acar/order/Orders.java)

This way you dont throw up and down a very complex Map of Map of List, and any special behaviors and data manipulation can be kept inside of it, so the user dont have to deal with complex things like `potentialTransports.get(SHIP_TRANSPORT.getTransportAssignmentType()).put(unit.getId(), new ArrayList<>());`

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
psikomonkie: Awesome, really good info Luana, thank you! I’ll get that added / updated soon for all my new files, and update the year for the existing ones.

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -294,27 +305,42 @@ public void run() {
...
+                // Try to load into transportShips first, then tacticalTransports
                 for (Unit unit : units) {
+                    boolean isTransported = false;
                     if (unit.hasTransportShipAssignment()) {
                         Unit transportShip = unit.getTransportShipAssignment().getTransportShip();
-                        if (scenario.getPlayerTransportLinkages().containsKey(transportShip.getId())) {
-                            scenario.addPlayerTransportRelationship(transportShip.getId(), unit.getId());
+
+                        if (potentialTransports.get(SHIP_TRANSPORT.getTransportAssignmentType()).containsKey(transportShip.getId())) {
+                            potentialTransports.get(SHIP_TRANSPORT.getTransportAssignmentType()).get(transportShip.getId()).add(unit.getId());
```
Scoppio, line 318: This is not a reasonable way to access to data, its unintuitive and hard to debug and understand. Follow the suggestion in the previous comment.

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -294,27 +305,42 @@ public void run() {
...
+
+                for(Map<UUID, List<UUID>> campaignTransportTypes : potentialTransports.values()  ) {
+                    Set<UUID> emptyTransports = new HashSet<>();
+                    for (UUID id : campaignTransportTypes.keySet()) {
+                        if (campaignTransportTypes.get(id).isEmpty()) {
+                            emptyTransports.add(id);
+                        }
+                    }
+                    for (UUID id : emptyTransports) {
+                        campaignTransportTypes.remove(id);
```
Scoppio, line 342: I'm unsure of what this is doing and where its going.

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -384,41 +410,44 @@ public void run() {
...
 
+                boolean alreadyResetTransport = false;
+
                 // All player and bot units have been added to the lobby
-                // Prompt the player to autoload units into transports
-                if (!scenario.getPlayerTransportLinkages().isEmpty()) {
-                    for (UUID id : scenario.getPlayerTransportLinkages().keySet()) {
+                // Prompt the player to autoload units into transport ships
+                if (!potentialTransports.get(SHIP_TRANSPORT.getTransportAssignmentType()).isEmpty()) {
+                    for (UUID transportId : potentialTransports.get(SHIP_TRANSPORT.getTransportAssignmentType()).keySet()) {
```
Scoppio, line 418: same as previous, hard to parse, the API of the new class should make this something very easy to read and obvious to any user. :)

MekHQ/src/mekhq/Utilities.java
```diff
@@ -1279,6 +1280,79 @@ public static void loadPlayerTransports(int trnId, Set<Integer> toLoad, Client c
...
+     */
+    public static void loadPlayerTransports(int trnId, Map<Integer, ? extends ITransportAssignment> toLoad, Client client,
+                                            boolean loadTactical, boolean isAlreadyReset) {
+        Set<Entity> alreadyTransportedEntities = new HashSet<>();
+
+        if (!loadTactical) {
+            // Nothing to do. Get outta here!
+            return;
+        }
+        Entity transport = client.getEntity(trnId);
```
Scoppio, line 1304: missing null check

MekHQ/src/mekhq/Utilities.java
```diff
@@ -1279,6 +1280,79 @@ public static void loadPlayerTransports(int trnId, Set<Integer> toLoad, Client c
...
+        // Reset transporter status, as currentSpace might still retain updates from
+        // when the Unit
+        // was assigned to the Transport on the TO&E tab
+        if (isAlreadyReset) {
+            alreadyTransportedEntities.addAll(transport.getLoadedUnits());
+        }
+
+        transport.resetTransporter();
```
Scoppio, line 1312: reset and already reset... those are unrelated?

MekHQ/src/mekhq/Utilities.java
```diff
@@ -1279,6 +1280,79 @@ public static void loadPlayerTransports(int trnId, Set<Integer> toLoad, Client c
...
+            } else {
+                if (transportAssignment.hasTransporterType()
+                        && Bay.class.isAssignableFrom(transportAssignment.getTransporterType())) {
+                    cargo.setTargetBay(selectBestBayFor(cargo, transport));
+                }
+            }
+        }
+
+        // Reset transporter status again so that sendLoadEntity can process correctly
+        transport.resetTransporter();
```
Scoppio, line 1339: why reset twice?

MekHQ/src/mekhq/Utilities.java
```diff
@@ -1279,6 +1280,79 @@ public static void loadPlayerTransports(int trnId, Set<Integer> toLoad, Client c
...
+            Entity cargo = client.getEntity(id);
+            if (cargo == null) {
+                continue;
+            }
+
+            ITransportAssignment transportAssignment = toLoad.get(id);
+
+            // Find a bay with space in it and update that space so the next unit can
+            // process, unless the unit isn't being loaded into a bay
+            if (transportAssignment.hasTransportedLocation()
```
Scoppio, line 1326: missing nullcheck

MekHQ/src/mekhq/Utilities.java
```diff
@@ -1279,6 +1280,79 @@ public static void loadPlayerTransports(int trnId, Set<Integer> toLoad, Client c
...
+                    && Bay.class.isAssignableFrom(transportAssignment.getTransportedLocation().getClass())) {
```
Scoppio, line 1327: `if (transportAssignment.getTransportedLocation() instanceof Bay)` would not work here?

MekHQ/src/mekhq/campaign/Campaign.java
Scoppio: I really wanted at least a way to make it very clear why each is used and when, with very short examples, so if we dont remove or consolidate any of them we know what they mean and when to use.
Another option would be to create a "class TransportService" and it initialize with the class, receiving a reference of the campaign itself and when you want to add things to transports, the functions are inside of it, and it can have all the specifics and docs inside of it.

MekHQ/src/mekhq/campaign/enums/CampaignTransportType.java
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * Enum for the different transport types used in MekHQ.
+ * Campaign Transports allow a unit to be
+ * assigned a transport (another unit).
+ * The different transport types primarily differ
+ * in the Transporters they allow.
+ * @see Unit
+ * @see Transporter
+ */
+public enum CampaignTransportType
```
Scoppio, line 18: This enum is way more complex than it needs to be, it is also trying to do the work of a factory and a validator.

It pains me to ask it, but I prefer that you split it into multiple files and behaviors. The thing you did is great, but the complexity of it is way more of an example of how far we can push the reflection features and enum but would not be something that I'd be confortable doing maintenance.

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,235 @@
...
+     * Main method to be used for unloading units from a transport
+     *
+     * @param transportedUnits Units we wish to unload
+     */
+    @Override
+    public void unloadTransport(Unit... transportedUnits) {
+        for (Unit transportedUnit : transportedUnits) {
+            unloadTransport(transportedUnit);
+        }
+    }
```
Scoppio, line 35: Prefer to go with `Iterable<Unit> transportedUnits`, if you have somewhere where you manually do `unloadTransport(u1, u2, u3, u3)` I suggest you just make a few helper functions, its very common tactic, other classes do it like `List.of` and `Map.of` both have up to 10 argument static constructors.

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Recalculates transport capacity - make sure you pass in all the transporters
+     * of a given type (class), or just all the transporters an entity has.
+     * @param transporters What transporters are we recalculating?
+     * @see Entity#getTransports()
+     */
+    @Override
+    public void recalculateTransportCapacity(@Nullable Vector<Transporter> transporters) {
+        if (transporters != null && !transporters.isEmpty()) {
```
Scoppio, line 52: add an early exit instead of wrapping everything in a nullcheck

```java 
if (transporters == null || transporters.isEmpty()) {
    return;
}
```

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Gets the different kinds of transporters the transport has
+     *
+     * @return Set of Transporter classes
+     */
+    @Override
+    public Set<Class<? extends Transporter>> getTransportCapabilities() {
+        return transportCapacity.keySet();
+    }
```
Scoppio, line 96: prefer to return an enum and not a set of classes, its not common to deal with classes as values unless you are making a framework.

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Returns true if the unit has capacity left for a transporter type
+     *
+     * @param transporterType Does the unit have free capacity in this type?
+     * @return True if the unit has capacity, false if not
+     */
+    @Override
+    public boolean hasTransportCapacity(Class<? extends Transporter> transporterType) {
+        return transportCapacity.containsKey(transporterType);
+    }
```
Scoppio, line 107: Again, try to use things like enum checks. Classes are not that bad, but can be avoided with simply doing thins like
```java
public <T extends Transporter> boolean hasTransportCapacity(T transporterType) {
    if (transporterType instanceof ...)
    //
```

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
```diff
@@ -0,0 +1,73 @@
...
+
+    Class<? extends Transporter> getTransporterType();
+
+    boolean hasTransporterType();
+
+    /**
+     * Where is this unit being transported?
+     * @return The transporter this unit is in
+     */
+    @Nullable Transporter getTransportedLocation();
```
Scoppio, line 51: I suggest you return optional instead of nullable, the result in the end is the same, but its harder to access an optional without checking it first.

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
```diff
@@ -0,0 +1,73 @@
...
+ * @see ITransportedUnitsSummary
+ * @see mekhq.campaign.enums.CampaignTransportType
+ */
+public interface ITransportAssignment {
+
+    /**
+     * The transport that is assigned, or null if none
+     * @return
+     */
+    @Nullable Unit getTransport();
```
Scoppio, line 39: I suggest you return optional instead of nullable, the result in the end is the same, but its harder to access an optional without checking it first.

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
```diff
@@ -0,0 +1,73 @@
...
+
+    boolean hasTransport();
+
+    Class<? extends Transporter> getTransporterType();
```
Scoppio, line 43: would prefer if this were an enum and not a class to be returned.

MekHQ/src/mekhq/campaign/unit/ITransportedUnitsSummary.java
```diff
@@ -0,0 +1,131 @@
...
+     * If this unit is capable of transporting another unit, return true
+     * @return true if the unit can transport another unit
+     */
+    boolean hasTransportCapacity();
+
+    /**
+     * Gets the different kinds of transporters the transport has
+     * @return Set of Transporter classes
+     */
+    Set<Class<? extends Transporter>> getTransportCapabilities();
```
Scoppio, line 56: same as before

MekHQ/src/mekhq/gui/menus/AssignForceToTacticalTransportMenu.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Constructor for a new tactical transport Menu
+     * @param campaign current campaign
+     * @param units selected units to try and assign
+     * @see CampaignTransportType#TACTICAL_TRANSPORT
+     */
+    public AssignForceToTacticalTransportMenu(Campaign campaign, Unit... units) {
+        super(TACTICAL_TRANSPORT, campaign, units);
+    }
```
Scoppio, line 34: prefer to use explicit type for list, so avoid varargs when building something that isnt for libs and frameworks as those may cause problems with memory leaks and things being stuck in the stack forever....

MekHQ/src/mekhq/Utilities.java
psikomonkie: It's poorly named, so I tried to rename it. This method is meant to be called in a sequence with other "transport" methods, like Ship transports. And when I went to look into renaming it, I realized it won't work properly if the player brings multiple transports. So this'll need reworked. Thanks!

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
psikomonkie: > prefer to return an enum and not a set of classes, its not common to deal with classes as values unless you are making a framework.

I'm struggling a bit with this one - not this particular usage, but switching the `Class<? extends Transporter>` to an enum. I feel like we're adding a lot of overhead to add a new Transporter class now by going down this route. In addition to adding it to the enum itself and the already existing map for converting an entity type into what it can be transported by, we'll need another map for converting Transporter classes to Transporter Enums and it's all just feels very... Redundant to map each Transporter subclass to an enum with nearly the same name.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: I agree, is this more sensible? 
```
if (potentialTransports.containsTransportKey(SHIP_TRANSPORT, transportShip.getId())) {
   potentialTransports.addTransportedUnit(SHIP_TRANSPORT, transportShip.getId(), unit.getId());
 }
```

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
psikomonkie: Done: replaced `Class<? extends Transporter>` with the `TransporterType` enum.

MekHQ/src/mekhq/campaign/unit/ITransportedUnitsSummary.java
psikomonkie: Done: replaced `Class<? extends Transporter>` with the `TransporterType` enum.

MekHQ/src/mekhq/campaign/enums/CampaignTransportType.java
psikomonkie: I took out what I think was the worst offender of this and moved it to `CampaignTransportUtilities`

MekHQ/src/mekhq/Utilities.java
psikomonkie: Thanks, fixed

MekHQ/src/mekhq/Utilities.java
psikomonkie: Thanks, fixed

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: >really wanted at least a way to make it very clear why each is used and when

That's the issue - they're doing literally the same thing. 

All three of the methods in my comment do the same thing. They add a Transport to the campaign's map of Transports. Two of the methods do it for specific CampaignTransportTypes, and the third one takes a CampaignTransportType as an additional parameter. 

**Should I keep the versions of the method that are specific (leaving us with 3 methods), or refactor them to use the one that takes a CampaignTransportType as a parameter (leaving us with 1 method)?** 

In either case, I'm going to improve the JavaDocs for this section to make sure it's more clear.

MekHQ/src/mekhq/AtBGameThread.java
Scoppio: I was thinking of something even simpler, like:
```java

if (transportRegistry.isRegisteredType(SHIP_TRANSPORT, transportShip)) {
    transportRegistry.embarkUnit(unit);
}
```

or 


```java

if (transportRegistry.isRegisteredAsTransportShip(transportShip)) {
    transportRegistry.embarkUnitInto(unit, transportShip);
}
```

MekHQ/src/mekhq/Utilities.java
Scoppio: Thanks

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
Scoppio: One way to do that is with factory classes.
like you have here a set of transport capabilities, what are those capabilities? just if something is a subclass of transporter? 
To dynamically decide if something is a class of something else I can check that is a simple `object instanceof Transporter`, dont even need the `<? extends Transporter>` in this case since all of them implement or extend transporter which means you should be tdine in here.

For something like a factory class, best way really is to return the correct type given some parameters. I like to use the interface itself as a factory sometimes.

```java

public interface Role {

    ASRange preferredRange();

    UnitRole getRole();

    boolean moveThroughCover(ASRange range);

    boolean tailTargets();

    boolean targetsLastAttacker();

    boolean dropToCoverIfDamaged();

    boolean disengageIfDamaged();

    boolean preferredTarget(UnitRole role);

    static Role getRole(UnitRole role) {
        return switch (role) {
            case AMBUSHER -> new Ambusher();
            case BRAWLER -> new Brawler();
            case JUGGERNAUT -> new Juggernaut();
            case MISSILE_BOAT -> new MissileBoat();
            case SCOUT -> new Scout();
            case SKIRMISHER -> new Skirmisher();
            case SNIPER -> new Sniper();
            case STRIKER -> new Striker();
            default -> new Brawler();
        };
    }
}
```

others I use a factory class with specific functions for that

```java
public class DamageApplierChooser {

    public static DamageApplier<?> choose(
        Entity entity, EntityFinalState entityFinalState) {
        if (entity instanceof Infantry) {
            return new InfantryDamageApplier((Infantry) entity, entityFinalState);
        } else if (entity instanceof Mek) {
            return new MekDamageApplier((Mek) entity, entityFinalState);
        } else if (entity instanceof GunEmplacement) {
            return new GunEmplacementDamageApplier((GunEmplacement) entity, entityFinalState);
        } else if (entity instanceof Tank) {
            return new TankDamageApplier((Tank) entity, entityFinalState);
        } else if (entity instanceof Aero) {
            return new AeroDamageApplier((Aero) entity, entityFinalState);
        }
        return new SimpleDamageApplier(entity, entityFinalState);
    }
    
}

File: MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java [hunk 32]
```diff
@@ -0,0 +1,257 @@
...
+     * Recalculates transport capacity - make sure you pass in all the transporters
+     * of a given type (class), or just all the transporters an entity has.
+     * @param transporters What transporters are we recalculating?
+     * @see Entity#getTransports()
+     */
+    @Override
+    public void recalculateTransportCapacity(@Nullable Vector<Transporter> transporters) {
+        if (transporters != null && !transporters.isEmpty()) {
+            // First let's clear the transport capacity for each transport type in transporters
+            for (Transporter transporter : transporters) {
```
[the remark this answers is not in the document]
Scoppio: codeql sometimes can be DENSE

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: We need to be able to differentiate between the two transport types at this point still, so passing in the CampaignTransportType is required. A unit could be assigned to be both a ship transport and a tactical transport for units, or a transport can have transported units of both types present, and it needs to be able to handle that. A convoluted example:
- The player is deploying a Leopard and 6 meks. 
- 4 of the meks are assigned the Leopard as a Tactical Transport. The other two meks are assigned that Leopard as a Ship Transport. 
- Once we figure out what needs loaded, we need to load them by transport type - if we load them all as one batch it'll be indeterminate which of the units will be loaded if there's more units assigned a transport than it has space. By keeping them separate up until we actually load them, we can assure our Ship Transported units are loaded first. 
- For this example, the two meks assigned the Leopard as a ship transport would be loaded, and two of the other Meks would be loaded. The remainder would be left unloaded as it's out of space now. 
- This weird use case came from how I like to use a smaller dropship as a "tactical" transport for a different group of meks once planet side. 

Generally the player won't be deploying a mix like this though, like I said it's a weird use case. And of course: the player is still able to edit this in the lobby if it's not what they want. 


---


anyway after typing all of that out, I think I figured out a way to implement what you're asking for - I'll get back to you later today with the results of that.

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -30,13 +31,14 @@
...
-     * Helps the menus need to check less when generating
+     * Helps the menus need to check less when generating Ship Transports can't use short-term
+     * transport types like InfantryCompartments or BattleArmorHandles. Use a Bay! Or DockingCollar
      *
      * @see Transporter
      * @param unit unit we want to get the Transporter types that could potentially hold it
      * @return Transporter types that could potentially transport this entity
      */
-    public Set<TransporterType> mapEntityToTransporters(Entity unit) {
+    public static Set<TransporterType> mapEntityToTransporters(CampaignTransportType campaignTransportType, Entity unit) {
```
Scoppio, line 41: I suggest another way to go here, I think you will like:


```java


    interface Visitor<T extends Entity> {
        boolean isInterestedIn(Entity entity);
        EnumSet<TransporterType> getTransporterTypes(T entity, CampaignTransportType campaignTransportType);
    }

    static List<Visitor> visitors = List.of(
        new Visitor<Tank>() {

            @Override
            public boolean isInterestedIn(Entity entity) {
                return entity instanceof Tank;
            }

            @Override
            public EnumSet<TransporterType> getTransporterTypes(Tank entity, CampaignTransportType campaignTransportType) {
                
                Set<TransporterType> transporters = new HashSet<>();

                if (unit.getWeight() <= 50) {
                    transporters.add(LIGHT_VEHICLE_BAY);
                }

                if (unit.getWeight() <= 100) {
                    transporters.add(HEAVY_VEHICLE_BAY);
                }

                if (unit.getWeight() <= 150) {
                    transporters.add(SUPER_HEAVY_VEHICLE_BAY);
                }
                return transporters;
            }
        },
        new Visitor<Infantry>() {

            @Override
            public boolean isInterestedIn(Entity entity) {
                return !(entity instanceof BattleArmor) && entity instanceof Infantry;
            }

            @Override
            public EnumSet<TransporterType> getTransporterTypes(Infantry entity, CampaignTransportType campaignTransportType) {
                Set<TransporterType> transporters = new HashSet<>();
                transporters.add(INFANTRY_BAY);

                //Ship transports can't use some transport types
                if (!(campaignTransportType.isShipTransport())) {
                    transporters.add(INFANTRY_COMPARTMENT);
                }
            
                return transporters;
            }
        }
    )
    
    public static Optional<Visitor<?>> getTransportTypeClassifier(Entity entity) {
        return visitors.stream().filter(v -> v.isInterestedIn(entity)).findFirst();
    }

    public static Set<TransporterType> mapEntityToTransporters(CampaignTransportType campaignTransportType, Entity unit) {
        return getTransportTypeClassifier(unit).map(v -> v.getTransporterTypes(unit, campaignTransportType)).orElse(Collections.emptySet());
    }
    
```

Of course the final implementation would have names that make more sense and would not rely on anonymous classes, but thats the idea.

Scoppio: So any change you want to do, there is a single file to change, if you want to add a new classifier you just add a new one, if there is a bug somewhere it is specific and in a 20 line file or less with very short functions.

psikomonkie: I do like it! I'll give it a go and see how it is. This block of code is one of my least favorite across everything I did, so anything to improve it I'll do.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: Done, there's a new `PotentialTransportsMap` for handling this.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: Now you can just pass a unit into `potentialTransports.tryToAddTransportedUnit(Unit)` and it will add it to the correct potential transport list.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: This has been moved into PotentialTransportsMap (and changed a little) as `removeEmptyTransports()`. It removes the transports from the map that don't have any transported units, so we don't need to keep checking it.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: This has been simplified to:
`if (potentialTransports.hasTransports(SHIP_TRANSPORT)) {
     for (UUID transportId : potentialTransports.getTransports(SHIP_TRANSPORT)) {`

MekHQ/src/mekhq/Utilities.java
psikomonkie: Okay, it's been reworked, and works now. So, if a Transport has both Ship & Tactical assigned transported units for some weird reason, this will prevent it from resetting the Ship transport's loaded units when doing the Tactical transported units.

If it was already reset, then let's save the currently transported Entities so we can restore them later. I added some comments to clarify this.

MekHQ/src/mekhq/Utilities.java
psikomonkie: It does, made the change, thanks!

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
psikomonkie: Done, didn't need the helper functions, it switched smoothly - just a lot of places to change :)

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
psikomonkie: Done

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
psikomonkie: Refactored `Class<? extends Transporter> transporterType` usages to a use a new Enum, `TransporterType`.

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
psikomonkie: Done

MekHQ/src/mekhq/campaign/unit/ITransportAssignment.java
psikomonkie: Done, thanks

MekHQ/src/mekhq/gui/menus/AssignForceToTacticalTransportMenu.java
psikomonkie: Done, thanks

MekHQ/src/mekhq/campaign/unit/ShipTransportedUnitsSummary.java
```diff
@@ -0,0 +1,190 @@
...
+     * and set
+     * both the target bay and the transport ship. Once in the MM lobby,
+     * this data
+     * will be used to actually load the unit into a bay on the transport.
+     *
+     * @param transportedUnits units being loaded
+     * @param transporterType type (Class) of bay or Transporter
+     * @return old transports; what were  the units' previous transport, if they had one?
+     */
+    public Set<Unit> loadTransportShip(Vector<Unit> transportedUnits, TransporterType transporterType) {
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/campaign/unit/ShipTransportedUnitsSummary.java
```diff
@@ -0,0 +1,190 @@
...
+        for (Unit transportedUnit : transportedUnits) {
+            Unit oldTransport = loadTransport(transporterType, transportedUnit);
+            if (oldTransport != null) {
+                oldTransports.add(oldTransport);
+            }
+        }
+
+        //Let's get matching transporters and then recalculate our transport capacity for this transporter type
+        Vector<Transporter> transporters = new Vector<>(transport.getEntity().getTransports().stream()
+            .filter(transporter -> TransporterType.getTransporterType(transporter) == transporterType).toList());
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/gui/menus/AssignForceToTransportMenu.java
```diff
@@ -0,0 +1,165 @@
...
+    protected abstract Set<TransporterType> filterTransporterTypeMenus(final Unit... units);
+
+    /**
+     * Different transporter type menus do different things when selected
+     * @param evt ActionEvent from the selection happening
+     * @param transporterType transporter type selected in an earlier menu
+     * @param transport transport (Unit) that will load these units
+     * @param units units being assigned to the transport
+     */
+    protected abstract void transportMenuAction(ActionEvent evt, TransporterType transporterType, Unit transport, Unit... units);
```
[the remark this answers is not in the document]
psikomonkie: intentional

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
psikomonkie: Done, thanks Luana!

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1672,13 +1394,11 @@
...
-                menuItem = new JMenuItem("Unassign Unit from Transport Ship");
-                menuItem.setActionCommand(TOEMouseAdapter.COMMAND_UNASSIGN_FROM_SHIP + unitIds);
-                menuItem.addActionListener(this);
-                menuItem.setEnabled(true);
-                popup.add(menuItem);
-            }
+            JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToShipTransportMenu(gui.getCampaign(), new HashSet<>(units)));
+            unassignShipTransportMenuClass(units, popup);
+
+            JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToTacticalTransportMenu(gui.getCampaign(), new HashSet<>(units)));
```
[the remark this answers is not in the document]
psikomonkie: Yes, it's replacing many deprecated calls with two

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1672,13 +1394,11 @@
...
             }
 
-            if (StaticChecks.areAllUnitsTransported(units) && !StaticChecks.areAnyUnitsDeployed(units)) {
-                menuItem = new JMenuItem("Unassign Unit from Transport Ship");
-                menuItem.setActionCommand(TOEMouseAdapter.COMMAND_UNASSIGN_FROM_SHIP + unitIds);
-                menuItem.addActionListener(this);
-                menuItem.setEnabled(true);
-                popup.add(menuItem);
-            }
+            JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToShipTransportMenu(gui.getCampaign(), new HashSet<>(units)));
```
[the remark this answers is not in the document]
psikomonkie: Yes, it's replacing many deprecated calls with two

MekHQ/src/mekhq/campaign/unit/AbstractTransportedUnitsSummary.java
```diff
@@ -0,0 +1,260 @@
...
+     * @see Entity#getTransports()
+     */
+    @Override
+    public void recalculateTransportCapacity(@Nullable Vector<Transporter> transporters) {
+        if (transporters == null || transporters.isEmpty()) {
+            return;
+        }
+
+        // First let's clear the transport capacity for each transport type in transporters
+        for (Transporter transporter : transporters) {
```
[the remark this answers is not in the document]
psikomonkie: Yes, bot, that's why I have the null guard.

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1169,149 +1162,12 @@
...
-                JMenuHelpers.addMenuIfNonEmpty(popup, menu);
-
-                if (StaticChecks.areAllUnitsTransported(unitsInForces)) {
-                    menuItem = new JMenuItem(TOEMouseAdapter.UNASSIGN_FORCE_TRN_TITLE);
-                    menuItem.setActionCommand(TOEMouseAdapter.COMMAND_UNASSIGN_FROM_SHIP + unitIds);
-                    menuItem.addActionListener(this);
-                    menuItem.setEnabled(true);
-                    popup.add(menuItem);
-                }
+                JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToTacticalTransportMenu(gui.getCampaign(), new HashSet<>(units)));
```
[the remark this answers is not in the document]
psikomonkie: Yes, it's replacing many deprecated calls with two

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -1169,149 +1162,12 @@
...
-                                        ship.getCurrentMekCapacity()));
-                                m_trn.setEnabled(true);
-                            }
-
-                            if (ship.getProtoMekCapacity() > 0) {
-                                pm_trn.add(transportMenuItem(ship.getName(), id, unitIds.toString(),
-                                        ship.getCurrentProtoMekCapacity()));
-                                pm_trn.setEnabled(true);
-                            }
+                JMenuHelpers.addMenuIfNonEmpty(popup, new AssignForceToShipTransportMenu(gui.getCampaign(), new HashSet<>(units)));
```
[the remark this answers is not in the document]
psikomonkie: Yes, it's replacing many deprecated calls with two

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -0,0 +1,244 @@
...
+     * In addition, CampaignTransportTypes that can't use certain TransporterTypes is handled,
+     * like Ship Transports not being able to use InfantryCompartments or BattleArmorHandles.
+     * Use a Bay! Or DockingCollar.
+     *
+     * @see TransporterType
+     * @param campaignTransportType type (enum) of campaign transport - some transport types can't use certain transporters
+     * @param unit unit we want to get the Transporter types that could potentially hold it
+     * @return Transporter types that could potentially transport this entity
+     */
+    public static EnumSet<TransporterType> mapEntityToTransporters(CampaignTransportType campaignTransportType, Entity unit) {
```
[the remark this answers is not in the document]
psikomonkie: Yes it is

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -384,41 +383,43 @@ public void run() {
...
+                        Unit transportShip = campaign.getUnit(transportId);
                         Set<Integer> toLoad = new HashSet<>();
                         // Let the player choose to load DropShips, Small Craft, fighters, and/or
                         // ground units on each transport
-                        if (transport.getTransportedUnits().stream()
+                        if (transportShip.getShipTransportedUnits().stream()
                                 .anyMatch(unit -> unit.getEntity().getUnitType() == UnitType.DROPSHIP)) {
                             loadDropShips = JOptionPane.YES_OPTION == JOptionPane.showConfirmDialog(null,
-                                    String.format(AtBGameThread.LOAD_DROPSHIP_DIALOG_TEXT, transport.getName()),
+                                    String.format(AtBGameThread.LOAD_DROPSHIP_DIALOG_TEXT, transportShip.getName()),
```
Scoppio, line 403: just marking to say that this should be using an i18n resource

MekHQ/src/mekhq/campaign/Campaign.java
Scoppio: CampaignTransportType param it is

MekHQ/src/mekhq/campaign/enums/CampaignTransportType.java
```diff
@@ -0,0 +1,113 @@
...
+     * Different Transport Types use different transported units summaries.
+     * @return Transported Unit Summary used by this transport type
+     */
+    public Class<? extends AbstractTransportedUnitsSummary> getTransportedUnitsSummaryType() { return transportedUnitsSummaryType; }
+    // endregion Getters
+
+
+    public String getName() {
+        return this.toString();
+    }
```
Scoppio, line 112:
```suggestion
```
the `Enum.name()` already exists and works fine, no need for a toString `getName`

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -427,23 +428,71 @@ public void run() {
...
+                                Unit transportedUnit = campaign.getUnit(cargoId);
+                                if (transportedUnit != null && transport.getEntity().canLoad(transportedUnit.getEntity())) { //transportedUnit.getTransportAssignment().getTransporterType()) {
+                                    // Convert the list of Unit UUIDs to MM EntityIds
+                                    toLoad.put(transportedUnit.getEntity().getId(), transportedUnit.getTacticalTransportAssignment());
+                                    if (scenario.getPlayerTransportLinkages().containsKey(transportId)) {
+                                        scenario.addPlayerTransportRelationship(transportId, transportedUnit.getId());
+                                    }
+                                    if (transportedUnit.getTech() != null && transportedUnit.getTacticalTransportAssignment().hasTransporterType()
+                                            && Bay.class.isAssignableFrom(transportedUnit.getTacticalTransportAssignment().getTransporterType().getTransporterClass())) {
+                                        cargoTechs.add(transportedUnit.getTech());
```
Scoppio, line 484: this is the only place I could find where a class instance comparision is being made, I and the result is not the best... maybe add this validations inside the class, something like `canTransportX`
Other than that, it looks great

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -0,0 +1,244 @@
...
+                    }
+
+                    if (entity.getWeight() <= 100) {
+                        transporters.add(HEAVY_VEHICLE_BAY);
+                    }
+
+                    if (entity.getWeight() <= 150) {
+                        transporters.add(SUPER_HEAVY_VEHICLE_BAY);
+                    }
+                    return EnumSet.copyOf(transporters);
```
Scoppio, line 114: no need to make a copy, you can just return the hashset

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -0,0 +1,244 @@
...
+public class CampaignTransportUtilities {
+    // region Static Helpers
+
+    interface Visitor<T extends Entity> {
+        boolean isInterestedIn(Entity entity);
+
+        EnumSet<TransporterType> getTransporterTypes(T entity, CampaignTransportType campaignTransportType);
+
+
+        List<Visitor> visitors = List.of(
```
Scoppio, line 40: move the list off from the interface, you can set as static final

Scoppio (approved):
Looks great as always :D
some points of attention but nothing major.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: It's a legacy string! I updated all 8 of these to i18n.

MekHQ/src/mekhq/AtBGameThread.java
psikomonkie: Made a new method in `TransportAssignment` for checking `isTransportedInBay`

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
psikomonkie: Nice, thanks, fixed.

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
psikomonkie: Fixed, thanks

MekHQ/src/mekhq/campaign/enums/CampaignTransportType.java
psikomonkie: Removed, thanks

psikomonkie:
Finished making changes requested during review, let's get this merged if possible :)

Saklad5:
Phenomenal work.

Saklad5:
> I think this could be reworked with a little more abstraction to better support assigning a unit to another unit but not for transport / transporter related reasons.

The main thing that comes to mind is stuff like paramedics and Mobile Field Bases, but that can probably be addressed better by the new MekHQ non-combat team designations.

Saklad5:
Actually, I can think of something: could C3 be refactored to use this code?

psikomonkie:
> > I think this could be reworked with a little more abstraction to better support assigning a unit to another unit but not for transport / transporter related reasons.
> 
> The main thing that comes to mind is stuff like paramedics and Mobile Field Bases, but that can probably be addressed better by the new MekHQ non-combat team designations.

I haven’t looked at Mobile Field Bases too closely but I did see ones that have Bays. That’d JustWork with this. 


> Actually, I can think of something: could C3 be refactored to use this code?

I don’t know how C3 works in MekHQ, haven’t looked into it, but if it works I don’t think it should be refactored. I was originally going to leave the Ship Transport code as-is but the bugs were such a pain to fix I finally gave in replaced it. Likewise, unless we have serious issues with C3 in MekHQ, I don’t think it’s worthwhile.

Saklad5:
> I haven’t looked at Mobile Field Bases too closely but I did see ones that have Bays. That’d JustWork with this.

Mobile Field Bases are basically transport bays without the transport, consisting solely of the assorted repair/refit gear.

> Likewise, unless we have serious issues with C3 in MekHQ, I don’t think it’s worthwhile.

C3 network assignments seem to get dropped when units are damaged, at least in 0.50.0, but it's out-of-scope for this particular merge either way.

PR #5644, opened 2025-01-04
Fixed Value Display in Resupplies

IllianiCBT:
Resupplies were incorrectly using sell value for both resupply cost and resupply evaluation. This made it seem like resupplies were a much worse value proposition than they actually are.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/GenerateResupplyContents.java
+++ b/MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/GenerateResupplyContents.java
@@ -208,11 +208,13 @@ public class GenerateResupplyContents {
      */
     private static void calculateConvoyWorth(Resupply resupply) {
         List<Part> convoyContents = resupply.getConvoyContents();
-        Money value = Money.zero();
+        Money sellValue = Money.zero();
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/GenerateResupplyContents.java
```diff
@@ -222,7 +224,7 @@ private static void calculateConvoyWorth(Resupply resupply) {
 
         // Smugglers always double the cost of the supplies they're offering
         if (resupplyType.equals(ResupplyType.RESUPPLY_SMUGGLER)) {
-            resupply.setConvoyContentsValueCalculated(value.multipliedBy(2));
+            resupply.setConvoyContentsValueCalculated(sellValue.multipliedBy(2));
```
IllianiCBT, line 227: We use 'sell value' as that is the lower of the two valuations. We use this to ensure the user will always get a deal from employer resupplies, while (at most) getting buy value from smugglers.

We use buy value for smugglers due to the risk of being scammed. This ensures that in the event the user is scammed they're not also paying out the nose for the privilege.

HammerGS (approved):

PR #5658, opened 2025-01-05
cleanup: Declaring a method parameter as final for an interface metho…

firefly2442:
…d is useless because the implementation may choose to not respect it.

https://docs.pmd-code.org/pmd-doc-7.9.0/pmd_rules_java_codestyle.html#finalparameterinabstractmethod


firefly2442:
Hmm, I might need some help on this one.  Should those functions the Github bot marked be public?

Scoppio:
> Hmm, I might need some help on this one. Should those functions the Github bot marked be public?

this bot is crazy.

But yes, they are supposed to be public, interfaces only allow for protected or private functions in the most modern versions os java (I believe java 18 or 21).

firefly2442:
PMD no longer flags these as issues.  Double check me here though, I'm pretty rusty with Java.

Issue #5663, opened 2025-01-06
[Issue] Out-of-Bounds Exception from New Finance Table Sorters


IllianiCBT:
### Brief Description *

The new finance table sorters added in #5642 do not account for the user having an empty finance table. Resulting in an out-of-bounds exemption if the finance table is ever empty. Such as when creating a new campaign, or if all transactions are deleted.

### 3. Steps to Reproduce

1. Create a new campaign
2. Start the campaign

or

1. Remove all transactions from an existing campaign

### Attach Files

[mekhq.log](https://github.com/user-attachments/files/18312756/mekhq.log)

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MekHQ Suite Version *

0.50.03

### Operating System *

macOS 15.1.1 Sequoia

### Java Version *

J17

### Final Verification

- [X] I confirm this is a single, unique issue that hasn't been reported before
- [X] I have included all necessary information and files to help reproduce this issue
- [X] I have asked on MegaMek Discord about this issue
- [X] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiCBT:
Closing as resolved

PR #5666, opened 2025-01-06
fix: addresses bug #5663 where the financeTable is empty (has zero ro…

firefly2442:
…ws) resulting in array out of bounds exception

Addresses #5663 

Sorry about that, I *think* this should fix it.  At least it did in my minimal testing.


IllianiCBT (approved):
Easy mistake to make. Lord knows I've done worse with less! :D

Looks good :)

HammerGS (approved):

PR #5693, opened 2025-01-07
Removed SP Cost for GM Reinforcement Attempts in StratCon

IllianiBird:
- Removed SP cost when using `Reinforce (GM)` option in StratCon.


Sleet01 (approved):
LGTM

Issue #5697, opened 2025-01-07
[MHQ - 50.03 Nightly CI1729] Deploying a 'Patrol' role force onto a hostile facility does not generate a facility scenario in the briefing roo,


UlyssesSockdrawer:
### Brief Description *

Screenshots:

Unit deployed to facility:
![image](https://github.com/user-attachments/assets/f177d244-a390-40df-8a49-1623a0e7d405)

No scenario gen'd in briefing room:
![image](https://github.com/user-attachments/assets/dde7b9e1-fd44-4403-9f24-ab56a53c7ed9)

Description:

As mentioned in Discord QA channel, when deploying a Patrol role force to a facility, no scenario is generated. The expected behaviour is that a scenario should be generated.

### 3. Steps to Reproduce

1. Add a hostile facility to the map via GM mode, or discover one naturally.
2. Deploy a 'Patrol' role force onto the facility hex.
3. No scenario generated in briefing room.

### Attach Files

Log:
[mekhq.log](https://github.com/user-attachments/files/18339355/mekhq.log)

Pre-deploy save with GM facility added to map:
[The Orphan Eater's Batshit Bastards24750518-PatrolonFacilityDeployTestPre.cpnx.gz](https://github.com/user-attachments/files/18339358/The.Orphan.Eater.s.Batshit.Bastards24750518-PatrolonFacilityDeployTestPre.cpnx.gz)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

50.03 CI1729 Nightly

### Operating System *

Win11

### Java Version *

 Java Vendor: Eclipse Adoptium     Java Version: 17.0.12

### Final Verification

- [X] I confirm this is a single, unique issue that hasn't been reported before
- [X] I have filled and provided all necessary information above
- [X] I have included any and all logs, custom units, and CPNX (save) files
- [X] I have asked on MegaMek Discord about this issue
- [X] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiCBT:
Closing as resolved

Issue #5603, opened 2025-01-02
Engage targets for Null duration


Tempest-56:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MekHQ documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

![Untitled](https://github.com/user-attachments/assets/6be79058-8b01-4ae1-a30c-da024615634f)

This is probably not supposed to happen.

### Operating System *

Windows 11

### Java Version *

17

### MekHQ Suite Version *

Free Text (type manually)

### Custom MekHQ Version

v0.50.02

### Attach Files

[Celestial Cavalry30230416.cpnx.gz](https://github.com/user-attachments/files/18287421/Celestial.Cavalry30230416.cpnx.gz)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

IllianiCBT:
Regenerating fixes it, so this is going to be a pain to properly resolve

IllianiCBT:
Closing as resolved

PR #5698, opened 2025-01-08
Fixed Multiple StratCon Bugs

### LINKED ISSUE #5636: [Issue] Jan 3/4th NI - Forces Regenerating whenever units are deployed
### Brief Description *

It may be possible that this is actually at least mildly intended. As to allow the player to use whatever lance they wish during a deployment; however, due to the ever persistent modifier duplication bug, this is having unintended consequences as it is, well, duplicating modifiers. Easy enough to workaround however by just deleting the associated bots.
Low Priority Issue

### 3. Steps to Reproduce

1. Load the attached save
2. GM Remove the active scenario [that way you can advance the calendar]
3. Advance to Sunday. This should spawn in one the pre allocated scenarios.
4. Assign any of the possible lances to the scenario and ignore leadership units. For simplicities sake deploy First Battle Lance as they dont have leadership values.
5. Click off of the mission in the briefing room and click back onto it. The forces should be regenerated.

Before Deployment
image

After Deployment
image

From all my tests it seems to regenerate regardless if the assigned lance is the designated lance or not. For situations without modifiers this isn't noticeable, but with modifiers involved the duplication bug comes into play and it becomes very obvious that it is happening. 

### Attach Files

[Issue Package.zip](https://github.com/user-attachments/files/18306035/Issue.Package.zip)

### Severity *

### MekHQ Suite Version *

Jan 3/4 NI

### Operating System *

Windows 10

### Java Version *

Java 17 Adoptium

### Final Verification

- [X] I have included all necessary information and files to help reproduce this issue

IllianiCBT:
- Stopped scenarios regenerating when reinforcements were applied.
- Stopped Patrol forces failing to generate scenarios if they land on top of an enemy facility
- Cleaned up some leftover code from prior to 50.02
- Ensured scenario objective time limits are recalculated when recalculating objectives

Fix #5636, #5697, #5603

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
@@ -1351,6 +1351,7 @@ public class AtBDynamicScenarioFactory {
         scenario.getScenarioObjectives().clear();
 
         for (ScenarioObjective templateObjective : scenario.getTemplate().scenarioObjectives) {
+            logger.info("templateObjective:"+ templateObjective.toString());
             ScenarioObjective actualObjective = translateTemplateObjective(scenario, campaign, templateObjective);
 
             scenario.getScenarioObjectives().add(actualObjective);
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
Looks good.

PR #5707, opened 2025-01-08
Corrected Available Force Check

IllianiCBT:
- Added missing isDeployed check to `getAvailableForceIDs` (first overload).
- Removed isDeployed check from `getAvailableForceIDs` (second overload). This will prevent Patrol forces from being over-emphasized as seed forces.
- Moved isDeployed check to 'generate weekly scenarios' so that only Integrated contracts care if a force has already been deployed. For all other contract command rights we're just using this information for the purposes of seeding the scenario, so whether the force is deployed or not is irrelevant.


Sleet01 (approved):
LGTM

PR #5708, opened 2025-01-08
Added a Soft Cap to Support Point Generation

IllianiCBT:
- Added a soft cap to Support Points, where Administrators will not add new Support Points (during weekly generation) if Support Points already meet or exceed the number of required forces for the Contract. We received feedback that Support Points were a bit generous, under the current implementation, because the user could just stockpile them indefinitely.


Sleet01 (approved):
LGTM

PR #5709, opened 2025-01-08
Added `MHQDialogImmersive` for Immersive Dialog Functionality that Immersively Improves Immersion

IllianiCBT:
- Introduced a new class, `MHQDialogImmersive`, to provide an immersive dialog interface with support for customizable speakers, messages, and buttons.
- Renamed `AbstractMHQDialog` to `AbstractMHQDialogBasic` to clarify its purpose and use-cases.

This new dialog is designed to allow developers to easily generate dialogs that conform with the new 'incoming transmission' style debuted in 50.02.

I wrote this class in a way that encourages developers to keep functionality code out of dialogs, which makes the functions a _lot_ easier to write Unit Tests for, as we test in a headless environment, so hiding code inside dialogs makes that code effectively untestable.

I've also ensured that any interaction with the dialog will cause it to be dismissed, rather than simply hidden. This ensures we're releasing memory when the dialog is closed, rather than keeping it in memory until the client restarts or such a time GC decides to finally eat it.

At time of writing this class is not used, as I wanted to be kind to whoever reviews this and allow them to focus on reviewing the new class, rather than needing to worry about the 101 implementations of said class.


MekHQ/src/mekhq/gui/baseComponents/MHQDialogImmersive.java
```diff
@@ -0,0 +1,417 @@
...
+import javax.swing.*;
+import java.awt.*;
+import java.util.List;
+import java.util.ResourceBundle;
+
+import static mekhq.campaign.force.Force.FORCE_NONE;
+import static mekhq.gui.dialog.resupplyAndCaches.ResupplyDialogUtilities.getSpeakerIcon;
+import static mekhq.utilities.ImageUtilities.scaleImageIconToWidth;
+
+public class MHQDialogImmersive extends JDialog {
```
IllianiCBT, line 39: I couldn't determine whether this should be called `MHQDialogImmersive` or `AbstractMHQDialogImmersive`. Guidance would be appreciated. :)

MekHQ/src/mekhq/gui/dialog/ContractAutomationDialog.java
```diff
@@ -1,3 +1,21 @@
+/*
```
IllianiCBT, line 1: Apparently, I forgot to include the copyright notice when I wrote this class in 2024, so adding it here.

MekHQ/src/mekhq/gui/baseComponents/MHQDialogImmersive.java
```diff
@@ -0,0 +1,417 @@
+/*
```
IllianiCBT, line 1: **Note for Reviewer:** this is the only class that has had functionality change/be introduced. The rest is just housekeeping.

MekHQ/src/mekhq/gui/baseComponents/MHQDialogImmersive.java
Sleet01: By our standards, `Abstract...` should only be used in the name of a class that _is_ Abstract.
Only a class that does not implement, or inherit from another class, its own methods should be Abstract, so
in this case `MHQDialogImmersive` is more correct.

Sleet01 (approved):
Looks good!

Sleet01:
@IllianiCBT Hit a conflict after merging the prior patches.

PR #5720, opened 2025-01-09
Added Handling for Unsupported Unit Types

IllianiCBT:
- Added handling for unsupported unit types (currently only Gun Emplacements).

Unsupported unit types are handled in the following manner:
1. Automatically sold on load (to handle older campaigns).
2. Blocked from being salvaged.
3. Blocked from purchase (including GM purchase).

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
@@ -21,6 +21,8 @@ package mekhq.gui.dialog;
 import megamek.client.generator.RandomCallsignGenerator;
 import megamek.client.generator.RandomNameGenerator;
 import megamek.client.ui.swing.util.UIUtil;
+import megamek.common.Entity;
+import megamek.common.GunEmplacement;
 import megamek.common.MekSummaryCache;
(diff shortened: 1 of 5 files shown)


Sleet01 (approved):

MekHQ/src/mekhq/gui/dialog/DataLoadingDialog.java
```diff
@@ -422,12 +427,60 @@ public Campaign doInBackground() throws Exception {
...
+                for (Unit soldUnit : soldUnits) {
+                    retVal.getQuartermaster().sellUnit(soldUnit);
+                    message.append("- ").append(soldUnit.getName()).append("<br>");
+                }
+
+                JOptionPane.showMessageDialog(null,
+                    String.format("<html>%s</html>", message),
+                    resources.getString("unsupportedUnits.title"),
+                    JOptionPane.WARNING_MESSAGE);
+            }
```
Scoppio, line 481: this can't be in a static function because it has reference to the resources, which is in the base instance.

IllianiCBT: That was fixed a while ago - if you're referring to what I think you are. If you update your master it should be good.

PR #5735, opened 2025-01-09
fix: compilation error due to a function being set as static but acessing instance data

Scoppio:
fix: compilation error due to a function being set as static but acessing instance data


IllianiCBT (approved):

Scoppio:
I just realised you already solved the problem

IllianiCBT:
I can confirm it was not.

IllianiCBT (approved):

PR #5741, opened 2025-01-10
Fixed Negative Shares Payouts; Removed `TotalEarnings` Tracking and Associated Functionality.

### LINKED ISSUE #5736: [Issue] Processing pay on Aug 1st 3050 seems to introduce massive unsustainable debt on seemingly random personnel. Y3050 error?
### Brief Description *

[logs.zip](https://github.com/user-attachments/files/18368930/logs.zip)
Campaign has before, autosave, after. Processing on 0.50.03 01/06 nightly reliably produced personnel with debt where previously there was none. Maybe they got in deep on the wrong side of a Solaris match, but it seems oddly specific that this one day seemed to incur debt. Afterwards (I advanced to December to see) the pay / debt remained stable and wages went out normally, so it doesn't seem to affect the unit. Tracking pay earned is an optional feature, and I understand it is resource intensive due to reoccurring reads of the personnel, but it seems like this has a very specific point where it occurs on aug 1 3050.

image

As an example of a randomly indebted personnel.

### 3. Steps to Reproduce

1. Load save.
2. Check Jesus and other personnel.
3. Advance day.
4. Check again. 

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

0.50.03 SNAPSHOT from 01/06 Nightly

### Operating System *

Windows 10

### Java Version *

17

### Final Verification

- [X] I have filled and provided all necessary information above
- [X] I have included any and all logs, custom units, and CPNX (save) files

IllianiCBT:
- Added a check to ensure end-of-contract Shares payouts are not made if they would be negative. This would result in the personnel paying the campaign.
- Removed the `TotalEarnings` field, related methods, and all references in the codebase. I opted to remove this data point as from what I can tell it was intended for further shares functionality that was never implemented.

While I expect this data point will be missed by some users, it resulted in a bunch of extra ancillary methods and extra process calls just so users could see how much Joe MechWarrior earned. By reducing the amount of personnel parsing we do we can save a little here and there to improve overall performance.

Fix #5736

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -7830,25 +7830,13 @@ public class Campaign implements ITechManager {
                 ResourceBundle financeResources = ResourceBundle.getBundle("mekhq.resources.Finances",
                         MekHQ.getMHQOptions().getLocale());
 
-                Money shares = remainingMoney.multipliedBy(contract.getSharesPercent()).dividedBy(100);
-                remainingMoney = remainingMoney.minus(shares);
+                if (remainingMoney.isGreaterThan(Money.zero())) {
+                    Money shares = remainingMoney.multipliedBy(contract.getSharesPercent()).dividedBy(100);
+                    remainingMoney = remainingMoney.minus(shares);
 
-                if (getFinances().debit(TransactionType.SALARIES, getLocalDate(), shares,
-                        String.format(financeResources.getString("ContractSharePayment.text"), contract.getName()))) {
-                    addReport(financeResources.getString("DistributedShares.text"), shares.toAmountAndSymbolString());
-
-                    if (getCampaignOptions().isTrackTotalEarnings()) {
-                        boolean sharesForAll = getCampaignOptions().isSharesForAll();
-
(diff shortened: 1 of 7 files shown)


psikomonkie:
> - Removed the `TotalEarnings` field, related methods, and all references in the codebase. I opted to remove this data point as from what I can tell it was intended for further shares functionality that was never implemented.
> 
> 
> 
> While I expect this data point will be missed by some users, it resulted in a bunch of extra ancillary methods and extra process calls just so users could see how much Joe MechWarrior earned. By reducing the amount of personnel parsing we do we can save a little here and there to improve overall performance.
> 
> 
> 
> Fix #5736

I actually really like the employee earnings feature. My favorite use for it is for MekWarrirors that own their own Mek - use their earnings to "fund" (GM add funds) pay for Mek upgrades/customizations. Like, we let people own their Meks so why am I responsible for paying to pimp their ride?

Anyway, where I'm going with all this: **Would you mind if I took a whack at the bug before resolving it this way?** We can chat in more detail on Discord if you want, I just want to put the offer out there.

Thanks,
Psi

IllianiCBT:
Go for it. I'll mark this as draft for now.

Issue #5715, opened 2025-01-09
[Issue] Typo/better grammar possible on personnel blurb


OrbMonky:
### Brief Description *

![image](https://github.com/user-attachments/assets/c4c58d3d-9a1f-46fb-a73f-875feada665b)
'Susan are' might be better as 'is' instead.

### Severity *

Low (Minor/Nuisance): Minor glitches or cosmetic issues that don't affect gameplay and occur rarely.

### MekHQ Suite Version *

0.50.03 SNAPSHOT from 01/06 Nightly

### Operating System *

Windows 10

### Java Version *

17

### Final Verification

- [X] I confirm this is a single, unique issue that hasn't been reported before
- [X] I have filled and provided all necessary information above
- [X] I have included any and all logs, custom units, and CPNX (save) files
- [X] I have asked on MegaMek Discord about this issue
- [X] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PR #5749, opened 2025-01-10
Minor grammar fixes

repligator:
Mostly "are" to "is". See https://github.com/MegaMek/mekhq/issues/5715.

### DIFF
--- a/MekHQ/resources/mekhq/resources/Personalities.properties
+++ b/MekHQ/resources/mekhq/resources/Personalities.properties
@@ -1015,7 +1015,7 @@ PersonalityQuirk.QUICK_TO_QUIP.description=%s often crack%s jokes or make sarcas
   intense combat situations, using humor as a coping mechanism, even if it irritates their lancemates.
 
 PersonalityQuirk.TECH_SKEPTIC.text=Distrusts New Technology
-PersonalityQuirk.TECH_SKEPTIC.description=%s are wary of new tech and equipment, preferring older, \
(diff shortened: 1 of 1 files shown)


IllianiCBT (changes requested):
This isn't an appropriate fix, as the value of %s can be either the character's name, or the gender neutral 'they'

repligator:
Quick update: English is hard. For some reason words can become plural when the subject is singular and singular when the subject is plural. Sometimes they even become singular when the subject is singular and plural when the subject is plural.

Sue fidget**s** with their hair.
They fidget with their hair.

This is in addition to is/are.

Sue **is** smart.
They **are** smart.

You can sort of dodge the first issue by awkwardly rewording every sentence. For example, the following only needs the first two words in each sentence swapped to switch from given name to they.

Sue is often seen fidgeting with their hair.
They are often seen fidgeting with their hair.

However reading a paragraph where every sentence is written like this (tense? voice?) sounds incredibly unnatural. An ideal solution should be extensible, elegant, require as few resources as possible, and have natural sounding output. I haven't thought of anything that meets those criteria yet.

IllianiCBT:
Converted this to 'draft' until a solution can be found, as we don't want it being merged by mistake

IllianiCBT:
At some point in 50.04 I'm going to switch to using Internationalization for the resource strings. That will give us a lot more flexibility when it comes to writing and editing these. When that happens I'll let you know so we can figure the best approach.

repligator:
Sounds good.

IllianiCBT:
This has been superseded by #5998. Closing - and thanks for the inspiration for solving this issue! <3

PR #5757, opened 2025-01-11
Added Active Contract Warning on Campaign Load

IllianiCBT:
### Requires: https://github.com/MegaMek/megamek/pull/6414

- Added a warning that requests the user complete all active contracts when updating to a new client version.
- This warning is triggered if the campaign being saved was last saved in an older version and has an active contract.
- Active contracts are defined as any contract which has already started, or has been accepted but not yet started (i.e. the user is still en route to the contract).
- The warning does not stop the campaign from loading, to ensure we can still load older saves when investigating bugs and what have you.


MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -581,11 +583,59 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
...
+                    "\n\nPlease complete the contract before updating to version %s.",
+                MHQConstants.VERSION);
+
+            JOptionPane.showMessageDialog(
+                null,
+                message,
+                "Error",
+                JOptionPane.ERROR_MESSAGE
+            );
+        }
```
Scoppio, line 636: I believe there are two variables here, `hasActiveContract` and `hasFutureContract`
then the message dialog should then show `if (hasActiveContract(campaign) || hasFutureContract(campaign))`

Make both variables static methods that return a boolean, this will make the calculation of the `hasFutureContract` be conditioned by the first, so if there is no current active contract it will check if there is a future active contract. The two new static functions are small and could be used for many things in the future.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -581,11 +583,59 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
                         }
                     });
 
+
+
+        if (MHQConstants.VERSION.isHigherThan(version)) {
+            triggerActiveContractWarning(retVal);
+        }
```
Scoppio, line 590: I think I'd prefer a trigger/warning in somewhere like MekHQ or in a GUI, not in the XML Parser. static checks like returning the information if it is in the correct version or not should be done in a place specific to the campaign + application.

I'm not hard on this, could go either way, just don't like calling a GUI think inside an XML parser. To me this kind of thing has to be handled by the application, otherwise we could have something like a GUI less application trying to run in the CMD and throwing errors when loading the XML because it requires a frame.

IllianiCBT:
- Updated to include feedback from Luana

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -581,6 +583,12 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
                         }
                     });
 
+        if (MHQConstants.VERSION.isHigherThan(version)) {
+            if (retVal.hasActiveAtBContract(true)) {
+                new ActiveContractWarning();
+            }
+        }
+
```
Scoppio, line 591: My suggestion would be something like the following, but inside the CampaignFactory file, with the CampaignXMLParser returning a container with the campaign and the relevant diagnostics:

Somewhere add this:
```java
    // Somewhere
    public enum CampaignProblemType {
        NONE,
        ACTIVE_CONTRACT,
        ACTIVE_FUTURE_CONTRACT,
        CANT_LOAD_FROM_VERSION,
        CANT_PARSE_FILE,
    }

    public record ParsedCampaign(Campaign campaign, String currentMekHQVersion, String fileMekHQVersion, CampaignProblemType problem, String diagnostics) { 
    public boolean hasProblems() {
            return problem != null;
        }
    }

```

In the CampaignFactory
```

    public ParsedCampaign createCampaign(InputStream is)
        throws CampaignXmlParseException, IOException, NullEntityException {
        if (!is.markSupported()) {
            is = new BufferedInputStream(is);
        }

        byte[] header = readHeader(is);

        // Check if the first two bytes are the GZIP magic bytes...
        if ((header.length >= 2) && (header[0] == (byte) 0x1f) && (header[1] == (byte) 0x8b)) {
            is = new GZIPInputStream(is);
        }
        // ...otherwise, assume we're an XML file.

        CampaignXmlParser parser = new CampaignXmlParser(is, this.app);
        ParsedCampaign parsedCampaign = parser.parse();
        return parsedCampaign;
    }

```

This way, the place that receives the file will be able to handle it accordingly.

DataLoadingDialog.java
```java

                // And then load the campaign object from it.
                try (FileInputStream fis = new FileInputStream(getCampaignFile())) {
                    ParsedCampaign parsedCampaign = CampaignFactory.newInstance(getApplication()).createCampaign(fis);
                    if (parsedCampaign.hasProblems()) {
                        // TODO: CALL ERROR HANDLING FUNCTION FOR EACH POSSIBLE PROBLEM
                    }
                    campaign.restore();
                    campaign.cleanUp();
                }
                // endregion Progress 6

```

Otherwise we keep with the problem where we have being imported stuff inserted inside a parser class.

Sorry for not being clear the first time.

Scoppio (changes requested):

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
Scoppio: The "parse()" and the diagnostics can run in two different steps, this way not disrupting any tests or old files and just adding new functionality.

IllianiCBT: So I had to wait until my brain had the cycles to process this, but now that it does what you're saying makes perfect sense and will be super maintainable. I would have never thought to use Enums like this. Thanks for the feedback! :)

IllianiCBT:
- incorporated Luana's suggestions

PR #5781, opened 2025-01-14
Feat/acar formations reintroduced

Scoppio:
# What it does

Add the concept of orders, the orders allow you to create a set of "orders" to give to your units, things like giving a priority target, retreating after X turns, leaving the battlefield through north or south, force enemy to retreat (focus attack only enemies not withdrawing) or stop enemies from retreating (focust attack on only enemies withdrawing).

The orders allows the player to set strategic objectives they want their formations to achieve during the simulation. It was primarily implemented for MekHQ AtBScenarios, to have each team follow the objectives that were set for them in the game.

Currently the feature transform scenario objectives into orders for the player, and load those into the simulation. The game also works fine without the orders.

## Related:
[#6402](https://github.com/MegaMek/megamek/pull/6407)


HammerGS (approved):
Co-Pilot checked and confirmed no issues

PR #5803, opened 2025-01-16
MegaMek Issue 6416: Ghostbusting - Reset isDestroyed for entities when returning to MekHQ

psikomonkie:
Fixes MegaMek/megamek#6416


psikomonkie:
Right, forgot master is messed up. The failing builds should be resolved once #5802 is merged.

IllianiCBT:
Looking now

IllianiCBT (approved):
I'm not super familiar with this code, but it's such a small thing that if it _does_ cause everything to catch fire and implode we can reset in seconds.

I say the benefit of killing the ghost bug far outweighs the risk that this might break other things. Especially as we have time to find and fix any issues between now and release.

PR #5824, opened 2025-01-19
Extract AtB Event Types to enum

Algebro7:
This simply moves the hardcoded ATB Event Type constants from static variables to an enum. This supports further refactoring work on AtBContract by reducing the number of public variables that other classes need to call, therefore simplifying the public interface. My understanding was also that there was an initiative to try and use enums instead of static ints throughout the codebase.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/enums/AtBContractType.java
+++ b/MekHQ/src/mekhq/campaign/mission/enums/AtBContractType.java
@@ -218,9 +218,9 @@ public enum AtBContractType {
      * {@link #generateStratConEvent()} method.</p>
      *
      * @param campaign the {@link Campaign} instance for which the event is being generated.
-     * @return an integer representing the event type.
+     * @return an AtBEvent enum representing the event type.
      */
-    public int generateEventType(Campaign campaign) {
+    public AtBEventType generateEventType(Campaign campaign) {
         if (campaign.getCampaignOptions().isUseStratCon()) {
             return generateStratConEvent();
         }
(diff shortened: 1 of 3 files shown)


IllianiCBT (commented):
Do we want to pull the event logic and enum into their own class? So that rather than living in `AtBContract.java` we just include a call to `AtBContractEvent.java` and have all the event logic handled there? The enum could be contained within `AtBContractEvent.java`, too. Keep it all together.

Algebro7:
Good thought--I'll look more into that this weekend when I have some time

Algebro7:
So looking at it a bit more, the event logic in `AtBContract.checkEvents()` reaches into the contract-specific details a fair bit (e.g., manipulating parts availability numbers, morale, etc) which I'm concerned will increase the coupling with `ATBContract` rather than decrease it. `AtBContract` will have less code in it, but other contract types won't necessarily be able to use the event types in `AtBContractEvent` because `AtBContractEvent` is designed to work with `AtBContract`.

Sorry for the long-winded and confusing explanation...in short, I'm concerned about splitting the logic into a different class while having it remain tightly coupled to the class we extracted it from. As a side note, I am trying different strategies for the CamOps Contract Market and one may actually involve just extending the AtBContract class after all, and having AtBContract just represent the base class for any campaign against the computer rather than a specific game type. If I go that route then your idea may be less of a concern, so maybe we can explore that then?

MekHQ/src/mekhq/campaign/mission/enums/AtBContractType.java
```diff
@@ -351,97 +351,97 @@ public int generateEventType(Campaign campaign) {
      *
      * @return an integer representing the event type.
      */
-    public int generateStratConEvent() {
+    public AtBEventType generateStratConEvent() {
         final int roll = Compute.randomInt(20) + 1;
```
Scoppio, line 355: this really should be a static map or something like it.

Algebro7: Thanks, that probably does make sense. I'm guessing we'd need to simplify the rolls to avoid a bunch of duplicate values (e.g., for keys 10-13 and 14-15), right?

By the way, with the current direction I'm heading with the Contract classes, this

MekHQ/src/mekhq/campaign/mission/enums/AtBContractType.java
```diff
@@ -237,107 +237,107 @@ public int generateEventType(Campaign campaign) {
...
                 } else if (roll < 19) {
-                    return AtBContract.EVT_REINFORCEMENTS;
+                    return AtBEventType.REINFORCEMENTS;
                 } else if (roll < 20) {
-                    return AtBContract.EVT_SPECIALEVENTS;
+                    return AtBEventType.SPECIALEVENTS;
                 } else {
-                    return AtBContract.EVT_BIGBATTLE;
+                    return AtBEventType.BIGBATTLE;
                 }
```
rjhancock, line 255: These, and all following if/else chains, can be simplified using switch/case with fall through.

Start at 20 and work down.

It'll increase readability for most.

rjhancock (commented):
Looks like a straight forward change but requesting some readability improvements.

HammerGS:
I have no clue if this is ready to go. Leaving it till it gets approved by a coder.

Algebro7:
Sorry @HammerGS , don't merge yet. I mentioned in a separate comment thread that this change could be more trouble than it's worth since I had to change direction on the Contract refactoring, but I'm also happy to implement the reviewers' suggestions if they think there is value in moving those static ints to an enum stored outside the class. 

Just let me know and I can either finish it or close it. Thanks!

HammerGS:
I'll flip it to a draft till things get sorted.

Algebro7:
I've gone ahead and implemented the requested changes (flipping the if/else chains to enhanced switch blocks). Flipping this to ready for review.

HammerGS (approved):

PR #5838, opened 2025-01-20
Issue 5836: Combat Force commanders should be updated when a Force's Commander is updated.

### LINKED ISSUE #5836: [Issue] StratCon: Units from a lance not assigned the Training role giving training messages, while units in the lance assigned the Training role seemingly not being trained.
### Brief Description *

Charlie Lance is assigned the Training role in the briefing room and deployed with "Remain Deployed" set in the AoO tab. Charlie Lance has a veteran LT SG mekwarrior, and three green Sergeant mekwarriors. I would expect those three green mekwarriors to gain skill.

Each Monday a warning is given that member of Reserves Lance "Sergeant Viveka "Raid" Weiss has learned all they can from Sereant Tora "Aardvark" Witmeyer [Also in Reserves lance] and should be reassigned" though Reserves Lance is a child of "Resupply" which is designated as a convoy, and Reserves Lance has the Auxiliary Role in the Briefing Room tab, and is not deployed.

Setting Reserves Lance to training and deploying with "Remain Deployed" made a *different* lance show a similar message. I was able to fix it by creating an entirely new lance "Training", moving the members of Charlie to that, setting it to Training Role, and deploying with remain deployed.

I noticed that the highest ranking member of each lance is not displayed in bolded text indicating they're the lance leader, and suspect it's related.

### 3. Steps to Reproduce

1. Load "Fuzzy's Last Minute Marauders30640509_wrong_training_lance.cpnx"
2. Note the message already shows in the Command Center Tab
3. Advance Day until a new training cycle and the message should appear again.

I attached the entire sequence of save games until the issue was fixed and the log file. In the last save things are working as expected.

### Attach Files

[training_issue.zip](https://github.com/user-attachments/files/18480856/training_issue.zip)

### Severity *

### MekHQ Suite Version *

v0.50.03-SNAPSHOT Build Date: 2025-01-14T03:17:55.982487281

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium Java Version: 17.0.13

### Final Verification

psikomonkie:
Fixes #5836 

A CombatForce should have its commander updated when its force's commander is updated. In addition, to fix broken saves, we'll refresh the CombatForce's commander when doing training.

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -533,7 +533,16 @@ public class Force {
         return forceCommanderID;
     }
 
-    public void setForceCommanderID(UUID commanderID) {
+    /**
+     * Sets the force commander ID to the provided UUID.
+     * You probably want to use
+     * setOverrideForceCommanderID(UUID) followed by
+     * updateCommander(campaign).
+     * @see #setOverrideForceCommanderID(UUID)
+     * @see #updateCommander(Campaign)
+     * @param commanderID UUID of the commander
+     */
+    private void setForceCommanderID(UUID commanderID) {
         forceCommanderID = commanderID;
     }
 
@@ -626,6 +635,16 @@ public class Force {
 
         if (getParentForce() != null) {
             getParentForce().updateCommander(campaign);
+        }
+
+        updateCombatTeamCommanderIfCombatTeam(campaign);
+    }
+
+    private void updateCombatTeamCommanderIfCombatTeam(Campaign campaign) {
+        if (isCombatTeam()) {
+            campaign.getCombatTeamsTable().values().stream().filter(
+                    combatTeam -> combatTeam.getForceId() == getId())
+                    .forEach(combatTeam -> combatTeam.setCommander((UUID) null));
         }
     }
 
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -600,17 +600,18 @@ public class Force {
         if (eligibleCommanders.isEmpty()) {
             forceCommanderID = null;
             overrideForceCommanderID = null;
+            updateCombatTeamCommanderIfCombatTeam(campaign);
             return;
         }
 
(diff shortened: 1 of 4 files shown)


psikomonkie:
@IllianiCBT This is fixed twice; Going forward, it shouldn't be possible to have a CombatForce with a different commander than its Force. If this is a problem, let me know and I can remove it. The other fix is for existing saves, let's just refresh the CombatForce's commander when training like we do when entering scenarios. 

Both of these fixes could be used on their own if you don't like one of them.

MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -627,6 +636,16 @@ public void updateCommander(Campaign campaign) {
         if (getParentForce() != null) {
             getParentForce().updateCommander(campaign);
         }
+
+        updateCombatTeamCommanderIfCombatTeam(campaign);
+    }
+
+    private void updateCombatTeamCommanderIfCombatTeam(Campaign campaign) {
+        if (isCombatTeam()) {
+            campaign.getCombatTeamsTable().values().stream().filter(
```
IllianiCBT, line 645: For the number of iterations, this will actually be more expensive than a traditional loop. In addition to being less intuitive to read at a glance.

How will this handle ComTs containing multiple forces?

psikomonkie: > For the number of iterations, this will actually be more expensive than a traditional loop. In addition to being less intuitive to read at a glance.

I can switch it around. 

> How will this handle ComTs containing multiple forces?

Should handle it fine. It should work the same way as the current way for getting the ComT's commander - currently it uses the ComT's forceId to get the force's commander. In this case we already have the force's commander and the force, but we need to find its ComT. The ComT's commander should be its forceId's Force's commander.

psikomonkie: It shouldn't even be a for loop, I misunderstood `getCombatTeamsTable`. Now it should be doing what we want it to do.

IllianiCBT (approved):

PR #5839, opened 2025-01-20
Refactored HPG Network Logic and Improved Visualization Clarity

IllianiCBT:
- Separated HPG radius values for A and B stations into constants.
- Adjusted HPG visualization colors for better distinction across ratings.


IllianiCBT:
- Changed uninhabited systems to only be visible if the show empty systems option is enabled.
- Changed the show empty systems option to be disabled by default

PR #5853, opened 2025-01-22
Remove Legacy AtB's Campaign Parts Availability System

IllianiCBT:
- Removed CPA
- Adjusted contract modifiers. These were retained as they add flavor to different contracts, but the rest of CPA had to go.

While I've not yet gone through all of the modifiers, this should ensure acquisition modifiers are RAW. I have plans to expand on this further, but there are some barriers that need to be resolved first and that won't be ready for 50.03.


rjhancock (approved):
So long as master updates pass, this is good to go.

PR #5868, opened 2025-01-23
Updated Several Scenario Effects to use `SupplyCache` over `SupportPointUpdate`

IllianiCBT:
- Replaced the `SupportPointUpdate` effect with `SupplyCache` in several scenario templates.
- Implemented new functionality in `ScenarioObjectiveProcessor` to handle the `SupplyCache` effect. This triggers a Resupply dialog at the end of the scenario allowing players to acquire parts and materials from these scenarios.

This was functionality intended for the release of Resupplies, so the code was all there I just forgot to actually hook it into our scenario templates. Whoops!


rjhancock (approved):
If tests pass after code update, good to merge.

PR #5870, opened 2025-01-23
Adjusted Randomness of Turning Points and Highlighted Them in Briefing Room

IllianiBird:
- Updated the random chance for setting attached unit modifiers and turning points to trigger around 33% more often. We've been seeing feedback from users that Turning Points were too rare.
- Corrected Strategic Objectives (scenarios required by the contract) to always be considered Turning Points. This will not be retroactive for scenarios that have already spawned.
- Improved the `ScenarioTableModel` to visually highlight scenarios that are turning points.
- Ensured we accounted for colorblind players.

In the below image the scenario in yellow (uses the user configurable 'warning' color) is a Turning Point:
image

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
@@ -1352,7 +1352,6 @@ public class AtBDynamicScenarioFactory {
         scenario.getScenarioObjectives().clear();
 
         for (ScenarioObjective templateObjective : scenario.getTemplate().scenarioObjectives) {
-            logger.info("templateObjective:"+ templateObjective.toString());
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/mission/AtBDynamicScenarioFactory.java
```diff
@@ -1352,7 +1352,6 @@ public static void translateTemplateObjectives(AtBDynamicScenario scenario, Camp
         scenario.getScenarioObjectives().clear();
 
         for (ScenarioObjective templateObjective : scenario.getTemplate().scenarioObjectives) {
-            logger.info("templateObjective:"+ templateObjective.toString());
```
IllianiCBT, line 1355: This was actually intended for a different PR. It was just a bit of troubleshooting logging that I forgot to remove.

PR #5872, opened 2025-01-23
The Great Campaign Option Cull - Death

IllianiCBT:
This is the second phase of an attempt to reduce the sheer volume of campaign options. This is going to be an iterative process over the next few versions.

Removed
- Keep Married Name on Spouse Death - now permanently enabled.
- Random Death Method - renamed to 'Use Random Death'. Options reduced to 'enabled' and 'disabled' (enabled uses the Exponential random death method).
- Enable Random Clan Death - now permanently enabled if random death is also enabled.
- Enable Random Prisoner Death - now permanently enabled if random death is also enabled.

Mechanical Changes
- Random death checks now occur on a weekly basis, not daily. This reduces the amount of random death related new day load by approximately 85% per week. This change was made to bring death in line with relationship checks - which were changed to be weekly a few versions ago.


rjhancock (approved):
Overall looks sane.

PR #5873, opened 2025-01-24
Added Force Type Enumeration

IllianiCBT:
- Added `ForceType` enum to allow to easy management of force classifications.
- Replaced previous boolean classifications (isCombat, isConvoy) with the new `ForceType` enums.
- Added handling for the `Security` `ForceType`.

The purpose of this PR is to make it easier for developers to add new force types. Previously we had a bit of a struggle as individual force types had to be written in individually and then interactions between the different force types had to be managed. By using an enum we ensure that all force types are exclusive and easily accessed.

Four `ForceTypes` are available at inception:
- Standard - previously known as 'combat'.
- Support - previously piggybacked off of 'combat'. If 'combat' was `false` then the force would be a Support force.
- Convoy
- Security - not currently used, but will be utilized by another 50.04 system.


MekHQ/src/mekhq/campaign/force/ForceType.java
```diff
@@ -0,0 +1,131 @@
...
+    CONVOY("Convoy"),
+
+    /**
+     * Security force type, typically used for protection and guarding operations.
+     */
+    SECURITY("Security");
+
+
+    // Fields
+    private final String name;
```
Scoppio, line 52: redundant, ForceType.STANDARD.name() would return `STANDARD`, while `.getName()` returns `Standard`, if you want to use it in text you should use i18n, so neither use would be good enough here

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
```diff
@@ -862,42 +880,45 @@ private void calculatePlayerConvoyValues() {
         totalPlayerCargoCapacity = 0;
 
         for (Force force : campaign.getAllForces()) {
-            if (!force.isConvoyForce()) {
+            if (!force.getForceType().isConvoy()) {
```
Scoppio, line 883: I think I like more a `force.getForceType().isConvoy()` as you already have the force here and the api already exists, so you only change the internals of the force object, not its API.

IllianiCBT: I'm not entirely sure what you mean by that, sorry

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -423,46 +423,27 @@ public void actionPerformed(ActionEvent action) {
...
-
-            final boolean subforces = command.contains(TOEMouseAdapter.CHANGE_COMBAT_STATUSES);
-            for (final Force force : forces) {
-                force.setCombatForce(combatForce, subforces);
-                force.setConvoyForce(false);
-
-
+            ForceType forceType = ForceType.STANDARD;
+            if (command.contains("SUPPORT")) {
+                forceType = ForceType.SUPPORT;
```
Scoppio, line 433:
```suggestion
            if (command.contains(ForceType.SUPPORT.name())) {
                forceType = ForceType.SUPPORT;
```

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -423,46 +423,27 @@ public void actionPerformed(ActionEvent action) {
...
             }
 
-            for (Force formation : gui.getCampaign().getAllForces()) {
-                MekHQ.triggerEvent(new OrganizationChangedEvent(formation));
+            if (command.contains("CONVOY")) {
+                forceType = ForceType.CONVOY;
```
Scoppio, line 437:
```suggestion
            if (command.contains(ForceType.CONVOY.name())) {
                forceType = ForceType.CONVOY;
```

Scoppio (changes requested):

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
IllianiCBT: This is much better, thank you

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/Resupply.java
Scoppio:
```suggestion
            if (!force.isConvoy()) {
```

you already had the function `boolean isConvoy()` in the class api, the change you made in the force type didnt required you to change the API of the object, which is a benefit of the object oriented design.

IllianiCBT: Ahhhh, yes, I see exactly what you mean! Sorry, the use of 'API' here threw me off as that's not a term I'm super familiar with in this context.

Scoppio: API just mean the "application interface" or something like that, the api of a class are all its public methods. The method is already there, so there is no need to change how the "world" interacts with it, if the world needs to know if a force is a convoy or not, let the world ask the force if its a convoy, internally the force deals with the question however it wants, without exposing its internal workings to the outside.

IllianiCBT: That makes a heap of sense, thank you :)

MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java
```diff
@@ -423,46 +423,27 @@ public void actionPerformed(ActionEvent action) {
...
             }
 
-            gui.getTOETab().refreshForceView();
-        } else if (command.contains(TOEMouseAdapter.CHANGE_CONVOY_STATUS)) {
-            if (singleForce == null) {
-                return;
+            if (command.contains("SECURITY")) {
+                forceType = ForceType.SECURITY;
```
rjhancock, line 441: Same adjustments here as @Scoppio has been suggesting.

rjhancock (changes requested):
We don't want to do direct string comparison for internal code. Want to compare to a defined value to prevent a bug of a misspelling.

Otherwise looks good.

IllianiCBT:
Should all be good now. Also added some other minor enhancements. Should make this much easier for future developers to add new ForceTypes.

MekHQ/src/mekhq/campaign/force/ForceType.java
```diff
@@ -20,56 +20,118 @@
...
+     * <p>If the current instance is {@code STANDARD}, an empty string is returned as the symbol.</p>
+     *
+     * @return The localized symbol associated with the current instance, or an empty string
+     *         if the instance is {@code STANDARD}.
+     */
+    public String getSymbol() {
+        if (this == STANDARD) {
+            return "";
+        }
+
```
rjhancock, line 110: Within the resource bundle just have this set to a blank string and remove the check. No reason to do an edge case when a single line can do the same thing.

Also allows for adding a symbol later if desired either in current localization or in another.

rjhancock (approved):
Just that one QOL change and this should be good for me.

PR #5888, opened 2025-01-26
Convoy context messages updates

UlyssesSockdrawer:
Some more edits/tweaks on the convoy messages file - trying to make them a little more military/comm's like.

Also the gitignore change Tex mentioned in Discord.

### DIFF
--- a/MekHQ/resources/mekhq/resources/Resupply.properties
+++ b/MekHQ/resources/mekhq/resources/Resupply.properties
@@ -911,13 +911,13 @@ statusUpdate5.text=Sorry for the blackout, %s, radio antenna was bent by a low-h
   \ additional issues. All units are maintaining contact, with regular check-ins confirmed. The\
   \ convoy is proceeding as planned, with no deviation from the established route. No delays are\
   \ expected at this time.
-statusUpdate6.text=Detour forced by fallen trees, %s. The arrangement seems too neat to be mere\
-  \ chance - perhaps it's nature, or perhaps it's enemy sappers. Crew members swear they've seen\
-  \ movement in the treeline, dark shapes that vanish before they can be confirmed. We're advancing\
-  \ cautiously, weapons ready, each rustle of leaves amplified by the silence that follows. There's\
-  \ a growing sense that we're being funneled somewhere, like prey unaware of the hunter. Spirits\
-  \ are uneasy, but discipline holds for now. ETA still stands at 18 hours, but the sense of dread\
-  \ is growing.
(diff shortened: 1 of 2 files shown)


MekHQ/resources/mekhq/resources/Resupply.properties
```diff
@@ -911,13 +911,13 @@ statusUpdate5.text=Sorry for the blackout, %s, radio antenna was bent by a low-h
...
-  \ movement in the treeline, dark shapes that vanish before they can be confirmed. We're advancing\
-  \ cautiously, weapons ready, each rustle of leaves amplified by the silence that follows. There's\
-  \ a growing sense that we're being funneled somewhere, like prey unaware of the hunter. Spirits\
-  \ are uneasy, but discipline holds for now. ETA still stands at 18 hours, but the sense of dread\
-  \ is growing.
+statusUpdate6.text=We've had to take a detour due to trees blocking the road, %s.\
+  \ Way too suspicious to just be a freak of nature. I've sent patrols out into the woods after\
+  \ crew members reported they saw movement in the treeline.\
+  \ I'm going to go out with the next patrol whilst the techs keep clearing the road.\
+  \ If I had to take a guess, %s, the OpFor is trying to canalize us onto an ambush route further\
```
IllianiCBT, line 918: So this will throw an error, because it will be looking for a replacement for the second instance of `%s` however one isn't defined in the code.

IllianiCBT (changes requested):
I'm going to look into a way to define commander address once and have it propagate across the whole resource string, but as things _currently_ stand this will throw an error at worst, or just have the second instance of `%s` appear as `%s` in the dialog.

IllianiCBT:
You can try replacing `%s` in the role-play addresses with `{0}`. Ideally you'd do this across all the roleplay events for consistency - those are all the resource strings which have their key begin `statusUpdate`. _In theory_ this will cause the first definition (commander address) in `String.format()` to be used in all instances of `{0}`. This will need to be thoroughly tested before merged.

IllianiCBT:
I'm going to update this to use the new It8n class, which will make what you're doing a lot easier. This won't undo the work you've done here but will introduce conflicts. I'm going to mark this as draft for now and I'll get the switch done as soon as I finish up on the Prisoner Rework (which is also a dialog-heavy system that needs to be switched to It8n - for my sins).

IllianiCBT:
With the exception of the gitIgnore change (which will need to be a separate PR) I've merged these changes into #5892.

Closing :)

If you're planning any more Resupply edits hit me up on discord and I'll take you through pulling #5892 so you're not working off out of date stuffs. :)

PR #5892, opened 2025-01-27
Refactored Resupply Messaging to use Internalization

IllianiCBT:
- Refactored Resupply Messaging to use Internalization
- Incorporated edits from #5888 (Credit: @UlyssesSockdrawer)


Scoppio:
internalization would using `.intern()` on the strings (a real java feature, look it up)

IllianiCBT:
Gah, I meant Internationalization. I am increasingly becoming very aware of why It8n is preferred and why I was wrong to object...

MekHQ/src/mekhq/campaign/mission/resupplyAndCaches/PerformResupply.java
```diff
@@ -177,12 +178,12 @@ public static void performResupply(Resupply resupply, AtBContract contract, int
             }
         }
 
-        logger.info("totalTonnage: " + totalTonnage);
+        logger.info(String.format("totalTonnage: %s", totalTonnage));
```
rjhancock, line 181: Logger .info should support VARRAGs so don't need to do this. Just pass them directly like you would String.format.

rjhancock (approved):
Overall decent.

PR #5895, opened 2025-01-27
Refactored Age Checks; Updated Unit Tests for Procreation and Marriage

### LINKED ISSUE #5884: [RFE] [EDUCATION] Liao Warrior House graduates can't be assigned to units for 2 years.
### Enhancement Type *

The way things work at the moment, children sent to be educated at the Warrior Houses graduate on their 16th birthday as 4/3 Mekwarriors. However, there is a hardcoded limit that you cannot assign anyone to combat roles until they're 18, in order to avoid any and all child soldier-related issues. This does leave a 2-year hole i fell into where such individuals are basically sitting on the roster doing nothing.

I am uncertain what can be done about this issue, since Warrior Houses canonically graduate at 16 and the age limit of combatants is a project policy, but i feel it should be noted for the record.

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

50.03 SNAPSHOT

### Final Verification

IllianiCBT:
- Updated methods and test cases to consistently use `isChild` with an 18-year-old threshold override where required.
- Clarified project policies on age restrictions for marriage, procreation, and academy-related events.
- Improved readability and maintainability of `canMarry` and 'canProcreate'.
- Rewrote `canMarry` and `canProcreate` Unit Tests

# Age Policy
Mid-2024 we settled on a policy of 18+ for marriage, procreation, and roles. This was to cut back on the child soldier problem that had been in mhq since ages were introduced. There were other reasons, but that's the main one. Since then, we've received feedback to reduce the strictness specifically for Roles, given how many 16 year old pilots and such there are in the lore.

After discussing it with Hammer we opted to update the policy so that marriage and procreation would still be limited to 18+, but professions could be assigned to 16+ characters. I added some other limiters in the code, surrounding academy destruction and dropout events, which have been commented accordingly.

For academy destruction events we're treating any character under 18 to be a 'child' for the purposes of whether they can be killed during such events. This was chosen due to real world problems that might cause players to become uncomfortable if such a thing happens and they did not explicitly opt-in.

# Unit Tests
I have had enough of every time I even _look_ at marriage or procreation the unit tests would explode in a mass fit of rage. I have tweaked them and adjusted them so many times I think I knew them by heart. So, with no great sorrow, I nuked the ones this PR touched and just rewrote them to be more modular and therefore a lot easier to maintain. Now, instead of a single monolithic unit test for `canMarry` and `canProcreate` that tested all conditionals, I added individual tests that did the same thing but in an easier to follow manner. I don't know if this is the right way to handle it, but it feels better to be.

Feedback, as always, welcomed and encouraged - I'm still learning what, when, and how to Unit Test. :)

Fix #5884

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -925,7 +925,7 @@ public class Utilities {
 
         // Clamp age, if necessary
         if (experienceLevel != EXP_NONE) {
-            age = max(18, age);
+            age = max(16, age);
         }
 
         return age;
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -1473,8 +1473,33 @@ public class Person {
         return id;
     }
 
+    /**
+     * Checks if the person is considered a child based on their age and today's date.
+     *
+     * <p>This method uses the default context where the person is not being checked
+     * for procreation-specific thresholds.</p>
+     *
+     * @param today the current date to calculate the age against
+     * @return {@code true} if the person's age is less than 16; {@code false} otherwise
+     */
     public boolean isChild(final LocalDate today) {
-        return getAge(today) < 18;
+        return isChild(today, false);
+    }
+
+    /**
+     * Checks if the person is considered a child based on their age, today's date, and procreation
+     * status.
+     *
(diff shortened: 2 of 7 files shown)


IllianiCBT:
@HammerGS, you've been assigned as a reviewer for this PR as it concerns project policy.

PR #5912, opened 2025-01-28
Implemented `NewsDialog` for Immersive News Display

IllianiCBT:
- Replaced `NewsReportDialog` with the new `NewsDialog` extending `MHQImmersiveDialog`.
- News Network will change based on year and (in the case of the Clans) faction.

Old...
image

New...
image


Sleet01 (approved):
LGTM!

Issue #5914, opened 2025-01-28
[RFE] Improvements to Scenario Resolution objective text for Stratcon Facilities


PhoenixHeart512:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

Currently the objectives part of scenario resolution for facilities is extremely confusing. This RFE is to track proposed improvements to the objective descriptions to improve readability and clarify.

Example of a confusing objective:

![Image](https://github.com/user-attachments/assets/09e1dff9-4df5-4949-96dc-59c52f3927e2)

In particular, this section says failing the "preserve" objective leaves the facility intact:

![Image](https://github.com/user-attachments/assets/ae93c4d9-21b1-4f64-8999-eff64ef7ddad)

### MekHQ Suite Version *

[0.50.03-SNAPSHOT]

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

PR #5915, opened 2025-01-28
Update scenario panel facility objective text

UlyssesSockdrawer:
Changes discussed in Discord for #5914 

When merged, this should fix that RFE by making the text that appears on the objecrtive panel clearer.

I've not used the i18n thing because frankly I don't know how.

Hopefully better for user understanding of what the objective will do.


Sleet01 (approved):
LGTM!

PR #5947, opened 2025-01-31
5932 Injury&fatigue

### LINKED ISSUE #5932: [Issue] Fatigue + Injury = No Medical Care
### Brief Description *

When an injured character reaches high levels of fatigue - say, from being significantly injured - it automatically sends them on leave to recover from the fatigue. This does not, however, account for the fact that they're injured and will remove them out of the infirmary untreated. (Since being On Leave means they're inactive and thus cannot be assigned a doctor.) 

As a result they will be restricted to natural healing only and are much more likely to worsen due to lack of treatment. This is also exceptionally silly in-universe as you're yanking somebody out of surgery to put them on a beach. "Never mind that sucking chest wound, you need a mai tai to relax."

The fatigue check to place characters on leave should probably exclude characters who are currently wounded.

### 3. Steps to Reproduce

1. Break every bone in a character's body
2. Set them to high fatigue
3. Advance day
4. Watch them go on vacation instead of seeing a doctor about their punctured lung

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

0.50.03 #1742

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

DM0000:
Add check for injuries that are not permanent before going on leave.

fixes #5932

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
+++ b/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
@@ -115,7 +115,8 @@ public class Fatigue {
         }
 
         if ((campaign.getCampaignOptions().getFatigueLeaveThreshold() != 0)
-                && (effectiveFatigue >= campaign.getCampaignOptions().getFatigueLeaveThreshold())) {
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #5949, opened 2025-02-01
Clarify Vocational XP Tooltip

IllianiCBT:
- Updated tooltip text for Vocational XP to specify that 'Relief Duty' contracts are excluded from the doubled XP exclusion. Previously the tooltip stated that all contracts ending in 'Duty' were excluded from granting double xp awards.

This was a visual bug only.


Sleet01 (approved):
LGTM

rjhancock (approved):

Issue #5922, opened 2025-01-29
[Issue] The Mass Personnel Training window "Experience Level" filter filters the next higher category.


GiovanniBertuccio:
### Brief Description *

When using the Mass Personnel Training dialog and selecting an experience level in the "Experience Level" dropdown it selects the group one higher
e.g. selecting "Ultra-Green" will pick "Green" personnel.

### 3. Steps to Reproduce

Open fuzzy's...30650213.cpnx.
Open the personnel tab and sort by skill level
Open the Mass Personnel Training Dialog
Select "Ultra-Green"
  Only Dae-Jin Stornes should appear, but 4 green personnel appear instead.

### Attach Files

[mekhq - mass training sorting issue.zip](https://github.com/user-attachments/files/18582025/mekhq.-.mass.training.sorting.issue.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

Starting MekHQ v0.50.03-SNAPSHOT     Build Date: 2025-01-14T03:17:55.982487281

### Operating System *

Windows 10

### Java Version *

Java Version: 17.0.13

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiCBT:
Looks like this has already been resolved. Closing

PR #5950, opened 2025-02-01
fix: addresses off by one issue with dialog filter for skills, addres…

firefly2442:
…ses issue: #5922

I'm not sure if this is the correct way to address this issue.  I'm a bit confused as to the consistency surrounding the skill levels.  In some places, ultra green starts at 0 and in some places it starts with 1.  This just hacks in a +1.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
@@ -394,7 +394,7 @@ public final class BatchXPDialog extends JDialog {
                 return false;
             } else if ((null != primaryRole) && (p.getPrimaryRole() != primaryRole)) {
                 return false;
-            } else if ((null != expLevel) && (p.getExperienceLevel(campaign, false) != expLevel)) {
+            } else if ((null != expLevel) && (p.getExperienceLevel(campaign, false)+1 != expLevel)) {
                 return false;
             } else if (onlyOfficers && !p.getRank().isOfficer()) {
                 return false;


MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
```diff
@@ -394,7 +394,7 @@ public boolean include(Entry<? extends PersonnelTableModel, ? extends Integer> e
                 return false;
             } else if ((null != primaryRole) && (p.getPrimaryRole() != primaryRole)) {
                 return false;
-            } else if ((null != expLevel) && (p.getExperienceLevel(campaign, false) != expLevel)) {
+            } else if ((null != expLevel) && (p.getExperienceLevel(campaign, false)+1 != expLevel)) {
```
IllianiCBT, line 397: I'm not sure about this fix. I think this should be held back until we have time to do a more thorough investigation. 50.03 launches in a few hours, so that likely won't be in time for launch.

IllianiCBT:
Marking this for New Dev Cycle as I think we need to take a proper look at this rather than putting in what feels like a ductile fix - a fix that likely works, but not one resolving the underlying issue.

Skills being Ultra Green at different skill levels is, however, working as intended. As different skills can have skill levels set to different points. One skill might be Green at level 3, while another might be Green at level 1 - for example.

rjhancock:
Hacks are not allowed. We want proper fixes. Sorry.

We appreciate the desire to fix, but this wont be accepted as skills have different growth paths.

PR #5994, opened 2025-02-06
fix: addresses the off by one issue presenting the wrong filtering re…

firefly2442:
…sults for skill level for mass training, fixes issue #5922

Second attempt here at a fix.

I assume `Elite` level personnel should be selectable as a dropdown in the filter for mass training?  Without this fix, Elite does not show up in the dropdown.  This also correctly shows the right personnel given the example save provided in the referenced issue.

If I'm misunderstanding something here, please correct me.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
@@ -170,8 +170,8 @@ public final class BatchXPDialog extends JDialog {
         choiceExp.setMaximumSize(new Dimension(Short.MAX_VALUE, (int) choiceType.getPreferredSize().getHeight()));
         DefaultComboBoxModel<PersonTypeItem> personExpModel = new DefaultComboBoxModel<>();
         personExpModel.addElement(new PersonTypeItem(resourceMap.getString("experience.choice.text"), null));
-        for (int i = SkillLevel.ULTRA_GREEN.ordinal(); i < SkillLevel.ELITE.ordinal(); i++) {
-            personExpModel.addElement(new PersonTypeItem(Skills.SKILL_LEVELS[i].toString(), i));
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
```diff
@@ -170,8 +170,8 @@ private JComponent getButtonPanel() {
         choiceExp.setMaximumSize(new Dimension(Short.MAX_VALUE, (int) choiceType.getPreferredSize().getHeight()));
         DefaultComboBoxModel<PersonTypeItem> personExpModel = new DefaultComboBoxModel<>();
         personExpModel.addElement(new PersonTypeItem(resourceMap.getString("experience.choice.text"), null));
-        for (int i = SkillLevel.ULTRA_GREEN.ordinal(); i < SkillLevel.ELITE.ordinal(); i++) {
-            personExpModel.addElement(new PersonTypeItem(Skills.SKILL_LEVELS[i].toString(), i));
+        for (int i = SkillLevel.ULTRA_GREEN.ordinal(); i <= SkillLevel.ELITE.ordinal(); i++) {
+            personExpModel.addElement(new PersonTypeItem(SkillLevel.parseFromInteger(i).toString(), SkillLevel.parseFromInteger(i).getAdjustedValue()));
```
IllianiCBT, line 174: We should probably be storing `SkillLevel.parseFromInteger` in a variable so we don't have to fetch it twice.

Something like...

```
final SkillLevel skillLevel = SkillLevel.parseFromInteger(i);
personExpModel.addElement(new PersonTypeItem(skillLevel.toString(), skillLevel.getAdjustedValue()));
```

IllianiCBT (changes requested):
A small change to reduce redundancy. Otherwise, while we'll need to test this, it _looks_ ok.

IllianiCBT (approved):

PR #6003, opened 2025-02-08
Issue 5979: Force players to use commit when deploying forces

### LINKED ISSUE #5979: [Issue]50.03 Preserve Victory Condition Does Not List Units Correctly
### Brief Description *

The endgame victory point scoring does not list friendly units that have been preserved.

Image

In this case, it is the dropship defense mission, with delta lance deployed and using beta as the seed to generate the scenario. It is under liaison command rights. The screenshot is immediately after the battle, but the result is the same if you use either PACAR, ACAR, or manually fight it. 

I do not know for sure if it is this scenario type exactly and will update later. 

### 3. Steps to Reproduce

1. Load up the campaign file, the dropship mission is setup to go.
2. Resolve the mission in any manner.
3. At the end in the VP screen, it lists no units under the preserve objective. 

### Attach Files

[Howling Gryffons30490224.cpnx.gz](https://github.com/user-attachments/files/18652332/Howling.Gryffons30490224.cpnx.gz)

[logs.zip](https://github.com/user-attachments/files/18652338/logs.zip)

### Severity *

### MekHQ Suite Version *

50.03

### Operating System *

Windows 11

### Java Version *

Eclipse Adoptium 17.0.13

### Final Verification

psikomonkie:
Fixes #5979 

When deploying forces in Stratcon, the player should be required to hit "commit" on the 2nd deployment window before their primary forces are comitted. This ensures the objective is properly updated.

I also added another message to the 2nd deployment window to make it more clear, and a cancel button per Illiani's request:
image

(displaying this even if there's no leadership units is intentional - it's a chance for the player to realize they don't get any free leadership units and opt to deploy a different force).

### DIFF
--- a/MekHQ/src/mekhq/gui/StratconPanel.java
+++ b/MekHQ/src/mekhq/gui/StratconPanel.java
@@ -120,7 +120,7 @@ public class StratconPanel extends JPanel implements ActionListener {
 
     private final Map<String, BufferedImage> imageCache = new HashMap<>();
 
-    private List<Force> pendingDeployments = new ArrayList<Force>();
+    private boolean commitForces = false;
 
     /**
      * Constructs a StratconPanel instance, given a parent campaign GUI and a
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/gui/StratconPanel.java
```diff
@@ -1049,7 +1053,17 @@ public void actionPerformed(ActionEvent evt) {
...
-            // Deliberate fall-through
+                if (selectedScenario != null &&  selectedScenario.getCurrentState() == PRIMARY_FORCES_COMMITTED) {
+                    scenarioWizard.setCurrentScenario(currentTrack.getScenario(selectedCoords),
+                        currentTrack, campaignState, isPrimaryForce);
+
+                    scenarioWizard.toFront();
+                    scenarioWizard.setVisible(true);
+                }
+                if (selectedScenario != null && !isCommitForces()) {
+                    selectedScenario.resetScenario(campaign);
```
psikomonkie, line 1064: In particular @IllianiCBT I wanted your thoughts on this line:

I haven't seen any issues with this in my testing. This should only be reachable during the initial deployment. isCommitForces should only return true if the user hit "Commit" on the 2nd deployment popup. So, if the user doesn't hit "commit" we're back where we started - nothing's deployed. It gets... Complicated if we instead try to delay deploying the initial force until after the 2nd deployment popup - a lot of the 2nd popup relies on a force already being deployed.

IllianiCBT: This all looks good to me. I'm going to approve and merge if but I'd like you to see if you can grab a volunteer or two to test it in tonight's nightly.

IllianiCBT (approved):

PR #6028, opened 2025-02-13
Cache finances balance

psikomonkie:
My MekHQ was running slow - taking over 10 seconds to process a new day for a thicc Battalion while off contract. IntelliJ's profiler suggested that `Finances::getBalance` was taking a lot of time, so it's now cached. 

Before caching:
image1
image

After caching:
image2


psikomonkie:
Hey @IllianiCBT this is that PR we briefly chatted about!

HammerGS (approved):
Checked with Co-Pilot and IDEA AI no issues identified

PR #6029, opened 2025-02-13
Linked scenario unit selection

DM0000:
Added checkbox to resolution screen so you can select which units continue to next scenario, instead of one button that sends force.
image

Open to ideas on how to improve the look of it. The "Continue to mission" part kind of long so wrapped it.

If whole force is selected then whole force will be deployed to next scenario otherwise force will still be undeployed from first scenario but just the units selected will be deployed to next linked scenario.

Checkboxes are not enabled if unit is totaled or not Functional.
Checkboxes are hidden if not a linked scenario.


HammerGS (approved):
Checked with Co-Pilot and IDEA AI no issues identified

PR #6072, opened 2025-02-23
Add source information to PlanetViewPanel

AaronGullickson:
This PR uses the new `SourceableValue` class introduced with the planet revision in PR #6031 to display canonicity information on the `PlanetViewPanel`. This information is now displayed via tooltip when a value is hovered over as shown in the screenshot. 

Screenshot 2025-02-23 at 8 45 35 AM

To do this, I created a `SourceValueLabel` class that extends `JLabel` and takes a `SourceableValue` as its main input. It can also take an optional formatting string that can be used to control how the value is displayed in the label. In several cases, the label text is derived implicitly from the `toString` method for the given class of `SourceableValue`, so I add a few overrides for that where needed.

The display will will look very similar to before, except for one change. Satellites and landmasses were previously shown as a comma-separated listed of names. Since each satellite and landmass needs to be specified separately as a label, they are now placed on different lines as shown in the screenshot above. The capital city is also a separate `SourceableValue`, so I handle that by indentation (when needed) and the prefix *Capital:*.


HammerGS (approved):

Issue #5719, opened 2025-01-09
[RFE] Add settings for minimum time spent by Tech on Mass Repairs


Dark-Hobbit:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

Add a setting in "Mass Repair/Salvage" menu that allows to put in a minimum of how many minutes a day a Tech must have after maintenance to be considered for repairs. Meaning that if a Tech has only 30 minutes/day after maintenance and the setting is put at "minimum 60 minutes/day", they  would never get used for Mass Repairs and only Techs with 60 minutes or more available would be considered.

It could be a general setting but would probably work better on by-category basis, added as an extra column to the right of "Max BTH" one.

Alternatively, it could be a setting of how many days at max a repair can take for that part. So if a part would take 300 minutes to get fixed and the setting is "maximum 5 days", only Techs with 60 minutes/day or more would be considered. The second implementation might be better but would require accounting for using more/less time in some way, so it would be harder to implement.

### Use Cases or Rationale

Right now "Mass Repair/Salvage" functionality has two prioritisation settings: Techs that can complete the task today and Techs assigned to the force. However, if no Tech can complete the repair today, any free Tech can be used regardless of how much free time they have after the daily maintenance. That can happen especially often with the use of "Extra Time", as individual repairs can last more than a day.

In worst case scenario, a Tech with 10-20 minutes/day left from maintenance can get assigned a repair/replacement task on a part with x4 Extra Time that would take them upwards of a month to complete (in case of 720/800/960/1080 minute repairs). That is often much less efficient than the Tech doing nothing at all or reloading ammo, leaving the part to wait for a different Tech to free up and fix it in a few days.

### MekHQ Suite Version *

0.50.02

### Final Verification

- [X] I confirm this request hasn't already been submitted (checked the tracker)
- [X] I've discussed or asked about this enhancement on MegaMek Discord
- [X] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

IllianiCBT:
This was resolved last week. Closing.

PR #6074, opened 2025-02-23
[RFE 5719] Add setting for minimum daily repair time to MRMS

Dark-Hobbit:
Implements #5719

Adds a set of editable control boxes that specify a minimum of free time (between 0 and 480 minutes) a Tech must have left to start repairing/salvaging an item of specific type. Mostly a way for players to avoid MRMS assigning Techs with 5/10/15 minutes a day left to something that could take them months to finish. Disabled by default (set at 0 minutes minimum).

Please check the code because I've never worked with GUI in Java before and might've messed up somewhere.

after


MekHQ/src/mekhq/gui/dialog/MRMSDialog.java
```diff
@@ -544,6 +544,12 @@ private JPanel createOptionsPanel() {
         gridBagConstraints.gridx = gridRowIdx++;
         pnlItems.add(maxBTHLabel, gridBagConstraints);
 
+        JLabel minDailyTimeLabel = new JLabel(resources.getString("minDailyTimeLabel.text"));
```
psikomonkie, line 547: Consider using the new text-value getting method: `MHQInternationalization.getTextAt`

MekHQ/src/mekhq/service/mrms/MRMSService.java
```diff
@@ -729,6 +729,12 @@ private static MRMSPartAction repairPart(Campaign campaign, IPartWork partWork,
 
             debugLog("Checking tech %s", "repairPart", tech.getFullName());
 
+            // If this tech has less time than required, ignore them for this part
+            if (tech.getMinutesLeft() < mrmsOptions.getDailyTimeMin()) {
```
psikomonkie, line 733: If a tech has 5 minutes left and the minimum time is 5 minutes, should they start work?

psikomonkie (commented):
Woah!! Great work! I took an initial look through it and it looks pretty good.

MekHQ/src/mekhq/service/mrms/MRMSService.java
psikomonkie: I reread this like 4 times and I still got it backwards 🤦🏻 ignore me

Dark-Hobbit: Perhaps I should rephrase/rename it then?

MekHQ/src/mekhq/gui/dialog/MRMSDialog.java
Dark-Hobbit: I was trying to stick with the same approach across this group of controls. Should I switch all of them to the new method?

MekHQ/src/mekhq/service/mrms/MRMSService.java
psikomonkie: No no, the comment is even perfectly describing it, I just goofed :(

MekHQ/src/mekhq/gui/dialog/MRMSDialog.java
psikomonkie: Yeah, this isn’t a must change from me just wanted to point it out - I think we should avoid adding new ones like this, but also I would probably do the same as you and emulate the local style.

Dark-Hobbit: I will leave it like this for now then, and potentially change it everywhere when I have the time to dig deeper into the MRMS code.

HammerGS (approved):
Tested with Co-Pilot and IDEA AI no issues detected

PR #6076, opened 2025-02-23
Added Reputation Sanity Options to Campaign Settings

IllianiCBT:
- Introduced options to clamp reputation-based pay, reduce mission performance score impact, and apply a cutoff for old missions in reputation calculations.

These changes aim to prevent excessive reputation inflation, particularly in long-running campaigns.

### Rampant Reputation and the 50.03 Update  
The issue of **Rampant Reputation** emerged in version 50.03 when we fixed a bug that had been preventing contract pay from correctly factoring in Reputation. This fix led to an extreme spike in contract pay—so significant that it was initially mistaken for a new bug. However, after investigation, we confirmed that the system was functioning according to **Rules as Written** and that the issue stemmed from the adaptation of **CamOps** mechanics into **MekHQ**.  

CamOps was designed for in-person groups who typically play only a few times a month, essentially a **D&D-style campaign with Meks**. A full campaign that would take an in-person group **a year or more** to complete can be finished in MekHQ in under an hour, especially with Luana's fantastic **auto-resolve** options. When combined with MekHQ’s **robust personnel simulation**, this encourages players to engage in **generational campaigns**—campaigns that span **decades of in-game time**, where personnel **age, marry, have children, and pass the campaign on to the next generation**.  

As a result, over time, **Reputation scores of 300, 600, or even higher** become inevitable. This becomes a serious issue when calculating **contract pay**, which uses an **uncapped multiplier** with no limiters. In a long-running campaign spanning from the **Age of War to ilClan**, Reputation could theoretically reach **1,000 or more**, with no upper boundary.  

Following the release of version 50.03, we extensively discussed possible solutions but struggled to find a single approach that effectively controlled the issue. Every proposal we considered failed to fully address the problem. Ultimately, we decided on a **three-pronged resolution**. 

### The Three-Pronged Solution  
1. **Capping Reputation’s Effect on Contract Pay**  
When this setting is enabled, the contract pay multiplier from Reputation is limited to a range between **0.5 and 2.0**. This ensures that no matter how high or low a campaign’s Reputation becomes, contract pay remains within a **reasonable range**.

The chosen band matches the **old FM:Mercenaries (revised) values**, keeping it in line with established mechanics.  

2. **Reducing Performance Impact on Reputation Gains**
When this option is enabled, the modifier to **Reputation gains from contract performance** is reduced by **80%**.  

- A **successful contract** will now increase Reputation by **1** instead of 5.  
- A **failed contract** will reduce Reputation by **2** instead of 10.  
- A **contract breach** now results in a **5-point reduction** instead of 25.  

By slowing the rate of Reputation progression, this change helps mitigate the impact of MekHQ’s accelerated contract completion speed.  

3. **Disregarding Old Contracts**
When enabled, **AtB and StratCon contracts that ended more than 10 years ago** are ignored when calculating Reputation modifiers. This ensures that campaigns are measured based on **current performance** rather than being artificially inflated by past successes. It also acts as a **sanity pass** for some of the **exceptionally long-running campaigns** that players have maintained over the years.  

### Implementation and Defaults  
All three of these options can be toggled in **Campaign Options** and apply **retroactively**.  

- In the **New Pilot Program** and **Complete Experience** presets, all three settings are **enabled by default**.  
- In contrast, they remain **disabled by default** in the two **CamOps presets**, as those presets are designed to adhere to **Rules as Written**. 

All three options are **disabled by default** in existing campaigns.

### DIFF
--- a/MekHQ/mmconf/campaignPresets/CampaignOperations.xml
+++ b/MekHQ/mmconf/campaignPresets/CampaignOperations.xml
@@ -7,6 +7,9 @@
 	<gm>true</gm>
 	<campaignOptions>
 		<manualUnitRatingModifier>0</manualUnitRatingModifier>
+		<clampReputationPayMultiplier>false</clampReputationPayMultiplier>
+		<reduceReputationPerformanceModifier>false</reduceReputationPerformanceModifier>
+		<reputationPerformanceModifierCutOff>false</reputationPerformanceModifierCutOff>
 		<logMaintenance>false</logMaintenance>
 		<defaultMaintenanceTime>4</defaultMaintenanceTime>
 		<mrmsUseRepair>true</mrmsUseRepair>
--- a/MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
+++ b/MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
@@ -7,6 +7,9 @@
 	<gm>true</gm>
 	<campaignOptions>
 		<manualUnitRatingModifier>0</manualUnitRatingModifier>
+        <clampReputationPayMultiplier>false</clampReputationPayMultiplier>
+        <reduceReputationPerformanceModifier>false</reduceReputationPerformanceModifier>
+        <reputationPerformanceModifierCutOff>false</reputationPerformanceModifierCutOff>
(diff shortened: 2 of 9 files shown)


Scoppio (approved):

HammerGS (approved):

PR #6080, opened 2025-02-24
Refactor/deprecate dead code

Algebro7:
While writing tests I found a `setEmployerCode()` overload which was only used by CO Contract Market and a dead/unused class called FactionMigrator. This overload also appeared to have a bug where it was setting the enemy camo instead of the ally camo.

This marks the method and class deprecated for later removal and switches the call in CO contract market to the correct version.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/AtBContract.java
+++ b/MekHQ/src/mekhq/campaign/mission/AtBContract.java
@@ -1344,6 +1344,7 @@ public class AtBContract extends Contract {
         return employerCode;
     }
 
+    @Deprecated
     public void setEmployerCode(final String code, final LocalDate date) {
         employerCode = code;
         setEmployer(getEmployerName(date.getYear()));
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -1344,6 +1344,7 @@ public String getEmployerCode() {
         return employerCode;
     }
 
+    @Deprecated
```
IllianiCBT, line 1347: Are we certain enemy camo is being set elsewhere?

Algebro7: It is being initialized to a random camo in the constructor and (disturbingly) in the GUI, as well as a few other places:

image

Scoppio: Be careful because intelliJ will reference **ANY** code/function/variable as being "used" as long as the name or the name and argument types match.

Issue #4364, opened 2024-07-10
[RFE] Add option to make portrait generation non-exclusive


Dark-Hobbit:
Currently, only one member of the unit can hold any given portrait - which means you can potentially ran out of portraits. That is especially the case with infantry, as they tend to add a lot of personnel to the unit.

Would it be possible to add a toggleable option in the settings that removes the limit of 1 person per 1 portrait? That way the algorithm of randomly generating portraits would keep working no matter how big the unit gets, without having to duplicate files in the portraits folder.

Dark-Hobbit:
Implemented in 0.50.04.

PR #6092, opened 2025-02-26
[RFE 4364] Add campaign option to allow duplicate portraits

Dark-Hobbit:
Implements #4364

When the option "Allow Duplicate Portraits" is enabled, random portrait generator ignores existing personnel and feels okay with potentially using the same portrait. Defaults to true.

Screenshot

### DIFF
--- a/MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
+++ b/MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
@@ -541,6 +541,7 @@
 		<limitLanceWeight>false</limitLanceWeight>
 		<limitLanceNumUnits>false</limitLanceNumUnits>
 		<assignPortraitOnRoleChange>false</assignPortraitOnRoleChange>
+        <allowDuplicatePortraits>false</allowDuplicatePortraits>
 		<allowOpForAeros>false</allowOpForAeros>
 		<allowOpForLocalUnits>false</allowOpForLocalUnits>
 		<opForAeroChance>5</opForAeroChance>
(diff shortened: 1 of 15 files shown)


MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
```diff
@@ -541,6 +541,7 @@
 		<limitLanceWeight>false</limitLanceWeight>
 		<limitLanceNumUnits>false</limitLanceNumUnits>
 		<assignPortraitOnRoleChange>false</assignPortraitOnRoleChange>
+        <allowDuplicatePortraits>false</allowDuplicatePortraits>
```
IllianiCBT, line 544: Honestly, I'd have this as `true`. I imagine players would be happier getting duplicates than they would no portraits. And those players who do want no duplicates can go into campaign options and disable the option.

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -1103,6 +1104,7 @@ public CampaignOptions() {
         Arrays.fill(usePortraitForRole, false);
         usePortraitForRole[PersonnelRole.MEKWARRIOR.ordinal()] = true;
         assignPortraitOnRoleChange = false;
+        allowDuplicatePortraits = true;
```
IllianiCBT, line 1107: This hooks back into my previous comment - that I accidentally made outside the review, whoops. If we're deciding that we allow duplicate portraits by default, then we might as well have it enabled in the campaign presets.

IllianiCBT (changes requested):
Looks great. One very minor bit of feedback that I have to mark as 'request changes' because Git doesn't have a 'suggested changes' option. Lemme know if you agree/disagree with my thoughts.

I'm also super keen to hear how you found the process of setting up the campaign option side of things. Something I really tried to do was improve the process so that it was more streamlined than prior CO IIC so looking for feedback now folks are starting to use it.

MekHQ/mmconf/campaignPresets/CampaignOperationsStratCon.xml
Dark-Hobbit: That's actually a typo - I meant for it to be set to `true`, but at some point inserted it incorrectly into the presets. And since I tested mostly on an existing campaign, where default value was set correctly, I didn't catch the mistake.

Thanks for spotting it!

IllianiCBT (approved):

PR #6095, opened 2025-02-28
Null Check for Random Dependents

psikomonkie:
I got a NPE in my nightly save. 

`genealogy.isDependent`'s return `Person` is tagged as `@Nullable`.  This will handle that. 

Child above shouldn't need it, if someone doesn't have children it'll return an empty list.


IllianiCBT (approved):
Whoopsie

Issue #6067, opened 2025-02-21
[Issue] unable to launch scenario from Mekhq


69chrisd69123:
### Brief Description *

I have stratcon turned off.  I am trying to just run a created scenario for my mekhq campaign. However I can't get it to launch

### 3. Steps to Reproduce

1. Scenario created
2. forces from TO&E deployed
3. start scenario

### Attach Files

[mekhq logs.zip](https://github.com/user-attachments/files/18916176/mekhq.logs.zip)

[Hawk's Ronin30490620.cpnx.zip](https://github.com/user-attachments/files/18916178/Hawk.s.Ronin30490620.cpnx.zip)

[customs.zip](https://github.com/user-attachments/files/18916181/customs.zip)

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

0.50.03

### Operating System *

windows 11

### Java Version *

Java Version 21.0.6

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

HammerGS:
```
3:25:06,793 ERROR [mekhq.gui.CampaignGUI] {AWT-EventQueue-0}
mekhq.gui.CampaignGUI.saveCampaign(CampaignGUI.java:1416) - 
java.lang.NullPointerException: Cannot invoke "mekhq.campaign.universe.Planet.getParentSystem()" because the return value of "mekhq.campaign.RandomOriginOptions.getSpecifiedPlanet()" is null
	at mekhq.campaign.RandomOriginOptions.writeToXML(RandomOriginOptions.java:204)
	at mekhq.campaign.CampaignOptions.writeToXml(CampaignOptions.java:4979)
	at mekhq.campaign.Campaign.writeToXML(Campaign.java:5995)
	at mekhq.gui.CampaignGUI.saveCampaign(CampaignGUI.java:1406)
	at mekhq.gui.CampaignGUI.saveCampaign(CampaignGUI.java:1377)
```

psikomonkie:
Duplicate of #5987 , this is fixed in the nightly. @HammerGS there was another error:
```
[...]

Issue #5987, opened 2025-02-05
[Issue] NPE error when starting a game from MekHQ


AaronGullickson:
### Brief Description *

NPE error when starting a game from MekHQ. Log file reports the following error:

```java
java.lang.NullPointerException: Cannot invoke "megamek.client.ui.swing.ILocalBots.getLocalBots()" because "this.localBots" is null
	at mekhq.GameThread.getLocalBots(GameThread.java:139)
	at mekhq.GameThread.run(GameThread.java:146)
```

I haven't had time to diagnose the issue further, but I would guess that adjustments were made to AtBGameThread and not GameThread. 

### 3. Steps to Reproduce

1. Put the custom unit in the mekfiles (remove .txt suffix).
2. Load the attached campaign.
3. Start the single scenario in the current mission on the Briefing Room tab. 

### Attach Files

[Dark Horse Irregulars31520705.cpnx.gz](https://github.com/user-attachments/files/18678030/Dark.Horse.Irregulars31520705.cpnx.gz)

[Marauder MAD-DH1.mtf.txt](https://github.com/user-attachments/files/18678038/Marauder.MAD-DH1.mtf.txt)

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

0.50.03

### Operating System *

Mac OS Sequia 15.2

### Java Version *

19

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

AaronGullickson:
Yes, I think my intuition was correct. The issue is a null `localBot` objet. Here is from the top of `AtBGameThread` run function:

```java
        client.addCloseClientListener(this);

        createController();
        if (autoResolveBehaviorSettings != null && minimalGUI)
        {
            var acarGui = new CommanderGUI(client, controller);
            localBots = acarGui;
            swingGui = acarGui;
            acarGui.start();
        } else {
            var clientGui = new ClientGUI(client, controller);
            localBots = clientGui;
            swingGui = clientGui;
            swingGui.initialize();
        }
```

`localBot` is assigned there in both if clauses. Here is the top of the `GameThread` run function:

[...]

PR #6104, opened 2025-02-28
Fixed `NullPointerException` In `RandomizeAroundSpecifiedPlanet`

IllianiCBT:
- Added a null check for `getSpecifiedPlanet()` in the condition to avoid potential `NullPointerException`.
- Ensured a fallback to the first available planet if no planet is selected from the dropdown.
- Prevented potential crashes caused by null values when setting the specified planet.

### Dev Notes
This issue was highlighted in the comments of #6067. Without knowing the exact source of the NPE, it's difficult to determine where to add null protection. However, since the only new functionality comes from CO IIC, it made sense that the issue originated there, so I added protection accordingly.

### DIFF
--- a/MekHQ/src/mekhq/campaign/RandomOriginOptions.java
+++ b/MekHQ/src/mekhq/campaign/RandomOriginOptions.java
@@ -18,16 +18,15 @@
  */
 package mekhq.campaign;
 
-import java.io.PrintWriter;
-
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-
 import megamek.common.annotations.Nullable;
 import megamek.logging.MMLogger;
 import mekhq.campaign.universe.Planet;
 import mekhq.campaign.universe.Systems;
 import mekhq.utilities.MHQXMLUtility;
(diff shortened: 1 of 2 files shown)


AaronGullickson:
That log file in #6067 does not match the error. The OP says the error occurred when launching a scenario from a non-stratcon game, which would suggest the same error as #5987, which is already fixed. The log file however is from an attempt to save the game.

AaronGullickson:
I just posted some comments on the issue. I suppose there is nothing wrong with an NPE check, but I am worried we are just chasing ghosts, if we can't replicate the error.

IllianiCBT:
I think you might have the right of it. I'm inclined to leave this PR up, because the extra null protection won't hurt anything. Unless you think that's just gilding the lily.

Lemme know either way. More than happy to close this off if deemed unnecessary. :)

AaronGullickson:
I mean the player got there *somehow* but its unclear to me how they got there and that bothers me.

HammerGS (approved):

PR #6124, opened 2025-03-02
Multiple Immersive Dialog Improvements

IllianiCBT:
- Added the ability to optionally stack user response buttons vertically, instead of horizontally.
- Updated prisoner events to take advantage of this change.
- Reduced the height of the dialog to (hopefully) better factor in the windows taskbar. Previously the bottom of the dialog could appear behind the taskbar. Once again Windows widdles in my cereal.

### DIFF
--- a/MekHQ/resources/mekhq/resources/PrisonerEvents.properties
+++ b/MekHQ/resources/mekhq/resources/PrisonerEvents.properties
@@ -61,11 +61,9 @@ warning.ooc=Your <a href='GLOSSARY:PRISONER_CAPACITY'>Prisoner Capacity</a> is g
 
 btnUnderstood.button=Understood
 
-btnDoNothing.button=Do Nothing
-btnDoNothing.tooltip=Let nature take its course.
+btnDoNothing.button=Do Nothing. Let nature take its course.
 
-free.button=Free {0} Prisoners
-free.tooltip=Lead by example.
+free.button=Free {0} Prisoners. Lead by example.
 free.report={0} has been freed.
 freeEvent0.message={0}, as ordered, we''ve released prisoners from the overcrowded sections. They\
   \ were processed swiftly, provided minimal supplies for their journey, and escorted beyond the\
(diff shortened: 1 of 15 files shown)


MekHQ/resources/mekhq/resources/PrisonerEvents.properties
```diff
@@ -61,11 +61,9 @@ warning.ooc=Your <a href='GLOSSARY:PRISONER_CAPACITY'>Prisoner Capacity</a> is g
 
 btnUnderstood.button=Understood
 
-btnDoNothing.button=Do Nothing
-btnDoNothing.tooltip=Let nature take its course.
+btnDoNothing.button=Do Nothing. Let nature take its course.
 
-free.button=Free {0} Prisoners
-free.tooltip=Lead by example.
+free.button=Free {0} Prisoners. Lead by example.
```
psikomonkie, line 66: Could you use choice format here so this shows correctly if you only free one prisoner?

Something like `Free {0} {0,choice,0#prisoners|1#prisoner|2#prisoners}. Lead by example.` to print out `Free 1 prisoner. Lead by example` and `Free 2 prisoners. Lead by example.`.

psikomonkie (approved):
Nice! I think this looks good. I have one comment about our favorite topic (pluralization) but it's not a must change, just wanted something I thought of. Thanks!

MekHQ/resources/mekhq/resources/PrisonerEvents.properties
IllianiCBT: I actually could. This implementation predates us having that whole discussion so I didn't even consider it. Lemme have a ponder.

IllianiCBT: To confirm '{0,choice,0#prisoners|1#prisoner|2#prisoners}' will mean the following:

0 prisoners (somehow) = prisoners
1 prisoner = prisoner
2**+** prisoners = prisoners?

psikomonkie: Yep! I agree it's 0 prisoners (somehow) but I mean we're already adding it so why not. The [Oracle docs on ChoiceFormat ](https://docs.oracle.com/javase/8/docs/api/java/text/ChoiceFormat.html) have the algorithm described as `X matches j if and only if limit[j] ≤ X < limit[j+1]`, followed by `If there is no match, then either the first or last index is used, depending on whether the number (X) is too low or too high. If the limit array is not in ascending order, the results of formatting will be incorrect. ChoiceFormat also accepts \u221E as equivalent to infinity(INF).`.

IllianiCBT: Answered my own question. This is _great_!

PR #6129, opened 2025-03-02
 Prevented New Contract Offers When an Active Contract Exists

IllianiCBT:
- Updated `AtbMonthlyContractMarket` and `CamOpsContractMarket` to restrict generating new contract offers if the player has an active contract (AtB or otherwise).
- Ensured MekHQ's limitation of supporting only one main contract (apart from subcontracts) is enforced.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/contractMarket/CamOpsContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/contractMarket/CamOpsContractMarket.java
@@ -69,10 +69,19 @@ public class CamOpsContractMarket extends AbstractContractMarket {
     @Override
     public void generateContractOffers(Campaign campaign, boolean newCampaign) {
         boolean isGrayMonday = isGrayMonday(campaign);
+        boolean hasActiveContract = campaign.hasActiveContract()
+            || campaign.hasActiveAtBContract(true);
 
         if (!(campaign.getLocalDate().getDayOfMonth() == 1) && !newCampaign) {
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/market/contractMarket/CamOpsContractMarket.java
```diff
@@ -69,10 +69,19 @@ public AtBContract addAtBContract(Campaign campaign) {
...
 
         if (!(campaign.getLocalDate().getDayOfMonth() == 1) && !newCampaign) {
             return;
         }
+
+        // If the player has an active contract, they will not be offered new contracts,
+        // as MekHQ doesn't support multiple contracts (outside of subcontracts).
+        if (hasActiveContract) {
+            return;
+        }
```
psikomonkie, line 83: Would this mean there'd be no contracts available right after completing a contract? I know it's a relatively small amount of time before the next contracts generate (monthly) but it would be nice if there were contracts available when the user finishes a contract.

IllianiCBT: There won't be by design. This is another nudge for the user to take breaks between contracts. You can logic it as there not being enough time to shop for new contract opportunities while the unit is in the middle of combat actions.

psikomonkie (approved):
Cool! Looks good. Just wanted to check about that one question. Thanks!

PR #6131, opened 2025-03-02
Enabled Autosave Before Missions By Default

IllianiBird:
- Updated `getAutosaveBeforeMissionsValue` to default to `true` instead of `false`.
- Ensures missions are autosaved by default.

### Dev Notes
I doubt this change will affect installs that were made prior to this change going into effect - namely, users running from master instead of an artifact.

We made this change as it increases the likelihood a player will have a pre-scenario save on hand when it comes to opening issue reports. It also improves user experience by allowing players to more reliably roll back disastrous scenarios, should they wish.

### DIFF
--- a/MekHQ/src/mekhq/MHQOptions.java
+++ b/MekHQ/src/mekhq/MHQOptions.java
@@ -18,17 +18,16 @@
  */
 package mekhq;
 
-import java.awt.Color;
-import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
-
-import javax.swing.UIManager;
-
 import megamek.SuiteOptions;
 import megamek.common.annotations.Nullable;
 import mekhq.campaign.universe.enums.CompanyGenerationMethod;
 import mekhq.gui.enums.ForceIconOperationalStatusStyle;
 import mekhq.gui.enums.PersonnelFilterStyle;
+
+import javax.swing.*;
+import java.awt.*;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
 
 public final class MHQOptions extends SuiteOptions {
     // region Display Tab
@@ -759,7 +758,7 @@ public final class MHQOptions extends SuiteOptions {
 
     public boolean getAutosaveBeforeMissionsValue() {
         return userPreferences.node(MHQConstants.AUTOSAVE_NODE).getBoolean(MHQConstants.SAVE_BEFORE_MISSIONS_KEY,
-                false);
+                true);
     }
 
     public void setAutosaveBeforeMissionsValue(boolean value) {


MekHQ/src/mekhq/MHQOptions.java
```diff
@@ -18,18 +18,17 @@
...
-import javax.swing.UIManager;
-
 import megamek.SuiteOptions;
 import megamek.common.annotations.Nullable;
 import mekhq.campaign.universe.enums.CompanyGenerationMethod;
 import mekhq.gui.enums.ForceIconOperationalStatusStyle;
 import mekhq.gui.enums.PersonnelFilterStyle;
 
+import javax.swing.*;
+import java.awt.*;
```
psikomonkie, line 28: I've been asked to avoid unnecessary wildcards in imports. What else are we using from `javax.swing` and `java.awt` that we want everything?

IllianiCBT: This was formatted using the set up Tex did. So I've been assuming it's correct. Lemme take a look.

IllianiCBT: swing components are used around 673 times. awt is used around 8. I believe our cut-off is around 6 uses.

psikomonkie: It’s not uses but the number of different packages used. It looks like there’s just `Color` and `UIManager`, one from each, being used in this class.

IllianiCBT: Ah, my apologies, lemme take another look.

IllianiCBT: So the issue I'm running into is that I can't un-wildcard it, because whenever I delete the import IDEA sees there are so many uses of the package that it automatically adds it back with wildcard. Any ideas?

IllianiCBT: And find usages just says...

image

IllianiCBT: Swing has.... JTextField, Slider, JCheckBox, JSpinner, JRadioButton, Frame, JTabbedPane, setName, Box, JLabel -- there are a bunch more.

awt is FlowLayout, Component, and a bunch of methods

psikomonkie: > Swing has.... JTextField, Slider, JCheckBox, JSpinner, JRadioButton, Frame, JTabbedPane, setName, Box, JLabel -- there are a bunch more.
> 
> awt is FlowLayout, Component, and a bunch of methods

None of those are being used in `MHQOptions` - your screenshot is of a different file, `MHQOptionsDialog`

IllianiCBT: For those following along at home. I was, indeed, in the wrong class.

psikomonkie (approved):
Nice! This is a great change. 

Follow your heart on the imports.

PR #6134, opened 2025-03-02
Refactored Campaign Force Value Calculation

### LINKED ISSUE #6027: [RFE] Mark Forces as Non Payable by Contract
### Enhancement Type *

Not all forces on the TOE should be considered when factoring contract pay (and number of forces requirements). My Jumpships and dropships are compensated by my transport pay. I have shuttles and passenger ships that my employer is likely not hiring. I might only want to manage a Company or Battalion of forces out of my regiment, so I mark the rest as support. My employer should not be paying for those.

This could be as simple as a right click option in the TO&E or as (Awesomely) complex as a new drag and drop interface in contract negotiations (where they say you need to provide 7 forces, you drag in the 7 forces you are hiring out, plus transport and convoy, and thats what they pay based on).

### Use Cases or Rationale

This helps mitigate the contract pay issue where they are currently paying for everything on your TO&E even if its not available for them to use or capitalize on.

Control of the contract pay helps give the player something to strive for rather than sitting on tens of billions wondering: What now?

Rulebook Reference: Campaign Ops specifically references combat units for contract pay. Units that are not used in combat should not factor into contract pay. But people do like having all their equipment in their TO&E.

### Attach Files

Image
Image

### MekHQ Suite Version *

50.03

### Final Verification

IllianiCBT:
- Refactored `getForceValue` method to remove reliance on a `CampaignOptions` object. Updated it with modern method calls and improved the layout for ease of human parsing.
- Refactored `getContractBase` to improve readability.
- Changed the `getAllUnits` call to only return units in 'STANDARD' forces. This will exclude any force with a non-combat assignment.
- Refactored the `Accountant` class to use a Java record for simplicity and immutability.
- Replaced getter methods with direct record accessor usage (`campaign()`).

Fix #6027

### Dev Notes
By including only combat forces we better align with CamOps RAW, which specifically references 'combat units', which would exclude forces assigned to non-combat formations. At that point they cease to be combat units.

Also, seeing that `Accountant.java` was being used in an immutable fashion, it made perfect sense to flip it into a `Record`.

### DIFF
--- a/MekHQ/src/mekhq/campaign/finances/Accountant.java
+++ b/MekHQ/src/mekhq/campaign/finances/Accountant.java
@@ -1,7 +1,7 @@
 /*
  * Accountant.java
  *
- * Copyright (c) 2020 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2020-2025 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
@@ -38,23 +38,14 @@ import java.util.UUID;
 /**
  * Provides accounting for a Campaign.
  */
-public class Accountant {
-    private final Campaign campaign;
-
-    public Accountant(Campaign campaign) {
-        this.campaign = campaign;
-    }
-
-    public Campaign getCampaign() {
-        return campaign;
-    }
+public record Accountant(Campaign campaign) {
 
     public CampaignOptions getCampaignOptions() {
-        return getCampaign().getCampaignOptions();
+        return campaign().getCampaignOptions();
     }
 
     public Hangar getHangar() {
-        return getCampaign().getHangar();
+        return campaign().getHangar();
     }
 
     public Money getPayRoll() {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -221,37 +243,55 @@ public Money getEquipmentContractValue(Unit u, boolean useSaleValue) {
...
+        }
         // And pay our pool
-        payRollSummary.put(null, Money.of(
-            (getCampaign().getCampaignOptions().getRoleBaseSalaries()
-                [PersonnelRole.ASTECH.ordinal()].getAmount().doubleValue()
-                * getCampaign().getAstechPool())
-            + (getCampaign().getCampaignOptions().getRoleBaseSalaries()
-                [PersonnelRole.MEDIC.ordinal()].getAmount().doubleValue()
-                * getCampaign().getMedicPool())));
+        payRollSummary.put(null, Money.of((campaign().getCampaignOptions().getRoleBaseSalaries()[PersonnelRole.ASTECH.ordinal()].getAmount().doubleValue() * campaign().getAstechPool()) + (campaign().getCampaignOptions().getRoleBaseSalaries()[PersonnelRole.MEDIC.ordinal()].getAmount().doubleValue() * campaign().getMedicPool())));
```
psikomonkie, line 294: When we have very long statements should we be including line breaks? I think it makes them easier to read.

IllianiCBT: Ugh, this was not an intended change.

psikomonkie (approved):
Nice! Another feature I'm very excited for. Thanks for getting that formatting fixed up quickly! It looks great.

PR #6139, opened 2025-03-03
 Enforced Skill Value Boundaries

IllianiCBT:
- Introduced `COUNT_UP_MAX_VALUE` and `COUNT_DOWN_MIN_VALUE` constants in `Skill` to define boundaries for skill values.
- Updated `Skill` methods to use `min` and `max` functions to enforce these boundaries when calculating skill values.
- Refactored `CreateCharacterDialog` and `CustomizePersonDialog` to utilize the new boundary constants and methods for consistent skill value calculations.
- Improved clarity of variable names in relevant methods for better readability and maintainability.

### Dev Notes
On this edition of 'Illiani Hates Fun', I decided to go ahead and clamp skills to a logical upper bound. While I do recognize that some players rather enjoy their -8/-8 übermensch, having skills fall below the expected minimum of 0/0 has previously been a problem (and will almost certainly be again).

For bonuses gained from Education (or elsewhere) this means that characters can reach the 'skill cap' faster than characters without the bonuses, but cannot beat them. With the XP cost increases we added last year such bonuses are still very powerful and I'm hoping players won't be too upset.

- Skills that count up cannot exceed 10.
- Skills that down cannot exceed 0.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/Skill.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Skill.java
@@ -20,15 +20,17 @@
  */
 package mekhq.campaign.personnel;
 
-import java.io.PrintWriter;
-
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-
 import megamek.common.Compute;
 import megamek.common.enums.SkillLevel;
 import megamek.logging.MMLogger;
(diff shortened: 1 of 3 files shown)


Sleet01 (approved):
LGTM

PR #6155, opened 2025-03-03
Added Mothball Option for Unit Market Deliveries

### LINKED ISSUE #5734: [RFE] Allow option for purchased units to arrive mothballed
### Enhancement Type *

New Feature

Currently you can GM mothball a unit even when it's on transit, or GM mothball once they arrive, which helps when you don't have enough techs to support the unit, but given that units slowly degrade in the hangar/while in transit/etc without being mothballed/attended, it makes more sense if they are able to be shipped mothballed at the campaign's option.

### Use Cases or Rationale

1. Reduce strain on units running at low tech capacity
2. Avoids new units suffering damage due to nags being turned off if a unit arrives and goes unnoticed
3. 'Makes sense' - verisimilitude increases as units aren't being shipped across the galaxy unmothballed, where players will recognize that it would likely arrive broken and degraded due to unattended transit.

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.03 SNAPSHOT from 01/06 Nightly

### Final Verification

- [X] I confirm this request hasn't already been submitted (checked the tracker)
- [X] I've discussed or asked about this enhancement on MegaMek Discord
- [X] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

IllianiCBT:
- Introduced a new campaign option: "Deliver Units Mothballed".
    - Units purchased from the unit market will now arrive in a mothballed state if enabled.
- Updated `Unit` logic to reflect the mothballed delivery state.
    - Units marked as mothballed but still in transit will display as "In transit (`n` days)".
- Enhanced `CampaignOptions` to handle the new `mothballUnitMarketDeliveries` flag.
    - Added appropriate getter, setter, and XML persistence updates.
- UI changes:
    - Added a checkbox in the Markets Tab of the campaign options dialog for enabling "Deliver Units Mothballed".
    - Updated tooltips and labels to describe the new functionality.

Fix #5734

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -1521,6 +1521,9 @@ lblUnitMarketRarityModifier.tooltip=This setting increases or decreases the roll
 lblInstantUnitMarketDelivery.text=Enable Instant Deliveries
 lblInstantUnitMarketDelivery.tooltip=Units bought from the unit market appear in the hangar\
   \ immediately rather than having to wait for delivery.
+lblMothballUnitMarketDeliveries.text=Deliver Units Mothballed
+lblMothballUnitMarketDeliveries.tooltip=Units bought from the unit market will arrive in a\
+  \ mothballed state.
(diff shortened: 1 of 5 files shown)


Sleet01 (approved):
LGTM

PR #6156, opened 2025-03-03
Added New Academies To Prestigious Academies Dataset

IllianiCBT:
- Added "New Avalon Institute of Science" and its revised version for 3079, detailing their history, qualifications, and curriculums. This data was provided by @LadyAdia.
- Added data for the reopening of the War Academy of Mars.

### Dev Notes
Prior to the ilClan Era, the last we heard about the War Academy of Mars was that it had been destroyed and there were no plans to have it reopened. However, we now know that that it reopened _at some point_ only to be destroyed (again) by Clan Jade Falcon during the opening salvos of the Battle for Terra (3151). We were unable to find a canon source for the re-opening year, so after discussing the situation with @HammerGS we opted to go with 3135.

### DIFF
--- a/MekHQ/data/universe/academies/Prestigious Academies.xml
+++ b/MekHQ/data/universe/academies/Prestigious Academies.xml
@@ -4557,6 +4557,81 @@
         <baseAcademicSkillLevel>0</baseAcademicSkillLevel>
     </academy>
     <academy>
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Nice!

PR #6157, opened 2025-03-03
 Updated Battlefield Control Logic in Scenarios

IllianiBird:
- Introduced `BattlefieldControlType` in `ScenarioTemplate` to handle battlefield control logic with options: `VICTOR`, `PLAYER`, and `ENEMY`.
- Updated dialogs to include detailed messages based on the battlefield control type from scenario templates.
- Modified `autoResolveConcluded` and other resolution methods to display control-specific logic and messages.
- Added XML serialization and deserialization support for `BattlefieldControlType` using `BattlefieldControlTypeAdapter`.
- Enhanced scenario cloning to include battlefield control configuration.
- Introduced a new <battlefieldControl> tag with appropriate values to all scenario templates.

image

### Dev Notes
Battlefield control is a crucial balancing tool that we’ve largely underutilized in recent months. When players don’t control the field during a scenario, they miss out on valuable salvage. The less salvage they can collect, the slower their overall progression—helping to maintain balance.

It also adds another layer of strategy when choosing non-Turning Point scenarios. If a scenario isn’t a Turning Point but offers salvage, players are more likely to take the risk. This encourages more thoughtful decision-making and strategic play.

At some point, we’ll be introducing RAW (or near-RAW) salvage, which will significantly reduce player access to salvage. This PR serves as a small step in that direction, helping to ease the transition so the final reduction doesn’t feel as abrupt.

### DIFF
--- a/MekHQ/data/scenariotemplates/Annihilation.xml
+++ b/MekHQ/data/scenariotemplates/Annihilation.xml
@@ -2,7 +2,10 @@
 <ScenarioTemplate>
     <name>Annihilation</name>
     <shortBriefing>Obliterate the enemy force.</shortBriefing>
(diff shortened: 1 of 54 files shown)


Sleet01 (approved):
LGTM

PR #6159, opened 2025-03-03
Renamed Method For Improved Clarity

IllianiCBT:
- Renamed `getAvailableForceIDs` to `getAvailableForceIDsForManualDeployment` for better clarity and specificity.
- Updated all method calls and references to use the new method name.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -2212,8 +2212,11 @@ public class StratconRulesManager {
      *                      state for further filtering of eligible forces.
      * @return a {@link List} of unique force IDs that meet all deployment criteria.
      */
-    public static List<Integer> getAvailableForceIDs(int unitType, Campaign campaign, StratconTrackState currentTrack,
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
```diff
@@ -2212,8 +2212,11 @@ public static List<Integer> getAvailableForceIDs(Campaign campaign, AtBContract
      *                      state for further filtering of eligible forces.
      * @return a {@link List} of unique force IDs that meet all deployment criteria.
      */
-    public static List<Integer> getAvailableForceIDs(int unitType, Campaign campaign, StratconTrackState currentTrack,
-            boolean reinforcements, @Nullable StratconScenario currentScenario, StratconCampaignState campaignState) {
+    public static List<Integer> getAvailableForceIDsForManualDeployment(int unitType, Campaign campaign,
```
IllianiCBT, line 2215: I should probably clarify for the confused: the reason this change was made is that we already have a method called `getAvailableForceIDs` and they are not for the same purpose. This has confused the heck out of me on more than on occasion.

PR #6160, opened 2025-03-03
 Replaced Magic Numbers With Constants For Scenario IDs

IllianiCBT:
- Added `NO_ASSIGNED_SCENARIO` as a new constant in the `Force` class.
- Replaced `-1` for unassigned scenario ID with the new `NO_ASSIGNED_SCENARIO` constant across relevant files.
- Updated all references and logic checks to use these new constants.
- Replaced `-1` for a couple of unassigned force IDs with the `FORCE_NONE`. This is far from exhaustive, the couple changed were just those I stumbled over while addressing Scenario ID.


Sleet01 (approved):
LGTM

PR #6162, opened 2025-03-04
Updated Manual Unit Rating Modifier Range

IllianiBird:
- Expanded the `ManualUnitRatingModifier` spinner range from [-200, 200] to [-1000, 1000].
- Allows for greater flexibility in campaign customization.

### Dev Notes
During the recent Reputation issues it was observed that [-200, 200] was too small an adjustment for some of our mega campaigns to be able to adjust Reputation in a meaningful way. This update should better factor in those players.


psikomonkie (approved):
Nice!

PR #6164, opened 2025-03-04
fix: PACAR auto-start only when using CommanderGUI

### LINKED ISSUE #6059: [RFE] Add option to set victory conditions for PACAR
### Enhancement Type *

Add an option to let you choose victory conditions for fights run by PACAR, for example what % of bv needs to be destroyed to declare one side a winner. Alternatively, try to somehow inherit them from scenario description.

### Use Cases or Rationale

Currently launching a scenario battle using PACAR tries to automatically start the MegaMek fight once all units have been added. However, Stratcon missions can have different victory conditions even for destroying units: 25%, 50% and 75%. Setting victory conditions in MegaMek allows to fight the battle according to Stratcon goals without having to keep track of bv and manually stop the fight when the target % is reached.

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.03

### Final Verification

Scoppio:
Stops PACAR from starting the game before the player has a chance to setup the scenario, game rules and units in the way they want when running with the legacy Client GUI.

- closes #6059 (it lists as an RFE, but after reflection, the behavior was wrong and should not be starting the game automatically, so I am treating as a bug instead)

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -543,15 +543,15 @@ public class AtBGameThread extends GameThread {
                         botClient.sendDone(true);
                     }
                     Thread.sleep(MekHQ.getMHQOptions().getStartGameBotClientDelay());
-                    if (swingGui != null && swingGui instanceof CommanderGUI commanderGUI) {
-                        commanderGUI.enableReady();
-                        commanderGUI.getLocalBots().put(bot.getName(), bot);
-                    }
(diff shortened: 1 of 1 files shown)


IllianiCBT (commented):
How do players set up the scenario parameters if they're using Commander GUI?

Scoppio:
> How do players set up the scenario parameters if they're using Commander GUI?

They dont, the Commander's GUI is experimental specifically because I don't know what is missing to consider it complete.
So now I have already 3 features I need to implement on it, but right now, thats what it is.

PR #6168, opened 2025-03-04
Improved the Distinguishing of Brand New Parts Handling

IllianiCBT:
- Introduced a flag to differentiate new vs. used parts during procurement and warehouse operations.
- Added a new `addPart` method in `Quartermaster` to explicitly handle brand-new parts, deprecating the older version.
- Updated warehouse merging logic to account for the `isBrandNew` property, ensuring compatibility during merges.
- Revised part quantity management in `Part` class, replacing `incrementQuantity` and `decrementQuantity` with the more flexible `changeQuantity` method and marking the former methods as deprecated.
- Ensured detailed logging and clarification in the process of adding parts, merging spares, and managing inventory across the codebase.
- Updated Unit Tests to match the new arrangement.

### Dev Notes
The logic here is that items that are explicitly purchased will have the 'Brand New' flag (as before), while all other sources of parts will cause the part to be classified as 'used'. Previously this only happened for parts that were explicitly acquired through salvage.


MekHQ/src/mekhq/campaign/parts/Part.java
```diff
@@ -732,7 +732,7 @@ public static Part generateInstanceFromXML(Node wn, Version version) {
                 } else if (wn2.getNodeName().equalsIgnoreCase("isTeamSalvaging")) {
                     retVal.isTeamSalvaging = wn2.getTextContent().equalsIgnoreCase("true");
                 } else if (wn2.getNodeName().equalsIgnoreCase("brandNew")) {
-                    retVal.brandNew = wn2.getTextContent().equalsIgnoreCase("true");
+                    retVal.brandNew = wn2.getTextContent().equalsIgnoreCase("false");
```
psikomonkie, line 735: How will this impact saves made prior to this - will this cause parts form older MekHQ saves to flip form new to old and vice versa?

IllianiCBT: Oh bugger, you’re right. Lemme think on how to handle that

psikomonkie: I think instead of assuming a missing `brandNew` tag is true/false it might be best to always save the tag. While it's slick not having to write the tag sometimes I think the ambiguity is more liable to cause problems than the slight increase in file size.

IllianiCBT: I've not had time to put any thought into this - I've been trapped in the Pit of Editing for the past 4 days - however, I do think we need to always save the tag.

I'm just not sure whether that will fix the issue of all parts switching state the first time a save is loaded into 50.04.

IllianiCBT: So what I've done is create #6225. The idea is that users will update to 50.04 (especially if it becomes a Milestone), at which point the 'Brand New' status of their Parts will be stored, no matter whether it's `true` or `false`. Then, in 50.05 we can merge this without issue, as users are meant to progress Milestone -> Milestone.

Sleet01:
If this PR is still active, it looks like the conflicts need resolving.

IllianiCBT:
It’s still active, but as it’s flagged for new dev cycle I’m just not bothering to fix the conflicts until after launch

IllianiCBT:
Updated for 50.05

PR #6169, opened 2025-03-04
 Updated `PartsTableModel.java` To Better Indicate Brand-New Status in Quality Column

IllianiCBT:
- Removed italicized formatting of brand-new parts in the Name column for a cleaner presentation.
- Introduced a new resource bundle (`PartsTableModel.properties`) to centralize localization strings for `PartsTableModel`.
- Replaced hardcoded column names with calls to `getFormattedTextAt` to dynamically fetch localized strings from the resource bundle
- Removed redundant hardcoded strings to improve maintainability and support for additional localization in the future.
- Simplified and modernized `getColumnName` logic using the `switch` expression.
- Appended "(Brand New)" and "(Used)" to the Quality column for improved clarity and visibility of part status.

### Dev Notes
This should cut back on the number of contacts we get from users asking what the italic name in their warehouse means.

### DIFF
--- a/MekHQ/resources/mekhq/resources/PartsTableModel.properties
+++ b/MekHQ/resources/mekhq/resources/PartsTableModel.properties
@@ -1,0 +1,13 @@
+label.COL_NAME=Name
+label.COL_COST=Value per Unit
+label.COL_TOTAL_COST=Total Value
+label.COL_QUANTITY=#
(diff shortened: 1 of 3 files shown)


psikomonkie (approved):
Lovely! I especially like changing the columns to be i18n keys.

PR #6173, opened 2025-03-05
Corrected Support Points Removal Logic

IllianiCBT:
- Refactored `btnGMRemoveSP` action listener to directly call `removeSP`, removing unnecessary `dispose()` call.
- Updated `removeSP` method to use `changeSupportPoints()` for modifying support points instead of directly setting them.
- Adjusted condition to allow removal of support points when greater than 0, instead of 1.


psikomonkie (approved):
Nice!

Issue #6123, opened 2025-03-02
edited Campaign with drones can not be loaded anymore


OctaviusValesius:
### Brief Description *

Once a campaign has been edited, it can no longer be loaded.

### 3. Steps to Reproduce

I have been getting this message (s. screen) since I last saved the campaign. What do I have to change in the file to fix this?
The last thing I did was the following:
Assigned a rank to all personnel.
I have converted remote controlled drones to personnel and assigned personnel to be able to assign them to the TOE. Without crew, they cannot be assigned to the TOE, not even as drones.
After I assigned them to the TOE, I threw the crew out of the campaign and re-equipped the drones. (State was "Crippeld" again)
I repeated the same procedure with an automated space station.
saved

### Attach Files

[Odysseus.zip](https://github.com/user-attachments/files/19042672/Odysseus.zip)

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

v0.50.03

### Operating System *

Windows 10

### Java Version *

Eclipse Adoptium

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

psikomonkie:
More details from Discord: https://discord.com/channels/458705327911731231/458706545379639296/1345302031543566436

rjhancock:
Please provide a meaningful title to the issue and all meaningful details within the issue.

psikomonkie:
@OctaviusValesius Please include all custom units. I am unable to load your save and I don't encounter the error you did: 

![Image](https://github.com/user-attachments/assets/d9c79dbe-d59d-4638-9734-c007c455cdc4)

OctaviusValesius:
Hello psikomonkie,

please excuse the missing files, I have just uploaded them. That's what happens when you save the equipment in different folders according to progress in an RPG campaign :D

[missing.zip](https://github.com/user-attachments/files/19083847/missing.zip)

psikomonkie:
Hey Octavius, 
[...]

PR #6177, opened 2025-03-05
Issue 6123: Mark drones as unsupported in MHQ.

psikomonkie:
Relates to #6123 

Drones are currently not supported by MekHQ. This change prevents the player from acquiring them.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignGUI.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignGUI.properties
@@ -48,7 +48,7 @@ miMHQOptions.toolTipText=This launches the MekHQ Options Dialog.
 menuThemes.text=Themes
 menuExit.text=Exit
 mekSelectorDialog.unsupported.gunEmplacement=Gun Emplacements are %s<b>not supported</b>%s by MekHQ.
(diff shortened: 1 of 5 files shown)


MekHQ/resources/mekhq/resources/CampaignGUI.properties
```diff
@@ -48,6 +48,8 @@ miMHQOptions.toolTipText=This launches the MekHQ Options Dialog.
 menuThemes.text=Themes
 menuExit.text=Exit
 mekSelectorDialog.unsupported.gunEmplacement=Gun Emplacements are %s<b>not supported</b>%s by MekHQ.
+mekSelectorDialog.unsupported.droneOs=Units with drone os are %s<b>not supported</b>%s by MekHQ.
```
IllianiCBT, line 51: My only comment is that, as a new player, I wouldn't know what 'drone os' was (also, it should probably be 'Drone OS'). Can we clear this up by rewording it at all? Maybe 'Drone units are not supported...'?

IllianiCBT (approved):
Just one comment, otherwise everything is good to go.

MekHQ/resources/mekhq/resources/CampaignGUI.properties
psikomonkie: I agree, that's a good change. It's what you suggested now.

PR #6190, opened 2025-03-07
Enforced Exchange Salvage for Contracts vs. Clan Factions Prior to Tukkayid

IllianiCBT:
- Implemented `clanTechSalvageOverride` method in `AtBContract` to enforce salvage exchange rules for Clan tech before the Battle of Tukayyid.
- Applied the `clanTechSalvageOverride` method to all relevant contract evaluation and processing points across `CamOpsContractMarket`, `AtbMonthlyContractMarket`, and `AbstractContractMarket`.
- Refactored method parameters in `rerollClause` for better consistency.


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -1662,6 +1662,26 @@ public AtBContract(Contract c, Campaign campaign) {
...
+    /**
+     * Applies a salvage override rule for Clan technology based on the contract timeline and faction
+     * involvement. This method checks the factions of both the enemy and employer and determines
+     * if a salvage exchange should be forced based on whether the battle occurs before the Battle
+     * of Tukayyid.
+     *
+     * <p>This rule was implemented to better match canon employer behavior during this period.</p>
+     */
+    public void clanTechSalvageOverride() {
+        final LocalDate BATTLE_OF_TUKAYYID = LocalDate.of(3052, 5, 21);
```
psikomonkie, line 1678: This will be the third spot in MekHQ where we're manually defining the Battle of Tukayyid date. Should we consider making it a constant so we don't need to manually enter it every time?

psikomonkie (approved):
Nice! One question, not a blocker for merging though.

IllianiCBT:
So I was actually thinking of exactly that. A class that exists just to store important dates.

MekHQ/src/mekhq/campaign/mission/AtBContract.java
IllianiCBT: So I was actually thinking of exactly that. A class that exists just to store important dates.

IllianiCBT: Outside the scope of this PR, mind, but yeah I do agree.

PR #6203, opened 2025-03-09
Added Suppress Inspection Annotations To `PrisonerEvents.properties`

IllianiCBT:
- Suppressed "UnusedMessageFormatParameter" inspection for the entire file.
- Suppressed "UnusedProperty" inspection for the entire file.

### Dev Notes
This just suppresses these warnings, so that it's easier to spot if there are actual issues at play.

The former is irrelevant because of how i18n works - you no longer need to have placeholders in the right order, or always include every placeholder.

The latter is irrelevant because we dynamically build the reference key, rather than direct referencing. So IDEA thinks the entries aren't being used.

### DIFF
--- a/MekHQ/resources/mekhq/resources/PrisonerEvents.properties
+++ b/MekHQ/resources/mekhq/resources/PrisonerEvents.properties
@@ -1,3 +1,5 @@
+# suppress inspection "UnusedMessageFormatParameter" for whole file
+# suppress inspection "UnusedProperty" for whole file
 # CapturePrisoners.java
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Great idea! I'll need to start doing this to property files I work on that have dynamically generated keys. I don't run into the unused property often, but it's good to know how to suppress it.

IllianiCBT:
I stumbled across the how entirely by accident. I use dynamic references a lot, nowadays, so it’s a tool I’ll be using a lot.

PR #6206, opened 2025-03-09
Reduced Attempt Count For Capturing Prisoners

IllianiCBT:
- Adjusted ATTEMPT_COUNT in `CapturePrisoners.java` from 3 to 2 to refine balance and gameplay mechanics.
- Updated related documentation `Prisoners of War & Abstracted Search and Rescue.pdf` to reflect the change.

### Dev Notes
This change was made to reduce the number of prisoners players need to contend with. Once we implement SAR forces in the TO&E this will be a dynamic value, however for now it's fixed.


MekHQ/docs/Stratcon and Against the Bot/Prisoners of War & Abstracted Search and Rescue.pdf
psikomonkie, line 1: Can you double check what's committed for this? I might be doing something wrong and looking at the wrong version, but I think it still has the old value here: 
image

psikomonkie (approved):
Nice! One question, might be me doing something wrong.

MekHQ/docs/Stratcon and Against the Bot/Prisoners of War & Abstracted Search and Rescue.pdf
IllianiCBT: Yeah, that's wrong. I'll get that corrected in a moment :)

PR #6208, opened 2025-03-10
Rebalanced Support Point Negotiation Logic

IllianiCBT:
- Removed the `calculateSupportPoints` method, simplifying the logic for awarding support points.
- Adjusted the `SupportPointNegotiation` process to increment `negotiatedSupportPoints` directly when the roll result meets or exceeds the admin's skill level.

### Dev Notes
The change here is to remove the 'margin of success' benefit from characters rolling really high. While implemented with the best of intentions this effectively resulted in a snowballing of Support Points once campaigns reached sufficient maturity that they had multiple Elite Admin/Transport characters.

I'm hoping that this change will result in a better reinforcement economy, however if it doesn't the next step will be to use a fixed TN with skill being applied as a modifier. Rather than the TN being the skill target number.


psikomonkie (approved):
Nice!

PR #6209, opened 2025-03-10
Updated Reinforcement Target Calculation Logic

IllianiCBT:
- Refactored `calculateReinforcementTargetNumber` method by removing unused parameters: `Campaign`, `StratconScenario`, and `StratconCampaignState`.
- Replaced facility-based modifier calculation with an enemy morale-based modifier for better balancing.
- Updated call to `calculateReinforcementTargetNumber` in `StratconScenarioWizard` to match the new method signature.

### Dev Notes
This was a change I actually thought I made the other week. It stops player reinforcements from becoming progressively easier (to the point of being nonsensically easy) as a contract progresses and the player captures more facilities.

The facility modifier was the primary cause being so many negative target numbers, with reinforcements.

The intention of the facility modifier was to reflect the OpFors control of a sector. However, really, Enemy Morale better reflects that and has far better checks and balances to ensure Target Numbers remain within reasonable bounds.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconRulesManager.java
@@ -35,7 +35,6 @@ import mekhq.campaign.event.StratconDeploymentEvent;
 import mekhq.campaign.force.CombatTeam;
 import mekhq.campaign.force.Force;
 import mekhq.campaign.mission.*;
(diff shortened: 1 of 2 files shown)


DM0000 (approved):
Looks good.  Just to confirm facilities no longer give reinforcement modifier just based off moral, liaison skill, and enemy skill right?

IllianiCBT:
More or less yeah, the doc has the updated details (in a separate PR)

PR #6240, opened 2025-03-11
Adjusted Resupply Tonnage Calculation to Enforce a Minimum Weight of 1 Ton

IllianiCBT:
- Added a minimum weight constraint (`CARGO_MINIMUM_WEIGHT`) for resupply cargo tonnage calculations.
- Updated the cargo tonnage computation to use the `max` function, ensuring it adheres to the minimum weight threshold.
- Included the `Math.max` static import to support the new calculation logic.

### Dev Notes
This prevents a bug where convoy size can become 0t.


Sleet01 (approved):
LGTM

PR #6243, opened 2025-03-11
Reduced the Reinforcement Arrival Time of Non-Spheroid AeroSpace Assets (Including DropShip Transported Units)

IllianiCBT:
- Added a x2 multiplier to ASF 'walk' speed, for the purposes of determining reinforcement speed.
- This modifier excludes spheroid DropShips and Small Craft.
- Updated method documentation to reflect changes.

### Dev Notes
The goal of this change is to encourage players to intelligently use their DropShip assets in a way that matches the canon. The humble Leopard is a lot more appealing if it can change the reinforcement speed of your assault lance from 8 to 2.


Sleet01 (approved):
LGTM

PR #6249, opened 2025-03-11
 Decoupled Campaign Dependencies in Nag Logic and Tests

IllianiCBT:
- Refactored nag dialog logic to decouple campaign-level dependencies by passing specific data (e.g., `List<Unit>`, `boolean flags`) instead of the full `Campaign` object.
- Updated related unit tests to use the new method signatures and removed irrelevant mock logic tied to the `Campaign` class.
- Removed unused fields and redundant mocks for clarity and efficiency in unit test classes.
- Enhanced maintainability by isolating responsibilities of nag checks from the overarching `Campaign` structure.

### Dev Notes
The aim of this PR is to reduce the amount of time the 'campaign' object is passed around while testing for whether a nag should be displayed or not. Due to the size of the campaign object reducing the hand-offs should give us a bit of a performance improvement, especially for very large campaign files. I doubt it'll be game changing, but given how unoptimized our new day code is, any improvement is worth the hassle.


MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -74,6 +74,11 @@ public class Force {
     // region Variable Declarations
     // pathway to force icon
     public static final int FORCE_NONE = -1;
+    /**
+     * This is the id of the 'origin node'. The force from which all other forces descend. Normally
+     * named after the campaign.
+     */
+    public static final int FORCE_ORIGIN = 0;
```
IllianiCBT, line 81: I also added this constant, because we check the origin node a lot, we might as well stop using magic numbers.

PR #6250, opened 2025-03-12
Add Admin Skills Document

stonewall072:
Adding a simple document explaining admin skills.

### DIFF
--- a/MekHQ/docs/history.txt
+++ b/MekHQ/docs/history.txt
@@ -127,7 +127,32 @@ MEKHQ VERSION HISTORY:
 + PR #6203: Added Suppress Inspection Annotations To PrisonerEvents.properties
 + PR #6204: Added Suppression Of UnusedProperty Inspection Comment to CampaignOptionsDialog.properties
 + PR #6205: Added Suppression Of UnusedProperty Inspection Comment to Resupply.properties
-
++ Fix #306: Added 'Can Use' Filter to Owned Unit Picker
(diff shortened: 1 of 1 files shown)


IllianiCBT:
Looks like your branch isn't updated to current master. Don't forget it may need to be updated on all three repos.

IllianiCBT:
In fact, I just noticed you pushed this from your own master. That's generally considered a bad idea and can cause all kinds of madness. You definitely want to be creating a working branch, moving forward.

MekHQ/docs/Stratcon and Against the Bot/Admin Skills.pdf
IllianiCBT, line 1: image
Can we get a signpost to direct the user to the Turnover documentation, if they want to know more.

image
Similarly, can we have a signpost to direct the user to the Combat Team documentation here.

image
Signpost to Combat Team documentation (that also covers reinforcements)

image
Signpost to the prisoner documentation, please

IllianiCBT (changes requested):
A couple of edits, just so users know where to look next. Otherwise, this is a much needed document. Great job. :)

stonewall072:
> In fact, I just noticed you pushed this from your own master. That's generally considered a bad idea and can cause all kinds of madness. You definitely want to be creating a working branch, moving forward.

Yeah sorry, hit was giving me fits last night and I forgot to switch my branch, I will make requested adjustments and resubmit.

Issue #3175, opened 2022-03-16
Enemy pilot SPAs not being loaded on save games where scenarios are already generated


PhoenixHeart512:
### Environment
MekHQ 0.49.7, but also observed in 0.48
Windows 10 21H1
OpenJDK11 (downloaded from link on MegaMek install FAQ)
Using customized AtB campaign ruleset (SPA's enabled, other changes, will attach)

### Description
When a scenario is generated, enemy pilots are randomly given SPA's as normal.
If you then save and load your MekHQ game (specifically, loading a save where the scenario exists) the SPA's will not be there anymore, seems like MekHQ is not keeping a list of pilot SPA's in the savegame data it has for the scenario?

Exact steps to repro:
1. Be in a mission/contract with SPA's enabled in campaign options
2. Scenarios generated on Monday
3. Save game after scenarios are generated
4. Launch scenario, verify enemy pilot SPA's (if none, maybe RNG just decided to not give them any, in which case generate new scenarios and repeat)
5. Close MegaMek (saving on exit from megamek is not necessary)
6. Close "you have become disconnected from the server" MekHQ popup box
7. Load MekHQ game to the save made in step 3
8. Launch scenario, observe that no enemy pilots have SPA's

### Files

[mekhq bug report files.zip](https://github.com/MegaMek/mekhq/files/8274887/mekhq.bug.report.files.zip)

Thom293:
I believe this is related:

If you load my scenario that includes OpFor portraits and launch the battle from MHQ, the OpFor shows up with portraits and SPAs, etc.  BUT if you load same scenario, save the game and close MHQ, and relaunch it, and the launch the battle into MM, the Opfor shows up with no portraits or SPAs.

Then, if you close that, and regenerate bot forces and relaunch, they show up as intended.  So I guess the save games do not save Opfor SPA or Portraits and perhaps other things in already generated scenarios of any type?

![Screenshot 2022-08-08 130349](https://user-images.githubusercontent.com/23645569/183489610-ccdb0947-a447-45f2-b262-5a5ead45b251.png)
![Screenshot 2022-08-08 130742](https://user-images.githubusercontent.com/23645569/183489662-61ba8b08-bfc6-4d4c-ada3-616e351867e5.png)

PhoenixHeart512:
Attaching new save file for version 0.50.03-SNAPSHOT
[NoSPA50-02.zip](https://github.com/user-attachments/files/18293460/NoSPA50-02.zip)

[...]

PR #6255, opened 2025-03-12
Fix MHQ 3175: load Campaign Options and Game Options before everything else

Sleet01:
Fixes the issue where saved enemy forces would not load with SPAs they had at campaign save time.
There may also be some other issues, such as Edge settings reverting to defaults for enemy crews, etc. that were also caused by this behaviour.

Simply put, we rely on the RPG_PILOT_ADVANTAGES game option to decide whether to load SPAs when loading units' crews from the campaign save.
At the same time, the `\<gameOptions\>` block comes after every other block in the save file, and if unread, we get a default value of `false` for RPG_PILOT_ADVANTAGES.
By moving the Campaign and Game Options loading into the first of three loops over the campaign save file, we can ensure that all options are properly populated prior to loading any parts that depend on those options, without spending noticeably more time on reading the file.

This PR also moves the two sets of  options together in the campaign write code for future-proofing.

Testing:
- Loaded campaign from #3175 and confirmed that the original SPA assignments persisted into the MM games.
- Stepped through campaign loading to verify that gameOptions are now loaded before all other blocks.
- Ran all 3 projects' unit tests.

Close: #3175

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -6020,11 +6020,13 @@ public class Campaign implements ITechManager {
         MHQXMLUtility.writeSimpleXMLCloseTag(pw, --indent, "info");
         // endregion Basic Campaign Info
 
-        // region Campaign Options
+        // region Options
         if (getCampaignOptions() != null) {
             getCampaignOptions().writeToXml(pw, indent);
         }
-        // endregion Campaign Options
+        getGameOptions().writeToXML(pw, indent);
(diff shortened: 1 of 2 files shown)


IllianiCBT (approved):
Thanks for looking into this Sleet. There's no way I would have thought to check load order.

Sleet01:
> Thanks for looking into this Sleet. There's no way I would have thought to check load order.

It was a fluke, really: I was looking at the Crew load section and decided to check the Game Options values while stepping through.  That pointed me to the load order part.

PR #6262, opened 2025-03-13
Removed Support For Commander And Individual Initiative Bonus Options

IllianiCBT:
- Eliminated `UseTactics` and `UseInitiativeBonus` options throughout the codebase.
- Removed associated checkboxes, properties, and configurations from GUI components.
- Deleted XML handling and file persistence logic for the removed options.
- Cleaned up related labels and tooltips from resource files.
- Updated campaign option initialization and synchronization logic to reflect the removal.

## Dev Notes  
Expecting some pushback on this one, so I’m laying it out in detail.  

### What Are the Tactics Initiative Modifiers?  
In MekHQ, players can enable one of two Campaign Options to get a bonus to Initiative rolls based on a character's _Tactics_ skill. For **Commander Initiative**, only one character needs to have _Tactics_, and the bonus is applied to all combatants in the scenario. For **Individual Initiative Bonuses**, each character must invest in _Tactics_ separately.

### Why Is This a Problem?  
First, this system undermines the value of **Command Meks** and **Communications Equipment**. Why invest in a Command Mek if you can spend 30 XP for the same bonus? For just 20 XP more, you could match the benefit of 7 tons of Communications Equipment — without the cost in tonnage, BV, or anything else. On top of that, the bonus applies to **any** unit you pilot. If you lose a Mek, you just jump into the next one and keep the same initiative modifier.  

Second, winning Initiative rolls is a massive advantage. If you consistently win initiative, you control the flow of combat.  
- Want to get behind an enemy Mek and stay there? **Initiative**.  
- Want to protect your damaged units from getting picked off? **Initiative**.  
- Want to bait the OpFor and kick their head off? **Initiative**.  

A good player can sometimes work around bad initiative rolls, but there’s no denying that the player who consistently wins initiative has a serious edge.  

Finally, this system is one-sided. We don’t track Tactics for 'entity' level units, which means the OpFor can’t benefit from stacking _Tactics_ to boost Initiative. Even if they could, it would quickly turn into a contest of who had the higher _Tactics_ modifier, not who rolled better. Facing an elite OpFor with +5 Initiative every round wouldn’t be fun — either you’d have to heavily invest in _Tactics_ or accept that you’d never win initiative. Neither scenario makes for a good experience.

**Edit:** I've been advised that I should also add that currently neither Campaign Option works. I'm not sure when it broke, or why it broke and I've been unable to fix it. I discovered this by accident while trying to make it less one-sided and give the OpFor Tactics.

### So, Is Tactics Useless Now?  
Not at all. _Tactics_ is still valuable. It allows you to reroll scenario attributes like weather, time of day, and map type. Avoiding an F4 tornado, skipping a pitch-black night battle, or securing a favorable map can have a huge impact on a scenario. Being able to shape the battlefield conditions is a strategic advantage that more than justifies investing a few points into _Tactics_.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -400,23 +400,6 @@ salariesTab.border="No job too small...no fee too high."\
 
 # createGeneralTab
 lblPersonnelGeneralTab.text=General Options
-
-lblUseTactics.text=Use Commander Initiative Bonus \u270E \u26A0
-lblUseTactics.tooltip=All initiative rolls are modified by the commander's Tactics skill.\
-  <br>\
-  <br><b>Warning:</b> Princess' units do not currently benefit from this modifier unless it is\
-  \ manually applied in MegaMek.\
-  <br>\
-  <br><b>Warning:</b> Should not be combined with 'Use Individual Initiative Bonus.'\
-  <br>\
-  <br><b>Requirement:</b> The commander initiative option must be enabled in MegaMek.
-lblUseInitiativeBonus.text=Use Individual Initiative Bonus \u270E \u26A0
-lblUseInitiativeBonus.tooltip=Each unit has their initiative rolls modified by their Tactics skill.\
-  <br>\
-  <br><b>Warning:</b> Princess' units do not currently benefit from this modifier unless it is\
-  \ manually applied in MegaMek.\
-  <br>\
-  <br><b>Warning:</b> Should not be combined with '=Use Commander Initiative Bonus.'
 lblUseToughness.text=Use Toughness \u270E
 lblUseToughness.tooltip=A character's Toughness score acts as a modifier to all consciousness checks.
 lblUseRandomToughness.text=Randomize Toughness \u270E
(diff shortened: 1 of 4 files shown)


IllianiCBT:
So there is a RAW rule - but it's ATOW, which insofar as I'm aware, while we may use ATOW for inspiration we don't support it.

rjhancock:
Then until there is a time where we support AToW (if we ever do), it's optional in my book.

rjhancock:
My only request... is scalpel, not sledge. Chance of issue is near 0 but not 0.

IllianiCBT:
The thing that gets me, is I have no idea how this ever worked, given that nowhere calls 'useTactics'. And I've gone back a good whiles and insofar as I can tell we never called 'useTactics'. So how did we have a campaign option that is being stored and loaded but never checked.

Any ideas? Because it clearly worked at some point, including earlier in the 50.04 dev cycle, I just don't see how or where.

pheonixstorm:
To add to this, not everyone uses ATB/Stratcon. Most of my campaigns don't. As for AToW, it is only used since every person is tracked. All infantry, every vehicle crew member, every *ship crew member. Do we need it? No, but people use it. It is an option, if if gives an unfair advantage to the player that is the users choice. The user set the skill defaults and xp cost. The user sets the xp costs for SPAs. WE, the developers/contributors just set the defaults. Find where it broke and fix it is the best option. I had to do that several times for other features that broke at some point.

PR #6271, opened 2025-03-13
Corrected One-Way Sync Between MegaMek and Campaign Options

IllianiCBT:
- Introduced `updateCampaignOptionsFromGameOptions` and `updateGameOptionsFromCampaignOptions` methods in `CampaignOptions` for clearer bidirectional synchronization between game and campaign options.  
- Deprecated `applyContinuousToCampaign` in `CampaignPreset` as it is no longer in use, following the arrival of CO IIC.
- Replaced the previous `updateCampaignOptionsFromGameOptions` logic in `Campaign` with a call to the new encapsulated method in `CampaignOptions`.

## Dev Notes  
Marking this as **high** because we really need this in 50.04. I’m genuinely surprised it’s taken this long for it to cause major problems.  

### Overview  
When the user updates MegaMek's options, some choices (such as tech level) are propagated back into Campaign Options. This makes sense — if the user selects "use edge" in MegaMek’s options, they probably want to use edge. However, this connection was entirely one-directional: MegaMek could affect Campaign Options, but Campaign Options could not affect MegaMek.  

This creates a major issue if you adjust MegaMek's options but don’t enable, say, Edge. Since Edge is not enabled by default in MegaMek, it would become disabled in Campaign Options — even though you might expect it to be enabled, because you yourself enabled it.

Meaning, if you *do* want to use Edge, you need to enable it in both Campaign Options and MegaMek’s options manually. If you forget to enable it in MegaMek, Edge won’t work.

### Fix  
This PR includes some minor refactoring but mainly ensures that option synchronization works in both directions. Now, enabling Edge in MegaMek will enable it in Campaign Options — and enabling Edge in Campaign Options will enable it in MegaMek.  

### Synchronized Options  
The following options are now synced between MegaMek and Campaign Options:  
- Use Commander Tactics Bonus to Initiative  
- Use Individual Tactics Bonuses to Initiative  
- Use Toughness  
- Use Artillery  
- Use SPAs  
- Use Edge  
- Use Implants  
- Use Quirks  
- Allow Canon Units Only  
- Maximum Tech Level

### DIFF
--- a/MekHQ/src/mekhq/CampaignPreset.java
+++ b/MekHQ/src/mekhq/CampaignPreset.java
@@ -294,11 +294,17 @@ public class CampaignPreset {
         return presets;
     }
 
+    /**
+     * @deprecated This method is scheduled to be removed following the next milestone
+     * @since 50.04
+     */
+    @Deprecated
     public void applyContinuousToCampaign(final Campaign campaign) {
         if (getGameOptions() != null) {
             campaign.setGameOptions(getGameOptions());
             if (getCampaignOptions() == null) {
-                campaign.updateCampaignOptionsFromGameOptions();
+                campaign.getCampaignOptions().updateCampaignOptionsFromGameOptions(campaign);
+                MekHQ.triggerEvent(new OptionsChangedEvent(campaign));
             }
         }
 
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
LGTM!

IllianiCBT:
Fantastic. I'm now going to merge with prejudice.

PR #6272, opened 2025-03-14
Fixed Typographical Error in GrayMonday Properties

IllianiCBT:
- Corrected a missing escape character for an apostrophe in the GrayMonday.properties file.


Scoppio (approved):
👍🏼

PR #6281, opened 2025-03-15
Fix 6267: Set each entity starting position explicitly

### LINKED ISSUE #6267: [0.50.04-SNAPSHOT] Opfor units using "any" deployment zone despite being set to Center
### Brief Description *

I'm not sure if this should go in the MegaMek tracker or MekHQ tracker, but this is occurring on a scenario spawned from Stratcon, so it's inheriting deployment details from MekHQ.

In this scenario, the Primary Opfor (base defender units) show as "center" deployment in MekHQ, as well as in the MegaMek lobby, both when viewed in the list as well as when you look at them in their "Configure" window.

When it's time to actually deploy, the units are being given an "any" deployment zone and the bot distributed them across the entire map.

I also verified a player would get this too, I connected a second client instead of the bot, and observed the full-map "any" deployment hex highlighting and was able to place units anywhere on the map as the second player.

The scenario is spawned from Stratcon, but I did confirm that if I load the lobby save from a standard MegaMek client I still see the same behavior, so that deployment data is being saved to the file.

I am also including the MekHQ campaign save since the scenario was generated from there originally.

----------------------

I also want to note that I DID use the GM tools -> regenerate bot forces on this scenario as part of some other testing I was doing. It's possible that by regenerating the bot forces, it broke their deployment data. If that ends up being the case, this may be a case of "use GM tools at your own risk". The "bug" in that case then would just be that the broken data isn't visible, and the deployment data APPEARS healthy right up until unit sprites appear on the map.

### 3. Steps to Reproduce

1. Load "atb1.sav.gz"
2. (optional) connect second client as "Republic of the Sphere OpFor"
3. Proceed to unit deployment and observe issue

### Attach Files

[DeploymentZoneAny.zip](https://github.com/user-attachments/files/19232742/DeploymentZoneAny.zip)

### Severity *

### MekHQ Suite Version *

0.50.04-SNAPSHOT     Build Date: 2025-03-13T03:29:17.678069598

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.6

### Final Verification

Sleet01:
Fixes bug where opfor units might deploy with "any" deployment zone instead of the zone specified for their owner's force in the scenario.

Testing:
- Ran all three projects' unit tests
- Started various scenarios and confirmed units were set to the correct deployment zones and actually deployed there.

Close #6267

### DIFF
--- a/MekHQ/src/mekhq/AtBGameThread.java
+++ b/MekHQ/src/mekhq/AtBGameThread.java
@@ -739,6 +739,7 @@ public class AtBGameThread extends GameThread {
             entity.setOwner(botClient.getLocalPlayer());
             String fName = String.format(forceName, lanceName, forceIdLance);
             entity.setForceString(fName);
+            entity.setStartingPos(botForce.getStartingPos());
             entities.add(entity);
             i++;
 


IllianiCBT (approved):
Sorry for the slow review, this one passed me by.

PR #6283, opened 2025-03-16
Added Blood Group Tracking For Personnel

IllianiCBT:
- Introduced `BloodGroup` enum for managing blood types with inheritance, randomization, and labeling logic.
- Updated `Person` to include blood group data with random generation and serialization support.
- Displayed blood type information in `PersonViewPanel`.
- Integrated blood group inheritance for newborns in `AbstractProcreation`.
- Added localized labels for blood types in resource files.
- Added unit tests for `BloodGroup` and inheritance logic to ensure correctness.
- Included `mockito-inline` in the build file for unit testing support.

### Dev Notes
I know players will ask, so it should be emphasized that there are no plans to actually do anything with blood types, beyond tracking and displaying them. And yes, because again - players will ask - we do properly handle blood type inheritance when characters produce offspring. Because of course we do! :D

I'll be honest, this mostly exists because I wanted to test adding static mocking capabilities to our Unit Tests, as that makes testing things like `d6()` and `randomInt()` _**SIGNIFICANTLY**_ easier. Taking testing methods which contain those methods from a right pain in the bum to effortless. Given how much of what I work with involves random chance, not being able to easily mock those two methods was really hindering my ability to write tests.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -1371,6 +1374,25 @@ public Gender getGender() {
...
+     * Retrieves the blood group of the person. If the blood group has not been set, it generates a
+     * random blood group using {@link BloodGroup#getRandomBloodGroup()}.
+     *
+     * @return The {@link BloodGroup} of the entity. If no blood group is previously assigned,
+     *         a random one is generated and returned.
+     */
+    public BloodGroup getBloodGroup() {
+        if (bloodGroup == null) {
+            bloodGroup = getRandomBloodGroup();
+        }
```
Scoppio, line 1391: this naive initialization will make random and "impossible" family trees.

IllianiCBT: This is only true for existing campaigns, all characters added once this feature is implemented will have correct blood genealogies. I implemented it this way so that we didn't have to travel through every single characters' genealogy looking for `null` blood types and then walking through that genealogy to update blood types every time a campaign is loaded.

If you have a better alternative, though, happy to implement it.

Scoppio: I have an idea, but would have to see when I get back, and it is not a GREAT idea and may lead to other bugs/problems. One option would be to make this funxtion recursive, it goes "up the chain" looking for its parents then once it finds the "final edges" of the graph it generates randomly and "consolidate" the children underneath it, a both BFS and DFS algorithms should work fine in this.

Scoppio: Its probably too much and I agree that your solution is enough/reasonable for pre-existing campaigns.

IllianiCBT: Yeah, I did a lot of thinking after reading through your suggestions and while something we _could_ do, very easily. I don't think the benefit warrants the increased load.

Sleet01 (commented):
LGTM

Sleet01 (approved):
And this time I'll remember to approve!

IllianiCBT:
Updated for 50.05

IllianiCBT:
Whoops

PR #6285, opened 2025-03-16
Added Mercenary Auction Events

IllianiCBT:
- Added `MercenaryAuction` system, including dialogs and logic, replacing old bonus unit functionality:
  - Implemented a bidding mechanism allowing campaigns to acquire units via auctions.
  - Deprecated `addBonusUnit` in favor of the new auction workflow.
- Updated logic handling auction-related UI and added new resource property files for auction text support.

### Requirements
Requires #6295

### Dev Notes
This PR updates the generic AtB 'ronin' event to be more immersive and narratively driven by allowing the player to interact with the Ronin before they are hired.


MekHQ/src/mekhq/campaign/randomEvents/MercenaryAuction.java
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * This class handles the logic for determining auction eligibility based on the player's finances
+ * and provides the interface for bidding in mercenary auctions. Successful auctions result in the
+ * unit being added to the campaign, while failures notify the player of the outcome.
+ */
+public class MercenaryAuction {
+    private static final String RESOURCE_BUNDLE = "mekhq.resources." + MercenaryAuctionDialog.class.getSimpleName();
+
+    private static final int AUCTION_MINIMUM_BID = 50;
```
Sleet01, line 57: Is there a good reason for these two values to be different types?  Changing the first to a double would obviate some casts later.

Also, it's customary to indicate percentage values via the variable or constant name.

Sleet01 (approved):
It would be nice to have the percent constants labelled and using the same types, just for consistency.  But it doesn't impact functionality. 
Otherwise looks good.

MekHQ/src/mekhq/campaign/randomEvents/MercenaryAuction.java
IllianiCBT: So, `AUCTION_MINIMUM_BID` is used to set up and manage the spinner, which uses integers as players are generally have a more intuitive understanding that 50 == 50% rather than 0.5 == 50%. While `AUCTION_MAXIMUM_BID` is only used internally, so can be `1.5` as we, as programmers, are used to seeing percents as doubles.

Whether that's a _good reason_ I'll leave to your judgement. It'd be easy to change.

Speaking of changes, I'll get those constants renamed :)

IllianiCBT:
- Renamed `AUCTION_MINIMUM_BID` and `AUCTION_MAXIMUM_BID` to `AUCTION_MINIMUM_BID_PERCENT` and `AUCTION_MAXIMUM_BID_PERCENT`, respectively, for better semantic accuracy.
- Added detailed inline documentation to clarify the purpose and usage of the constants.
- Updated all references in logic and UI code to use the newly named constants, ensuring consistency.

IllianiCBT:
Updated to 50.05

PR #6286, opened 2025-03-16
Increased Minimum Skill Requirement For Replacement Limb Procedures to Match RAW

IllianiCBT:
- Updated the minimum Doctor skill level required for 'Replacement Limb Types 3, 4, and 5' from to total 5+.
- Adjusted related logic in `PersonnelTableMouseAdapter` to use the `getFinalSkillValue` method for accurate skill evaluation.
- Refactored `ReplacementLimbDialog` to dynamically inject the new skill level requirement into the message description.
- Adjusted the message text in `ReplacementLimbDialog.properties` to make the skill level requirement configurable.


Sleet01 (approved):
LGTM

PR #6293, opened 2025-03-17
Added Level Attribute To Intelligence Enum

IllianiCBT:
- Assigned integer `level` attribute to each `Intelligence` enum value for more precise representation.
- Updated constructor to accept and assign the new `level` parameter.
- Added `getLevel` method to retrieve the `level` of an `Intelligence` enum.
- Refactored `getIntelligenceScore` to use the new `level` attribute instead of relying on `ordinal`.

### Dev Notes
This was done to reduce our reliance on ordinals, in favor of more robust arbitrary values.

Marking this as for new cycle, as it adds nothing of value to the user experience so can wait for 50.05.


Scoppio (approved):
Great improvement

IllianiCBT:
Updated to 50.05

PR #6295, opened 2025-03-18
Enhanced Dialog System With Configurable Options

IllianiCBT:
- Updated dialog constructors across various classes to include new parameters for spinners and vertical layout control.
- Improved immersion by introducing a `GenericImmersiveMessageDialog` for simple, in-character interactions.
- Centralized reusable components and logic in `MHQDialogImmersive` for streamlined dialog creation.
- Added "Understood" button localization in GUI resources.
- Refined button panel creation to support spinners and additional configurations.

## Important
This PR consists of the Immersive Dialog improvements originally pushed as part of #6285. No changes have been made to the implementation found in that PR. I split it out because I needed the functionality for another PR.

### GenericImmersiveMessageDialog
The most important part of this PR is the introduction of `GenericImmersiveMessageDialog`. This dialog allows us to very quickly implement basic Immersive Dialogs in instances where we don't need any special functionality and we're just presenting information to the player. An example of which I'll include below.

image

So in the above dialog all we're doing is informing the player that they can't afford the auction, so rather than going through the whole process of setting up a unique immersive dialog instance, we just feed in the necessary values into `GenericImmersiveMessageDialog` and everything else is already good to go. This will help us avoid a situation where we have 101 immersive dialog classes that all do the same thing.

image

### MHQDialogImmersive
`MHQDialogImmersive` has also had some new functionality added. I added the ability to pass in a JPanel, ostensibly containing a JSpinner, but really you can add in whatever you like and it'll appear above the response options, as show below.

image

Beyond that, I went ahead and moved some things around so it's easier to `Override` them. Stuff like this is added as it becomes relevant, so if you find yourself in a position where you can't override something you need, then consider refactoring Immersive Dialog. Remember: any functionality you add to the core class is extended to all classes that inherit from Immersive Dialog, no need to go through and update every other class.

Finally, now - when you initialize an instance of Immersive Dialog - you can specify whether you want it modal or not.


Sleet01 (approved):
LGTM

IllianiCBT:
Updated for 50.05

PR #6297, opened 2025-03-18
Removed Timer Delay for Filter Panel in Interstellar Map Tab

IllianiBird:
- Changed the `optionPanelTimer` delay value from 50 to 0 in `InterstellarMapPanel.java`.
- Likely improves responsiveness of the option panel functionality.

### Dev Notes
This affects the little blue filters tab in the bottom-right of the Interstellar Map Tab. When clicked the panel will expand and contract. Previously we were waiting 50ms per expansion/contraction step. This made the process feel very laggy and non-responsive. This PR changes that so there is no delay between steps, making the whole thing feel _much_ smoother.


Sleet01 (approved):
LGTM

PR #6299, opened 2025-03-18
Disabled Remove CVP and Remove SP Options if GM Mode is Disabled; Allowed GMs to Take SP and CVP into Negative

IllianiCBT:
- Changed `btnRemoveCVP` and `btnGMRemoveSP` to enable based on `gmMode` instead of point counts.
- Ensured consistency in button enabling logic for GM mode in the Campaign Management Dialog.

### Dev Notes
This just ensures these options are only usable if GM Mode is enabled, while also allowing GMs to take CVP and SP into negative, should they wish.


Sleet01 (approved):
LGTM

PR #6303, opened 2025-03-19
Fixed Deployment Turn Calculation For Mounted Reinforcements

IllianiCBT:
- Updated `setDeploymentTurnsForReinforcements` to include campaign and scenario context.
- Improved handling of transported units by creating and utilizing player transport relationships in scenarios.
- Refined logic for skipping transported units during deployment turn calculations.
- Introduced `addPlayerTransportRelationships` in `StratconScenario` to manage transport linkages directly within the campaign.
- Ensured proper initialization of player transport linkages in `Scenario` class.

### Dev Notes
This fixes an issue where mounted units (such as infantry) were being considered unmounted, because Transport ID wasn't set at that point. It appears this value is only set once the unit is loaded into MegaMek. I fixed this by hooking into Player Transport Linkages. This fix will not likely address scenarios spawned prior to the fix going in.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratcon/StratconScenario.java
+++ b/MekHQ/src/mekhq/campaign/stratcon/StratconScenario.java
@@ -32,12 +32,15 @@ import jakarta.xml.bind.annotation.XmlElementWrapper;
 import jakarta.xml.bind.annotation.XmlTransient;
 import jakarta.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
 import megamek.common.annotations.Nullable;
+import megamek.logging.MMLogger;
 import mekhq.MekHQ;
 import mekhq.adapter.DateAdapter;
 import mekhq.campaign.Campaign;
+import mekhq.campaign.enums.CampaignTransportType;
 import mekhq.campaign.event.DeploymentChangedEvent;
 import mekhq.campaign.force.Force;
 import mekhq.campaign.mission.*;
+import mekhq.campaign.unit.ITransportAssignment;
 import mekhq.campaign.unit.Unit;
 
 import java.time.LocalDate;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/stratcon/StratconScenario.java
```diff
@@ -137,6 +156,34 @@ public void addUnit(Unit unit, String templateID, boolean useLeadership) {
...
+     */
+    private void addPlayerTransportRelationships(Unit unit) {
+        for (CampaignTransportType transportType : CampaignTransportType.values()) {
+            ITransportAssignment transportAssignment = unit.getTransportAssignment(transportType);
+            if (transportAssignment != null) {
+                Unit transport = transportAssignment.getTransport();
+
+                if (transport == null) {
+                    logger.warn(
+                          "Unit {} has a transport assigned, but the transported unit doesn't exist",
```
IllianiCBT, line 175: @psikomonkie when you take a look at this later, can you advise what we should be doing in this situation. Whether just skipping the unit is the correct action, or whether we should do something to correct the presumably malformed transportation assignment?

psikomonkie (approved):
Nice! There's some weird edge cases I'm curious about especially involving mixed forces, like if a player doesn't deploy a transport, or if they're in different forces, but I think this is a good fix. 

I think we might want to rework the scenario reinforcement turn to instead recalculate the reinforcement turn for all player units when reinforcements are added - this may better handle cases where units and their transports are in different forces, or if a transport isn't used. But that's a larger change with more potential issues so I don't want to request it now.

psikomonkie (approved):
Nice! Thanks for fixing that issue quickly. I think this looks good now.

PR #6309, opened 2025-03-20
Updated Deprecation Tags to include Version

rjhancock:
Also includes an updated EditorConfig file created with IDEA, additional deprecations in related files for methods that have been indicated as not used, updated JavaDocs.

Additional charges are related to the EditorConfig updates as well as fixing issues brought up from IDEA including simplifying if/else, converting switch statements, etc.


.editorconfig
```diff
@@ -1,33 +1,449 @@
...
-[*.blk]
-end_of_line = lf
-trim_trailing_whitespace = true
-insert_final_newline = true
-
-[*.mtf]
-end_of_line = lf
-trim_trailing_whitespace = true
-insert_final_newline = true
+ij_continuation_indent_size                                     = 6
```
Sleet01, line 18: Will this override the * setting, or vice-versa?

MekHQ/src/mekhq/CampaignPreset.java
```diff
@@ -85,19 +96,19 @@ public class CampaignPreset {
...
-    private Planet planet;
-    private RankSystem rankSystem;
-    private int contractCount;
-    private boolean gm;
+    private LocalDate                date;
+    private Faction                  faction;
+    private Planet                   planet;
+    private RankSystem               rankSystem;
+    private int                      contractCount;
+    private boolean                  gm;
```
Sleet01, line 104: Pretty sure this is not correct.

MekHQ/src/mekhq/CampaignPreset.java
```diff
@@ -425,9 +457,9 @@ public static List<CampaignPreset> loadCampaignPresetsFromDirectory(final @Nulla
         }
 
         return Arrays.stream(Objects.requireNonNull(directory.listFiles()))
-                .map(CampaignPreset::parseFromFile)
-                .filter(Objects::nonNull)
-                .collect(Collectors.toList());
+                     .map(CampaignPreset::parseFromFile)
+                     .filter(Objects::nonNull)
+                     .collect(Collectors.toList());
```
Sleet01, line 462: Looks like new tabs?

MekHQ/src/mekhq/CampaignPreset.java
```diff
@@ -443,22 +475,22 @@ public static List<CampaignPreset> loadCampaignPresetsFromDirectory(final @Nulla
         final Element element = xmlDoc.getDocumentElement();
         element.normalize();
 
-        final String versionString = element.getAttribute("version");
-        final Version version = new Version(versionString);
+        final String  versionString = element.getAttribute("version");
+        final Version version       = new Version(versionString);
```
Sleet01, line 479: What is causing this?

Sleet01:
Just looking at the first few files, there are a number of issues:
1. do we need the subsequent sections of the EditorConfig if [*] covers all our needs?
2. tabbed variable declarations.
3. random new tabs on unrelated lines.

I think the EditorConfig and reformatting updates (if intended) should be separate from the deprecation updates.

.editorconfig
rjhancock: The global * is for all files then each section overrides it for THOSE specific files.

rjhancock:
> Just looking at the first few files, there are a number of issues:
> 
> 1. do we need the subsequent sections of the EditorConfig if [*] covers all our needs?

It wont cover all of our needs as XML, YAML, and other file times need different formatting (including ignoring line wrappings in XML).

> 2. tabbed variable declarations.
> 3. random new tabs on unrelated lines.

Both resultant of the EditorConfig changes. My OCD kicked in a bit.

> I think the EditorConfig and reformatting updates (if intended) should be separate from the deprecation updates.

PR #6317, opened 2025-03-21
Fixed Bot Force Template Mapping Issue in AtB Scenarios

IllianiCBT:
- Added `reviewBotForceTemplateCompleteness` to ensure all bot forces are correctly mapped to their templates.
- This method checks for missing template mappings and restores them using force and template names.
- Fix addresses rare cases where templates became inconsistent, ensuring data integrity and proper tracking of post-generation bot forces.

Fix #6313

PR #6319, opened 2025-03-21
Added Early Exit for Objective Determination Self-Correction Loop

IllianiCBT:
- Added a `break` statement to stop unnecessary iterations after finding the correct force template.
- Improved performance and prevented potential overwrites by halting further template checks once a match is found.

### Dev Notes
The correction added in #6317 works perfectly fine, but paranoia called me to double-check everything before launch and I spotted that I missed an opportunity for an early exit. It's not needed, the size of the array we're looping through is tiny, but if it's worth doing, it's doing properly. Or however the saying goes.


Sleet01 (approved):
LGTM

PR #6389, opened 2025-03-24
Updated Insufficient Medics Nag To Use Immersive Dialog

IllianiCBT:
- Refactored `InsufficientMedicsNagDialog` to utilize `ImmersiveDialogSimple` for improved modularity and customization.
- Replaced inheritance from `AbstractMHQNagDialog` with a standalone class to simplify and streamline the implementation.
- Updated NagController logic to replace `wasAdvanceDayCanceled()` with `shouldCancelAdvanceDay()` to reflect the updated behavior.

image

### DIFF
--- a/MekHQ/resources/mekhq/resources/GUI.properties
+++ b/MekHQ/resources/mekhq/resources/GUI.properties
@@ -309,12 +309,6 @@ NoCommanderNagDialog.text=Please be advised that no commanding officer is curren
   <br>\
   <br><i>You can assign one of your personnel as the campaign commander by right-clicking on that\
   \ character, navigating to Flags and selecting the Commander flag.</i>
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
Looks good to me.
I this feels like an opportunity to extract a middle-level class, say something like an `ImmersiveNagDialog`, with a similar layout but more parameterization.
Nothing that needs doing now, just something to think about if more immersive nag dialogs are needed.

IllianiCBT:
Yeah, I concur. It's something that will almost certainly be added once I spot a good use-case. I need that so I can tell what functionality is required.

As things stand `ImmersiveDialogSimple` exists for cases where we just want to present information to the user, and maybe track their responses - like nags. Then there is `ImmersiveDialogCore` for when we want access to all the bells and whistles.

PR #6401, opened 2025-03-24
Created `ImmersiveDialogNag` Base Class; Rolled Out Base Class to Previously Edited Nags

IllianiCBT:
- Consolidated the common nag dialog logic into a newly created `ImmersiveDialogNag` base class to reduce code duplication.
- Updated `AdminStrainNagDialog`, `EndContractNagDialog`, and `InsufficientAstechsNagDialog` to extend `ImmersiveDialogNag`.

### Dev Notes
Nothing super fancy here, just some consolidation of code so we don't have so much duplication.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/nagDialogs/NagController.java
+++ b/MekHQ/src/mekhq/gui/dialog/nagDialogs/NagController.java
@@ -174,11 +174,7 @@ public class NagController {
               isOvertimeAllowed,
               possibleAstechPoolOvertime)) {
             InsufficientAstechTimeNagDialog insufficientAstechTimeNagDialog = new InsufficientAstechTimeNagDialog(
-                  campaign,
-                  units,
-                  possibleAstechPoolMinutes,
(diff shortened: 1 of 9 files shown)


MekHQ/src/mekhq/gui/dialog/nagDialogs/NagController.java
```diff
@@ -174,11 +174,7 @@ public static boolean triggerDailyNags(Campaign campaign) {
               isOvertimeAllowed,
               possibleAstechPoolOvertime)) {
             InsufficientAstechTimeNagDialog insufficientAstechTimeNagDialog = new InsufficientAstechTimeNagDialog(
-                  campaign,
-                  units,
-                  possibleAstechPoolMinutes,
-                  isOvertimeAllowed,
-                  possibleAstechPoolOvertime);
+                  campaign);
```
IllianiCBT, line 177: We have to construct this information locally, within the class, rather than inserting it due to the current need to have `super()` be the first call within the constructor.

PR #6430, opened 2025-03-26
Added Skill Deprecation Tool For Managing Deprecated Skills; Deprecated the 'Scrounge' Skill

IllianiBird:
- Introduced `SkillDeprecationTool` to handle deprecated skills for personnel in campaigns.
- Added functionality for identifying deprecated skills, providing XP refunds, and managing skill removal.
- Implemented Immersive Dialogs to alert user to skill deprecation.
- Deprecated the "Scrounge" skill.

### Dev Notes
Around 50.02 we removed the Clan Tech Knowledge SPA. This was in response to official errata. At the time I noted that the way we handled the removal wasn't great. As SPAs are expensive, and that was potentially a lot of XP we just poof'd away. Not a great user experience.

image

Following a poll on Discord (see above) we decided to officially deprecate the 'Scrounge' skill. Not wanting to repeat the mistakes of the past I have created the `SkillDeprecationTool` to handle skill deprecations. Now, whenever we want to Deprecate a skill we just add it to the `DEPRECATED_SKILLS` list. Users loading campaigns containing characters with a deprecated skill are then prompted to refund any xp spent on the deprecated skill.

image

Originally, the intent was to refund Scrounge when the skill was removed, however that created an order of operations issue. How can we refund a skill that has been completely removed? In short, we can't. At least not easily.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
+++ b/MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License (GPL),
+ * version 3 or (at your option) any later version,
+ * as published by the Free Software Foundation.
+ *
+ * MekHQ is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
IllianiCBT, line 1: When reviewing this is the only class you really need to pay attention to. The reset are just bog standard deprecations or method calls.

rjhancock (approved):

PR #6450, opened 2025-03-27
Added Dynamic Contract Difficulty Option; Fixed Clan OpFor Skill Level Generation

IllianiBird:
- Introduced campaign dynamic difficulty settings by considering skill levels in contract rating methods.
- Fixed OpFor veterancy bug

### Dev Notes
So, first the bug. When we were adjusting the veterancy of the OpFor we were actually adjusting the veterancy of the players' ally. So that wonderful clause that switched Clan OpFors to always be Veteran or Regular? Yeah, that was actually making the employer forces Veteran or Regular. Apparently the Inner Sphere decided to field their best! Well, not any more, that's been fixed. You're welcome.

Onto the more interesting stuff. I added a campaign option that will adjust the veterancy and quality of the allied and enemy forces based on the player's experience rating.

- A high experience rating means Allied forces are going to be of a lower quality (and veterancy), while the enemy is likely to be higher quality and veterancy.
- A low experience rating means allied forces are going to be of a higher quality (and veterancy), while the enemy is likely to be a lower quality and veterancy.

In either case it is still possible to get a nightmare of a contract with a low experience rating. And, conversely, to be able to enjoy clubbing baby seals with a high experience rating. All that has changed is that contract generation will trend towards one direction or the other.


rjhancock (changes requested):

IllianiCBT:
- Added `@Deprecated` annotations to `setAllyRating(AtBContract, int)` and `setEnemyRating(AtBContract, int)` methods.

PR #6464, opened 2025-03-27
Enhanced Deprecation Skill Handling

IllianiBird:
- Added options to "Skip All" and "Refund All" deprecated skill checks for streamlined personnel management.
- Introduced boolean flags (`skipAllDeprecationChecks` and `refundAllDeprecatedSkills`) to control deprecation behaviors.
- Updated `SkillDeprecationTool` to handle new dialog choices and flag propagation.
- Prevented deprecated skills from being assigned during skill generation in `DefaultSkillGenerator`.
- Adjusted UI labels and localization to support the new deprecation handling options.

### Dev Notes
This PR adds two new buttons to the Deprecated Skills dialog: Skip All, and Refund All. This functionality is introduced to better support campaigns with a large volume of personnel.

I also added a check to character creation that prevents new characters from generating with a Deprecated skill.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
+++ b/MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
@@ -50,6 +50,8 @@ import mekhq.gui.baseComponents.immersiveDialogs.ImmersiveDialogSimple;
 public class SkillDeprecationTool {
     private final String RESOURCE_BUNDLE = "mekhq.resources." + getClass().getSimpleName();
 
+    private final int SKIP_ALL_DIALOG_OPTION_INDEX = 1;
+    private final int REFUND_ALL_DIALOG_OPTION_INDEX = 3;
     /**
      * A list of deprecated skills.
      *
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
```diff
@@ -75,13 +79,22 @@ public class SkillDeprecationTool {
      * @param campaign the {@link Campaign} instance that provides context for the operation
      * @param person   the {@link Person} whose skills will be checked for deprecation
      */
-    public SkillDeprecationTool(Campaign campaign, Person person) {
+    public SkillDeprecationTool(Campaign campaign, Person person, boolean refundAll) {
```
rjhancock, line 82: Signature change. Need alternative one with the original and can pass it to this one with a default value for `refundAll`.

rjhancock (changes requested):

MekHQ/src/mekhq/campaign/personnel/SkillDeprecationTool.java
IllianiCBT: Oo, good catch. I'm still trying to get into the habit of doing that

rjhancock (approved):

PR #6505, opened 2025-03-31
Refactored Advanced Medical Edit Injury Dialog To Improve Error Handling

IllianiCBT:
- Wrapped logic in `btnOKActionPerformed` with a try-catch block to handle potential errors, adding logging for failure cases.
- Replaced `this.setVisible(false)` with `this.dispose()` to properly close the dialog on action.

Fix [Sentry Log](https://sentry.tapenvy.us/organizations/tapenvyus/issues/4137/?alert_rule_id=13&alert_type=issue&notification_uuid=60647ca8-d3a8-4c54-ac53-9814dd2d0a89&project=10&referrer=discord)

### Dev Notes
We had a couple of unchecked casts from String to Integer in the AM edit injury dialog. I wrapped the block in a try-catch, with the idea being that if we hit an issue, we log it and then block progression until the user fixes their mistake or closes the dialog.

I also changed setVisible to dispose, because we're unlikely to need to keep this dialog in memory once it's been finished with.


MekHQ/src/mekhq/gui/dialog/EditInjuryEntryDialog.java
```diff
@@ -328,16 +328,20 @@ private void setUserPreferences() {
...
+            injury.setLocation(((BodyLocationChoice) Objects.requireNonNull(ddLocation.getSelectedItem())).loc);
+            injury.setType(((InjuryTypeChoice) Objects.requireNonNull(ddType.getSelectedItem())).type);
+            injury.setPermanent(ddPermanent.getSelectedIndex() == 0);
+            injury.setWorkedOn(ddWorkedOn.getSelectedIndex() == 0);
+            injury.setExtended(ddExtended.getSelectedIndex() == 0);
+            injury.setUUID(UUID.randomUUID());
+            this.dispose();
+        } catch (Exception ex) {
+            logger.info("Failed to update injury entry", ex);
+        }
```
rjhancock, line 344: Update to use the MathUtility methods for parseInt with default values where needed. That should take care of the exception portion.

rjhancock (approved):

PR #6535, opened 2025-04-02
Implemented ATOW-lite Aging Effects

IllianiCBT:
- Centralized access to `Campaign` in `NewRecruitDialog` by introducing `getCampaign` method, reducing repetitive calls to `hqView.getCampaign`.
- Added `AgingMilestone` enum to handle aging-related attribute modifications in a structured and scalable manner.
- Updated `CustomizePersonDialog` UI and logic to display and apply age-based skill modifiers.
- Adjusted newborn initialization in `AbstractProcreation` to reset and re-roll skills, reflecting age-based adjustments.
- Improved formatting and code consistency across modified classes, reducing redundant imports and ensuring proper indentation.

### Requirements
Requires #6534

### Dev Notes
In short, this adds aging modifiers to skill Target Numbers based on the aging effects outlined in ATOW.

In long, we make some changes to the ATOW rules. Specifically due to how mhq handles skills. In ATOW aging modifiers apply a direct modifier to the xp spent on affected skills, based on their Linked Attributes (see ATOW for linked attributes). So if a character has spent 300 xp on Small Arms, and aging says -100 xp, they now have whatever skill level they'd have if they only had 200 xp in that skill.

In mekhq that would have been an absolute _nightmare_ to implement. Not to mention very difficult to communicate to the player.

So what we do is use the table of modifiers and then divide the effects by 100. That value is then applied as a direct modifier to the skill's skill level - like how bonuses work. So a modifier that would have reduced xp spent by 100 instead decreases the character's skill level by 1.

However, we go a step further. During testing I found that sticking to the above quickly piled on penalties until such a point that the character was effectively worthless. While it's realistic that a 150 year old is a pretty rubbish MekWarrior, it's not very fun. Players enjoy having their militant OAPs flying about kicking ass.

Instead, rather than combining the modifiers from both tied attributes (as per ATOW) we instead average them. So if a character has a skill that uses both Body and Strength and aging says they should get -100 xp in each, in mhq they will only have a -1 to that skill and not -2.

I'm not going to list the effects of aging here, as it would be duplicating official rules a tad too much for my taste. So if anyone is interested check ATOW pg. 300.


IllianiCBT:
# Added Skills
- Acrobatics
- Acting
- Animal handling
- Appraisal
- Archery
- Art/Dancing
- Art/Drawing
- Art/Painting
- Art/Writing
- Climbing
- Communications
- Computers
- Cryptography
- Demolitions
- Disguise
- Escape Artist
- Forgery
- Interest/History
- Interest/Literature
- Interest/Holo-Games
- Interest/Sports
- Interrogation
- Investigation
- Languages
- Martial Arts
- Perception
- Sleight of hand
- Protocols
- Science/Biology
- Science/Chemistry
- Science/Mathematics
- Science/Physics
- Security Systems/Electronic
- Security Systems/Mechanical
- Sensor Operations
- Stealth
- Streetwise
- Survival
- Tracking
- Training

PR #6542, opened 2025-04-03
Fixes Compatibility for AC Ammo in MHQ

psikomonkie:
In MekHQ, `AmmoBin::isSamePartType` does not properly recognize AC ammo as being the same part type. Two AC/5 ammo bins, for example. This led to any refit in MHQ for a unit that had AC ammo would show as needing changed in the refit even if nothing was changed. 

In the below screenshots, nothing was changed for this unit. I right clicked the unit in the Hangar and selected customize > customize in mek lab. I made no changes to the unit before taking the screenshots.
Before Fix:
image

After Fix:
image

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/equipment/AmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/AmmoBin.java
@@ -490,8 +490,9 @@ public class AmmoBin extends EquipmentPart implements IAcquisitionWork {
         // AmmoType and number of rounds of ammo (i.e. they are the same
         // irrespective of "munition type" or "bomb type").
         return (getClass() == part.getClass()) &&
-               getType().isCompatibleWith(((AmmoBin) part).getType()) &&
-               (((AmmoBin) part).getFullShots() == getFullShots());
+            ((getType().isCompatibleWith(((AmmoBin) part).getType())) ||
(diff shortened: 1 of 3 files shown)


IllianiCBT (approved):
Great find!

MekHQ/src/mekhq/campaign/parts/equipment/AmmoBin.java
```diff
@@ -490,8 +490,9 @@ public boolean isSamePartType(Part part) {
         // AmmoType and number of rounds of ammo (i.e. they are the same
         // irrespective of "munition type" or "bomb type").
         return (getClass() == part.getClass()) &&
-               getType().isCompatibleWith(((AmmoBin) part).getType()) &&
-               (((AmmoBin) part).getFullShots() == getFullShots());
+            ((getType().isCompatibleWith(((AmmoBin) part).getType())) ||
+                   (getType().equals(((AmmoBin) part).getType()))) &&
```
Sleet01, line 494: OMFG :face_holding_back_tears:

Sleet01: Sorry, just wanted to echo Illiani's comment.  That must have been super-annoying to track down.

PR #6551, opened 2025-04-04
Fixed Inability to GM Activate Self-Crewed Units (such as DropShips)

IllianiCBT:
- Added a check to ensure only non-GM users are prevented from activating non-CI self-crewed units.


rjhancock (approved):

rjhancock:
Wow. What broke with the CI? Looks like a type mismatch.

IllianiCBT:
Actually, I'm not entirely sure. It's throwing a wobbly because it can't find a method - a method which absolutely exists. Lemme rerun the tests.

PR #6552, opened 2025-04-05
Added ATOW Attribute Tracking and Utility Methods

IllianiCBT:
- Introduced localized label, short name, and description retrieval methods in `SkillAttribute`.
- Implemented `Attributes` in `Person` to track character attribute scores with getter, setter, and modifier methods.
- Updated XML read/write functionality in `Person` to handle `Attributes` persistence.
- Added comprehensive unit tests for `SkillAttribute` and `Attributes` classes to ensure new methods' correctness.

### Dev Notes
This PR doesn't really do much on its' own and is more setup for future stuff. Essentially all it does is process and store a character's Attribute scores. There is no way to access, or influence these. Characters just get assigned the default of `5` in every attribute, which is stored and loaded alongside the rest of their stats.

Of special note, mhq handles Edge very differently and I really have no interest in implementing an ATOW-accurate Edge system, frankly I'm not sure it's possible without a _lot_ of work. So the Edge attribute isn't tracked by mhq and probably won't be.

TLDR: Edge remains unchanged from its current implementation.

### DIFF
--- a/MekHQ/resources/mekhq/resources/SkillAttribute.properties
+++ b/MekHQ/resources/mekhq/resources/SkillAttribute.properties
@@ -1,0 +1,25 @@
+# suppress inspection "UnusedProperty" for whole file
+NONE.label=None
+NONE.shortName=None
+NONE.description=THIS SHOULDN''T BE VISIBLE
+STRENGTH.label=Strength
+STRENGTH.shortName=STR
+STRENGTH.description=Represents the character''s physical strength or power.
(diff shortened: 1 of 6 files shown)


IllianiCBT:
### Attributes Tracked
(For those who don't want to pick up ATOW)
- Strength
- Body
- Reflexes
- Dexterity
- Intelligence
- Willpower
- Charisma

rjhancock (approved):

PR #6554, opened 2025-04-05
Added the Ability to Purchase Attribute Improvements; Added the GM Ability to Mass Set Attributes

IllianiCBT:
### Dev Notes
This adds the ability to increase Attributes as per ATOW. It also adds the ability to set Attributes to specific values, while in GM mode.


psikomonkie (approved):
Nice!

PR #6527, opened 2025-04-01
Issue 6178: [RFE] Add warehouse stock level when salvaging

psikomonkie:
Fixes #6178

Issue #6178, opened 2025-03-06
[RFE] Add warehouse stock level when salvaging


BTSS88:
### Enhancement Type *

New Feature

### Brief Description of Enhancement *

Would it be possible when stripping down salvages, we get to see how many of the item is already in the warehouse? We already have that in repair and it's annoying to flip between the warehouse tab and the repair tab to see if it's worth salvaging X item.

### Use Cases or Rationale

Save player time by reducing the need to check inventory before salvaging. 
Keep warehouse bloat down.

### Attach Files

![Image](https://github.com/user-attachments/assets/41b738f7-8514-48aa-b6fa-a4a8362630f0)
![Image](https://github.com/user-attachments/assets/06aeafac-2f7e-4f33-8cb3-6f042391cec1)

### MekHQ Suite Version *

50.03

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

IllianiCBT:
I took a look into this and it's a lot more complicated than it initially appears, so I'm going to throw in the towel for now.

It's a great idea, though.

PR #6557, opened 2025-04-05
Issue 6178: [RFE] Improved Display for in stock parts while salvaging

psikomonkie:
Improved the display for in stock parts when salvaging: 
image

Relates to #6178 and #6527

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Part.java
+++ b/MekHQ/src/mekhq/campaign/parts/Part.java
@@ -450,19 +450,27 @@ public abstract class Part implements IPartWork, ITechnology {
         String details = getDetails();
 
         if (details != null && !details.isBlank()) {
-            toReturn.append("</b><br/>").append(getDetails()).append("<br/>");
+            toReturn.append("</b><br>").append(getDetails()).append("<br>");
         }
 
-        if (this instanceof mekhq.campaign.parts.equipment.AmmoBin ammoBin) {
-            if (campaign.getQuartermaster() != null) {
-                toReturn.append("<br> <b>In Warehouse:</b> ")
-                      .append(campaign.getQuartermaster().getAmmoAvailable(ammoBin.getType()))
-                      .append(' ');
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/parts/Part.java
```diff
@@ -450,19 +450,27 @@ public String getDesc() {
...
 
-        if (this instanceof mekhq.campaign.parts.equipment.AmmoBin ammoBin) {
-            if (campaign.getQuartermaster() != null) {
-                toReturn.append("<br> <b>In Warehouse:</b> ")
-                      .append(campaign.getQuartermaster().getAmmoAvailable(ammoBin.getType()))
-                      .append(' ');
+        if (this.isSalvaging()) {
+            int inStock = 0;
+            if (this instanceof mekhq.campaign.parts.equipment.AmmoBin ammoBin) {
+                if (campaign.getQuartermaster() != null) {
```
IllianiCBT, line 459: let's cache quartermaster, we're using it a few times

MekHQ/src/mekhq/campaign/parts/Part.java
```diff
@@ -450,19 +450,27 @@ public String getDesc() {
...
+                    inStock = campaign.getQuartermaster().getAmmoAvailable(ammoBin.getType());
+                }
+            } else if (campaign.getWarehouse() != null) {
```
IllianiCBT, line 462: And warehouse

IllianiCBT (approved):
Two nitpicks. Nothing to stop this being merged, just things to consider :)

MekHQ/src/mekhq/campaign/parts/Part.java
psikomonkie: It feels wrong to only do this to Part. Should I refactor all of the classes that inherit Part to also use the cached warehouse and quartermaster? I feel like that should be a separate PR at that point.

PR #6558, opened 2025-04-05
Show engine tonnage in parts-in-use table

psikomonkie:
Engines with the same rating but different unit tonnages cost different and are not interchangeable. This updates the parts in use table to reflect that. Prior to this change it only showed engine tonnage for Clan engines. It now shows & differentiates for IS engines too. 

An engine's actual tonnage doesn't matter when differentiating two engines, it's the unit's tonnage that matters.

image

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/MissingEnginePart.java
+++ b/MekHQ/src/mekhq/campaign/parts/MissingEnginePart.java
@@ -30,9 +30,6 @@ package mekhq.campaign.parts;
 
 import java.io.PrintWriter;
 
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-
 import megamek.common.Aero;
 import megamek.common.CriticalSlot;
 import megamek.common.Engine;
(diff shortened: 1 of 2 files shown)


IllianiCBT:
Could we use `part.isUnitTonnageMatters()` to display the weight for all parts where tonnage is a thing we care about? Or was it only engines that had this issue?

psikomonkie:
> Could we use `part.isUnitTonnageMatters()` to display the weight for all parts where tonnage is a thing we care about? Or was it only engines that had this issue?

Other parts where unit tonnage matters should already have the tonnage displayed when necessary. This was a strange interaction between the change made to support IS vs Clan DHS (the parenthesis bit) and Engines showing the wrong tonnage. If there are any other parts where there is this issue it'll probably need the parts "getAcquisitionName" or details updated. 

For example, check out Aero Sensors or actuators - they show the weight as required already. 
image

IllianiCBT (approved):

PR #6561, opened 2025-04-05
Added Dedicated Skill Check Class Utility

IllianiCBT:
#### The Problem
Under our current implementation to perform a skill check a developer needs to perform the following actions:

- Check whether the character even has the skill. Failing to do this will result in an NPE.
- Determine what the final target number is, manually adding modifiers as necessary.
- Make the actual roll.
- Manually check whether the check was passed or failed. If necessary, add handler code so we can factor in Margins of Success.
- Determine whether Edge is used and if so, manually factor that in.

Speaking from extensive experience, this is a right pain in the bum and fraught with easy to make mistakes. Not checking whether the character _has_ the skill being the biggest.

All of this becomes exacerbated as characters become increasingly more complex. With interactions between ATOW Attributes, and SPAs, and other such similar exceptions and special rules.

#### Introducing `SkillCheckUtility`
The Skill Check Utility class allows you to pass in the character, the skill name, and any special modifiers and it will handle all of the above for you. Either returning a binary pass/fail, or a more nuanced 'margin of success', as desired. It's also possible to pass in a boolean dictating whether Edge should be rolled in the event of a failure.

Finally, this class will generate a skill check string showing who made the check, what the target number, and roll were, and finally what the margin of success was.

This is fully supported with a margin of success enum, ensuring things are kept consistent.

#### Testing & Integration
This class is not currently integrated into any other classes, so currently isn't accessed. My next challenge will be going through current skill checks and moving them to use `SkillCheckUtility`, if appropriate.

As this is a new feature and liable to be fairly critical, I went ahead and added 31 new Unit Tests to ensure the class works as intended.

#### ATOW & Attributes
`SkillCheckUtility` is fully set up so that skill checks made through it will incorporate any modifiers from that skill's linked attributes. It also implements ATOW's 'untrained' skill check functionality. Both of these follow the official rules outlined in ATOW. I went with the ATOW implementation here as it has a fairly robust margin of success system and clear interaction with attributes in the event a skill is untrained.

#### A Living Class
I'm not expecting this class to be perfect out the gate. There will be use-cases I didn't account for and implementations that need expanding or adjusting. Much like my Immersive Dialog pet classes, `SkillCheckUtility` will be expanded and improved as it's used.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
+++ b/MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
@@ -320,20 +320,22 @@ public class SkillType {
      * Default constructor for the {@code SkillType} class.
      *
      * <p>Initializes a default skill type with placeholder values, primarily for testing or fallback purposes.</p>
+     *
+     * <p><b>Usage:</b> Generally you don't want to be calling this, outside of loading from xml or in Unit Tests.
+     * Instead, you want to use the full constructor.</p>
      */
     public SkillType() {
-        new SkillType("MISSING_NAME",
-              7,
-              false,
-              COMBAT_GUNNERY,
-              REFLEXES,
-              DEXTERITY,
-              1,
-              3,
-              4,
-              5,
-              new Integer[] { 16, 8, 8, 8, 8, 8, 8, 8, DISABLED_SKILL_LEVEL, DISABLED_SKILL_LEVEL,
-                              DISABLED_SKILL_LEVEL });
+        this.name = "MISSING_NAME";
+        this.target = 7;
+        this.countUp = false;
+        this.subType = COMBAT_GUNNERY;
+        this.firstAttribute = REFLEXES;
+        this.secondAttribute = DEXTERITY;
+        this.greenLvl = 1;
+        this.regLvl = 3;
+        this.vetLvl = 4;
+        this.eliteLvl = 5;
+        this.costs = new Integer[11];
     }
 
     /**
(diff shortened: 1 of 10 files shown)


MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
```diff
@@ -316,6 +316,12 @@ public static String[] getSkillList() {
 
     /** Creates new SkillType */
     public SkillType() {
+        name = "MISSING_NAME";
+        target = 7;
+        countUp = false;
+        subType = COMBAT_GUNNERY;
+        firstAttribute = REFLEXES;
+        secondAttribute = DEXTERITY;
```
psikomonkie, line 324: This feels weird. I know that the static methods later manually set these values, but I feel like it would be more straightforward to have this just be a constructor like `public SkillType(String name, int target, boolean countUp,....)`. Idk, it's late, I'll take another look at this PR in the morning, or maybe someone else has a thought.

rjhancock: Add a constructor for them and have this call it with defaults. Keep in mind not all skills have sub types or secondary attributes.

MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
```diff
@@ -408,6 +437,42 @@ public void setEliteLevel(int l) {
...
+     *                        existing value is preserved.
+     *
+     * @author Illiani
+     * @since 0.50.05
+     */
+    public void setSecondAttribute(SkillAttribute secondAttribute) {
+        if (secondAttribute != null) {
+            this.secondAttribute = secondAttribute;
+        }
+    }
```
rjhancock, line 474: Not all skills have secondary attributes. Allow null values.

rjhancock (changes requested):
Add the constructor like you talked about.

MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
IllianiCBT: The idea here isn't to define the values, but to define the default values. Similar to what happens if your initialize a new `Campaign` object, or a `Person` object. The expectation is that these values would be assigned elsewhere, but what I wanted to do here is ensure that we had some reliable default values so there wasn't any risk of having a `null` value.

MekHQ/src/mekhq/campaign/personnel/skills/SkillType.java
IllianiCBT: Yes they do?

IllianiCBT: Or am I missing something, because I'm fairly certain I went through and updated every skill to have linked attributes set.

IllianiCBT:
Updated incorporating null acceptance in `setFirstAttribute` and `setSecondAttribute` and a general class constructor. I also went ahead and updated all skill initializers to use this new constructor.

PR #6564, opened 2025-04-06
Refactored Phenotype Enum; Added ATOW Attributes to Character Generation; Added 'Extra Attribute Randomization' Campaign Option

IllianiBird:
### Requirements
Requires #6607

#### Phenotypes
Certain phenotypes increase or decrease attributes. Some phenotypes also provide specific Traits (SPAs, in our nomenclature). Both of these cases are now fully implemented. I also went ahead and cleaned up the enum a bit, switching it over to i18n and updating the Unit Tests.

#### Professions
ATOW includes a few archetypes which offer suggested starting points for each attribute score. I've set up the code so that characters will generate (on average) with those attribute scores. However, a few of the examples provided in the manuals are straight up _awful_.

~~The Tanker archetype is the worst example. If I were to implement that Archetype most vehicle crew would spawn with a -1 penalty to both gunnery and piloting. Needless to say, that shan't be happening.~~

~~So while the code is set up to handle role-based attribute starting points, currently they're all set up to infer no modifier. I'm hoping to encourage a knowledgeable community member into providing some suggested starting attributes for our different professions. Once that's been sourced I can go in an update this (possibly in a separate PR).~~

_Edit: This turned out to be an error in the manuals. I've since gone through and updated the Professions to have appropriate starting Attribute scores._

#### Campaign Options & Character Generation
When a character is generated, they are done so with 5 in every Attribute. A d6 is then rolled for each Attribute. On a roll of 1 the Attribute is decreased by 1; on a roll of 6 the Attribute is increased by 1.

In Campaign Options players can enable 'Extra Random Attributes'. This works identical to the above, however the d6 will 'explode'. What this means is if you roll a 1, you decrease the Attribute by 1 and then roll again. And keep going until you stop rolling 1s (or hit the minimum Attribute score of 1 - this isn't Traveller, nobody is dying in character creation). Conversely, if you roll a 6 you keep increasing the Attribute score by 1 until you stop rolling 6s (or hit the maximum Attribute score of 10).


IllianiCBT:
Updated to incorporate the ATOW and ATOW:Companion attributes.

I'm going to mark this as draft, as there is one last thing I want to do.

IllianiCBT:
Added a personnel right-click menu GM option to set ATOW Attributes for existing characters.

IllianiCBT:
Flipping this back to live from draft

IllianiBird:
Added Attribute Maximums Based on Phenotype and Exceptional Attribute SPAs

- Introduced a new `AttributeScoreSorter` for comparing "x / y" formatted strings, sorting by primary and secondary values.
- Updated attribute display and sorting in `PersonnelTableModelColumn` to represent both current scores and caps ("x / y").
- Added new "Exceptional Attribute" SPAs to increase attribute caps.
- Modified phenotype definitions in `Phenotype` to include attribute caps specific to each phenotype.
- Adjusted attribute handling logic to respect both phenotype caps and SPA adjustments.
- Enhanced unit tests to verify attribute caps and ensure adherence to defined ranges.

Issue #6560, opened 2025-04-05
[Issue][0.50.04] 100% reduction in prisoner capacity after prisoner event


goron111:
### Brief Description *

I have a security force of 1 SRM Platoon
on 3022-08-30 i  I captured  may (wounded) prisoners in a "Irregular Force Suppression" scenario.

At that point i had
26 healthy, 25 in the Infirmary, 27 Waiting for treatment 1 permanetly wounded.
The Capacity was show as 104/120 

On 3022-09-02 I had an prisoner event. that did result in 1 dead soldier, 1 wounded soldier, and if i remember correctly 2 dead PoWs und 2 wounded PoWs.

I recuited an other Platoon as Security force, but the capacity after the event is "0", even on 3022-09-09 
it has not recovered.

- unsure if the 100% reduction is intended
- newly hired guards should provide some capacity
- there should be some information and/or recovery over time 

 

### 3. Steps to Reproduce

[Save Games](https://gigamove.rwth-aachen.de/de/download/f025687ca44eb618cddce29ad0704d28) 
before and after the event
Download available till 2025-04-19

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

0.50.04

### Operating System *

Windows 10

### Java Version *

OpenJDK Runtime Environment Temurin-17.0.14+7 (build 17.0.14+7)

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

goron111:
I had an other PoW event (food poisening) on 3022-09-16, 
which did resolve the a "Security is back under control" notification,  but the Capacity is still "0"

goron111:
I did try to debug the problem by myselfe. here are my findings:

- In my campaing safefile after the event the _temporaryPrisonerCapacity_ is "-8"
- In _degradeTemporaryCapacity()_ the _degreeOfChange_ will be negative if the is   _temporaryPrisonerCapacity_ negative, so the _newCapacity_ will move in the wrong direction. 
[...]

PR #6567, opened 2025-04-06
Issue 6560: Fix for negative temporary prisoner capacity

goron111:
Fix for #6560 
- prevent temporaryPrisonerCapacity from going below MINIMUM_TEMPORARY_CAPACITY
  negative values should not be possible at all, but with 0% it would still be 
  impossible to contain the prisoners with any number of additional forces.
- MINIMUM_TEMPORARY_CAPACITY = 25 (25% normal prisoner capacity) 
  is an arbitrary number. I tried to balance being able to recover from an bad event
  and being to easy to recove. This value should be subject to further testing
- Changed the formular for degreeOfChange so that a lower value for temporaryPrisonerCapacity
  is not slower degrading than a higher temporaryPrisonerCapacity
- Updated the UnitTest to reflect the changes in degreeOfChange

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -6407,7 +6407,7 @@ public class Campaign implements ITechManager {
     }
 
     public void setTemporaryPrisonerCapacity(int temporaryPrisonerCapacity) {
-        this.temporaryPrisonerCapacity = temporaryPrisonerCapacity;
+        this.temporaryPrisonerCapacity = max(PrisonerEventManager.MINIMUM_TEMPORARY_CAPACITY, temporaryPrisonerCapacity);
     }
 
     public RandomEventLibraries getRandomEventLibraries() {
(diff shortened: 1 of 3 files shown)


IllianiCBT:
Hi, please update the PR description with a brief summary of the changes you made and (ideally) why. It'll make reviewing a lot easier and means if users click on the PR number in the change log they don't need to understand Java to know what changes were made.

IllianiCBT:
Also, sorry for the comment spam, but you _really_ want to be making changes in a working branch and not `master`. Master should be kept 'pure' at all times.

IllianiCBT:
Ignore my last comment. I'm pre-coffee and apparently woke up dumb.

IllianiCBT (approved):
Thanks for taking a look at this. It was on my TODO, but I'm not sure I'd have done as thorough a job, so great work. :)

If you can update the PR description that would be fantastic. But otherwise, nothing stopping this from being merged.

PR #6572, opened 2025-04-06
Added Randomize Traits Option For Character Creation

IllianiBird:
- Introduced a new option to randomize traits during character creation, including Connections, Reputation, Wealth, and Unlucky scores.
- Updated `CampaignOptionsDialog.properties` with labels and tooltips for the new setting.
- Extended `RandomSkillPreferences` to support `randomizeTraits` with default settings and XML integration.
- Modified personnel generators to calculate and assign random trait values when the option is enabled.
- Added methods to the `Person` class for modifying traits with proper clamping to ensure values remain within defined limits.
- Updated `AdvancementTab` to include a checkbox for enabling or disabling the new randomization feature.

### Dev Notes
This is the last of the random character creation features needed to finish off Traits and Attributes. While I was working on it I also adding clamping to the various Trait setters and changers to ensure they can't fall outside of expected values.


IllianiCBT:
- Added a personnel right-click menu to reset a character's Traits.

rjhancock (approved):

PR #6575, opened 2025-04-07
Updated Aging Effects to Include Clan Reputation Modifiers & Glass Jaw Gain

IllianiBird:
### Requirements
Requires #6607

### Dev Notes
This PR expands our aging implementation to inflict Glass Jaw & Slow Learner on characters whenever they hit certain milestones. As per RAW this SPA gain is inflicted every valid milestone. So while the character can mitigate it with repeated purchases of the Toughness and Quick Learner SPAs, but that gets expensive _quick_.

I also went and implemented Reputation penalties for non-blood named characters in Clan Campaigns once they hit specific Milestones. This required refactoring how Reputation is fetched, which also exposed a couple of minor bugs which I cleaned up as I went.


IllianiBird:
This PR now requires #6574 && #6605

rjhancock (approved):

PR #6578, opened 2025-04-07
Added Skill Check Dialog Using New Centralized Skill Check Utility

IllianiBird:
### Dev Notes
As mentioned in the title, this PR adds the ability for players to trigger Skill Checks at will. This is done by right-clicking on the character (or characters) and selecting 'Perform Skill Check', at which point the player is presented with a dialog where they can choose the skill and modifiers. They can also choose whether to use Edge, in the event of a failure.

image

Once the check has been set up the player is then presented with a second dialog showing the results of the check.

image

This functionality all uses our new centralized Skill Check utility class, ensuring that SPAs, traits, and attribute scores are all correctly factored into the target number.

#### Placeholder Art
This PR uses placeholder AI Generated art assets. A human artist has already been commissioned and expects to have the pieces ready by the launch of 50.05. Three pieces have been commissioned, the first is for the page where the user selects skill. Then there are two follow-up pieces one for when the check passes, another for when the check fails.

#### Ulterior Motives
Really I just wanted to test the new centralized Skill Check class and fix any bugs. That we get a fancy new dialog out of it is just a cherry on top of an unintentional cake.


MekHQ/src/mekhq/gui/dialog/SkillCheckDialog.java
```diff
@@ -0,0 +1,293 @@
+package mekhq.gui.dialog;
+
+import static megamek.common.Compute.randomInt;
+import static mekhq.campaign.personnel.skills.SkillCheckUtility.determineTargetNumber;
+import static mekhq.utilities.MHQInternationalization.getFormattedTextAt;
+
+import java.awt.GridBagConstraints;
```
rjhancock, line 7: Copyright header.

rjhancock (changes requested):

PR #6585, opened 2025-04-07
Fix #6563: Fixed In Transit Sorting in Warehouse Tab

### LINKED ISSUE #6563: [Issue] Leading 0's needed in delivery times to sort correctly 01,02,03 days vs 1,20,2,35,37,38,3 days
### Brief Description *

Bug for Mekhq Should have a leading 0 on Warehouse "Status" column for delivery times....2 gets put AFTER 20 days.... if you had leading 0's 01 days, 02 days, 03 days, they'd sort correct I think :)

I dont like how complex it is to report something.   I help bug fix for a variety of games, I screen shot, markup, drop into bug channels typically on discord.  Maybe someone would like to offer that and then transfer it here.  I likely wont report any more in the future.

### 3. Steps to Reproduce

_No response_

### Attach Files

Image

### Severity *

### MegaMek Suite Version *

50.04

### Operating System *

Window11

### Java Version *

?? no idea one of the newest

### Final Verification

- [x] I have included all necessary information and files to help reproduce this issue

IllianiBird:
- Implemented a new `StringAndThenNumberSorter` comparator to handle sorting of strings with embedded numeric values.
- Updated `WarehouseTab` to use `StringAndThenNumberSorter` for the `COL_STATUS` column in the parts table.
- Enhanced sorting behavior to properly order entries with mixed alphanumeric content, i.e. `In Transit (6 days)` will appear before `In Transit (10 days)`, but both will appear after `Functional`.

Fix #6563

### Dev Notes
RegEx is freakin' sorcery, I swear.

I stole this pattern from the internet so don't ask me _how_ it works, I just know it _does_.


MekHQ/src/mekhq/gui/sorter/StringAndThenNumberSorter.java
```diff
@@ -0,0 +1,140 @@
...
+     * A compiled RegEx pattern used to match and separate numeric and non-numeric parts of a string in the order they
+     * appear.
+     *
+     * <p>The pattern alternates between sequences of digits (numeric parts) and non-digit characters (non-numeric
+     * parts). This pattern is useful for parsing strings that need to be interpreted in a manner where numeric and
+     * non-numeric parts are processed differently.</p>
+     */
+    private static final Pattern NUMBER_PATTERN = Pattern.compile("(\\d+|\\D+)");
+
+    /**
```
rjhancock, line 68: Pattern is incorrect. that will match all numbers and non-numbers.

To match all numbers it is `\d+` (or for that variation `\\d+`)

rjhancock (changes requested):

MekHQ/src/mekhq/gui/sorter/StringAndThenNumberSorter.java
rjhancock: If it's only displayed in days, the `\\d+` will work. Otherwise you'll need to adjust it to be `\\(((\\d+) (\\D+)\\)` to capture the number of days AND interval to compare days, weeks, months, years...

IllianiCBT: I'm not sure what I need to do here, could you provide a little more information?

For context, the table column uses a mix of both numbers and non-numbers. For example, the column might have 'damaged', 'functional', 'in transit (5 days)', 'in transit (353 days)'. The intention is that it will sort based on letters and then, when it hits a number, start sorting based on that.

Right now, the implementation visually works correctly, but if I move it to "\\d+" I run into `Comparison method violates its general contract!` errors, presumably because I'm left with empty strings to compare because all the non-digits get filtered out.

IllianiCBT: Also, if it helps, it only appears in batches of 'days', we don't convert that to months or similar

IllianiBird:
- Replaced `StringAndThenNumberSorter` with a new `WarehouseStatusSorter`.
- Updated the comparator logic sort strings numerically for "day(s)" and alphabetically otherwise.
- Renamed and refactored the sorter to align with its updated functionality and naming conventions.
- Adjusted the `WarehouseTab` to utilize the new `WarehouseStatusSorter`.

MekHQ/src/mekhq/gui/sorter/WarehouseStatusSorter.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public class WarehouseStatusSorter implements Comparator<String> {
+    private static final MMLogger logger = MMLogger.create(WarehouseStatusSorter.class);
+
+    /**
+     * Regular expression pattern to locate the number of "days" in a string.
+     *
+     * <p>Matches strings like "(5 days)" or "(1 day)" and captures the numeric part.</p>
+     */
+    private static final Pattern DAYS_PATTERN = Pattern.compile("\\((\\d+)\\s*day(s)?\\)");
```
psikomonkie, line 59: You can make a group a "non-capturing" group by starting it with ?: `day(?:s)?`. Though in this case you don't need the capture group at all, you should get the same result with `days?`.

psikomonkie (commented):
Nice! 

I think I prefer c872c04fe14d88d0479de8172bd9896b70cb731e. While both currently work, I think the older commit version would be more robust. The older commit implementation will natively support translations as long as it follows the pattern of <Meaningful letters to sort by><meaningful number to sort by>. The older commit implementation will better support if a second warehouse part status gets "days" added to it, like days remaining for repairing something broken. I know this contradicts what we discussed yesterday but after comparing the two and taking translation support more seriously I think the original commit will be more robust.

Does anyone else have any thoughts on this?

IllianiBird:
> Does anyone else have any thoughts on this?

rjhancock (approved):

PR #6595, opened 2025-04-08
Fix #6341: Improved Name Escaping when Saving Unit in MekHQ's MekLab

### LINKED ISSUE #6341: [Issue] HQ Meklab, Putting '/' in the name of a refit throws an error.
### Brief Description *

Error message is thrown whenever you're using the MHQ meklab and you include a '/' in the model of a unit. This appears to include units that already have '/' in their models such as Weapons Carrier (AC/20)

Issue isn't gamebreaking in any way, but the error message tells the user to report it so here I am. Easy to work around so long as the user knows that putting a '/' in the model is a nono. (Also note, there may be other characters that throw similar issues, I only noticed it because I was making an AC/10 refit of a unit and got the error on that)

### 3. Steps to Reproduce

1, Have a unit in your hangar in HQ.
2, Put the unit into the lab for refit
3, Modify the unit a bit and set the Model to something including a '/' in it. For example, refit a unit to add an AC/2, and call it AC/2 refit.
4, Receive the error message after trying to order the refit.

### Attach Files

Image
Image

### Severity *

### MekHQ Suite Version *

0.50.04

### Operating System *

Windows 10

### Java Version *

17.0.12

### Final Verification

IllianiBird:
- Fixed a bug where users were unable to save units if the unit name included `/` or `\`.

Fix #6341


rjhancock:
Instead of escaping the file name, place it inside of a CDATA block.

IllianiBird:
CDATA doesn't work here because this isn't an xml entry, it's a filename. Wrapping it in a CDATA doesn't fix the issue, we're just using the xml escaping to ensure we don't need to input a manual escaping rule. The only issue being that `/` and `\` aren't escaped by the stock method, which is why I've added a special case.

PR #6598, opened 2025-04-08
Fixed Refitting Units With Bays

IllianiBird:
- Changed `oldUnitBays` and `newUnitBays` from immutable lists to mutable ArrayLists.

### Dev Notes
Because these lists were immutable we'd run into errors any time a user tried to refit them.

I'm marking this as 'high' because it completely prevents the refitting of affected units. I'm surprised we didn't have any issue reports.


psikomonkie (approved):
Nice!

PR #6602, opened 2025-04-08
Fixed 24-hour Work Day in Refit Display

IllianiBird:
- Replaced hardcoded time calculations with the `TECH_WORK_DAY`.
- Added singular/plural handling for "day" in the refit status message.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -266,11 +266,10 @@ public class Unit implements ITechnology {
         } else if (!isPresent()) {
             return "In transit (" + getDaysToArrival() + " days)";
         } else if (isRefitting()) {
-            int minutesInHour = 60;
-            int hoursInDay = 24;
-            int minutesInDay = hoursInDay * minutesInHour;
-            int days = (int) ceil((double) getRefit().getTimeLeft() / minutesInDay);
-            return "Refitting" + " (" + days + " days)";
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -266,11 +266,10 @@ public String getStatus() {
         } else if (!isPresent()) {
             return "In transit (" + getDaysToArrival() + " days)";
         } else if (isRefitting()) {
-            int minutesInHour = 60;
-            int hoursInDay = 24;
-            int minutesInDay = hoursInDay * minutesInHour;
-            int days = (int) ceil((double) getRefit().getTimeLeft() / minutesInDay);
-            return "Refitting" + " (" + days + " days)";
+            int days = (int) ceil((double) getRefit().getTimeLeft() / TECH_WORK_DAY);
```
psikomonkie, line 269: Consider how this is handled in `Campaign`:
```
int daysLeft = (int) Math.ceil((double) theRefit.getTimeLeft() / (double) tech.getDailyAvailableTechTime(campaignOptions.isTechsUseAdministration()));
```
I believe you could swap out `tech` by using `getRefit().getTech()`. This would ensure it has the true value in case the player is using a tech that has tech as a secondary role.

IllianiBird: Great call, consider it done :)

IllianiBird:
- Updated `Unit.java` to calculate refit duration dynamically based on the assigned technician's available work time per day.

rjhancock (approved):

PR #6634, opened 2025-04-12
Fixed NPE When Loading Units Without an Engine

IllianiBird:
- Enhanced `Accountant` class to handle null engine scenarios, including logging a debug message and skipping null-engine units during fuel calculations.


psikomonkie (approved):
Amazazing!

PR #6648, opened 2025-04-12
Tractor transport types can now be used for leadership bonus unit selections

psikomonkie:
When selecting bonus units from leadership, there is a dropdown that lets the UI automatically select assigned tactical or ship transport units. This expands that to include tractors.

image


MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -289,6 +292,8 @@ public static Vector<Pair<String, CampaignTransportType>> getLeadershipDropdownV
         retVal.add(new Pair<>(MHQInternationalization.getTextAt("mekhq.resources.AssignForceToTransport", "CampaignTransportUtilities.selectTransport.null.text"), null));
```
IllianiBird, line 292: Is it worth defining this resource bundle as a class static?

psikomonkie: I don't particularly think so, but I did it anyway :)

IllianiBird (approved):

PR #6665, opened 2025-04-14
Fix #6654: Added New Player Quickstart Option

### LINKED ISSUE #6654: [RFE] Replace old Campaigns with "The Learning Ropes"
### Enhancement Type *

The Download Packages in include 3 Campaigns (Version 0.47.15) 

- "Fist and Falcon" 
- "Sword and Dragon" 
- "Wolf and Blake"

These are not supported by 0.50.x and can not resonable be upgraded.
Keep them in the Repo but remove them from the Download file.

As an alternative Capmpaign "The Learning Ropes" used in the "docs/0_New Player Guide.pdf" 

### Use Cases or Rationale

The old Campaigns 
- can not be used
- can lead to irreleant bug reports

Removing the file reduces the download size

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.05-SNAPSHOT

### Final Verification

IllianiBird:
Fix #6654

### Dev Notes
This PR adds a new 'New Player Quickstart' option to the MekHQ start screen.

image

Selecting this will present the user with a dialog setting the scene and highlighting the New Player Guide and where to find it.

image

Once this dialog is confirmed MekHQ loads a stock save which has a campaign all set up with a _very_ nice starting contract already prepped in the Contract Market. At that point the user is free to do as they will.


IllianiBird:
The lines changed looks crazy, but only because I moved the original starter campaigns into a new folder

MekHQ/src/mekhq/gui/panels/StartupScreenPanel.java
```diff
@@ -155,15 +174,29 @@ protected void initialize() {
...
-                UIComponents.MainMenuButton.getComp(), true);
+              UIComponents.MainMenuButton.getComp(),
+              true);
         btnLoadLastCampaign.setEnabled(lastSaveFile != null);
         btnLoadLastCampaign.addActionListener(evt -> {
             btnLoadLastCampaign.setEnabled(false);
             startCampaign(lastSaveFile);
         });
 
+        MegaMekButton btnNewPlayerQuickstart = new MegaMekButton(resources.getString("btnNewPlayerQuickstart.text"),
```
IllianiBird, line 185: This block is the only code changes in this file, the rest is EC

IllianiBird:
After several hours of mucking about, I'm finally marking this 'ready for review'

psikomonkie (approved):
Nice! I wanted something like this. 

This is approved, but I'm a little concerned around the maintainability of this. I think it'd be really neat if we could explore programmatically creating the new player campaign instead of providing a campaign file, so hopefully as changes are made to the campaign or saving format we won't need to keep updating the new player campaign. I think we should consider that for the next release cycle so we can spend less time maintaining the new player campaign file.

PR #6669, opened 2025-04-14
Fix/acar move entity by copy

Scoppio:
Quick fix for ACAR, should preserve the unit "as-is" and still work in a multi-threaded environment with hundreds or thousands of simulations running witohut causing havoc in the original entities of the player or bot.

A better solution is under development on MM, but this fix is more important as the problem doesn't seen to be too destructive on MM alone.

- depends on MM [#6883](https://github.com/MegaMek/megamek/pull/6883)


psikomonkie (approved):
Nice!

PR #6675, opened 2025-04-14
Fixed Command Skills Counting Down not Up

IllianiBird:
- It appears 50.05 picked up a brief bug where the three command skills (Strategy, Command, and Leadership) started counting down, not up. This corrects that.


psikomonkie (approved):
Nice!

PR #6674, opened 2025-04-14
Implement RFE #6058 display cost of orderd parts

goron111:
Implement alternative to RFE #6058 

I did not implement add the "Total Cost" to the "Parts in Use" window 
as that is already overcrounded and needs to be redone.

- [x] display total cost of Parts in procurement in Comand Certer Tab
- [x] NAG if total cost > Funds  
- [x] refactor calculation of totalBuyCost for IAcquisitionWork  
        ground work to implement "Total Cost" in "Parts in Use" window

- [x] fixed sorting the procurement table by "Total Cost"  
  extracted in PR #6690 

- [x] Update Total Cost Cell on inreasing decreasing part quantity in Table  
   extracted in PR #6690 

- [x] Added UnitTests

DisplayCostOfOrderdParts1
DisplayCostOfOrderdParts2
UnableToAffordShoppingList

PR #6690, opened 2025-04-15
Fix 2 unreported problems in procurement table

goron111:
While working on #6674 i found and fixed 2 unreported problems.
- Fixed sorting for the "Total Cost" column  in the procurement table
- Fixed cell update for "Total Cost" column if quantity is incremented / decremented


IllianiBird:
So something I've noticed is that the table is only updating when it's interacted with, not when the items are added. This can be seen by going into the purchase unit or part dialog, adding a bunch of the same unit and observing the table. Unless the table is interacted with - or the tab is changed - the total will not update.

Is that something you could look into?

IllianiBird (approved):
LGTM, though we should take a look at why the table isn't updating when incrementing items via the purchase dialogs.

PR #6725, opened 2025-04-22
Fix #6400: Corrected Cargo Capacity of Cargo Bays to Account for Transported Units

### LINKED ISSUE #6400: [Issue] Infantry occupying a cargo bay don't effect cargo capacity
### Brief Description *

It is now possible for infantry to be loaded into cargo bays. This currently doesn't reduce the convoy's cargo capacity. 

### 3. Steps to Reproduce

1. Have a unit with cargo capacity and an infantry unit weighing less than the cargo capacity.
2. Assign the unit with cargo capacity to a convoy force
3. Assign the infantry unit to have the unit with cargo capacity as a tactical transport.

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

Latest

### Operating System *

W11

### Java Version *

Adoptoum

### Final Verification

IllianiBird:
- Introduced `getCapacityModifiedForTransportedUnits` to dynamically adjust cargo bay capacity based on transported units.


HammerGS (approved):
Check with AI and no issues found,.

psikomonkie (changes requested):
Amazing! It really bothered me how infantry could hitch a ride for free. This is a great step towards totally preventing that, and makes it go from something a player can accidentally do to something that a player has to go out of their way to exploit. 

Anyway, while writing up the comments and inquiring about your intent, I did come across a bigger question:

If (when, I'm optimistic) units are persistent on the MekHQ stratcon map I think there will be a place for _actual_ transported stuff. Assignments will still live, to assist the player with not having to manually load units, but I think there will be a point where we need to care what a unit in MekHQ is transporting at a given moment. 

At that point, should this method still use the assignments (therefore it gives you the ideal or expected tonnage used) or should this method look at the _actual_ space being used?

Or I guess the answer could be "both" if you're running this both to generate the expected convoy weight and again if there's an interception to make sure the player isn't being sneaky. 

Requested changes if this is intended to be used for assignments: see the comments I left. 

Request changes if this is intended to be actual capacity or if you're confident you're going to use it as "both" in the meantime: 
 - Ignore the two comments related to the method signature and method JavaDoc, they were specifically written assignment oriented version of this method in mid
 - Update the JavaDoc to at least briefly mention it's currently using transport assignments. 
 - Add a comment (TODO? Probably premature) within the method mentioning that this method will need updated once that's implemented. 
 - And add a variable `TransporterType transporterType = TACTICAL_TRANSPORT` in the method and use that - directly passing a specific transport type has always wound up coming back to bite me. I don't think you need to make it a whole parameter like if this is intended to always be used for assignments, but having it mapped to a variable'll make it a bit easier to refactor & hopefully isn't much work for you to make the change 😀 


Thanks Illiani!!

psikomonkie (approved):
Nice!! This'll be a great feature to include in 0.50.05, I like how it ties together the (relatively) new Tactical Transport Assignments, the ability for units to be transported in cargo, and the MHQ convoy system. 

I especially appreciate the changes to the JavaDoc, I think transport & cargo is very confusing in MHQ.

One new comment, it's optional, it shouldn't break anything if you leave it but it caught my eye. Lmk if I didn't describe it well enough, we can chat on Discord.

PR #6726, opened 2025-04-22
Fix #1914 : Refactored Contract Terms Negotiations to Be Closer to RAW

### LINKED ISSUE #1914: RFE: Negotiator Flag
"Additionally, being able to set a negotiator flag would be beneficial. I can imagine having a team of negotiators rather than a single administrator handling all contract negotiation."

This was originally requested in #1911, and was split off to ensure a single item per issue.

IllianiBird:
- More consistently use the Negotiation skill for negotiations.
- The most senior Admin characters in each roll (Command, Logistics, and Transport) are used, not necessarily the best per skills. This brings negotiation in line with other systems while also giving the player more control and predictability over who makes the rolls.
- Reduced the number of negotiation rerolls to 1 (rather than 1 per rank in Negotiation). This matches RAW.
- Labeled the renegotiation buttons with who is negotiating

Fix #1914 (by providing the player a way to directly influence who performs the negotiations)

### Dev Notes
These changes mark another step down in the power of the player and further slow progression by reducing the frequency of 'perfect' contracts. Now choosing to negotiate is a gamble, rather than players being able to negotiate their way into ideal conditions for most contracts offered.

### DIFF
--- a/MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
+++ b/MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
@@ -29,6 +29,9 @@
 package mekhq.gui.view;
 
 import static megamek.client.ui.WrapLayout.wordWrap;
+import static mekhq.campaign.Campaign.AdministratorSpecialization.COMMAND;
+import static mekhq.campaign.Campaign.AdministratorSpecialization.LOGISTICS;
+import static mekhq.campaign.Campaign.AdministratorSpecialization.TRANSPORT;
 
 import java.awt.Cursor;
 import java.awt.GridBagConstraints;
@@ -80,6 +83,10 @@ public class ContractSummaryPanel extends JPanel {
     private JLabel txtStraightSupport;
     private JLabel txtBattleLossComp;
 
+    private Person commandNegotiator;
+    private Person transportNegotiator;
+    private Person logisticsNegotiator;
+
     private final ResourceBundle resourceMap = ResourceBundle.getBundle("mekhq.resources.ContractMarketDialog",
             MekHQ.getMHQOptions().getLocale());
     private ContractPaymentBreakdown contractPaymentBreakdown;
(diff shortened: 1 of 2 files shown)


HammerGS (approved):
Checked with AI no issues found

MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
```diff
@@ -572,37 +579,45 @@ private boolean hasSupportRerolls() {
...
+    @Deprecated(since = "0.50.05", forRemoval = true)
     private String generateRerollText(int rerolls) {
-        StringBuilder text = new StringBuilder(resourceMap.getString("lblRenegotiate.text"));
-        if (method != ContractMarketMethod.CAM_OPS) {
-            text.append(" (")
-                .append(rerolls)
-                .append(')');
-        }
-        return text.toString();
+        return generateRerollText("");
```
psikomonkie, line 616: If we're going to deprecate a method it should retain the same function if possible. Could this return `generateRerollText(" (" + rerolls + ')')` so it returns the same value as it did before this change, and check if the contract market method matches?

If that's not possible I think you should delete it. Deprecating a method while also removing its functionality doesn't feel right.

psikomonkie (approved):
Nice! I do have one request regarding the deprecated method, but not pressing enough to require it.

MekHQ/src/mekhq/gui/view/ContractSummaryPanel.java
IllianiBird: I can. For the last point, take that up with Tex. I was instructed to deprecate every method I changed the signature on and that's what I'm doing.

IllianiBird: Done

rjhancock: The change both changes the method signature AND the way it functions. If all other ContractMarketMethod's are removed and ONLY the CAM_OPS option is left, that's fine. If they still exist, the new method should still account for it and the value passed to it.

As for depreciation of methods, @IllianiBird is doing it as requested (mostly). Deprecate the old method and have it call the new method passing appropriate values to it along with a JavaDoc denoting which method to use going forward.

That being said, these are private methods so it is less of an issue and a straight up removal would not be a bad thing either. Protected and Public methods need a longer period. Private methods would prefer this way, but not out right required as the only place they'll be called is in this one file. Much easier to determine if they are in use.

PR #6736, opened 2025-04-22
Simplified Roleplay Skills Randomization

IllianiBird:
- Roleplay skills now work off a single modifier, rather than being adjusted based on the experience level of the character they're being generated for.

### Dev Notes
This reduces the number of options users need to engage with, should they wish to make adjustments to roleplay skill frequency (or disable them).

It also reduces the amount of effective xp investment skipped when hiring characters from the market.


IllianiBird:
Psi, I'm requesting your review on this, as it was your comments which prompted the change. This will ensure Dependents and other low xp characters can still get role-play skills.

HammerGS:
@IllianiBird

psikomonkie (approved):
Nice! It's a shame to see all the stuff you did for having it be configurable go to waste - a thought hit me while reading this, it could be nifty to have the configurable rolls be tied to age - older characters having more/higher roleplay skills... Anyway, that's another RFE for another time.

PR #6737, opened 2025-04-22
Adjusted Maintenance Cycle Hangar Display

IllianiBird:
- Changed next maintenance cycle display to count down rather than display 'x / y days', counting up. This makes it much more intuitive at first glance.


HammerGS:
@IllianiBird

PR #6741, opened 2025-04-22
Fix #3963: Added the Ability to GM Increase the Number of Each Part in the Warehouse.

### LINKED ISSUE #3963: RFE: Expand GM Options in Warehouse to match existing functionality
### Environment
49.xx Branches
Windows 10
Java 11.0.12

### Description
When right clicking a part in the warehouse while GM mode is enabled, the only option available to the player under the GM actions is to remove all of a part, or change quality. This prevents easily allowing inventory adjustments to support house rules or other GM fiat in a campaign, as you either have to manually re-add the new desired count of items via the Purchase Parts menu (As removal removes it from warehouse entirely) or you have to use the sales feature, then match the credit with a debit to negate the gains. Neither are particularly intuitive or smooth to execute.

Requested enhancement would be a "Remove #" option, that provides a dialog allowing you to select how many of an item you wish to remove, same as the buy and sell # of type options that are presented. It might be a good idea to provide a matching "Add #" as well, as you can buy/sell from warehouse already, but its not as necessary as GM add is available through the Purchase Parts menu already.

IllianiBird:
- While in GM mode users can 'add n parts' in addition to the ability to remove n parts added earlier in 50.05.
- Ammo is increased by n shots. Armor is increased by n points, other parts are just increased by n.

Fix #3963

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/PartsTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/PartsTableMouseAdapter.java
@@ -170,6 +170,32 @@ public class PartsTableMouseAdapter extends JPopupMenuAdapter {
                 int quantity = dialog.getValue();
                 gui.getCampaign().getWarehouse().removePart(selectedPart, quantity);
             }
+        } else if (command.equalsIgnoreCase("ADD_N")) {
+            if (null != selectedPart) {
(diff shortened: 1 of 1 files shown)


HammerGS:
@IllianiBird

PR #6743, opened 2025-04-22
Fix #4086: Added Parts In Use Button to Warehouse Tab

### LINKED ISSUE #4086: RFE: Inventory Management improvement
Ideas for inventory management improvement:

Used parts report inside warehouse so all inventory management can be done in one tab
     Having the command center one is still good for overview purposes

Exportable Warehouse and Used Parts Lists (.csv or .xls)
     Having the Parts in Use list be the same format as Warehouse would help with this (if possible)

Ammo/Armor Nag for low inventory
     Inventory drop below set number send Nag to user.
     Potentially linked to an "auto" purchase system
     Could be static set level or with auto-update based on a given "need" parameter
          Example: Need 35 days worth LRM ammo times ([Used LRM Launchers]) = "need" number

IllianiBird:
- Introduced a "Parts In Use" button to the `WarehouseTab` GUI.

Fix #4086


HammerGS:
@IllianiBird

PR #6754, opened 2025-04-23
Issue #6753: Check if `Game` is null when replacing a Unit's Entity & Transfer C3 During Refit if Possible

psikomonkie:
An `Entity` needs to be in a game to copy its C3. If it's not a `Game`, let's not copy its C3 network. Fortunately, when refitting, we can still restore the C3 info later once the new entity is added to the game. Neat!


IllianiBird (approved):
Great job resolving this so fast :)

PR #6784, opened 2025-04-26
Fix: Stopped Mercenary Auctions Gifting a Hundred Support Points (No Really)

IllianiBird:
### Dev Notes
We were incorrectly adding the total bid success chance to the players available Support Points instead of deducting the support points spent.


HammerGS:
@IllianiBird Co-pilot

IllianiBird:
This should be in 50.06 because it resolves a problem with SP economics and us accidentally gifting the player with 20+ SP per event

Scoppio (approved):

PR #6792, opened 2025-04-27
Fix: Fixed Error When Attempting to Fetch Bloodname for Characters Without a Bloodname

IllianiBird:
Fix [Sentry Report A](https://sentry.tapenvy.us/organizations/tapenvyus/issues/4610/events/8879ee6b92fa453586fa405c00629436/), [Sentry Report B](https://sentry.tapenvy.us/organizations/tapenvyus/issues/4610/events/8879ee6b92fa453586fa405c00629436/)

### Dev Notes
As part of CamOps Reputation we need to test whether a character has a Bloodname. By all accounts bloodname should initialize as a blank string, however Sentry reports show it can also be null. So we need to test for null state.


IllianiBird:
This should be in 50.06 as it fixes a constant source of NPEs following prisoner capture.

PR #6806, opened 2025-04-28
Fix: Fixed Exception When Viewing Past-Scenarios While Blind Drop or True Blind Drop is Enabled

IllianiBird:
This fixes an exception that would occur when a player tried to view an already resolved scenario while having Blind Drop or True Blind Drop enabled.

### Work Around
Restarting the client fixes the issue in 50.05

### DIFF
--- a/MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
@@ -284,6 +284,10 @@ public class AtBScenarioViewPanel extends JScrollablePanel {
             DefaultMutableTreeNode top = new DefaultMutableTreeNode(botStub.getName());
             List<String> allEntries = botStub.getEntityList();
             for (String entry : allEntries) {
+                if (!scenario.getStatus().isCurrent()) {
+                    continue;
+                }
+
                 if ((team != 1) && isTrueBlindDrop) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
```diff
@@ -284,6 +289,10 @@ private void fillStats() {
             DefaultMutableTreeNode top = new DefaultMutableTreeNode(botStub.getName());
             List<String> allEntries = botStub.getEntityList();
             for (String entry : allEntries) {
+                if (!scenario.getStatus().isCurrent()) {
+                    continue;
+                }
```
Scoppio, line 294: I will need some convincing that this is not just hiding all entries for all non-current scenarios.
Actually, if `isCurrent` would be important here, than it should be moved 2 lines up and just not even bother with getting the entityList at all if its not current scenario.

Scoppio (changes requested):

MekHQ/src/mekhq/gui/view/AtBScenarioViewPanel.java
IllianiBird: It's definitely hiding units for non-current scenarios, and that's intentional. Your point about moving the conditional up a couple of lines was a good call.

Scoppio (approved):

PR #6807, opened 2025-04-28
Fix: Significantly Improved TO&E Reading Speed by Removing C3 Calculations

### LINKED ISSUE #6794: [Issue] [0.50.05] HQ hangs when opening TO&E tabs
### Brief Description *

When trying to open a force tab in the TO&E in my campaign, the game hangs for several minutes while my computers fans work hard. It does eventually expand the tab, but the issue repeats when I double-click to open one of that force's subforces.

This campaign file was imported from 50.03, had its settings updated according to the update instructions, was saved in 50.05, and was then reopened. This issue is occurring with the new save.

### 3. Steps to Reproduce

1. Open HQ file.
2. Go to TO&E tab.
3. Double-click on 12th Taskforce force tab.

### Attach Files

[battle_armor_maintenance_bug.zip](https://github.com/user-attachments/files/19930781/battle_armor_maintenance_bug.zip)
(Don't mind the file name, I found this issue while reporting a different bug.)

### Severity *

### MekHQ Suite Version *

0.50.05

### Operating System *

Windows 11

### Java Version *

Java Vendor: Eclipse Adoptium  Java Version: 17.0.14

### Final Verification

IllianiBird:
Fix #6794

We previously received feedback from a user that C3 was not being factored into BV calculations in the TO&E tab. We then went in and fixed this, however that caused the every C3 equipped unit needing to read the ammo loadout of every unit in the TO&E any time the TO&E was interacted with. This resulted in a lock-up for large campaigns.

### DIFF
--- a/MekHQ/src/mekhq/gui/view/ForceViewPanel.java
+++ b/MekHQ/src/mekhq/gui/view/ForceViewPanel.java
@@ -189,6 +189,7 @@ public class ForceViewPanel extends JScrollablePanel {
         for (UUID uid : force.getAllUnits(false)) {
             Unit unit = campaign.getUnit(uid);
             if (null != unit) {
+                // Never factor in C3 in this check. It will cause the TO&E to lock up for large campaigns.
                 bv += unit.getEntity().calculateBattleValue(true, !unit.hasPilot());
                 cost = cost.plus(unit.getEntity().getCost(true));
                 ton += unit.getEntity().getWeight();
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/view/ForceViewPanel.java
```diff
@@ -182,16 +187,16 @@ private void fillStats() {
...
-
-            Unit u = campaign.getUnit(uid);
-            if (null != u) {
-                cost = cost.plus(u.getEntity().getCost(true));
-                ton += u.getEntity().getWeight();
-                String utype = UnitType.getTypeDisplayableName(u.getEntity().getUnitType());
+            Unit unit = campaign.getUnit(uid);
+            if (null != unit) {
+                // Never factor in C3 in this check. It will cause the TO&E to lock up for large campaigns.
+                bv += unit.getEntity().calculateBattleValue(true, !unit.hasPilot());
```
psikomonkie, line 193: Why do we need to calculate the entity's battle value every time? Could we just use the value from the entity's `getInitialBV()`? Could that be used in `force.getTotalBV(campaign, true)`?

Scoppio: This could be cached.

Scoppio (approved):
I agree with the rollback, this is not the time for large refactors and redesigns for big problems.

PR #6810, opened 2025-04-29
Improvement: Improved Robustness of Automatic Unmothballing in Contract Automation

IllianiBird:
### Dev Notes
During a recent user stream it was noted that units failed to unmothball upon arriving at the contract. I investigated, and the most likely cause was us storing units instead of unit IDs. If, for some reason, a unit changed they would no longer match what was stored. While this shouldn't happen I went ahead and switched to storing unit IDs - which are both unique and will never change.

### DIFF
--- a/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
+++ b/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
@@ -95,7 +95,6 @@ import mekhq.campaign.market.contractMarket.AtbMonthlyContractMarket;
 import mekhq.campaign.mission.AtBContract;
 import mekhq.campaign.mission.Mission;
 import mekhq.campaign.mission.Scenario;
-import mekhq.campaign.personnel.medical.advancedMedical.InjuryTypes;
 import mekhq.campaign.parts.EnginePart;
 import mekhq.campaign.parts.MekActuator;
 import mekhq.campaign.parts.MekLocation;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1095,10 +1095,10 @@ private static List<Person> processPersonnelWhoAdvancedInXP(Node workingNode, Ca
         return personWhoAdvancedInXP;
     }
 
-    private static List<Unit> processAutomatedMothballNodes(Node workingNode, Campaign campaign) {
+    private static List<UUID> processAutomatedMothballNodes(Node workingNode) {
         logger.info("Loading Automated Mothball Nodes from XML...");
 
-        List<Unit> mothballedUnits = new ArrayList<>();
+        List<UUID> mothballedUnits = new ArrayList<>();
```
Scoppio, line 1101: This is the only thing I dislike. Do not remove the reading mothball `Unit`, instead, remove the persist of Unit and save UUID as you did, I believe you are correct in doing that, but for reading, keep the old one and make it do a fallback in the `Unit` if it cant find the new `automatedMothballIds`. Otherwise you could see something nonsense like opening campaing with a mothballed unit and it just disappears, etc.

Scoppio (changes requested):

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
IllianiBird: So in that instance the unit wouldn't disappear, because we're not loading the unit in this instance, the unit is loaded elsewhere.

What we were doing was loading the unit (elsewhere, during normal unit loading) and then loading a second instance of the same unit here and then just hoping they matched up. I don't think keeping the double-processing here is helpful.

IllianiBird: Ok, so I just went and triple-checked this and the units definitely, 100% do _not_ vanish. Everything works as intended, the unit unmothballs and all is good. It's worth noting, in case it wasn't obvious, that this is _not_ a list of all mothballed units, it's just a list of UUIDs for units that have been auto mothballed during the contract acceptance process, that list is then used so we have a record of what units we should unmothball once the campaign arrives on contract.

Scoppio (approved):

PR #6813, opened 2025-04-29
Improvement: Reworked Integrated Command Rights to No Longer AutoAssign Forces and to Allow Interactions with StratCon

IllianiBird:
### Dev Notes
Integrated command rights have long been a pain point for players and generally considered a 'newbie trap'. Under integrated command rights the player does not deploy forces, instead MekHQ will pick a force at random and deploy that for them.

This means StratCon players under integrated command rights _do not play StratCon_ they play "advance day until something interesting happens and hope MekHQ doesn't deploy an infantry company against an assault lance for the 4th time this contract."

This PR changes Integrated command rights so that they work like House with some minor differences. Every scenario is still considered a Turning Point, under integrated, but unlike House command rights you will be supported by allied forces in every scenario.

This makes integrated rights a good choice for starting campaigns - the time in which players are most likely to be posted to integrated contracts.


IllianiBird:
I have tested this change both in a standard contract (non-integrated) and in an integrated contract and I've not noticed any issues. The bulk of this PR consist of removing special handlers for Integrated command rights, rather than adding new content.

While I recognize the reduced scope of 50.06 I think the inclusion of this PR will be a net positive. During a recent user stream I spotted the user encounter a terrible play experience where they accidentally selected an integrated contract. They only discovered this after traveling for close to a year (to reach the contract site). This meant their autosave had overwritten their pre-contract save. They were left with the choice of canning the contract and hopefully finding another one close enough. Or playing through the contract and hoping for the best.

Landmines - problems users only know not to step on if they already know not to step on them - are poor design, especially if there is nothing the user can do to fix the issue.

psikomonkie (approved):
I like it! I might even use integrated next time I'm starting a new game. It always seemed like a trap to me!

IllianiBird:
I'm second-guessing myself and moving this to 50.07

IllianiBird:
I'm second guessing my second guessing. I think this should be ok for 50.06. @HammerGS what are your thoughts?

IllianiBird:
Flipping this to live

IllianiBird:
This accidentally still had the new dev cycle tag. It has since been removed.

PR #6814, opened 2025-04-29
Fix: Fixed Arrival Time in Several Scenario Templates

IllianiBird:
### Dev Notes
This removes the arrival time delay for several scenarios. The addition of the arrival time deplay made sense of paper, less so in practice.

I also removed an errant reinforcement force that didn't need to be in the scenario.


DM0000 (approved):
LGTM.  That reinforcement force being generated elsewhere or just gone from skirmish?

IllianiBird:
Juat gone from skirmish

Scoppio (approved):

PR #6816, opened 2025-04-29
Fix: Fixed Ammo Not Stacking in the Warehouse

IllianiBird:
### Dev Notes
This issue was caused by the code that checks whether an item already exists in the warehouse, prior to attempting to stack the parts. What was happening is we checked the warehouse and if the first eligible item we found didn't have the same newness as the part we were comparing MekHQ would mistakenly think there were no eligible items. This resulted in the part (usually ammo) not stacking.

We also had a couple of lines of code that were trying to specially handle ammo. However, we were already handling ammo later in the process. So I removed it.


MekHQ/src/mekhq/campaign/Warehouse.java
```diff
@@ -263,13 +271,52 @@ private Part mergePartWithExisting(Part part) {
      * @return The matching spare part or null if none were found.
      */
     public @Nullable Part checkForExistingSparePart(Part part) {
-        Objects.requireNonNull(part);
+        if (part == null) {
+            logger.error("checkForExistingSparePart(Part): Part is null");
```
Scoppio, line 275: This is not an error, it's a debug, Maybe an `info`, but it will show up in the log as an error and it will mean nothing, it won't have any kind of information, wont show an ID, nothing, so its an error log that will exist but no one will be able to do anything to act upon this error being logged.

MekHQ/src/mekhq/campaign/Warehouse.java
```diff
@@ -263,13 +271,52 @@ private Part mergePartWithExisting(Part part) {
...
+     */
+    public @Nullable Part checkForExistingSparePart(Part part, boolean includeNewnessCheck) {
+        if (!includeNewnessCheck) {
+            return checkForExistingSparePart(part);
+        }
+
+        if (part == null) {
+            logger.error("checkForExistingSparePart(Part, boolean): Part is null");
+            return null;
+        }
```
Scoppio, line 313: Check for null first unless `checkForExistingSparePart` accepts @Nullable.

Also, this log is another instance of an error state that isn't really an error, since there is nothing that can be done about it other than say "something happened". Either you raise that as a special exception so somewhere "upstream" you can manually handle the error, or you just accept that if you receive a null you will return a null. In either case, you can remove this error log.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -146,11 +150,6 @@ public void addPart(Part part, int transitDays, boolean isBrandNew) {
             return;
         }
 
-        // don't keep around spare ammo bins
-        if ((part instanceof AmmoBin) && (null == part.getUnit())) {
-            return;
-        }
```
Scoppio, line 152: Are we 100% certain that this piece of code is unnecessary? I won't judge, just want a second confirmation from you @IllianiBird that this is 100% as intended.

Scoppio (changes requested):

MekHQ/src/mekhq/campaign/Quartermaster.java
IllianiBird: It shouldn't be necessary, but you know what? Let's add it back. If it's not necessary it won't do any harm, and if it is somehow necessary (despite being handled elsewhere in the pipeline) it's sitting pretty.

MekHQ/src/mekhq/campaign/Warehouse.java
IllianiBird: Surely we should be logging unexpected `null`? Otherwise we end up in situations where erroneous nulls are being swallowed silently with nothing in the logs to show it happened. As occurred in 50.05's dev cycle which nearly led us to shipping 50.05 with completely broken OpFor generation.

Scoppio: If you log `null`  you dont get any meaningful information other than that a null arrived there, there is no ID that generated the null, no entity, nothing, meaning it is useless from a bug-fixing standpoint. If you want to log that something went wrong, sure, add a warning with something like `expected part ID {} but received null`, which won't give you any information unless you had an info log prior to that which gave something like `logger.info("Looking at unit {}", unit);`

Scoppio (approved):

MekHQ/src/mekhq/campaign/Warehouse.java
IllianiBird: What I ended up doing has logging a handled NPE, so that we'd get the stack dump which will tell us where the error was coming from.

PR #6820, opened 2025-04-29
PR: Moved SPA Info in Preparation for Future Changes

IllianiBird:
### Dev Notes
I'm about to move some logic out of `AbilitiesTab.java` connected to the categorization of SPAs. This will allow that logic to be used anywhere, not just campaign options. However, to do that I needed to move the enum values out of `CampaignOptionsAbilityInfo.java`. I was advised to make a PR that just did the move and renaming, and then to follow up with the logic changes.


Scoppio (approved):
I suggest to avoid doing just the static import of enum values once the class is moved to another places.

PR #6825, opened 2025-04-29
Improvement: Switched Medical Daily Notices from Reports to Log Entries

IllianiBird:
### Dev Notes
Changed several medical notices from posting to the daily report to instead post to the log. This reduces daily report spam, without losing the information if we ever need to troubleshoot. After we're certain this system is working correctly we can change the log level from 'info' to 'debug'

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/medical/MedicalController.java
+++ b/MekHQ/src/mekhq/campaign/personnel/medical/MedicalController.java
@@ -24,11 +24,23 @@
  *
  * Catalyst Game Labs and the Catalyst Game Labs logo are trademarks of
  * InMediaRes Productions, LLC.
+ *
+ * MechWarrior Copyright Microsoft Corporation. MekHQ was created under
+ * Microsoft's "Game Content Usage Rules"
+ * <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
+ * affiliated with Microsoft.
  */
 package mekhq.campaign.personnel.medical;
 
+import static mekhq.campaign.personnel.medical.advancedMedical.InjuryUtil.resolveDailyHealing;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/medical/MedicalController.java
```diff
@@ -152,7 +160,8 @@ public boolean checkNaturalHealing(Person patient) {
      * @param doctor  the {@link Person} performing the treatment
      */
     private void healPerson(Person patient, Person doctor) {
-        campaign.addReport(getFormattedTextAt(RESOURCE_BUNDLE, "MedicalController.report.intro",
+        // TODO change logging level from info to debug in 50.08
+        logger.info(getFormattedTextAt(RESOURCE_BUNDLE, "MedicalController.report.intro",
```
Scoppio, line 164: I get it, but I would prefer to have an option in campaign options where you can re-enable it to show in the campaign report, start with it disabled so if anyone complains they just have to click a button.

Scoppio (changes requested):

MekHQ/src/mekhq/campaign/personnel/medical/MedicalController.java
IllianiBird: Even with small campaigns (I tested with just a few people in the infirmary) it quickly becomes really spammy. We need the information, but the player has never had access to this information so nobody will notice it missing.

IllianiBird: Coming back to this later, it's also worth mentioning that posting lots of messages to daily report on New Day actually slows down new day progression so it's something we should be avoiding.

Scoppio (approved):

PR #6839, opened 2025-05-01
Improvement: Added Tooltips that Explain the Skill When Viewing a Person in the Personnel Tab or Purchasing/Improving Skills

IllianiBird:
image

### Dev Notes
We frequently get contacted by users querying what one skill or another does. With the introduction of both aging effects, SPAs that affect skills, traits, _and_ attribute scores this was only going to get worse.

I went ahead and created a set of tooltips that should significantly reduce confusion. Now, whenever a player is purchasing (or improving) a skill, or viewing a character in Person View, simply hovering over the skill will show not just what the skill does but what modifiers are affecting it.


IllianiBird:
Removed 'For New Dev Cycle' tag. Updated to be suitable for 50.06

Scoppio (approved):

Scoppio:
its alot of new code, i'd be happier if at least the newly added code in Skills was fully covered in tests instead of just 30% of it.

IllianiBird:
It really isn't a lot of code if you consider that the past majority of it is in guy components or resource bundles.

IllianiBird:
However, as much as I don't want to write even more tests, if there is specific things you think need testing lemme know. As it is, I tested what I thought needed to be tested.

IllianiBird:
This PR was required by #6894 as that has been merged and this is approved, I'm going to go ahead and merge this.

PR #6843, opened 2025-05-02
Fix: Fixed Default Attribute Scores; Fixed Attribute Score Randomness; Reworked Attribute Score Randomness Option; Added Starting Attribute Score Documentation

IllianiBird:
### Dev Notes
There were a few issues with starting attribute scores. Namely that some of the scores were wrong (0, in some places, or 41 charisma for LAM Pilots).

Attribute score randomization also wasn't correctly resetting attribute scores to their default value before randomizing. What this would mean is that character attributes would frequently spike towards one extreme or the other.

Regarding attribute score randomization, I didn't like the play feel of the extra randomness option. Especially as we didn't have a non-random attribute option. So I reworked the 'extra randomness' option. That now just toggles randomness on/off.

Marking this as 'high' as it could have significant impact on a players' campaign if they upgrade a campaign from 50.05 or 50.06 to 50.07.


Scoppio (approved):
Could add a type of point-buy in a next iteration, and allow players to set the base point-buy stack in the campaign options. This can help with the `all random` stuff

IllianiBird:
That'll be coming in 50.07/50.08, with the new company generator. But yeah, definitely on my radar.

PR #6846, opened 2025-05-02
Improvement: Improved Speed of Log Entry Saving in Large Campaigns

IllianiBird:
### Dev Notes
While profiling a user save I noticed that new day advancement was taking far longer than expected. Checking the results I found a significant portion of the process was being taken up by Save on New Day. Digging deeper I found the culprit to be the saving of log entries. This is especially true for generational campaigns where characters can accumulate 50+ years worth of log entries.

I moved the saving of log entries from using an escaping method to instead use CDATA (which removes the need for escaping). This changed the process from 1,183ms/day to 570ms/day with no loss of security.

The string I used to test whether the CDATA worked was `5 < 10 & x > 3; description: "The user's input: <Hello & Goodbye> & <World>"`

Tex, I'm going to tag you for review as you're the one who brought CDATA to my attention and know far more about it than I.


MekHQ/src/mekhq/campaign/log/LogEntry.java
```diff
@@ -83,18 +83,16 @@ public void setType(LogEntryType type) {
...
         if (date != null) {
-            sb.append("<date>").append(date).append("</date>");
+            printWriter.println(MHQXMLUtility.indentStr(indent + 1) + "<date><![CDATA[" + date + "]]></date>");
         }
-        sb.append("<desc>").append(MHQXMLUtility.escape(desc)).append("</desc>");
+        printWriter.println(MHQXMLUtility.indentStr(indent + 1) + "<desc><![CDATA[" + desc + "]]></desc>");
         if (type != null) {
-            sb.append("<type>").append(MHQXMLUtility.escape(type.toString())).append("</type>");
+            printWriter.println(MHQXMLUtility.indentStr(indent + 1) + "<type><![CDATA[" + type + "]]></type>");
         }
```
rjhancock, line 94: Looks good but unless it's actually needed on `date` and `type` (as in internally they aren't some other type other than string or an ISO 8601 Date), then don't need CDATA on them.

But personally all String based entries SHOULD be "escaped" with `CDATA`.

IllianiBird: I figured I'd include them as they're strings and the extra time taken to CDATA them was negligible so figured I might as well.

PR #6857, opened 2025-05-03
Improvement: Switched Faction Logos to Use Same Logos as Campaign Options IIC

IllianiBird:
Previously I set us up to use the colored icons found in the force icon section of data images. However those icons were designed for force icons and ended up with some really weird scaling.

This PR moves us to use the same icons that we're already using in Campaign Options IIC. This has a few benefits: more consistent scaling, and a unity of design. A player will see the faction icon for Clan Wolf in CO IIC and then, when they encounter Clan Wolf in the field, will see the same icon again.

This PR is required for #6858 and I would like to see both merged into 50.06.


MekHQ/src/mekhq/campaign/universe/Factions.java
```diff
@@ -242,214 +259,129 @@ public static Factions load(String factionsPath)
...
+            case "EF" -> "logo_mercenaries"; // Elysian Fields
+            case "GV" -> "logo_mercenaries"; // Greater Valkyrate
+            case "JF" -> "logo_mercenaries"; // JarnFolk
+            case "MSC" -> "logo_mercenaries"; // Marik-Stewart Commonwealth
+            case "OP" -> "logo_mercenaries"; // Oriente Protectorate
+            case "RA" -> "logo_mercenaries"; // Raven Alliance
+            case "RCM" -> "logo_mercenaries"; // Rim Commonality
+            case "NIOPS" -> "logo_mercenaries"; // Niops Association
+            case "AXP" -> "logo_mercenaries"; // Axumite Providence
+            case "NDC" -> "logo_mercenaries"; // New Delphi Compact
```
IllianiBird, line 377: I'm currently working with an artist to add art for the missing icons, but for now we use mercenaries as a fallback.

Scoppio (approved):

PR #6868, opened 2025-05-04
Fix for Unit Sale Costs being generally incorrect - Mekhq part

savanik:
Updated functions in TankLocation.java to correct sale price. Location costs now calculate for internal structure / controls. Still needs updates for chassis costs per support vehicles in Tech Manual p280, maybe?


savanik:
Updated the commit comments to properly reflect what was being updated. No other changes.

PR #6883, opened 2025-05-06
Fix/acar npe on resolution

Scoppio:
# What it does?

It refactors one function that resets crew hit status in order to be simpler to check what it should do, it is also clear in what should throw exceptions, and also made the exception easier to read and with more information to the investigative person.

### DIFF
--- a/MekHQ/src/mekhq/MekHQ.java
+++ b/MekHQ/src/mekhq/MekHQ.java
@@ -45,8 +45,7 @@ import java.io.FileNotFoundException;
 import java.io.ObjectInputFilter.Config;
 import java.lang.management.ManagementFactory;
 import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.Objects;
 import javax.swing.InputMap;
 import javax.swing.JOptionPane;
 import javax.swing.KeyStroke;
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/MekHQ.java
```diff
@@ -821,59 +821,39 @@ public void autoResolveConcluded(AutoResolveConcludedEvent autoResolveConcludedE
...
         }
     }
 
+    private void resetPersonsHits(ResolveScenarioTracker tracker) {
+        var peopleStatus = tracker.getPeopleStatus();
+        Objects.requireNonNull(peopleStatus, "getPeopleStatus() returned null");
+        Objects.requireNonNull(getCampaign(), "getCampaign() returned null");
+        for (var entry : peopleStatus.entrySet()) {
+            Person person = getCampaign().getPerson(entry.getKey());
+            Objects.requireNonNull(person, "getPerson() returned null for Person ID=" + entry.getKey());
```
IllianiBird, line 853: What happens if we hit a null person? Will that obstruct progression or will it just throw an error and then carry on?

Scoppio: It will abort everything, the original idea is that breaking one character or many characters is the same problem so it exits early.

But reflecting on this... yeah, make it run it fully and collect all exceptions then re-raising the exception is better than exploding on the first.
Thanks

IllianiBird (approved):

PR #6892, opened 2025-05-06
Improvement: Added Alphabetical Sorting of Skills

IllianiBird:
The alphabetical sorting of skills is a bit of a double-edged sword. Veteran players hate it because they're used to how things are, newer players love it because it makes finding individual skills easier.

Previously we removed this functionality, but honestly I don't think we should have. With the new changes coming for the Person Pane having skills sorted alphabetical makes more sense than ever. It also means that we, as developers, don't need to maintain an unspoken order whenever new skills are added - and we've added quite a few.


HammerGS:
Alphabetical is kinda the human standard.

Scoppio (approved):

PR #6893, opened 2025-05-06
fix: Armor will now show in its name if its Clan or IS, allowing for warehouse automation to work properly

Scoppio:
# What this does?

Armor will now show in its name if its Clan or IS, allowing for warehouse automation to work properly. The armor pieces origin faction only matter for construction, and by extension they matter for repair and stock on MekHQ.

Before:
Screenshot 2025-05-06 at 20 48 25

After:
Screenshot 2025-05-06 at 20 34 17

This enables the parts in use report to show IS and Clan armor individually, however this also shows that we need to fix that in the Parts in use report itself, it needs to have the faction (IS/Clan) exposed and being used as part of a composed index.


IllianiBird (approved):
To think it was this simple

PR #6895, opened 2025-05-07
Improvement: Deprecate Archery

IllianiBird:
I've decided to deprecate archery as it's an example of a specialized weapon skill and we don't support combat to that scale, nor have we included any of the other specialized weapon skills such as Melee Weapons or Throwing Weapons.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/skills/SkillDeprecationTool.java
+++ b/MekHQ/src/mekhq/campaign/personnel/skills/SkillDeprecationTool.java
@@ -28,6 +28,7 @@
 package mekhq.campaign.personnel.skills;
 
 import static java.lang.Math.round;
+import static mekhq.campaign.personnel.skills.SkillType.S_ARCHERY;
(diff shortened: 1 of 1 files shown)


Scoppio:
maybe just keep those around so people can have the scorekeeping for their ATOW character sheets?

Scoppio (approved):
I understand all had that S_* in the beginning, but it really makes no sense to have that S in the name of the skills.

IllianiBird:
So I thought about that - and that's what stayed my hand - but it's just going to open the door for people requesting we use the correct skills at an MM-level and that's just going to open up a while can of worms I don't think any of us are interested in dealing with.

If you disagree, though, I can add the remaining specialist weapon skills to my other missing skill PR: Melee Weapons, Throwing Weapons, and Support Weapons.

Scoppio:
I disagree, but lets review this decision at a later date

IllianiBird:
Closing this following discussion in the developer channel.

PR #6896, opened 2025-05-07
Improvement: Added Additional Testing for Skills

IllianiBird:
With the amount of new skills that have been recently added I'm really aware of how fragile our skill addition is. It's a lot of manual entry with a fair amount of stuff that can go wrong.

With a mind to making the process more secure I went ahead and added a handful of tests. These tests will run on every skill type and ensure full skill validity.


Scoppio (approved):

IllianiBird:
Thank you Past Illiani for adding these tests. They proved _immediately_ useful when I accidentally deleted two skills and we'd have likely not noticed for months.

PR #6900, opened 2025-05-07
Improvement: Added GM Option to Remove All Roleplay Skills From A Character

IllianiBird:
Not everyone is a fan of roleplay skills, but roleplay skill generation is enabled by default in most stock presets. This adds a GM option to the personnel right-click menu that allows players to remove all currently generated roleplay skills from a character. It supports selecting multiple characters at once.

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
@@ -238,6 +238,7 @@ public class PersonnelTableMouseAdapter extends JPopupMenuAdapter {
     private static final String CMD_PERSONALITY = "PERSONALITY";
     private static final String CMD_ADD_RANDOM_ABILITY = "ADD_RANDOM_ABILITY";
     private static final String CMD_GENERATE_ROLEPLAY_SKILLS = "GENERATE_ROLEPLAY_SKILLS";
+    private static final String CMD_REMOVE_ROLEPLAY_SKILLS = "REMOVE_ROLEPLAY_SKILLS";
     private static final String CMD_GENERATE_ROLEPLAY_ATTRIBUTES = "GENERATE_ROLEPLAY_ATTRIBUTES";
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -1507,6 +1508,20 @@ public void actionPerformed(ActionEvent action) {
...
+                    List<Skill> allSkills = new ArrayList<>(person.getSkills().getSkills());
+                    for (Skill skill : allSkills) {
+                        SkillType skillType = skill.getType();
+
+                        if (skillType.isRoleplaySkill()) {
+                            person.removeSkill(skillType.getName());
+                        }
+                    }
+                }
+                break;
```
Scoppio, line 1523: This could be moved to `Person` then all you have to do is
```java
people.forEach(Person::removeRolePlaySkills);
```

Scoppio (approved):

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
IllianiBird: While true, I can't think of any reason why we would need this method outside of the personnel context menu.

Scoppio: Because the object should be the one dealing with its internal implementations. If you change the Skills to be an enum or not to have the second layer of `getStkill().getSkill()` then you suddenly dont have to change code just inside the Skill and Person, now you have to change code in the GM command, and then you will have to change field in the parser, and maybe in the persisting process, etc... A object that is not just a data transfer object, should be the one responsible of doing any kind of "work" with its internals, this makes easier to do refactors later, or create extensions, extracting interfaces, etc. Today, if you want to check if the GM Command works with a unit test you have to mock person and the getSkills response.

IllianiBird: That seems reasonable

PR #6914, opened 2025-05-07
Improvement: Introduced Profession SubTypes for Easier Categorization

IllianiBird:
Previously we were using a boolean to determine whether a profession was Combat or Not Combat; and then doing some additional checks to determine if the profession is Not Combat and Also Civilian.

Now we just use an enum with three entries: Combat, Support, and Civilian.

This is prep work for a minor GUI adjustment coming separately.


HammerGS:
@IllianiBird 

The failure occurred in the PersonnelRoleTest.testIsSupport() unit test. The logs indicate that an AssertionFailedError occurred at PersonnelRoleTest.java:560, but I was unable to retrieve the PersonnelRoleTest.java file for further analysis. However, I can provide guidance based on the context available.

Analysis
Unit Test Failure:

The testIsSupport() method failed, likely due to an assertion mismatch.
The error in the logs does not provide the specific assertion that failed, so it is necessary to inspect the test code to identify the issue.
Workflow Observations:

The workflow checks out specific branches from related repositories (MegaMek and MegaMekLab), falling back to master if the branch does not exist.
There might be a misalignment between the versions of MegaMek, MegaMekLab, and MekHQ used in the test.
Suggested Solution
Retrieve the PersonnelRoleTest.java file and check the logic in the testIsSupport() method, particularly at or around line 560.
Ensure that the test aligns with the current functionality of the dependent code.
Double-check the configurations and compatibility of related repositories (MegaMek and MegaMekLab) by verifying the branches and commits being used.

PR #6930, opened 2025-05-09
Fix: Stopped Personnel From Skipping Their Paycheck

IllianiBird:
Due to the way we were fetching personnel for salary calculations anyone who didn't have a status of 'active' was not receiving a salary. Now personnel will more correctly receive salaries so long as they have a status where it would be logical for them to do so.


psikomonkie:
Wait, we don’t pay people that’re sent away for school? That’s silly, if my employer is sending me for education they better still be paying me. Is this functionality firm or open to being changed?

IllianiBird:
Very open to change. I’ll see it done

IllianiBird:
- Updated so that all students get a paycheck not just homeschooled ones

IllianiBird:
- Updated to include other statuses that should be eligible to still receive a salary

PR #6938, opened 2025-05-09
feat: add actual board for acar

Scoppio:


IllianiBird:
As this is marked for 50.07 I'm gonna flip it to draft so there's no risk of it being accidentally merged

MekHQ/src/mekhq/utilities/ScenarioUtils.java
```diff
@@ -0,0 +1,155 @@
...
+ * @author Luana Coppio
+ */
+public class ScenarioUtils {
+
+    private static final MMLogger LOGGER = MMLogger.create(ScenarioUtils.class);
+
+    private ScenarioUtils() {}
+
+    /**
+     * Creates a game board based on the settings in the provided Scenario.
```
IllianiBird, line 58: This is cool, I'm guessing this will allow you to better track the location of units for things like range and closing? Plus planetary effects?

PR #6943, opened 2025-05-09
Improvement: #6867 Added Weapons as a separate autologistics category

### LINKED ISSUE #6867: [RFE] Add a weapons category to AutoLogistics and improve default settings values
### Enhancement Type *

With current default settings AutoLogistics doesn't stock any weapons at all. This looks like an oversight, as this is equipment that may be required for repairs relatively frequently. There is also no separate setting to define how many weapons need to be restocked. Right now they fall into "other" category (which is 0% by default), along with some equipment that may not be needed in similar quantities. I suggest to:
- Separate weapons into it's own category in AutoLogistics settings
- Tweak default AutoLogistics settings values to include weapons (i'm not experienced enough to suggest exact value)
- Also maybe check what else falls into "Other" category right now, and review if it's reasonable to keep default setting at 0% (i believe it has some equipment that may be needed for repairs sometimes, but not very often)

### Use Cases or Rationale

AutoLogistics is a great feature that can reduce micromanagment and make MekHQ more enjoyable for a lot of players. Most people probably won't look into tweaking its settings too much, since the point of reducing micromanagement is to not force players to bother with the details too much. Newer players especially probably just gonna assume that AutoLogistics can take bulk of parts management for them, as it kind of supposed to. 
For these reasons, it's important to make sure that default settings are good enough for an average player. Right now they are not quite good enough, and could easily lead new players into a bad spot in their first campaign.

### Attach Files

N/A

### MekHQ Suite Version *

0.50.05

### Final Verification

psikomonkie:
Fixes #6867

Weapons on units were part of the "Other" category. A player wanting to keep a stock of backup weapons would need to manually change the value for each weapon they use,  or change the default for "other". This adds "Weapons" as its own category of autologistics.


psikomonkie:
Should infantry small arms be excluded? Currently they're counted as weapons and will be restocked with this change. Should they stay as other?

IllianiBird:
Infantry equipment *should* be already excluded. If it isn't, it should be.

psikomonkie:
> Infantry equipment _should_ be already excluded. If it isn't, it should be.

Stuff like their basic rifles, yeah. It’s being included now as a weapon. I’ll make an update to this to keep them as ‘other’ & add a test for that

psikomonkie:
All's good regarding infantry weapons. They only showed on the table because I had some in the warehouse - it's an old save :) This is ready for review.

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -5706,6 +5717,8 @@ public static CampaignOptions generateCampaignOptionsFromXml(Node wn, Version ve
                     retVal.autoLogisticsJumpJets = Integer.parseInt(wn2.getTextContent().trim());
                 } else if (nodeName.equalsIgnoreCase("autoLogisticsEngines")) {
                     retVal.autoLogisticsEngines = Integer.parseInt(wn2.getTextContent().trim());
+                } else if (nodeName.equalsIgnoreCase("autoLogisticsWeapons")) {
```
IllianiBird, line 5720: Let's change these to use `MathUtility.parseInt(wn2.getTextContent().trim())`. That has built in error handling in the event we, somehow, pass something that isn't an int into this method.

If you're feeling fancy, `parseInt` can accept a second param that will be the value used in the event `parseInt` fails, no param means `0`, but you could put the default value for each of the autoLogi options.

Again, only if you're feeling fancy.

IllianiBird (approved):
One small point of feedback, but not a deal breaker

MekHQ/src/mekhq/campaign/CampaignOptions.java
psikomonkie: Sounds good! Love easy improvements like that. 

I won’t be able to make the change until later today, please don’t merge this until this comment chain’s resolved. 

Thanks!

psikomonkie: Done!

MekHQ/unittests/mekhq/campaign/CampaignTest.java
```diff
@@ -813,6 +832,37 @@ public void testGetDefaultStockPercentEngines() {
...
+                WeaponType mockWeaponType = mock(WeaponType.class);
+                part = new EquipmentPart(1, mockWeaponType, Entity.LOC_NONE, 1.0, false, campaign);
+
+                // Act
+                try {
+                    initialStockPercent = (int) method.invoke(campaign, part);
+                    initialAllPercents = getAllDefaultStockPercents();
+
+                    // Let's change it and make sure that it uses the new value
+                    when(mockCampaignOptions.getAutoLogisticsWeapons()).thenReturn(DESIRED_STOCK_LEVEL); //TODO
```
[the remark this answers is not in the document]
IllianiBird: Ha, it's nice to see Copilot bullying someone else for a change! :P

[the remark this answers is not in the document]
psikomonkie: No that was actually something I missed. The //TODO tag itself just needed removed. It was originally commented out during the development process. I left the //TODO flag so I'd uncomment it once it's implemented it. Just forgot to remove the todo :clown:

PR #6951, opened 2025-05-10
Fix: Fixed Unclear Wording in Procurement when Out of Acquisition Attempts

IllianiBird:
The procurement error display for when a campaign has no procurement personnel is reused when the campaign's procurement personnel have run out of acquisitions for the period. Specifically, it stated that nobody in the campaign was capable of acquiring parts. Which is factually correct, however gives the impression that there are no acquisitions personnel, as opposed to them just being out of acquisition attempts.

I reworded this message to specifically state that the personnel are out of acquisitions. The message is still reused if there are no acquisitions personnel in the campaign, but even then it's factually correct and far less likely to cause confusion than the original message.

In the land of wishes and fairies we'd rework this entire system, but that's out of scope for this PR.


Sleet01 (approved):
LGTM

PR #6960, opened 2025-05-11
Fix: Added Clan Labels and Descriptions to Profession Descriptions and Labels

IllianiBird:
Profession labels were always using IS terms. Now they don't. For example the BA trooper profession will show as Elemental.


Sleet01 (approved):
LGTM!

PR #6962, opened 2025-05-12
Fix: Dependents Now Correctly Contribute to Administration Requirements when Calculating Unit Reputation

IllianiBird:
Thanks to @Saklad5 for pointing this out.

CamOps considers all non-techs, non-combatants to be 'admin personnel' for the purposes of Unit Reputation. MekHQ, unlike CamOps, has distinct admin roles.

When writing the CamOps Reputation my eyes slipped over the importance of the wording in the below section:

image

The paragraph clearly shows that Dependents (which would cover things like 'lawyers' and 'bean counters') should count towards the unit reputation. At least in their current iteration. In 50.07 we will probably need to adjust this to exclude 'unemployed' dependents, but otherwise this will be a significant boost towards making dependents feel like they have a use.

While I was at it I closed a couple of bugs where we were counting prisoners for some roles.

### DIFF
--- a/MekHQ/src/mekhq/campaign/rating/CamOpsReputation/AverageExperienceRating.java
+++ b/MekHQ/src/mekhq/campaign/rating/CamOpsReputation/AverageExperienceRating.java
@@ -114,7 +114,7 @@ public class AverageExperienceRating {
         int personnelCount = 0;
         double totalExperience = 0.0;
 
-        for (Person person : campaign.getActivePersonnel(true)) {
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM

Saklad5:
Are Doctors being counted as Support Vehicle crew right now? The ones that operate in Medical Teams normally would be, so it's important not to double-count.

IllianiBird:
They have been for a while, maybe 3+ versions, you just need to give them the Vehicle Crewmember role (secondary or primary)

PR #6963, opened 2025-05-12
Improvement: Rolled Back Renaming of 'Dependent' Role to 'Unemployed'

IllianiBird:
'Unemployed' is such a weighted term, emotionally, and would undermine our effects to make players more attached to their civilians.


Sleet01 (approved):
LGTM

PR #6968, opened 2025-05-13
Improvement: Added Hint Panel to TO&E Tab; Updated Glossary With New TO&E-Related Entries

IllianiBird:
image

This adds a panel to the bottom of the TO&E tab that includes a key for the informatics we use in the TO&E as well as hyperlinked terms.

These links bring up the Glossary, allowing the user instance access to information about the TO&E itself, combat teams, force types, resupplies, a whole bunch of other stuff. If this proves successful I'm inclined to add something like this to the other tabs.

While I was at it I did a grammar sweep of the glossary, added a couple of extra entries, and renamed 'standard' forces to 'combat' as the standard term was causing confusion.

image

image


Sleet01 (approved):
LGTM

PR #6969, opened 2025-05-13
Improvement: Added Scenario Briefing Dialog to Briefing Room on Scenario Start

IllianiBird:
When the player selects 'Start Game' in the Briefing Room instead of going straight to the connection dialog they are presented with the scenario briefing.

The briefing will be delivered by a representative of the contact employer. This representative will deliver all briefings for the duration of the contract, becoming a familiar name the player can latch onto.

The briefing will be delivered to the most senior member of the drop.

~~This feature is for Legacy AtB and StratCon enabled campaigns only due to a reliance on code that is exclusive to the Digital GMs.~~

The persistent employer representative feature is available for Legacy AtB and StratCon enabled campaigns only.

image


IllianiBird:
- Updated to include code that means this isn't restricted to AtB and StratCon only

Sleet01 (approved):
LGTM!

PR #6973, opened 2025-05-13
Improvement: Changed Logic for Mass Repair Mass Salvage

Dark-Hobbit:
How it works now:
1. You have "Preferred TN" and "Max TN" settings. "Preferred TN" must be equal or lower than the "Max TN".
2. No job gets carried out if it is harder than "Max TN".
3. If  "Use Extra Time" option is enabled and task's difficulty is above the "Preferred TN", it will try to lower the difficulty by increasing the time spent. If it cannot reach teh "Preferred TN" even with highest extra time settings, the job can still get carried out. This allows jobs to execute even if they were originally above the "Max TN", as long as the adjusted TN does not exceed it. 
4. If "Use Rush Job" option is enabled and the task's difficulty is below the "Preferred TN", it will likewise try to increase the difficulty in order to save the time under the same rules as with "Use Extra Time"
5. Default settings are: "Use Extra Time" and "User Rush Job" both on, "Preferred TN" is equal to 4, "Max TN" is equal to 6.

image


IllianiBird:
As this is marked for New Dev Cycle I'm going to go ahead and switch it to Draft so there is no risk of it being accidentally merged.

IllianiBird:
Bring this out of draft at your convenience. Let's get this merged! :)

Dark-Hobbit:
> Can we remove the rush jobs being on by default?

Could you please give some examples (preferably with numbers) when it would be preferential? Right now using rush jobs are essentially negative levels of extra time with no other side effects. @cooshoo

rjhancock:
Is this ready for merge?

psikomonkie (approved):
I love this! One comment that applies a few times about methods being deprecating, but that's minor. 

I especially appreciate your improvements to the debug logging for this & updating the unit tests properly. I hope both of these will prevent MRMS from being as opaque and confusing as it was.

I'm looking forward to using this revised MRMS in my campaign!

Dark-Hobbit:
> Is this ready for merge?

Yep, ready to merge.

PR #6974, opened 2025-05-13
Fix: Added Missing Commander Address from Convoy Roleplay Events

IllianiBird:
We were missing the commander's address from convoy roleplay events. Now we're not.


Sleet01 (approved):
LGTM

PR #6984, opened 2025-05-13
Improvement: Updated Life Events Options to More Clearly Indicate That the Options are for RP Only; Updated Ronin Events to More Clearly Show Results of Each Option

IllianiBird:
We received feedback that it wasn't clear what options had an effect and what were just for roleplay. This addresses that.


psikomonkie (approved):
Nice! To be entirely honest I was really unsure what most of the options meant on things like the Birth Announcement & Coming of Age Announcement. It's good to know that they're just for fun. 

My heart is also wishing these were tool-tips and not something that always displayed but I don't know if it's possible and don't need it done now.

IllianiBird:
It's entirely possible and Immersive Dialog is already set up to do that. However, when we were developing Prisoner Events the general feedback was that critical information shouldn't be hidden in tooltips as they'll just be missed by players. So we're using the low-tech solution here.

PR #6992, opened 2025-05-14
Fix: Restored Missing Character Profession in Person View

IllianiBird:
Apparently the section that displays what profession/s a character has got removed during my rework of the Person View panel. This restores it.


Sleet01 (approved):
LGTM

PR #6993, opened 2025-05-14
Feature: Added Glossary Button to Campaign GUI to Allow Users to Quickly Access the Glossary

IllianiBird:
We've had a glossary in MekHQ for a few versions now, but largely its entries were hidden in nag dialogs and similar. Last night, as I was asleep the Good Ideas Fairy came to me in a dream and said "Illiani, why the heck do we have a Glossary the users can't access?"

And you know what? My nocturnal dream walker was right, why do we? No good reason, so now we have a glossary button.

https://github.com/user-attachments/assets/0f9e764e-e6eb-4e51-89f2-64fe774aa6de

This is a super basic implementation but with the expectation that 50.06 is going to be a Milestone (and therefore used for a while) it made sense to whip something up nice and quick and then make it pretty later.


MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1156,10 +1157,23 @@ private void initTopButtons() {
         gridBagConstraints.insets = new Insets(3, 10, 3, 3);
         btnPanel.add(lblLocation, gridBagConstraints);
 
-        btnAdvanceMultipleDays.addActionListener(e -> new AdvanceDaysDialog(getFrame(), this).setVisible(true));
```
IllianiBird, line 1159: I don't know why GitHub is saying this line was edited, because it wasn't. And unlike what GitHub would have you believe I didn't delete `btnAdvanceMultipleDays` you can see it on line 1173

Sleet01 (approved):
LGTM

Sleet01:
@IllianiBird The one and only conflict so far.

PR #6994, opened 2025-05-14
Fix: Added Missing Campaign Options Iconography to New Weapons Restock Percentage Option; Changed Color of Iconography to Stand Out

IllianiBird:
image


Sleet01 (approved):
LGTM

PR #7009, opened 2025-05-15
Fix: #7007 Fixed Cancel Deployment Producing Scenario with 0 BV Budget

### LINKED ISSUE #7007: [Issue] 0.50.06-nightly-2025-05-15 - Manual deployment of force to attack facility on integrated contract has seed BV of 0 in log
### Brief Description *

I deployed my command lance to attack a facility on the stratcon map during this integrated contract, and it generated an abnormally low BV opfor. When I checked the log, it says the seed force BV is 0.

Attaching save files from before and after deployment.

### 3. Steps to Reproduce

Deploy force to facility on hex 0004

### Attach Files

[0bv Seed Force.zip](https://github.com/user-attachments/files/20231857/0bv.Seed.Force.zip)

### Severity *

### MekHQ Suite Version *

0.50.06-nightly-2025-05-15

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.6

### Final Verification

IllianiBird:
Fix #7007

When a player deploys to a hidden scenario we generate the scenario, including the BV budget of forces assigned to that scenario. However, if the player cancels deployment - at that moment - the budget was getting set to 0 because we were not correctly resetting the scenario. This was due to the conditional that checked for the deployment being canceled was not functioning fully. However, as the force didn't get deployed it gave the impression that it _was_ fully functional.

We fixed this by added a new variable to the deployment wizard that explicitly states whether the deployment was canceled. If that event comes to pass the scenario is reset fully. However, the originally generated force is not, preventing the 0 BV bug.

### Design Notes
The cancel deployment bug is a hacky solution to a UX problem and one we should really rethink. However, such a task isn't on my schedule so would need to be picked up by another developer. Basically the problem we have to solve is that players sometimes want to backtrack out of deployment. However, if the player force is 'ambushed' (i.e. the scenario is populated at point of deployment) the player should - at that point - be barred from canceling the deployment as their force has been ambushed.

### DIFF
--- a/MekHQ/src/mekhq/gui/StratconPanel.java
+++ b/MekHQ/src/mekhq/gui/StratconPanel.java
@@ -1098,8 +1098,12 @@ public class StratconPanel extends JPanel implements ActionListener {
                     scenarioWizard.toFront();
                     scenarioWizard.setVisible(true);
                 }
-                if (selectedScenario != null && !isCommitForces()) {
+                if (selectedScenario != null && scenarioWizard.isWasCanceled()) {
+                    logger.info("Scenario wizard was cancelled. Resetting scenario to default state.");
                     selectedScenario.resetScenario(campaign);
+
+                    // We currently retain the wizard in memory, so need to make sure we reset the canceled state
+                    scenarioWizard.setWasCanceled(false);
                 }
 
                 setCommitForces(false);
--- a/MekHQ/src/mekhq/gui/stratcon/StratconScenarioWizard.java
+++ b/MekHQ/src/mekhq/gui/stratcon/StratconScenarioWizard.java
@@ -24,6 +24,11 @@
  *
  * Catalyst Game Labs and the Catalyst Game Labs logo are trademarks of
  * InMediaRes Productions, LLC.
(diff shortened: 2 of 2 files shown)


Sleet01:
Does the system support a two-step process:
1. a scenario is _generated_, but not added to the mission, when the player starts the deployment.
2. the generated scenario is added to the mission _if and only if_ the player does not cancel
or something?

IIRC this is how the MegaMek unit configuration dialog works: the player's changes are recorded in the dialog state itself, but only applied to the unit(s) if the player clicks "OK".

Sleet01 (approved):
LGTM

IllianiBird:
The Scenario is added to the Mission the moment it's spawned. The issue was around the BV Budget allocated to forces within that scenario.

What we would need to do, to solve the 'backing out of an ambush' problem is have this dialog predate the spawning of the scenario but with sufficient knowledge to know that a scenario *has* been spawned. The problem is that, insofar as I can see, StratCon was never built with this interaction in mind.

PR #7010, opened 2025-05-15
Fix: Prevented Unscrupulous from Performing Extreme Expenditure on Prisoners to Rob Them of Their Wealth Trait

IllianiBird:
Added checks to ensure characters are both Active and Free (or Bondsman) before Extreme Expenditure can be used on them.


Sleet01 (approved):
LGTM

PR #7011, opened 2025-05-15
Fix: Fixed Formatting Error in Title of Some Campaign Options Tabs

IllianiBird:
We were missing the html tags from a handful of tab headers. Now we're not.


Sleet01 (approved):
LGTM

PR #7013, opened 2025-05-15
Fix: Fixed Age Modifiers Applying Incorrectly for Ages 31-40; Added Unit Tests for Age Modifiers

IllianiBird:
The age modifiers for the 31-40 bracket were incorrectly only applying to some magical person that manages to be both >= 31+ and < 14. Instead of >= 31 and < 41.

Also, we previously didn't have any unit tests for this mechanic. Why I didn't test this I'll never know, but now I have. As I should have done when this was first developed. -_-


Sleet01 (approved):
LGTM

PR #7015, opened 2025-05-15
Fix: Fixed Weight Calculation Visual Bug in StratCon Campaigns

IllianiBird:
We were factoring in the 'Player Vehicles Weigh 50%' Legacy AtB option, even when StratCon is enabled. Now that option is only used if StratCon is disabled.

This is a visual bug only.


Sleet01 (approved):
LGTM

PR #7030, opened 2025-05-16
Updates the dropship acquisition TNs in UnitOrder

savanik:
This should more closely represent the rules in CO 4th p17.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/UnitOrder.java
+++ b/MekHQ/src/mekhq/campaign/unit/UnitOrder.java
@@ -30,10 +30,6 @@ package mekhq.campaign.unit;
 
 import java.io.PrintWriter;
 
-import org.w3c.dom.Element;
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-
 import megamek.common.*;
 import megamek.common.loaders.EntityLoadingException;
 import megamek.logging.MMLogger;
@@ -44,6 +40,9 @@ import mekhq.campaign.parts.Part;
 import mekhq.campaign.personnel.Person;
 import mekhq.campaign.work.IAcquisitionWork;
 import mekhq.utilities.MHQXMLUtility;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 /**
  * We use an extension of unit to create a unit order acquisition work
@@ -114,6 +113,7 @@ public class UnitOrder extends Unit implements IAcquisitionWork {
 
     /**
      * @param quantity - the number of parts of this type
+     *
      * @return a string that gives a grammatical correct name based on the quantity
      */
     @Override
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -219,7 +224,88 @@ public TargetRoll getAllAcquisitionMods() {
...
         // TODO: Fix weight classes
-        // TODO: aero large craft
+        if (entity instanceof Dropship || entity instanceof Jumpship) {
+            //FIXME ?
+            /* In theory, these should be the base target numbers of these vehicles. The base TNs in the book are
+             * static and don't adjust for Admin skill or anything. This doesn't match our usage of skills such as
+             * Administration to find vehicles. Instead, we're going to just kind of assume they 'intended'
+             * a base TN of 7 and adjust for windage.
+             *
+             */
```
Scoppio, line 234: I suggest this be discussed before implemented.

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -219,7 +224,88 @@ public TargetRoll getAllAcquisitionMods() {
...
+                    target.addModifier(3, "Rare");
+                    break;
+                case 4:
+                    target.addModifier(6, "Very Rare");
+                    break;
+                case 5:
+                default:
+                    target.addModifier(10, "Unique");
+
+            }
```
Scoppio, line 270: Another one that needs discussion, coming up with a solution is fine, but any unnoficial value should really be at least talked about with other devs onto why you are thining of adding it this way, and sometimes also it is worth to bring up those as really valid questions on how to most closely to implement the rules as written on MekHQ, alot of contributors can give great advice and sometimes even have access to erratas on how things changed its workings since release/publication

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -219,7 +224,88 @@ public TargetRoll getAllAcquisitionMods() {
...
+                target.addModifier(-6, "Era");
+            } else if (year > 2651) {
+                target.addModifier(-6, "Era");
+            } else if (year > 2571) {
+                target.addModifier(-5, "Era");
+            } else if (year > 2412) {
+                target.addModifier(-4, "Era");
+            } else {
+                target.addModifier(-3, "Era");
+            }
```
Scoppio, line 306:
```java
            int year = getCampaign().getGameYear();
            int eraModifier = getTargetModifierForEra(getCampaign().getGameYear());
            target.addModifier(eraModifier, I18n.getText("unitOrder.eraModifier")); 
            return target;
        }    
        
        private int getTargetModifierForEra(int year) {
            if (year > 3084) {
                modifier = 0;
            } else if (year > 3068) {
                modifier = -2;
            } else if (year > 3050) {
                modifier = -1;
            } else if (year > 2901) {
                modifier = +0;
            } else if (year > 2821) {
                modifier = -2;
            } else if (year > 2751) {
                modifier = -6;
            } else if (year > 2651) {
                modifier = -6;
            } else if (year > 2571) {
                modifier = -5;
            } else if (year > 2412) {
                modifier = -4;
            } else {
                modifier = -3;
            }
            return modifier;
        }
```

Scoppio (changes requested):

Scoppio:
Too many decisions on how to implement book rules were made but the comments seen to imply that there were no discussion around the decisions that were taken

PR #7037, opened 2025-05-17
 Finishing TODO: Conventional Fighter avionics parts cost fix

savanik:
Updated Avionics part with costs from Conventional Fighter table. Costs based on Tech Manual p238.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/Avionics.java
+++ b/MekHQ/src/mekhq/campaign/parts/Avionics.java
@@ -37,17 +37,6 @@ import mekhq.campaign.finances.Money;
 import mekhq.campaign.personnel.skills.SkillType;
 import org.w3c.dom.Node;
 
-import megamek.common.Aero;
-import megamek.common.Compute;
-import megamek.common.CriticalSlot;
-import megamek.common.Dropship;
-import megamek.common.Entity;
-import megamek.common.EquipmentType;
-import megamek.common.IAero;
-import megamek.common.Jumpship;
-import megamek.common.LandAirMek;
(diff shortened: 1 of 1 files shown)


psikomonkie:
PR description has incorrect page citation - TN 283 no 238.  

In addition,TM 277:
> As a conventional fighter, the ’Buster uses 5 tons of avionics equipment (rather than an aerospace fighter cockpit and control systems) with no VSTOL gear (though as a fighter it does include landing gear and attitude thrusters).

ASF simply don’t have avionics. It’s part of the ASF cockpit. Consider adding this page reference too so future contributors don’t wonder why ASF is missing from the table on 283 and therefore return 0 in the method.

Sleet01 (approved):
LGTM

Sleet01:
@psikomonkie @savanik Any objection to merging this to get it in the nightly for the last round of testing? 
I would say that anyone following up would be able to get here from the Git history.

psikomonkie:
> @psikomonkie @savanik Any objection to merging this to get it in the nightly for the last round of testing? I would say that anyone following up would be able to get here from the Git history.

I think I’m okay with this. Are we sure this isn’t overvaluing CFs by the Avionics $$? that’s my only concern.

Sleet01:
> > @psikomonkie @savanik Any objection to merging this to get it in the nightly for the last round of testing? I would say that anyone following up would be able to get here from the Git history.
> 
> I think I’m okay with this. Are we sure this isn’t overvaluing CFs by the Avionics $$? that’s my only concern.

It's correct as far as I can tell; it looks like ASFs pay for their Avionics through the Cockpit systems.

PR #7038, opened 2025-05-17
 Finishing TODO: Remove inline file path

savanik:
Updated CurrencyManager to remove inline file path and move reference to MHQConstants.


Scoppio (approved):

Scoppio:
I think your VS Code is set to reformat code on save automatically (which explain your single-line edits comming up with dozens of lines changed.

IllianiBird:
Ah, and if you're using VS Code and not IntelliJ IDEA it won't be formatting according to our style guide, which would explain why it's hitting files I'm fairly certain were set up according to CE.

Scoppio:
> Ah, and if you're using VS Code and not IntelliJ IDEA it won't be formatting according to our style guide, which would explain why it's hitting files I'm fairly certain were set up according to CE.

Maybe VS Code is more compliant with editor config than intelliJ? I recall most of the problem we had with it was making intelliJ comply... dont really remember the conversations we had.

PR #7042, opened 2025-05-18
Fix: Fixed Commander's Day Suppression

IllianiBird:
We weren't checking whether celebration announcements were enabled when triggering Commander's Day.


Sleet01 (approved):
LGTM

PR #7043, opened 2025-05-18
Leases!

savanik:
Needs heavy review. Would like assistance writing units tests for (at least some) parts of it.

Design document can be found here: https://tinyurl.com/4erz5r9y


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -4490,8 +4503,7 @@
                   getCampaign().getCampaignOptions().isUseInitiativeBonus()) {
             // Tactics command bonus. This should actually reflect the unit's commander
             if (null != commander && commander.hasSkill(SkillType.S_TACTICS)) {
-                entity.getCrew()
-                      .setCommandBonus(commander.getSkill(SkillType.S_TACTICS).getTotalSkillLevel());
+                entity.getCrew().setCommandBonus(commander.getSkill(SkillType.S_TACTICS).getTotalSkillLevel());
```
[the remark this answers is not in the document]
savanik: I didn't change that - that's just a reformat.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@
...
+        getCampaign().removeUnit(unit.getId());
+    }
+
+    /**
+     * Leases a unit.
+     *
+     * @param en   An entity representing the unit to be leased.
+     * @param days How many days it will take to arrive.
+     **/
+    public boolean LeaseUnit(Entity en, int days) {
```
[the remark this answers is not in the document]
Scoppio: Code QL is right

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@ public boolean buyUnit(Entity en, int days) {
...
+
+        if (getCampaignOptions().isTrackLeases()) {
+            Unit unit = new Unit(en, getCampaign());
+            unit.addLease(new Lease(getCampaign().getLocalDate(), unit));
+            getCampaign().addNewUnit(en, false, days, quality);
+            return true;
+        } else {
+            getCampaign().addNewUnit(en, false, days, quality);
+            return true;
+        }
```
Scoppio, line 638: Here is a mistake due to how things are implemented.
"trackOverhead" which has a variable named "payOverhead" means that overhead is something that is constantly being calculated, but you dont pay and you dont track it on finances unless you enable the variable "payOverhead".

Here you created the "trackLeases" which is enabled when the variable "payLeases" is enabled, but differently from overheaa which is an extra expense that you can choose to pay or not, leases is a new syste, one which you either use it or you dont have access to it.

The way the function is currently set means "if I turn off the track leases, this means that I can get units for free".

When you have a system like this one disabled, it should not have a "fallback", it should simply not happen, should not be called, and calling it while it is disabled should be an error.
And in both cases it is returning "true" for no reason at all since there is no situation where it would ever return false.
 This should not return anything since the return doesnt mean anything.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@ public boolean buyUnit(Entity en, int days) {
...
+
+    public void cancelUnitLease(Unit unit) {
+        Objects.requireNonNull(unit);
+        Money lastMonthLease = unit.getUnitLease().getFinalLeaseCost(getCampaign().getLocalDate());
+
+        getCampaign().getFinances()
+              .debit(TransactionType.UNIT_CANCEL_LEASE,
+                    getCampaign().getLocalDate(),
+                    lastMonthLease,
+                    "Final Monthly Lease of " + unit.getName());
```
Scoppio, line 649: You are requiring the unit not null, but you are not requiring the unit to have a lease when you are getting it, you are just assuming that the lease is there, but the lease can be a null variable.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -697,14 +787,18 @@ public void sellArmor(Armor armor, int points) {
 
         Money cost = armor.getActualValue().multipliedBy(saleProportion);
 
-        getCampaign().getFinances().credit(TransactionType.EQUIPMENT_SALE, getCampaign().getLocalDate(),
-                cost, "Sale of " + points + " " + armor.getName());
+        getCampaign().getFinances()
+              .credit(TransactionType.EQUIPMENT_SALE,
+                    getCampaign().getLocalDate(),
+                    cost,
+                    "Sale of " + points + " " + armor.getName());
```
Scoppio, line 794: You really should not be reformatting the whole file while working on it. Keep reformats to only the code you are working on.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -107,6 +107,16 @@ public Money getMaintenanceCosts() {
         return Money.zero();
     }
 
+    public Money getLeaseCosts() {
+        if (getCampaignOptions().isTrackLeases()) {
+            return getHangar().getUnitsStream()
+                         .filter(u -> u.hasLease())
```
Scoppio, line 113:
```suggestion
                         .filter(Unit::hasLease)
```

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (c) 2009 Jay Lawson (jaylawson39 at yahoo.com). All rights reserved.
+ * Copyright (C) 2013-2025 The MegaMek Team. All Rights Reserved.
```
Scoppio, line 3: Im pretty sure Jay lawson has not worked on this file and that it has not existed before 2025.
Remove the line of copyright for Jay Lawson and the copyright line need to contain the actual year span of existance of this file (so, only 2025)

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
...
+import megamek.common.Entity;
+import megamek.logging.MMLogger;
+import mekhq.campaign.unit.Unit;
+import mekhq.campaign.unit.UnitOrder;
+import mekhq.utilities.MHQXMLUtility;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+public class Lease {
+    private static final MMLogger logger = MMLogger.create(UnitOrder.class);
```
Scoppio, line 43: logger is static, should be upper case LOGGER, the MMLogger.create should receive `Lease.class`

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
...
+    private Money leaseCost;
+    private LocalDate acquisitionDate;
+
+    /**
+     * Leases are nominally attached to units while they are in the hanger.
+     */
+    public Lease(LocalDate currentDay, Unit unit) {
+        acquisitionDate = currentDay;
+        // Campaign Operations 4th., p43
+        leaseCost = unit.getSellValue().multipliedBy(0.005);
```
Scoppio, line 54: Magic number need to be a constant.
It's also important to put somewhere the limit to what types of units are available for lease, as the leasing for transport actually only applies for dropships and transport ships, not for meks, ASF, etc.

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Gets the lease cost, for the accountant. Lease cost is prorated for the first month, so we need to check if
+     * yesterday was the first month. Should only be called on the 1st. It's also possible we get a lease object before
+     * the lease actually starts, if it takes a while to deliver the unit. Cost will be zero in this case.
+     *
+     * @param time The current campaign LocalDate
+     */
+    public Money getLeaseCostNow(LocalDate time) {
```
Scoppio, line 70: In the book the transport terms apply as a "discout" or rebate from your employer and it should only cover for the amount of ships you need for the job, so a "100% transport terms" would not allow you to just build a navy of 500 Leopards and 100 Overlords.

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
...
+            for (int x = 0; x < nl.getLength(); x++) {
+                Node wn2 = nl.item(x);
+
+                if (wn2.getNodeName().equalsIgnoreCase("leaseCost")) {
+                    retVal.leaseCost = Money.fromXmlString(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("acquisitionDate")) {
+                    retVal.acquisitionDate = LocalDate.parse(wn2.getTextContent());
+                }
+            }
+            return retVal;
```
Scoppio, line 155: you could create one constructor for the lease cost + acquisitionDate.
```java
Lease lease = new Lease(leaseCost, acquisitionDate);
```

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,162 @@
...
+        } catch (Exception ex) {
+            logger.error("Could not parse lease for unit {}", u.getId(), ex);
```
Scoppio, line 157: wrong use of logger error
Should be:
```java
LOGGER.error(ex, "Could not parse lease for unit {}", u.getId());
```

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright (c) 2009 Jay Lawson (jaylawson39 at yahoo.com). All rights reserved.
+ * Copyright (C) 2013-2025 The MegaMek Team. All Rights Reserved.
```
Scoppio, line 3: Copyright is wrong

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+import megamek.logging.MMLogger;
+import mekhq.campaign.Campaign;
+import mekhq.campaign.unit.UnitOrder;
+import mekhq.utilities.MHQXMLUtility;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+public class LeaseOrder extends mekhq.campaign.unit.UnitOrder {
+    private static final MMLogger logger = MMLogger.create(UnitOrder.class);
```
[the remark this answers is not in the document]
Scoppio: Copilot is correct here, logger should also be LOGGER

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /*
+     * LeaseOrders is in the shopping list, it doesn't have a unit yet to attach to.
+     */
+    public LeaseOrder(megamek.common.Entity entity, mekhq.campaign.Campaign campaign) {
```
Scoppio, line 47: use proper imports instead.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+        super(entity, campaign);
+        this.entity = entity;
```
Scoppio, line 49: The entity is already set by the super class, this line is redundant

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // For the XML reader.
+    private LeaseOrder() {
+    }
+
+    @Override
+    public void decrementDaysToWait() {
+        super.decrementDaysToWait();
+    }
```
Scoppio, line 59: You dont have to implement a function that is already implemented by the super class

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public int getDaysToWait() {
+        return super.getDaysToWait();
+    }
```
Scoppio, line 64: You dont have to implement a function that is already implemented by the super class

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public String find(int transitDays) {
+        super.getCampaign().getQuartermaster().leaseUnit((megamek.common.Entity) getNewEquipment(), transitDays);
```
Scoppio, line 68: Dont the "leaseUnit" creates a new lease?

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+        return "<font color='" +
+                     mekhq.MekHQ.getMHQOptions().getFontColorPositiveHexColor() +
+                     "'><b> unit found for leasing</b>.</font> It will be delivered in " +
```
Scoppio, line 71: missing i18n, and lots of hardcoded full absolute paths to classes which is not recomended in code.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+    // Leases don't actually COST anything...
+    @Override
+    public Money getTotalBuyCost() {
+        return Money.zero();
+    }
+
+    @Override
+    public Money getBuyCost() {
+        return Money.zero();
+    }
```
Scoppio, line 94: No upfront costs? Hum, interesting, I would imagine players would have to pay the month of use upfront.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+                    retVal.quantity = Integer.parseInt(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("daysToWait")) {
+                    retVal.daysToWait = Integer.parseInt(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("entity")) {
+                    retVal.entity = MHQXMLUtility.parseSingleEntityMul((Element) wn2, c);
+                }
+            }
+        } catch (Exception ex) {
+            logger.error("", ex);
+        }
```
Scoppio, line 125:
```suggestion
        } catch (Exception ex) {
            LOGGER(ex, "Exception while parsing lease order from {} -> {}", nl, wn2);
        }
```

MekHQ/src/mekhq/campaign/market/ShoppingList.java
```diff
@@ -182,6 +183,8 @@
                 ((Part) shoppingItem).writeToXML(pw, indent);
             } else if (shoppingItem instanceof UnitOrder) {
                 ((UnitOrder) shoppingItem).writeToXML(pw, indent);
+            } else if (shoppingItem instanceof LeaseOrder) {
```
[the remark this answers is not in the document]
Scoppio: Super serious bug, this would basially make it impossible to save the lease order

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -6804,4 +6816,29 @@ public static PartQuality getRandomUnitQuality(int modifier) {
...
+        }
+        return false;
+    }
+
+    public Lease getUnitLease() {
+        if (hasLease()) {
+            return unitLease;
+        }
+        return null;
+    }
```
Scoppio, line 6842: this should be just `return unitLease;` as it will return null if there is no lease.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -6804,4 +6816,29 @@ public static PartQuality getRandomUnitQuality(int modifier) {
...
+    public void removeLease() {
+        unitLease = null;
+    }
+
+    public boolean hasLease() {
+        if (unitLease != null) {
+            return true;
+        }
+        return false;
+    }
```
Scoppio, line 6835: this should be just `return unitLease != null;`

Scoppio (changes requested):
I have not finished reviewing, but there are lots of stuff to fix, others to change, and alot of redundancy you add.

IllianiBird (changes requested):
So I know this is still WIP but I thought I'd take a nosey around. I'm going to leave quite a bit of feedback, some minor some not so minor. I am fully onboard with this feature, but we need to be making sure we're doing it right.

MekHQ/src/mekhq/campaign/Quartermaster.java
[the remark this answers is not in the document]
savanik: Yeah I need to just change the entire name in Quartermaster

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@ public boolean buyUnit(Entity en, int days) {
...
+        getCampaign().removeUnit(unit.getId());
+    }
+
+    /**
+     * Leases a unit.
+     *
+     * @param en   An entity representing the unit to be leased.
+     * @param days How many days it will take to arrive.
+     **/
+    public boolean LeaseUnit(Entity en, int days) {
```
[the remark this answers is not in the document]
savanik: Resolved by removal.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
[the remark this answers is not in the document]
savanik: Fixed!

MekHQ/resources/mekhq/resources/Finances.properties
```diff
@@ -107,6 +111,10 @@ Loan.title=loan payment to %s
 Loan.text=Your account has been debited for %s in loan payment to %s
 Loan.insufficient.report=<b>You have insufficient funds to service the debt on loan %s!</b>%s Funds required: %s
 Loan.paid.report=You have fully paid off loan %s
+# Leases
+LeaseCosts.title=Lease payments
+LeaseCosts.text=Your account has been debited for %s in lease payments.
+LeaseCosts.insufficient.report=<b>You have insufficient funds to service your leases! Funds Required: %s
```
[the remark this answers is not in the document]
savanik: Also fixed

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: Fixed!

MekHQ/campaigns/The Learning Ropes.cpnx.gz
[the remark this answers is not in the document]
savanik: Should not have been pushed - didn't think I'd made any changes.

MekHQ/src/mekhq/campaign/Quartermaster.java
savanik: In theory, we shouldn't be trying to delete a Unit that doesn't have a Lease set with this function - it would just function as a normal deletion in that case. The lease is disposed of by the garbage collector once it's no longer being referenced. However, in case another coder calls this function mistakenly, I added a test to check for a null lease.

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: Yes, but I feel that is a problem with the process used in contract generation code for calculating 'transport costs'. Employers should know what is a reasonable transportation cost is for the unit and not change that based on the player leasing excessive ships.

Technically speaking, I think the book states that 'Transport terms' are only offered when you don't have Dropships and Jumpships attached to the unit. It also states (or at least heavily implies) that most units are supposed to have them already, since reputation is penalized, etc. There's definitely some ambiguity here. I would say we should update the contract generator with a 'reasonable approximation' by the employer of what your transport costs are, based on your 'leased dropship/jumpship' cost, plus a fee for uncovered units.

savanik: Hrm, that still wouldn't cover the '100% transport' case, though. Maybe we could add a check for 'if this units bays have nothing in them, the employer won't pay for it'?

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: Don't we want the exception trace? (This is a valid function call, and is used in other places.)

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Weird - I think that got added by the IDE for some reason. Fixed.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Yep, fixed.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Hrm... yeah, I think that was to resolve something I had earlier that got fixed in a refactor. Fixed.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Hrm... yeah, I think that was to resolve something I had earlier that got fixed in a refactor. Fixed.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: The lease gets prorated at the end of the month. Currently, no 'deposits' are required for leasing dropships, as this matches book rules.

MekHQ/src/mekhq/campaign/unit/Unit.java
savanik: Fixed

MekHQ/src/mekhq/campaign/unit/Unit.java
savanik: Fixed

MekHQ/resources/mekhq/resources/Finances.properties
```diff
@@ -83,6 +85,8 @@ TransactionType.UNIT_PURCHASE.text=Unit Purchase(s)
 TransactionType.UNIT_PURCHASE.toolTipText=A financial transaction where a unit was or multiple units were purchased.
 TransactionType.UNIT_SALE.text=Unit Sale(s)
 TransactionType.UNIT_SALE.toolTipText=A financial transaction where a unit was or multiple units were sold.
+TransactionType.UNIT_CANCEL_LEASE.text=Cancel Unit Lease
```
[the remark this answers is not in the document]
savanik: I felt like closing a lease was different enough from a regularly monthly payment to have a different entry?

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@ public boolean buyUnit(Entity en, int days) {
         }
     }
 
+    /**
+     * Leases a Unit.
+     *
+     * @param
+     */
+    public boolean leaseUnit(Entity en, int days) {
```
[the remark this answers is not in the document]
savanik: Fixed

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
```diff
@@ -0,0 +1,40 @@
...
+        Money testCost2 = mockUnit.getUnitLease().getLeaseCostNow(dateForTest2);
+        Money testCost3 = mockUnit.getUnitLease().getFirstLeaseCost(dateForTest3);
+        Money testCost4 = mockUnit.getUnitLease().getLeaseCostNow(dateForTest4);
+
+        //Assert
+        assertEquals(Money.of(3000.00), testCost); // this should be the base cost of the lease
+        assertEquals(Money.of(3000.00), testCost2);// This should be a full month's lease
+        assertEquals(Math.round(Money.of(3000.00 / 31).getAmount().doubleValue()),
+              Math.round(testCost3.getAmount().doubleValue())); // This should be a single day's lease
+        assertEquals(Money.of(0.00), testCost4); // This checks leases that have not started yet
```
psikomonkie, line 38: I think this should be several separate tests. If one of these has an issue, it will be more clear which one because the test can be named more descriptively. Consider `@Nested`. 
```
@Nested
class LeaseCostForDatesTests {
	Unit mockUnit;
	
	@BeforeEach
	void beforeEach() {
		Unit unit = new Unit();
		mockUnit = Mockito.spy(unit);
		Mockito.doReturn(Money.of(600000.00)).when(mockUnit).getSellValue();
		
		Lease testLease = new Lease(LocalDate.parse("3025-03-31"), mockUnit);
		mockUnit.addLease(testLease);
	}
	
	@Test
	void testBaseLeaseCost() {
		//Act
		Money testCost = mockUnit.getUnitLease().getLeaseCost();
		
		//Assert
		assertEquals(Money.of(3000.00), testCost); // this should be the base cost of the lease
	}
	
	@Test
	void test fullMonthCost() {
	//Arrange
	LocalDate dateForTest = LocalDate.parse("3025-05-01");
	
	//Act
	Money testCost = mockUnit.getUnitLease().getLeaseCostNow(dateForTest2);
	
	//Assert
	assertEquals(Money.of(3000.00), testCost);// This should be a full month's lease
	}
//TODO etc.
}
```

This class would be "nested" within LeaseTest. If you search the repo for `@Nested` you should find some other tests that are similar.

MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -510,6 +512,22 @@ public void newDay(final Campaign campaign, final LocalDate yesterday, final Loc
             }
         }
 
+        // Leases - on the 1st, go get the cost of units in the hanger looking for leases. Charge money if possible.
+        if (today.getDayOfMonth() == 1 && (campaign.getCampaignOptions().isTrackLeases())) {
```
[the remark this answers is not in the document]
savanik: I was staying away from refactoring a whole method as part of this update, but something like that would probably be more efficient, yes.

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -219,7 +224,83 @@ public TargetRoll getAllAcquisitionMods() {
             target.addModifier(getCampaign().getCampaignOptions().getIsAcquisitionPenalty(), "Inner Sphere tech");
         }
         // TODO: Fix weight classes
-        // TODO: aero large craft
+        if (entity instanceof Dropship || entity instanceof Jumpship) {
```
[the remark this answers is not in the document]
savanik: So when attempting to lease a dropship or jumpship, you have TNs that you're trying to get to (detailed in Campaign Operations). There should have been a comment right beneath that line ... explaining those?

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: I'm not sure I understand - the other classes I've looked at create an empty constructor and use the XML to populate the variables from the save file into the constructor. Is that not correct?

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: It... does? I've update the function name in Quartermaster for more clarity.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: I will need more guidance to accomplish internationalization. I also don't know what you mean by hardcoded paths? Do you mean the nested function calls?

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +577,83 @@ public boolean buyUnit(Entity en, int days) {
...
+        Objects.requireNonNull(en);
+
+        PartQuality quality = PartQuality.QUALITY_D;
+
+        if (campaign.getCampaignOptions().isUseRandomUnitQualities()) {
+            quality = Unit.getRandomUnitQuality(0);
+        }
+        //We don't want to start the new lease until the unit arrives.
+        Unit newUnit = getCampaign().addNewUnit(en, false, days, quality);
```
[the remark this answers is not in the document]
savanik: Should I code up a new function in place of addNewUnit that has a flag whether or not to mothball the unit?

MekHQ/resources/mekhq/resources/Finances.properties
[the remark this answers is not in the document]
IllianiBird: I prefer just 'lease' because it's more versatile. This isn't a requirement to change though, just how I'd do it.

MekHQ/src/mekhq/campaign/Quartermaster.java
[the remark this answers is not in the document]
IllianiBird: I think that would be wise, yeah.

MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -5317,6 +5327,7 @@ public void writeToXml(final PrintWriter pw, int indent) {
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "payForOverhead", payForOverhead);
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "payForMaintain", payForMaintain);
         MHQXMLUtility.writeSimpleXMLTag(pw, indent, "payForTransport", payForTransport);
+        MHQXMLUtility.writeSimpleXMLTag(pw, indent, "trackLeases", trackLeases);
```
Scoppio, line 5330: if you disable this variable, it just stop tracking leases? Leases are still enabled?

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +580,81 @@ public boolean buyUnit(Entity en, int days) {
...
+     * Leases a Unit by creating a new Unit from the entity specified and adding it to the hangar with attached lease.
+     *
+     * @param leaseEntity The unit to lease
+     * @param days        The number of days before the lease starts (typically days to arrival)
+     */
+    public boolean createLeasedUnit(Entity leaseEntity, int days) {
+        if (leaseEntity == null) {
+            LOGGER.error(new NullPointerException(), "null unit passed into createLeasedUnit");
+            return false;
+        }
```
Scoppio: This is not a nullpointer exception, you are actively handling the state and you are returning a valid result. This is not even an error the way you are handling it in this function. If you want leaseEntity to never be null you make it explicit that it cannot be null and you make it so that the function calling it never passes a null to it.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +580,81 @@ public boolean buyUnit(Entity en, int days) {
...
+        if (!campaign.getCampaignOptions().isTrackLeases()) {
+            LOGGER.error("attempted to call createLeasedUnit in a campaign where isTrackLeases is disabled");
+            return false;
+        }
```
Scoppio: so, tracking leases actually means "enable Lease System"?
Also, not an error, trying to use a system that is disabled should at max log a warning.

Error is to be used only when something is wrong and requires engineering/coding time/work to be dedicated to fix it. Warning means something is wrong, should not be wrong, but it is manageable, and probably the result wont be the expected, but nothing broke or is irrecoverable.

This here looks more like a debug to me, not an error nor a warning being honest.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -558,23 +580,81 @@ public boolean buyUnit(Entity en, int days) {
...
+        PartQuality quality = PartQuality.QUALITY_D;
+
+        if (campaign.getCampaignOptions().isUseRandomUnitQualities()) {
+            quality = Unit.getRandomUnitQuality(0);
+        }
+        //We don't want to start the new lease until the unit arrives.
+        LocalDate leaseStart = campaign.getLocalDate().plusDays(days);
```
Scoppio, line 604: if days means daysToArrival then you should name it daysToArrival

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * This function is called from Campaign newDay when successfully acquiring equipment, or when using the GM function
+     * to acquire things in the Acquisition panel.
+     *
+     * @param transitDays How long it takes for the unit to arrive.
+     * @return The string for a successful find report.
+     */
+    @Override
+    public String find(int transitDays) {
+        super.getCampaign().getQuartermaster().createLeasedUnit((megamek.common.Entity) getNewEquipment(), transitDays);
```
Scoppio: You use `super` to access the function you are overriding, here you are trying to access `getCampaign` which should ne available in the current function without the need for `super`.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public int getDaysToWait() {
+        return super.getDaysToWait();
+    }
+
+    @Override
+    public String find(int transitDays) {
+        super.getCampaign().getQuartermaster().leaseUnit((megamek.common.Entity) getNewEquipment(), transitDays);
+        return "<font color='" +
```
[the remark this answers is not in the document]
Scoppio: You need to resolve this

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Displays the correct name for the Procurement List, so they're not confused with standard orders.
+     *
+     * @return {String} Lease for UnitName
+     */
+    @Override
+    public String getAcquisitionName() {
+        return "Lease for " + getName();
```
Scoppio, line 86: missing i18n

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,142 @@
...
+                String nodeName = childNode.getNodeName();
+                String nodeText = childNode.getTextContent();
+
+                if (nodeName.equalsIgnoreCase("quantity")) {
+                    savedLeaseOrder.quantity = Integer.parseInt(nodeText);
+                } else if (nodeName.equalsIgnoreCase("daysToWait")) {
+                    savedLeaseOrder.daysToWait = Integer.parseInt(nodeText);
+                } else if (nodeName.equalsIgnoreCase("entity")) {
+                    savedLeaseOrder.entity = MHQXMLUtility.parseSingleEntityMul((Element) childNode, campaign);
+                }
```
Scoppio, line 132: personal preference, I prefer to have a reference to the entity be persisted instead of serializing the entire entity here.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        } catch (Exception ex) {
+            LOGGER.error("Exception while parsing lease order from {}", childNodeList);
+        }
```
Scoppio, line 136:
```suggestion
        } catch (Exception ex) {
            LOGGER.error(ex, "Exception while parsing lease order from {}", childNodeList);
        }
```

exception is missing from the error log

MekHQ/src/mekhq/campaign/market/ShoppingList.java
```diff
@@ -186,40 +187,47 @@ public void writeToXML(final PrintWriter pw, int indent) {
...
-                    p.setCampaign(c);
-                    if (p instanceof IAcquisitionWork) {
-                        retVal.getShoppingList().add((IAcquisitionWork) p);
+            for (int x = 0; x < childNodeList.getLength(); x++) {
+                Node childNode = childNodeList.item(x);
+                if (childNode.getNodeName().equalsIgnoreCase("part")) {
+                    Part savedPart = Part.generateInstanceFromXML(childNode, version);
+                    savedPart.setCampaign(campaign);
+                    if (savedPart instanceof IAcquisitionWork) {
+                        savedShoppingList.getShoppingList().add((IAcquisitionWork) savedPart);
```
Scoppio, line 211:
```suggestion
                    if (savedPart instanceof IAcquisitionWork acquisitionWork) {
                        savedShoppingList.getShoppingList().add(acquisitionWork);
```

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,191 @@
...
+import mekhq.campaign.unit.UnitOrder;
+import mekhq.utilities.MHQXMLUtility;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import java.io.PrintWriter;
+import java.time.LocalDate;
+
+public class Lease {
+    private static final MMLogger LOGGER = MMLogger.create(UnitOrder.class);
```
Scoppio:
```suggestion
    private static final MMLogger LOGGER = MMLogger.create(Lease.class);
```

MekHQ/src/mekhq/campaign/finances/Lease.java
Scoppio: still not resolve the magicc unit here.

MekHQ/src/mekhq/campaign/finances/Lease.java
Scoppio: Still need a solution

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,191 @@
...
+        if (today.getDayOfMonth() == 1) {
+            int startDay = acquisitionDate.getDayOfMonth();
+            LocalDate yesterday = today.minusDays(1);
+            int daysInMonth = yesterday.lengthOfMonth();
+            int daysElapsed = yesterday.getDayOfMonth() - startDay + 1;
+            float fractionOfMonth = (float) daysElapsed / daysInMonth;
+            return leaseCost.multipliedBy(fractionOfMonth);
+        } else {
+            LOGGER.error("getFirstLeaseCost(today) cannot be called on other days of the month than the 1st.");
+            return null;
```
Scoppio, line 151: Either this is an error or it is not. If it is not then it will be handled correctly and will return a response, if it is an error then it will throw an exception and someone will have to fix it so it stop calling this method when it should not happen.

MekHQ/src/mekhq/campaign/finances/Lease.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * This function checks to see if the Entity is of a leasable type. This is currently hardcoded to restrict it to
+     * Dropships and Jumpships only.
+     *
+     * @return True if unit is leasable
+     */
+    public static boolean isLeasable(Entity check) {
+        return check.isDropShip() || check.isJumpShip();
```
Scoppio, line 162: Duplicated code, entity has a method called `isLargeCraft` which does exactly this.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1391,8 +1394,11 @@ public boolean isDamaged() {
     }
 
     public Money getSellValue() {
-        Money partsValue = Money.zero();
+        if (this.hasLease()) {
+            return Money.zero();
+        } // don't count leased vehicles
```
Scoppio, line 1399: I dont think I agree with this. The unit still have resale value. You not being able to sell is different from something having no sell value.

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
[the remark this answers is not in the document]
Scoppio: `entity.isLargeShip()` does the exact same check as this line

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -225,7 +229,83 @@ public TargetRoll getAllAcquisitionMods() {
...
+            } else if (entity instanceof Dropship) {
+                target.addModifier((int) Math.ceil(entity.getCost(false) / 50000000) + 5 - 6, "Dropship Base");
+            } else {
+                int collars = ((Jumpship) entity).getDockingCollars().size();
+                target.addModifier((int) Math.ceil(entity.getCost(false) / 100000000) + collars - 6, "Jumpship Base");
+            }
+
+            // Since the actual rules are ambiguous, I'm using the tech level as a stand-in for 'uniqueness'
+            // 1 is 'common' and 5 is 'experimental' according to enums
+            int techLevel = entity.getTechLevel(getCampaign().getGameYear());
```
Scoppio: tech level by game year is only used if variable tech level is being used. it is also ignoring extinction rules

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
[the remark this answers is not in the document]
Scoppio: also.... you jus added a large FIXME here for dropship and jumpship?

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
```diff
@@ -225,7 +229,83 @@ public TargetRoll getAllAcquisitionMods() {
...
+                target.addModifier(-6, "Era");
+            } else if (year > 2651) {
+                target.addModifier(-6, "Era");
+            } else if (year > 2571) {
+                target.addModifier(-5, "Era");
+            } else if (year > 2412) {
+                target.addModifier(-4, "Era");
+            } else {
+                target.addModifier(-3, "Era");
+            }
```
Scoppio: move to a function.

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -279,6 +280,20 @@ public void actionPerformed(ActionEvent action) {
...
+        } else if (command.equals(COMMAND_UNLEASE)) {
+            for (Unit unit : units) {
+                if (!unit.isDeployed() && unit.hasLease()) {
+                    Money unleaseValue = unit.getUnitLease().getFinalLeaseCost(unit.getCampaign().getLocalDate());
+                    String text = unleaseValue.toAmountAndSymbolString();
+                    if (0 ==
+                              JOptionPane.showConfirmDialog(null,
+                                    "Do you really want to cancel the lease on " + unit.getName() + " for " + text,
+                                    "Cancel Unit Lease?",
+                                    JOptionPane.YES_NO_OPTION)) {
```
Scoppio, line 292: missing i18n, also, "text" means amount?

MekHQ/src/mekhq/gui/dialog/MekHQUnitSelectorDialog.java
```diff
@@ -118,6 +123,30 @@ public void updateOptionValues() {
...
 
+    public void leaseButtonHandler(boolean isGM) {
+        // Block acquisition if the unit type is unsupported
+        if (getSelectedEntity() == null) {
+            return;
+        }
+        Entity entity = selectedUnit.getEntity();
+        // Remove this check if we ever allow other units to be leased.
+        if (!(Lease.isLeasable(entity))) {
+            return; //TODO: Add a dialog box for why it doesn't work
```
Scoppio, line 134: add this functionality, dont forget i18n

MekHQ/src/mekhq/gui/view/UnitViewPanel.java
```diff
@@ -291,7 +292,11 @@ private void fillStats(ResourceBundle resourceMap) {
         pnlStats.add(lblCost, gridBagConstraints);
 
         txtCost.setName("lblCost2");
-        txtCost.setText(unit.getSellValue().toAmountAndSymbolString());
+        if (unit.hasLease()) {
+            txtCost.setText(unit.getUnitLease().getLeaseCost().toAmountAndSymbolString() + " (Monthly Lease)");
```
Scoppio: missing i18n

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+import megamek.common.Dropship;
+import megamek.common.Mek;
+import mekhq.campaign.unit.Unit;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+@Nested
```
Scoppio: You are not nesting this test, Nested should be used only for nested classes

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
```diff
@@ -0,0 +1,165 @@
...
+    @Test
+    void getLeaseCostNow_WhenFullMonthCost_CalculatesBaseLeaseCostCorrectly() {
+        //Arrange
+        LocalDate testDate = LocalDate.parse("3025-05-01");
+
+        //Act
+        Money testCost = mockUnit.getUnitLease().getLeaseCostNow(testDate);
+
+        //Assert
+        assertEquals(Money.of(3000), testCost);
```
Scoppio: remove the comments

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        //Act
+        Money testCost = mockUnit.getUnitLease().getLeaseCostNow(testDate);
+
+        //Assert
+        assertEquals(Money.zero(), testCost);
+    }
+
+    @Test
+    void getFirstLeaseCost_WhenInFirstMonth_ShouldGivePartialLeaseAmount() {
```
Scoppio: standard in this project is camelcase, it also applies for tests.

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
```diff
@@ -0,0 +1,165 @@
...
+        LocalDate date = LocalDate.parse("3025-04-01");
+
+        //Act
+        Money testCost = mockUnit.getUnitLease().getFirstLeaseCost(date);
+
+        //Assert
+        assertEquals(Math.round(Money.of(3000.00 / 31).getAmount().doubleValue()),
+              Math.round(testCost.getAmount().doubleValue())); // This should be a single day's lease
```
Scoppio:
```suggestion
        assertEquals(Math.round(Money.of(3000.00 / 31).getAmount().doubleValue()),
              Math.round(testCost.getAmount().doubleValue()), "This should be a single day's lease");
```

You should use the `message` field in the assert to add the explanation of why something is wrong.

Scoppio (changes requested):

MekHQ/src/mekhq/campaign/CampaignOptions.java
savanik: Updated code - if you disable this, you will no longer be able to acquire new leases. Existing leases will continue to be tracked until the player manually removes those units.

MekHQ/src/mekhq/campaign/Quartermaster.java
savanik: No, in this context it means when the lease starts. If we ever change the math on that function, it'll still mean when the lease should start.

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: Updated all these.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Added internationalization, I think.

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: So the solution will be that the employer will pay a 'reasonable amount' for transport as a lump sum, as per the current rules. I do need to go and review how that is calculated to make sure it lines up with expectations.

MekHQ/src/mekhq/campaign/finances/LeaseOrder.java
savanik: Fixed the error reporting, generally, to debug.

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: Fixed

MekHQ/src/mekhq/campaign/finances/Lease.java
savanik: The return of this function may change in the future if we decide to make other units leasable, and as it serves an independent purpose, it should not be tied to the other code.

MekHQ/src/mekhq/campaign/unit/Unit.java
savanik: You don't own it - it doesn't have resale value _for the unit_. This value is used in other places for things like calculating how much of a loan you can take out. It obviously shouldn't count for those purposes.

MekHQ/src/mekhq/campaign/unit/UnitOrder.java
savanik: Since this section of the rules is ambiguous and up to GM discretion, I guess we'll just ignore it for now. Maybe when we have a canon table of how many dropships of what types exist in any particular time period, we can codify this.

MekHQ/unittests/mekhq/campaign/finances/LeaseTest.java
IllianiBird: I think the comments are fine. They're helpful in making tests clearer and also encouraging developers to write well constructed tests.

IllianiBird:
This has been in development for a while and I don't want it to die on the vine. What still needs to be done to bring it across the finish line? :)

PR #7044, opened 2025-05-18
Fix: Stopped Random Death Not Randomly Causing People to Die

IllianiBird:
Random death is initialized when the campaign is initialized, however it relies on campaign options. The default death multiplier is 0.

All of this combined to make death chance 0. Which is somewhat less than intended.

Now we correctly reinitialize random death both when the campaign is loaded and when campaign options is exited.


Sleet01 (approved):
LGTM

PR #7050, opened 2025-05-19
Finance NewDay cleanup. Variable and function condensing.

savanik:
Tidying up the Finance NewDay function by adding some variables and a function call for insufficient funds


MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -310,9 +311,23 @@ public void addLoan(Loan loan) {
         loans.add(loan);
     }
 
+    public void addReportInsufficientFunds(Campaign campaign, String report) {
+        campaign.addReport(String.format("<font color='" +
```
IllianiBird, line 315: Use the ReportingUtilities methods here, please.

IllianiBird (changes requested):
One very small change needed, then we can get this merged :)

Scoppio (approved):

PR #7051, opened 2025-05-19
Fixed: Fixed Exception During Avionics Evaluation

IllianiBird:
Recent changes to `Avionics#getStickerPrice` prevent existing campaign saves from being loaded if that campaign has a unit that uses Avionics. Or, if the campaign has an Avionics part in their warehouse.

This is because `unit` is `null` both during loading and while a part is loose in the warehouse.

Added null protection to the method which resolves the issue.


Scoppio (approved):
This makes the part effectively zero cost lol

PR #7059, opened 2025-05-20
Fix: Fixed Processing of Profession Salaries in the Event New Profession Added Since Campaign Last Loaded

IllianiBird:
When we load a campaign we transcribe profession salaries to the campaign. If a new profession were added this could potentially fail. The end result would be the remaining professions would all receive default salaries.

This isn't a huge issue currently as, concerning the recent addition of professions, we want them to have default salaries anyway. However, this would become a problem if a new profession was added in the middle of the profession enum, rather than at the end.

This PR resolves the problem.


MekHQ/src/mekhq/campaign/CampaignOptions.java
```diff
@@ -5529,1179 +5530,1187 @@ public void writeToXml(final PrintWriter pw, int indent) {
...
                     Money[] mergedSalaries = new Money[PersonnelRole.values().length];
                     for (int i = 0; i < mergedSalaries.length; i++) {
-                        mergedSalaries[i] = (newSalaries[i] != null) ? newSalaries[i] : defaultSalaries[i];
+                        try {
+                            mergedSalaries[i] = (newSalaries[i] != null) ? newSalaries[i] : defaultSalaries[i];
+                        } catch (Exception e) {
+                            // This will happen if we ever add a new profession, as it will exceed the entries in
+                            // the child node
+                            mergedSalaries[i] = defaultSalaries[i];
+                        }
```
IllianiBird, line 5943: While I was in this neck of the woods I took the opportunity to clean up xml parsing. This is the only section with actual functionality changes.

PR #7078, opened 2025-05-21
Fix: Fixed Required Combat Teams Returning Lower Than Intended Values

IllianiBird:
Removes unnecessary line in calculations. While me and Psi were futzing about trying to make this approach work I managed to leave in a line from a draft that made it into the final merge.


psikomonkie:
What is this fixing? I thought this was intentional.

PR #7092, opened 2025-05-23
Improvement: Added Faction Standing Support for Non-StratCon Campaigns

IllianiBird:
When a StratCon campaign is concluded faction standings are updated automatically. However, this cannot happen for non-StratCon campaigns as we do not have sufficient employer or enemy information. Instead we prompt the user with a dialog asking them to manually declare who their employer and enemy were.

https://github.com/user-attachments/assets/17554e55-d768-4bfe-a751-2648ba676b6c


IllianiBird:
Flipping this to live now that the main gui report has been merged

rjhancock (approved):

Issue #7116, opened 2025-05-26
[RFE] Add toggle for OpFor Callsigns


theyoungster21:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

Add a toggle for the recent added feature of OpFor generating callsigns

### Use Cases or Rationale

While neat that OpFor can generate with a little more personality now, some people (myself included) may prefer for them to keep their unit names for quick ease of recognition without having to mouse over. This also can help when multiple units stack into a tile and only one sprite can reasonably be seen. A toggle that allows one to choose whether OpFor generates with these names will let us get the best of both worlds, with those preferring the flavour to get it, and those preferring the ease of recognition to get it as well.

### MekHQ Suite Version *

0.50.06

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

theyoungster21:
Yeah that looks like it would work perfectly! Love that you can tie it to experience level too

PR #7123, opened 2025-05-28
RFE: Allow toggling OpFor callsigns on or off, or setting them to only apply to pilots at or above some skill level

Sleet01:
Since the auto-gen OpFor callsigns went in we've had a few requests to disable or gate the feature via some option.
This adds the option to toggle randomized OpFor callsigns on and off, as well as setting the lower skill limit for applying callsigns (if desired).

Testing:
- Added tests for the new options and callsign generation
- Tested in MekHQ with live enemy forces.

Requires: MegaMek/megamek#7153
Close: #7116


MekHQ/unittests/mekhq/campaign/mission/AtBDynamicScenarioFactoryTest.java
```diff
@@ -0,0 +1,141 @@
+package mekhq.campaign.mission;
+
+import megamek.common.Entity;
+import megamek.common.Game;
+import megamek.common.MekSummary;
+import megamek.common.MekSummaryCache;
+import megamek.common.Player;
+import megamek.common.enums.SkillLevel;
+import mekhq.campaign.Campaign;
```
rjhancock, line 9: Copyright Header

rjhancock (changes requested):
Copyright Header....

MekHQ/unittests/mekhq/campaign/mission/AtBDynamicScenarioFactoryTest.java
Sleet01: I should just write a test I can run locally, save us all some time -_-;

HammerGS (approved):

PR #7129, opened 2025-05-29
Fix: stop checking a misc flag against equipment class

Scoppio:
Fix: stop checking a misc flag against equipment class


psikomonkie (approved):
Nice!

PR #7152, opened 2025-06-05
Add page to hanger to show what unit is assigned to each transport

wildfire142:
QoL addition to allow quick look up of which unit is assigned to which transports.

Had issues with a large unit remembering which unit was on which dropship, so made this view.

No way to add units to transports on the hangar menu currently - don't have the skill to code it yet.

### DIFF
--- a/MekHQ/src/mekhq/gui/model/UnitTableModel.java
+++ b/MekHQ/src/mekhq/gui/model/UnitTableModel.java
@@ -85,7 +85,9 @@ public class UnitTableModel extends DataTableModel {
     public static final int COL_SITE = 19;
     public static final int COL_QUIRKS = 20;
     public static final int COL_RSTATUS = 21;
-    public static final int N_COL = 22;
+    public static final int COL_SHIP_TRANSPORT = 22;
+    public static final int COl_TAC_TRANSPORT = 23;
+    public static final int N_COL = 24;
 
     private Campaign campaign;
     //endregion Variable Declarations
@@ -130,6 +132,8 @@ public class UnitTableModel extends DataTableModel {
             case COL_SITE -> "Site";
             case COL_QUIRKS -> "Quirks";
             case COL_RSTATUS -> "Mode";
+            case COL_SHIP_TRANSPORT -> "Ship Transport";
+            case COl_TAC_TRANSPORT -> "Tactical Transport";
             default -> "?";
         };
     }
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -85,7 +85,9 @@ public class UnitTableModel extends DataTableModel {
     public static final int COL_SITE = 19;
     public static final int COL_QUIRKS = 20;
     public static final int COL_RSTATUS = 21;
-    public static final int N_COL = 22;
+    public static final int COL_SHIP_TRANSPORT = 22;
+    public static final int COl_TAC_TRANSPORT = 23;
```
psikomonkie: This should be `COL_TAC_TRANSPORT` not `COl_TAC_TRANSPORT`

MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -318,6 +322,10 @@ public Object getValueAt(int row, int col) {
             case COL_SITE -> Unit.getSiteName(unit.getSite());
             case COL_QUIRKS -> entity.countQuirks();
             case COL_RSTATUS -> unit.isSalvage() ? "Strip" : "Repair";
+            case COL_SHIP_TRANSPORT -> (unit.getTransportShipAssignment() != null) ?
+                                             unit.getTransportShipAssignment().getTransportShip().getName() : "-";
+            case COl_TAC_TRANSPORT -> (unit.getTacticalTransportAssignment() != null) ?
+                                            unit.getTacticalTransportAssignment().getTransport().getName() : "-";
```
psikomonkie: Want to support tractors too? Consider using `Unit.getTransportAssignment(CampaignTransportType)` instead of `getTransportShipAssignment()` and `getTacticalTransportAssignment`. Then just pass in the proper CampaignTransportType enum (`SHIP_TRANSPORT`, `TACTICAL_TRANSPORT`, `TOW_TRANSPORT`).

MekHQ/src/mekhq/gui/model/UnitTableModel.java
wildfire142: Fixed the typo and pushed the changes

MekHQ/src/mekhq/gui/model/UnitTableModel.java
wildfire142: Was only thinking about bays and shipping not towing, also late here and early start tomorrow will add it another day.

psikomonkie (approved):

Issue #7045, opened 2025-05-18
[RFE] personnel report to show secondary roles held


wildfire142:
### Enhancement Type *

New Feature

### Brief Description of Enhancement *

A report similar to the personnel report showing how many people hold each secondary role.

### Use Cases or Rationale

In larger units it hard to quantify how many people hold secondary roles and the number of each role.

### MekHQ Suite Version *

MekHQ v0.50.06-nightly-2025-05-14

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

wildfire142:
Closed via PR#[7170](https://github.com/MegaMek/mekhq/pull/7170)

PR #7158, opened 2025-06-07
Added secondary role sections to personnel report

wildfire142:
This is what I came up with to answer my own RFE #[7045](https://github.com/MegaMek/mekhq/issues/7045)

added a count of personnel performing secondary roles split by combat and support along with a breakdown of a count of each type of role.

Located just below the primary role counts and above the extra if an the bottom of each section.

current personnel report
new personnel report

Not that elegant just added the total for each secondary role under the current ones, have not added to the salary or other parts of the report at the moment,

### DIFF
--- a/MekHQ/src/mekhq/campaign/report/PersonnelReport.java
+++ b/MekHQ/src/mekhq/campaign/report/PersonnelReport.java
@@ -34,6 +34,7 @@
 package mekhq.campaign.report;
 
 import java.time.LocalDate;
+import java.util.EnumMap;
 
 import mekhq.campaign.Campaign;
 import mekhq.campaign.finances.Money;
@@ -244,15 +245,15 @@ public class PersonnelReport extends AbstractReport {
     }
 
     public String getSecondarySupportPersonnelDetails() {
-        final PersonnelRole[] personnelRoles = PersonnelRole.values();
-        int[] countPersonByType = new int[personnelRoles.length];
+        EnumMap<PersonnelRole, Integer> countPersonByType = new EnumMap<>(PersonnelRole.class);
         int countSecondary = 0;
         for (Person person : getCampaign().getPersonnel()) {
             // Add them to the total count
             final boolean secondarySupport = person.getSecondaryRole().isSupport(true);
 
             if (secondarySupport && person.getPrisonerStatus().isFree() && person.getStatus().isActive()) {
-                countPersonByType[person.getSecondaryRole().ordinal()]++;
+                countPersonByType.put(person.getSecondaryRole(),
+                      (countPersonByType.getOrDefault(person.getSecondaryRole(), 0) + 1));
                 countSecondary++;
                 }
             }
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
 
         return sb.toString();
     }
+
+    public String getSecondarySupportPersonnelDetails() {
+        final PersonnelRole[] personnelRoles = PersonnelRole.values();
+        int[] countPersonByType = new int[personnelRoles.length];
```
Scoppio: An EnumMap<PersonelRoles, Integer> will give you the same result, but with a more explicit intent, also less fragile as it wont be subject to out of bound errors nor nullpointers

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
...
+            // Add them to the total count
+            final boolean secondarySupport = person.getSecondaryRole().isSupport(true);
+
+            if (secondarySupport && person.getPrisonerStatus().isFree() && person.getStatus().isActive()) {
+                countPersonByType[person.getSecondaryRole().ordinal()]++;
+                countSecondary++;
+                }
+            }
+
+        StringBuilder sb = new StringBuilder("\nSecondary Role Support Personnel\n\n");
```
Scoppio, line 265: Missing i18n support.

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
...
+
+        sb.append(String.format("%-30s%4s\n", "Total Secondary Role Support Personnel", countSecondary));
```
Scoppio, line 267: Missing i18n support

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
...
+
+        for (PersonnelRole role : personnelRoles) {
+            if (role.isSupport(true)) {
```
Scoppio: An enummap will help avoid this kind of validation

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
...
+            // Add them to the total count
+            final boolean secondaryCombat = person.getSecondaryRole().isCombat();
+
+            if (secondaryCombat && person.getPrisonerStatus().isFree() && person.getStatus().isActive()) {
+                countPersonByType[person.getSecondaryRole().ordinal()]++;
+                countSecondary++;
+            }
+        }
+
+        StringBuilder sb = new StringBuilder("\nSecondary Role Combat Personnel\n\n");
```
Scoppio, line 295: Missing i18n support

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -239,4 +242,62 @@ public String getSupportPersonnelDetails() {
...
+
+        sb.append(String.format("%-30s %4s\n", "Total Secondary Role Combat Personnel", countSecondary));
```
Scoppio, line 297: Missing i18n support

Scoppio (changes requested):
Really good overall, I suggest to use enumMap instead of array as it makesnit more expressive, debugging becomes natural as it is not dependent of indices and ordinal positions but instead of enums. This is not the reason why Im asking for changes, but I'd appreciate if you did it this way instead, as it will make my life easier when debugging and refactoring code

About i18n, it's really not something optional, and once it is added please keep in mind that any "format" that is necessary should be done using the i18n tools and parameters, meaning that you should not use i18n with string builder or string concat, but instead use it with parametric format, like I18n.getFormatesTextAt("bundle", "text.key", param1, param2...)

If you need to format the numbers you can use a NumberFormatter.

wildfire142:
> Please add a more descriptive comment body in this PR and a screenshot showing how the report is currently. Adding a before / after is also cool because it gives an easy point of comparison for who is revieweing.

Done added some info tthe orginal commnet is it enough or do you need more info?

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: Made the change

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: Made the change and now also only shows roles when people are assigned them, quirk not planned

wildfire142:
The report looks a little different now.

with no secondary roles assigned
updated new report part 1
With some secondary roles assigned
updated new report part 2

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: added suport

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: added suport

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: added suport

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
wildfire142: Added support using getFormattedTextAt() in the only way I could figure out apologies if its not right never attempted this before :)

wildfire142:
Going to close this as I've had a better stab at i18n after a kind soul explained it to me better and open a new PR for that atempt.

PR #7173, opened 2025-06-10
Fix/rolling logs

Scoppio:
fix: added rolling logs, also added two fixes, one for phenotype parsing and one for misc type testing on ammo type


Sleet01 (approved):
LGTM

PR #7182, opened 2025-06-11
fix: multiple small and simple fixes, reenable planet search, fix npe…

Scoppio:
# What it does?

- Removes one function marked for deletion
- Improves logger on campaign xml parser (not all entries, but most)
- Naming convention applied for some instances of "mek" incorrectly named in the code
- A few javadocs improvements
- SkillPerquisite renamed as SkillPrerequisite (probably need to be renamed to SkillPreRequisite)?
- Refactored SkillPrerequisite to use a more robust clone method.
- Missing "VeeStabiliser" in parts deserializer restored with special migration process that is "rename hardened".
- Fix NPE that is caused when setting up the config on a new campaign.
- Fix NPE on SPA Utilities.
- Restore funcitonality to Planet Search on map tab
- Remove the blue border when the planet search is focused.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,8 @@ jobs:
       - name: Test All
         working-directory: mekhq
         run: ./gradlew testAll  --stacktrace --scan
+        env:
+          mm.profile: dev
 
       - name: Upload Test Logs on Failure
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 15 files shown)


Sleet01 (approved):
LGTM

Sleet01:
@Scoppio Recommend adding:

```java
    // TODO: rework Systems singleton to allow unit testing
    @Disabled("Disabled due to OOM")
```
above `loadDefaultTest` in `SystemsIntegrationTest.java`

PR #7214, opened 2025-06-15
Improvement: New Personnel Market Dialog Will Now Remember Last Size

IllianiBird:
Adds preferences to personnel market dialog so that last size is restored when reopening dialog.


rjhancock (approved):

rjhancock:
Has conflict. Need fixing.

PR #7231, opened 2025-06-17
Improvement: Replaced Placeholder Glossary with New Layout and Functionality

IllianiBird:
Requires https://github.com/MegaMek/megamek/pull/7255

The original Glossary was something thrown together in a few hours just before the launch of 50.06. At the time we didn't want to put too much complex functionality into the feature as we didn't want to risk it being broken. This PR replaces the original Glossary with a less utilitarian layout, as well as improving the ease with which new glossary entries can be added.

### Old Glossary
image

### New Glossary
image

Individual glossary entries have also been given a glow-up. One of the issues with the original glossary system is that it was very easy to get several entries deep with no real way to get back out, short of spamming left-right. Now it should be a lot easier to have several entries open and referenced at ease.

Finally, neither the glossary and individual entries are 'modal', meaning they can be left open (on a second monitor, for example) while the user plays. This should make it much easier to reference entries during play.

### Old Entry
image

### New Entry
image

On the left is a list of all tabs that are currently open. Using Drake's new Enhanced Tab functionality it is possible to drag and drop these tabs to re-order them. While on the right is a list of all Glossary entries available.

### Documentation
Finally I added the ability to access MekHQ's documentation from within the Glossary

image


IllianiBird:
Updated description to reflect new functionality

rjhancock (approved):

rjhancock:
Updated with Master and waiting on CI to pass.

PR #7241, opened 2025-06-19
Improvement: Implements Faction Standing-based Batchall Modifiers

IllianiBird:

If a campaign's Faction Standing falls too low with one or more Clans, it's possible those Clans will no longer offer Batchalls.


Sleet01 (approved):
LGTM

PR #7242, opened 2025-06-19
Improvement: Implements Faction Standing-based Food & Housing Cost Modifiers; Fixed Bug with Food & Housing Costs

IllianiBird:

This implements the resupply food and housing cost modifiers gained from positive or negative faction standing.

It also fixes an embarrassing bug where I set up food and housing costs, fully tested them, wrote bundle entries... ...but then forgot to actually hook it up to anything so those costs were never deducted...


Sleet01 (approved):
LGTM

PR #7244, opened 2025-06-19
Improvement: Implements Faction Standing-based Support Point Generation Modifiers

IllianiBird:

A high or low Faction Standing with a faction can influence how many Support Points are generated at the beginning of a contract and how easy it is to generate support points while on contract.


Sleet01 (approved):
LGTM

PR #7246, opened 2025-06-19
Improvement: Implements Faction Standing-based Recruitment Modifiers

IllianiBird:

A high or low Faction Standing with a faction can influence how many recruitment rolls a campaign gets and how frequently they get applicants from specific factions


Sleet01 (approved):
LGTM

PR #7265, opened 2025-06-21
Fix: Adjusted Active Faction Getter to Optionally Exclude 'Command' Factions

IllianiBird:
Commands (units associated with a faction, such as the Sword of Light) were added with the recent faction data overhaul. However, as these are considered factions, they were being included whenever we fetched a list of all active factions. This PR adjusts that method so now we can optionally exclude Commands, if desired.


MekHQ/src/mekhq/campaign/universe/Factions.java
```diff
@@ -106,14 +106,38 @@ public Collection<Faction> getFactions() {
...
+     * these Command factions are included in the results.</p>
+     *
+     * @param date            the date for which to check faction activity
+     * @param includeCommands whether to include Command factions (those whose short name contains a dot)
+     *
+     * @return a collection of active factions for the specified date, optionally including Commands
+     */
+    public Collection<Faction> getActiveFactions(LocalDate date, boolean includeCommands) {
+        return getFactions().stream()
+                     .filter(f -> f.validIn(date)
```
SJuliez: Chaining multiple filters instead of packing it into one makes this easier to format. Also the last line can be simplified.

```
        return getFactions().stream()
                     .filter(f -> f.validIn(date))
                     .filter(f -> !f.isInactive())
                     .filter(f -> (includeCommands || !f.getShortName().contains(".")))
                     .toList();
```

or even

```
        return getFactions().stream()
                     .filter(f -> f.validIn(date))
                     .filter(Faction::isActive)
                     .filter(f -> (includeCommands || !f.getShortName().contains(".")))
                     .toList();
```
if an inverted isActive() method is added

IllianiBird: Oh I didn’t know you could do that, that makes things so much more readable!

IllianiBird:
I want to incorporate Juli's suggestion, so please hold off merging this. I'll merge myself once done.

HammerGS (approved):

PR #7275, opened 2025-06-24
Fix: Fixed a Handful of Calls to Unsafe Faction Standing Getters

IllianiBird:
We should be using the safe variants, as that avoids needing to check whether faction standing is being tracked every time we call the getter


Sleet01 (approved):
LGTM

PR #7278, opened 2025-06-26
Improvement: Added Faction Standing Level Descriptions for all Factions

IllianiBird:
Previously I chose to obscure standing level descriptions for factions that aren't given special attention, however that didn't feel right so I've removed the comments that blocked faction standing level descriptions for all.

The fact that I commented this code out, instead of just deleting, shows I wasn't certain it was the right call at the time it was implemented.


Sleet01 (approved):
LGTM

PR #7282, opened 2025-06-29
Improvement: Updated Immersive Dialog to Better Handle Large Volumes of Text

IllianiBird:
The updates the 'simple' version of immersive dialog to optionally accept a new enum that will dictate display width. This is designed to be used for immersive dialogs where we are displaying significant amounts of text (such as in the upcoming Faction Standings system).


Sleet01 (approved):
LGTM

PR #7292, opened 2025-07-03
Improvement: Improved Faction Standing Defection with New Dialogs and the Ability for At Will Defection

IllianiBird:
The biggest change here is a pile more dialog text for when a player defects to another faction. It also loosens up the requirements for defection so that more factions are available for defection.

Finally, I went ahead and added some ease-of-access buttons to the gui. This left us with a weird gap, so I added an ease-of-access button for the quick start campaign generator. That change is mostly for us developers who use that generator a _lot_, but I know players will also find it useful.


Sleet01 (approved):
LGTM

PR #7296, opened 2025-07-05
Improvement: Added Faction Standing Ultimatum Events for the Onset of the FedCom Civil War and ComStar Schism.

IllianiBird:
This PR is the final feature update for the new Faction Standing system, introducing 'ultimatum events'. These are major canonical events where the player is forced to pick between two factions. The included events are the onset of the FedCom Civil War and the ComStar Schism. These were chosen because I knew enough about them to write the events. I don't doubt others can be added - I imagine people would _love_ if we added in the Arano Restoration as an ultimatum...

Ultimatums follow a distinct pattern. First, the player is presented with the inciting incident...

image

...then their personnel offer arguments for each side...

image

...the the player is given the ultimatum. They must choose one side or the other...

image

...finally, a news bulletin discusses the players choice giving the decision weight...

image


Sleet01 (approved):
LGTM

PR #7297, opened 2025-07-05
Improvement: Made the Arano Restoration Faction Playable in StratCon; Added Faction Standing Ultimatum for Arano Restoration War

IllianiBird:
# Requires https://github.com/MegaMek/mekhq/pull/7296

Previously we had a bit of a data problem that caused the Arano Restoration faction to be effectively unplayable during the Arano Restoration war. This was a limitation from AtB's contract generation system that resulted in the Restoration being unable to generate contracts.

I fixed this by giving the Restoration a presence in the Magistracy of Canopus. Not as a system owner, per se, but as a government in exile. I made this decision based on the existing faction data which had the Restoration's capital moved into the MoC during the war period.

This then required me to update some of the faction relation stuff so that two things happened: the Arano Restoration wouldn't try and attack the Magistracy, and so that the Directorate wouldn't attack the Taurians. Given the Restoration Wars' effective status as a proxy war for those states it didn't make a whole lot of sense for the various factions to be trying to beat up their sponsors.

Finally, I updated planetary ownership data ever so slightly so that the switch from Coalition happened on the 1st February (the coronation) rather than a month earlier on the 1st January.

With the popularity of HBS's Battletech I made the decision to add a 'Restoration War' date bookmark to the date picker so that users can easily jump to the schism should they wish to relive their HBS adventures in MekHQ.

With that out of the way I was able to set up a Faction Standing Ultimatum so that when the coup occurs the player will have the option to support the Directorate or the Restoration.

image


MekHQ/src/mekhq/campaign/universe/factionStanding/FactionStandingUltimatum.java
```diff
@@ -0,0 +1,195 @@
...
+ */
+public class FactionStandingUltimatum {
+    /**
+     * Represents the earliest possible date for when Lord Espinosa overthrew the legitimate heir of the Arano
+     * Coalition
+     *
+     * <p>This date was determined by comparing dates in the House Arano manual, as well as dates in HBS'
+     * Battletech.</p>
+     */
+    private static final LocalDate ESPINOSA_COUP = LocalDate.of(3022, Month.FEBRUARY, 1);
```
Sleet01, line 76: If we add any more of these events, I think we should discuss making a loadable class and a yaml file format for these configurations.

Sleet01 (approved):
LGTM

PR #7298, opened 2025-07-05
Improvement: Faction Leader Data

IllianiBird:
This is the first batch of faction leader data for use in Faction Standings. Included are a handful of data fixes, as I spotted them.


Sleet01 (approved):
LGTM

PR #7301, opened 2025-07-06
Improvement: Updated Faction Starting Location Logic for Mercenary and Pirate Campaigns

IllianiBird:
Earlier in the 50.07 dev cycle I introduced some logic to allow pirate campaigns to start on any playable faction's capital. The intent was for this to introduce a degree of randomness to a play start. However, I really wasn't happy with the code and it didn't really work as I intended.

So I went ahead and updated the logic, improved it for robustness and extended the random start logic to mercenary campaigns.

The logic is thus:
- If the campaign isn't mercenary or pirate start on the relevant faction capital
- If the campaign is mercenary, start on the mercenary capital 33% of the time, otherwise start on a random playable faction's capital (excluding Clan factions and those in the Deep Periphery). No weighting.
- If the campaign is pirate, start in the Tortuga Dominions capital 33% of the time (if TD is around, otherwise start on the Mercenary capital), otherwise start on a random playable faction's capital as before, only this time with additional weighting for periphery (non-deep periphery) nations.

Combined with the new faction standing system, this should help give individual campaigns more character. Especially mercenary campaigns, as it means campaigns will now start embedded in nations other than the Lyran Commonwealth.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -251,8 +251,8 @@ import mekhq.campaign.universe.factionStanding.FactionAccoladeEvent;
 import mekhq.campaign.universe.factionStanding.FactionAccoladeLevel;
 import mekhq.campaign.universe.factionStanding.FactionCensureEvent;
 import mekhq.campaign.universe.factionStanding.FactionCensureLevel;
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #7302, opened 2025-07-06
Improvement: Moved Faction Standing Ultimatum Data into YAML Format

IllianiBird:
This was based on a suggestion from @Sleet01 requesting that we move the ultimatum data from hardcoded values and into the data folder.


MekHQ/src/mekhq/campaign/universe/factionStanding/FactionStandingUltimatumsLibrary.java
```diff
@@ -0,0 +1,112 @@
...
+     * Returns an unmodifiable map of all loaded Faction Standing ultimatums.
+     *
+     * <p>The map is keyed by {@link LocalDate} and values are {@link FactionStandingUltimatumData} instances.</p>
+     *
+     * @return unmodifiable map of ultimatums by date
+     *
+     * @author Illiani
+     * @since 0.50.07
+     */
+    public Map<LocalDate, FactionStandingUltimatumData> getUltimatums() {
```
Sleet01: This will only support one ultimatum per given date; do you think we will ever want to have multiple events on the same date?

Sleet01 (approved):
Looks great, nice work!
Just one question about future capabilities but it can be answered or updated later

MekHQ/src/mekhq/campaign/universe/factionStanding/FactionStandingUltimatumsLibrary.java
IllianiBird: Yeah, that occurred to me. I'm torn, because by using dates as keys we can very cheaply and reliably confirm whether an ultimatum is occurring in a given day for a given faction without using a loop.

If you don't mind taking a look at the ultimatum code in src/campaign/universe/faction standing I'd welcome your thoughts on alternative models.

The key thing is that I have a method that just checks whether an ultimatum is occurring before we call the class constructor. As that allows us to avoid passing in Campaign unnecessarily. It's all fully commented, so if you have the time to look I would appreciate your insight.

IllianiBird: I think I've got a workable solution.

PR #7337, opened 2025-07-15
followup: Rename variables `p` and `e` to verbose names

Abhiram98:
I noticed that in a recent [commit](https://github.com/MegaMek/mekhq/commit/925dd6acfcab6187649d73817e0a4a67816de41d), multiple variables were renamed in a similar fashion:

- `Person p` -> `Person person` (2 renames in the previous commit)
- `LogEntry e` -> `LogEntry logEntry` (also 2 renames in the previous commit)

Here's a patch to make those changes consistent with the rest of the file!


Abhiram98:
Requesting a review from @IllianiBird, as you authored the previous commit

IllianiBird (approved):

PR #7373, opened 2025-07-19
Improvement: Implemented Illiterate & Scrounge SPAs

IllianiBird:
-4 to Intelligence linked skill checks unless the character's Languages/Any skill level is 4 or greater.

Scrounge was partially implemented, so I finished it up while working on this PR as they interacted.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -2465,63 +2465,64 @@ public class Campaign implements ITechManager {
         int bloodnameTarget = 6;
         PersonnelOptions options = person.getOptions();
         Attributes attributes = person.getATOWAttributes();
+        boolean isIlliterate = person.isIlliterate();
         if (!ignoreDice) {
             switch (person.getPhenotype()) {
                 case MEKWARRIOR: {
                     bloodnameTarget += person.hasSkill(SkillType.S_GUN_MEK) ?
                                              person.getSkill(SkillType.S_GUN_MEK)
-                                                   .getFinalSkillValue(options, attributes) :
+                                                   .getFinalSkillValue(options, attributes, isIlliterate) :
                                              TargetRoll.AUTOMATIC_FAIL;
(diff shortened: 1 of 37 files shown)


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -2465,63 +2465,64 @@ public void checkBloodnameAdd(Person person, boolean ignoreDice) {
...
         PersonnelOptions options = person.getOptions();
         Attributes attributes = person.getATOWAttributes();
+        boolean isIlliterate = person.isIlliterate();
         if (!ignoreDice) {
             switch (person.getPhenotype()) {
                 case MEKWARRIOR: {
                     bloodnameTarget += person.hasSkill(SkillType.S_GUN_MEK) ?
                                              person.getSkill(SkillType.S_GUN_MEK)
-                                                   .getFinalSkillValue(options, attributes) :
+                                                   .getFinalSkillValue(options, attributes, isIlliterate) :
```
Sleet01, line 2474: This looks like an absolutely massive amount of changes for one SPA.
Wouldn't it make more sense to store "isIlliterate" state in the ATOWAttributes data, rather than passing it to `getFinalSkillValue()`?
This seems cleaner to me, both for this change, and for any future changes we might need to make to that function.

Sleet01 (changes requested):
At first glance this looks like a lot of unnecessary churn just to pass one boolean around.
Can the literacy state be wrapped into Attributes?

MekHQ/src/mekhq/campaign/Campaign.java
IllianiBird: Lemme think on this. I *might* be able to handle that, but RAW the status of being illiterate isn't static.

Illiteracy is determined by the character having both the Illiterate 'SPA' and the Languages/Any skill at less than level 4. This is RAW.

What that means, however, is that we cannot just store a value in Attributes and then expect it to be correct unless any time the Languages/Any skill is updated we also recheck the stored value in Attributes. Which *might* be better, but would also be a different kind of churn.

Thoughts?

PR #7378, opened 2025-07-19
Fix: Fixed Soft-Lock in Faction Standings When Both Employer and Enemy Are Untracked Factions

IllianiBird:
If, while manually resolving a non-AtB Contract, both the employer and enemy were untracked factions the player would be stuck in a loop of constantly being told the update was invalid.

Now we have a skip button.


Sleet01 (approved):
LGTM

PR #7394, opened 2025-07-22
Fix: #7338 Fixed Numerous Instances Where We Were Requesting all Units & Only Retrieving Those Units in Combat Forces

### LINKED ISSUE #7338: [Issue] [50.06] Units not properly registering for HQs post battle resolutions
### Brief Description *

Me and another user [AceOfBlades on the discord] have encountered issues where units were not properly handling in resolution after an HQ battle. 

Image

This image is of Ace of Blades ToE force. Apparently they entered a battle and their BA units did not end up in the resolution.

Image

This is mine. In my case I dont believe any of these units ended up in the resolution. Mine was post a convoy battle. Only the support/reinforcing units showed up in the resolution.

In both cases there seems to be something going on that is causing units to fail to register for post battle resolution. IF more information/testing is required, just ask me here or in discord and I can try my best to run tests when I get the chance.

### 3. Steps to Reproduce

_No response_

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.06

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

IllianiBird:
Fix #7338

Across the codebase we use a method called `getAllUnits` to fetch all units in a given force. However, in numerous cases we were explicitly requesting to only include units in combat forces.

This means that if a force passed into the method isn't a combat force - for example, a convoy - those units were not being always factored into several locations. Most noticeably the scenario resolution dialog. This PR addresses that.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -5008,7 +5008,7 @@ public class Campaign implements ITechManager {
                         // If any unit in the force is under repair, don't deploy the force
                         // Merely removing the unit from deployment would break with user expectation
                         boolean forceUnderRepair = false;
-                        for (UUID uid : forceIds.get(forceId).getAllUnits(true)) {
+                        for (UUID uid : forceIds.get(forceId).getAllUnits(false)) {
                             Unit u = getHangar().getUnit(uid);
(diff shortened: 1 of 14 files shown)


Sleet01 (approved):
LGTM

PR #7395, opened 2025-07-23
Feature: #7316 Removed Mid-Contract Version Change Restrictions; Added Automated Campaign Version Upgrader

### LINKED ISSUE #7316: [RFE] Version Upgrade Contract Completion/Regeneration
### Enhancement Type *

New Feature

The idea is that there would be a GM Mode way to end a contract with a version modifier (A,B,C . . . 0.1,0.2,0.3 . . .) and placeholders, and then, after upgrading to the newest MekHQ version, a "new" or continuation contract with the remaining objectives and information migrated/regenerated as needed.

### Use Cases or Rationale

There is a requirement that one complete a contract before migrating MekHQ versions because regression testing and old contracts don't play well together. This would allow faster adoption of new MekHQ versions and functionality.

As apocryphal personal experience, I have been months behind the latest version multiple times trying desperately to get my contract complete so I can get all the new stuff. And because I play MekHQ in solo mode, and also play with my BIL, I have to ensure that both files can be a position to complete the contracts.

Finally, I see with some regularity, people on the forums with version info several instances behind. I have to presume, since it is my personal experience, that they are forestalling the upgrade because of contract incompleteness and not any of the other completely reasonable reasons. 

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.06

### Final Verification

IllianiBird:
Fix #7316

This PR updates the process by which campaigns from older versions are loaded into MekHQ.

Previously, we had a number of restrictions in place. Users needed to wait until they were out of contract, had to open and then manually apply campaign options, and then advance day. This was a pain in the bum and as a result we know for a fact that players were not following the upgrade process.

Now, when a player loads a campaign from a prior version they are presented with this dialog:

image

Here, players can select their preferred stock (or custom) preset, or select 'manual'. If the player selects 'manual' they will immediately be transferred to the Campaign Options dialog. Just as if they had opened Campaign Options in a loaded campaign.

If the player selects a stock preset we handle the upgrade process in the background. Essentially, we're silently loading the Campaign Options dialog and then closing it again, but the player just sees this:

image

The version number shown is dynamically updated based on the current MekHQ version. So no maintenance required. Once the upgrade process is completed the campaign finishes loading like normal.

The entire process is shown in this video (which also gives a rough estimate of how long the upgrade takes).

https://github.com/user-attachments/assets/5c45e8a7-48be-4781-b678-197063ce4d7c


Sleet01 (approved):
Copilot suggestions might have value, but looks fine as-is.

PR #7396, opened 2025-07-23
Fix: Fixed NPE When Exporting Campaign

IllianiBird:
When exporting a campaign with a null CamOps Reputation an NPE would occur. This PR addresses that, as well as removing the risk of a NPE Reputation Controller moving forward.


Sleet01 (approved):
LGTM

PR #7398, opened 2025-07-23
Improvement: Added the Ability to Only Consider the SPAs of a Unit's Commander, When Determining What SPAs a Multi-Crewed Unit Has

### LINKED ISSUE #6934: [RFE]Need to adjust several maneuvering pilot-only SPAs pre-req so multi-crew vees can use them
### Enhancement Type *

Several maneuvering SPAs are pilot-only but multi-crew vees require at least 50% crews to have the SPA for it to activate. Please include ground vehicle gunnery to the list. I have to manually add the pre-req after each update.  

Image
Image
Image
Image
Image
Image
Image

### Use Cases or Rationale

Multi-crew vees can't access these SPAs because the gunnery crews can't take it so most tanks are out of the question. 

### Attach Files

_No response_

### MekHQ Suite Version *

50.05

### Final Verification

IllianiBird:
This addresses a long term pain point that exists with multi-crewed units, making it much easier to acquire SPAs for those units.

There are a lot of balancing concerns with this option that were raised when I brought up the idea in 50.04. It mostly revolves around it making it much easier to 'complete' the SPAs needed for a unit. Ultimately, I decided to ignore that as a problem in favor of user experience. By keeping this as a Campaign Option we ensure that players can pick whichever experience best suits their play style.

That this will mean a single character, in a crew, may be required to skill into gunnery/piloting skills that are not required for their role - to gain access to the various SPAs - is a feature, not a bug. It helps mitigate the decreased SPAing-up costs, now that SPAs are only required by a single character.

Close #6934

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -374,6 +374,9 @@ lblUseArtillery.tooltip=Characters can possess the Artillery skill which is used
   \ for specific weapons.
 lblUseAbilities.text=Use Special Pilot Abilities (SPAs)
 lblUseAbilities.tooltip=Characters can purchase and benefit from SPAs.
+lblUseCommanderAbilitiesOnly.text=Commander SPAs Only <span style="color:#C344C3;">\u2605</span>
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
LGTM

PR #7402, opened 2025-07-24
Improvement: Added Faction Standing Speed Campaign Option

IllianiBird:
This adds a campaign option that allows users to apply a multiplier to all Regard gains and losses. Allowing users to increase or decrease the Faction Standing speed.


Sleet01 (approved):
LGTM

PR #7417, opened 2025-07-26
Improvement: Decreased Support Point Cost of Ronin Hires by 50%

IllianiBird:
I wasn't happy with the cost of personnel acquired through Ronin events. At their current price point they were almost always a losing consideration.

This PR reduces the price to Required Combat Teams / 2, or 1 Support Point. Whichever is higher.


Sleet01 (approved):
LGTM

PR #7421, opened 2025-07-26
Fix: Added Missing Label for Coming of Age Options in Campaign Options

IllianiBird:
The label for the Coming of Age options in the Biography tab was missing.


Sleet01 (approved):
LGTM

PR #7423, opened 2025-07-26
Improvement: Moved Two Large Campaign Options Utility Methods Out of Campaign Options

IllianiBird:
We recently started seeing test failures due to our testing system running out of memory. The only consistent thing was that the test was running out of memory trying to parse `CampaignOptions`. The most likely culprits were the marshalling and unmarshalling methods.

This PR refactors those methods into their own discrete classes and moves CampaignOptions and those two new classes into their own package.

Functionality is largely unchanged except where it needed to be adjusted to account for the new location of the method.


MekHQ/src/mekhq/Utilities.java
```diff
@@ -1012,7 +1012,7 @@ public static Money[] readMoneyArray(Node node) {
     }
 
     public static Money[] readMoneyArray(Node node, int minimumSize) {
-        String[] values = node.getTextContent().split(",");
+        String[] values = node.getTextContent().trim().split(",");
```
IllianiBird, line 1015: I'm unsure if I just got lucky, or if this has been causing problems in the background for years, but the lack of 'trim' here was causing campaign options to fail to parse finances properly.

rjhancock (approved):

PR #7426, opened 2025-07-27
Improvement: Added Version History to Campaign Save

IllianiBird:
This was a request from @HammerGS.

All this PR does is keep a log of what versions a campaign was saved into, preserved in the campaign save.

image

This allows developers, at a glance, see how old a campaign is when debugging.

Nothing is done with this information, it is purely for ease of reference.

I should add that this function will not have any memory prior to the campaign being saved in 50.07, the screenshot is for illustration purposes only.


Sleet01 (approved):
LGTM

PR #7427, opened 2025-07-28
Fix: Fixed Minor Bug Where Clan Faction Welcome Events Were Being Given by Merchants and Not MekWarriors

IllianiBird:
As per title. Just a copy-paste error.

This was a visual bug only.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/CompanyGenerationDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/CompanyGenerationDialog.java
@@ -32,6 +32,8 @@
  */
 package mekhq.gui.dialog;
 
+import static mekhq.campaign.universe.Faction.MERCENARY_FACTION_CODE;
+
 import java.awt.Container;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/CompanyGenerationDialog.java
```diff
@@ -156,14 +158,14 @@ private void confirmationActionListener(final ActionEvent evt) {
 
         Faction campaignFaction = campaign.getFaction();
         String campaignFactionCode = campaignFaction.getShortName();
-        if (campaignFactionCode.equals("MERC")) {
+        if (campaignFactionCode.equals(MERCENARY_FACTION_CODE)) {
```
IllianiBird, line 161: This constant didn't exist when this code was written, but there's no reason not to use it instead of a magic string.

Sleet01 (approved):
LGTM

Issue #4602, opened 2024-08-10
Battle loss compensation is payed to the player for damaged units even if they weren't damaged in that battle


repligator:
### Environment
Java 17
Linux
Master (latest PR https://github.com/MegaMek/mekhq/commit/5ab5959011ea22901321b5a4a0792da4d43d65ad)

### Severity

Low (Minor/Nuisance): Minor glitches or cosmetic issues that don’t affect gameplay and occur rarely.

### Description
The player receives battle loss compensation for damaged units at the end of a battle. However, if the player deploys a unit that is already damaged, they will still be compensated for it, as though the damage was new. There is no difference between manual and automatic resolution.

### Files
[Manual-Omega Galaxy31521015.cpnx.gz](https://github.com/user-attachments/files/16570587/Manual-Omega.Galaxy31521015.cpnx.gz)
[automatic-Omega Galaxy31521015.cpnx.gz](https://github.com/user-attachments/files/16570589/automatic-Omega.Galaxy31521015.cpnx.gz)
[mekhq.log](https://github.com/user-attachments/files/16570591/mekhq.log)

IllianiCBT:
This will be because we're just checking whether the unit is damaged and not whether its new state differs from the state currently stored in the campaign. I'm not entirely sure how complicated this would be to resolve.

repligator:
Yeah, it's not exactly what I'd call a high priority thing, but I'd figured I'd open an issue regardless.

IllianiBird:
This was resolved back in August, closing

PR #7458, opened 2025-08-04
Fix: battle loss compensation is payed to the player for units damaged before the battle

Raugharr:
Fixes issue #4602


MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
```diff
@@ -1873,7 +1873,6 @@ private void editUnit(UUID id, int unitIndex, boolean isSalvage) {
 
         UnitEditorDialog med = new UnitEditorDialog(frame, unitStatus.getEntity());
         med.setVisible(true);
-        unitStatus.getUnit().runDiagnostic(false);
```
Raugharr, line 1869: This should be safe to remove, any unit that is edited in this dialog should have `runDiagnostic` performed on them in the `ResolveScenarioTracker`

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -86,6 +86,8 @@
  * @author Jay Lawson (jaylawson39 at yahoo.com)
  */
 public class ResolveScenarioTracker {
+    public static final double DAMANGED_PART_COMPENSATION_MODIFIER = 0.2;
```
Raugharr, line 100: I found a campaign option called "Damaged Parts" in operations/price multipliers. I'm not sure if `getValueOfAllDamagedParts` should be multiplied by that option instead of this constant.

IllianiBird: It probably should, I'll double-check once things are stable enough for me to do a full review (see my comment on #7463)

Sleet01:
Can you describe how this actually fixes the bug?  A cursory reading makes it look like the previous logic has just been wrapped into a function.
How does this differentiate pre-existing damage from damage taken during the most recent scenario?

Sleet01 (approved):
LGTM, thanks for the rundown.

IllianiBird:
I went ahead and fixed the conflict. It was a minor thing and I really don't want to mess OP around any more than we already have.

PR #7517, opened 2025-08-17
Feature: Implemented 'Extra-Income' A Time of War Trait

IllianiBird:
This implements the ATOW Trait 'Extra-Income' and incorporates it into our monthly checks.

Each month the campaign will gain, or lose, finances based on the commander's Extra Income Trait.

If the character is not the campaign commander (and is not a child) then the characters' Total Earnings will be adjusted by this value, instead (if tracked).

The funds gained/lost match those values found in ATOW pg116 and aren't worth the xp investment, but hey, it's a Trait we didn't have and now we do: progress!

I'll probably optionally buff the Trait later, but for now it's RAW.


MekHQ/unittests/mekhq/campaign/personnel/ExtraIncomeTest.java
```diff
@@ -0,0 +1,404 @@
...
+    @EnumSource(value = ExtraIncome.class)
+    void testAllFinancesImprove(ExtraIncome extraIncome) {
+        int currentTraitLevel = extraIncome.getTraitLevel();
+
+        int priorTraitLevel = currentTraitLevel - 1;
+        Money priorFunds = Money.of(-100000000);
+        try {
+            ExtraIncome otherExtraIncome = ExtraIncome.extraIncomeParseFromString(priorTraitLevel + "");
+            priorFunds = otherExtraIncome.getMonthlyIncome();
+        } catch (Exception ignored) {}
```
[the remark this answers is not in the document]
IllianiBird: Its fine in this instance

IllianiBird:
Apparently I forgot to take this off 'for new dev cycle'. Updated and doing so now

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5569,7 +5576,7 @@ public void processNewDayPersonnel() {
 
                 if (currentDay.getMonthValue() % 3 == 0) {
                     if (person.hasDarkSecret()) {
-                        String report = person.isDarkSecretRevealed(true, false);
+                        report = person.isDarkSecretRevealed(true, false);
                         if (report != null) {
```
psikomonkie: Should this be `if (!report.isBlank()) {` or `if (!StringUtility.isNullOrBlank(report)) {`? It looks like `String report = person.isDarkSecretRevealed(true, false);` will never result in a null value for `report` as it returns an empty string if the secret isn't revealed.

MekHQ/src/mekhq/campaign/personnel/enums/ExtraIncome.java
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * <p>Each enum instance corresponds to a trait level ranging from -10 to +10, with an associated {@code lookupKey}
+ * for identification and a {@code monthlyIncome} value that determines the expected extra income (positive) or cost
+ * (negative) for that level.</p>
+ *
+ * <p>The enum provides methods for retrieving lookup information, trait levels, and monetary values, as well as
+ * resolving instances by their lookup key, enum name, or trait level.</p>
+ *
+ * @author Illiani
+ * @since 0.50.07
```
psikomonkie: Should this be `@since 0.50.10`? This won't exist in any version calling itself `0.50.07` so I think this might be misleading.

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -3430,7 +3431,7 @@ protected Optional<JPopupMenu> createPopupMenu() {
             menuItem.setToolTipText(String.format(resources.getString("spendOnUnlucky.tooltip"), target));
             menuItem.setActionCommand(makeCommand(CMD_BUY_TRAIT,
                   UNLUCKY_LABEL,
-                  String.valueOf(-traitCost),
+                  String.valueOf(traitCost),
```
psikomonkie: Why is the lucky trait's cost being switched - it looks like currently a character gains XP when they become more "unlucky". This looks to be reverting that - should it cost XP to become more unlucky, and reward XP when the character becomes less unlucky?

Same w/ Bloodmark

IllianiBird: I'm glad you caught this. It looks like this is reverting a fix that was made during the 07 cycle. Basically a sign of the PRs' age.

MekHQ/src/mekhq/campaign/personnel/enums/ExtraIncome.java
IllianiBird: Taking this to the dev channel.

IllianiBird: Done

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
IllianiBird: I went ahead and updated the various issues like this to match current `main`. I've got a mental note to go in and verify `main` is correct and if not I'll update it in a separate PR.

MekHQ/src/mekhq/campaign/Campaign.java
IllianiBird: Yeah, there was a bunch of weirdness here, which I went and fixed. Thanks for calling it out :)

PR #7544, opened 2025-09-20
Fix 50.07 MHQ tests part 3

Sleet01:
Fixes ~~majority~~ all of failing MHQ tests.
Disables one test which makes use of multiple deprecated functions and needs a Subject Matter Expert to fix.


MekHQ/unittests/mekhq/campaign/market/ContractMarketAtBGenerationTests.java
```diff
@@ -502,6 +510,15 @@ public void addMercWithoutRetainerEmployerNeutralAtWarAtBContractSucceeds(final
...
 
+    /**
+     * This appears to be a test that a Mercenary campaign faction employer
+     * allows mercenary sub-faction contracts.
+     * Currently this test fails for reasons that are unclear; disabling temporarily.
+     * @param gameYear  see generateData() above
+     * @param unitRating
+     * @param isClanEnemy
+     */
+    @Disabled("XXX SME: Needs deprecated methods replaced.")
```
IllianiBird, line 521: This is the exact ATB test I was talking about in the dev channel when I said there is a test which, if broken, is getting disabled.

The long and short is that if you make _any_ changes to contract generation this test will fail with very little indication as to what the issue is. It's a very fragile test and is a pain to fix.

In an ideal world we'd replace it with a more robust test suite, rather than trying to cram everything into a single test. However, we don't live in an ideal world and I fully support disabling.

PR #7554, opened 2025-09-21
Fix #7552: Fixed Contract Generation on New Campaign

### LINKED ISSUE #7552: [Issue] Unable to Launch New AtB/StratCon Campaign
### Brief Description *

Launching a new campaign, with AtB or StratCon enabled, will cause campaign creation to fail after Campaign Options. For some reason mhq is failing to fetch local factions, as if the universe was empty.

As this causes campaign creation to fail, we're bounced back to the start screen so it's not possible to see the state of the universe to confirm my theory.

Loading into an existing campaign does not have this issue and contracts can be generated as normal.

### 3. Steps to Reproduce

1. launch a new campaign with the veteran preset
2. Observe the error

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.07

### Operating System *

Mac

### Java Version *

J17

### Final Verification

IllianiBird:
Fix #7552

During the recent refactor a change was made to an AtB Contract Generation method so that it accepted String (faction code) instead of the full Faction class. However, the source of the Faction (which we use to fetch the code) is Nullable. No null safety was included, resulting in an NPE and contract creation failure in the event RandomFactionGenerator returned a null value (returning a null value here is working as programmed).

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/contractMarket/AtbMonthlyContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/contractMarket/AtbMonthlyContractMarket.java
@@ -289,11 +289,15 @@ public class AtbMonthlyContractMarket extends AbstractContractMarket {
                 int retries = MAXIMUM_GENERATION_RETRIES;
                 AtBContract contract = null;
                 while ((retries > 0) && (contract == null)) {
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

Issue #7562, opened 2025-09-23
[MekHQ-0.50.07-nightly-2025-09-22] Campaign & MegaMek option tooltip errors


Tzahr:
### Brief Description *

Very simple, missing tooltips on the following:

**Campaign options:**

Operations > Markets > Contracts.
Contract Max Salvage tooltip:

<img width="331" height="46" alt="Image" src="https://github.com/user-attachments/assets/9dc3c653-df6c-4b87-a4d2-46fce4b5b9d5" />

Operations > Digital GMs > General
Automatically generate skills for OPFOR:
<img width="559" height="96" alt="Image" src="https://github.com/user-attachments/assets/e2815380-07cc-49b0-9dba-331e63210fcc" />

MegaMek Options: 
Advanced Rules: 
<img width="431" height="56" alt="Image" src="https://github.com/user-attachments/assets/8962cb5f-71bd-40a8-b077-96f3fcacf357" />

Also worth checking the log for generic issues. 

### Attach Files

[unified_log.log](https://github.com/user-attachments/files/22483331/unified_log.log)

### Severity *

Low (Minor/Nuisance): Minor glitches or cosmetic issues that don't affect gameplay and occur rarely.

### MekHQ Suite Version *

v0.50.07-nightly-2025-09-22

### Operating System *

Windows 11 Home Version	10.0.26100 Build 26100

### Java Version *

21.0.3

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PR #7564, opened 2025-09-23
Fixes missing strings in Campaign Options dialog and broken Army Creator

Sleet01:
It looks like not all Resources got renamed evenly during the text normalization process.
This patch aligns all property values used by the Auto-Generated Call Signs panel and options.

Previously the sub-panel within Campaign Options had mangled strings, and the Army Generator couldn't open because it tried to read a non-existent option value.

Testing:
- Verified Campaign Options strings were correct
- Verified the Army Generator tool could now be opened


IllianiBird:
@Sleet01 did this close off #7562?

PR #7566, opened 2025-09-25
Make Campaign more testable with dependency injections

Sleet01:
Campaign is a very large, very heavy class that has resisted unit testing due to some egregious issues:
- Large amount of in-constructor logic and dependency instantiation,
- Lack of lightweight versions of some dependencies (which couldn't be injected anyhow),
- Lack of argument options for Campaign constructor

This patch attempts to resolve as many of those issues as possible without creating too much extra burden for developers, or increasing memory usage / instantiation time from the current default.
It also attempts to make Campaign unit tests easier to set up and more lightweight, so that existing tests that depend on Campaign (or Systems, or PartsStore) can be re-enabled - or so that new tests can be written quickly and easily.

Adds:
- A `CampaignConfiguration` class for passing, storing, and adjusting `Campaign` class configuration with the ability for injecting test classes as dependencies (this helps with unit testing).
- A few `CampaignFactory` static methods to assist in constructing `CampaignConfiguration` objects, or whole Campaign instances (this helps with writing unit tests).
- Lightweight versions of a few of the heavier singletons like Systems and PartsStore, to reduce memory usage and test runtime (this helps with running tests)
- Utility methods for creating Campaigns that use the lightweight classes (this helps with writing unit tests)
- The ability to load a single PlanetarySystem file in Systems and TestSystems (this helps with conserving space and memory)

Changes:
- Several market / manager classes so that they can be passed to Campaigns at instantiation instead of created inside the Campaign constructor (this helps with writing unit tests).  Note that Campaign instances are now responsible for registering themselves with these market / manager instances!

Testing:
- Ran all MekHQ unit tests, including previously-disabled suites, in 1 second under the current test time (with the larger Campaign-related tests disabled).

Still to test:
- User experience in-game with the new Campaign class.


Sleet01:
Taking this to Draft until the Ranks thing is ironed out.

MekHQ/src/mekhq/campaign/CampaignFactory.java
```diff
@@ -120,6 +156,147 @@
...
+
+        // A starting CurrentLocation is required
+        PlanetarySystem starterSystem = systems.getSystems().get("Galatea");
+        CurrentLocation location;
+        try {
+            location = new CurrentLocation(starterSystem, 0);
+        } catch (Exception ex) {
+            String message = String.format(
+                  "Unable to set location to %s. If this wasn't during automated testing this must be investigated.",
+                  starterSystem
```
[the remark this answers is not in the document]
Sleet01: PlanetarySystem probably needs a `toString()` of its own but that is for later; I'll just get the name here.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -502,102 +494,144 @@ public enum AdministratorSpecialization {
...
+        retirementDefectionTracker = retDefTracker;
+        autosaveService = autosave;
+        autoResolveBehaviorSettings = behaviorSettings;
+        this.partsStore = partsStore;
+        this.newPersonnelMarket = newPersonnelMarket;
+        this.randomDeath = randomDeath;
+        this.campaignSummary = campaignSummary;
+
+        // Members that take `this` as an argument
+        // (TODO Sleet01: refactor to allow lazy linking / setting post-instantiation)
```
IllianiBird: Rogue TODO

MekHQ/src/mekhq/campaign/CampaignFactory.java
```diff
@@ -120,6 +156,147 @@ public static CampaignFactory newInstance(MekHQ app) {
...
+        campaignConfig.setLocation(location);
+        campaignConfig.setPartsStore(partsStore);
+        campaignConfig.setNewPersonnelMarket(newPersonnelMarket);
+        campaignConfig.setRandomDeath(randomDeath);
+        campaignConfig.setCampaignSummary(campaignSummary);
+
+        return campaignConfig;
+    }
+
+    public static @Nullable Campaign createCampaign() {
```
IllianiBird, line 313: If `createCampaignConfiguration` is analogous to `new Campaign()` which should be used, `createCampaignConfiguration()` or `createCampaign()`? It might be worth adding a Java Doc to `createCampaign()` as I can see that being a source of confusion.

MekHQ/src/mekhq/campaign/market/personnelMarket/markets/NewPersonnelMarket.java
```diff
@@ -169,11 +165,12 @@ public static NewPersonnelMarket generatePersonnelMarketDataFromXML(final Campai
...
-                        case PERSONNEL_MARKET_DISABLED -> new NewPersonnelMarket(campaign);
-                        case MEKHQ -> new PersonnelMarketMekHQ(campaign);
-                        case CAMPAIGN_OPERATIONS_REVISED -> new PersonnelMarketCamOpsRevised(campaign);
-                        case CAMPAIGN_OPERATIONS_STRICT -> new PersonnelMarketCamOpsStrict(campaign);
+                        case PERSONNEL_MARKET_DISABLED -> new NewPersonnelMarket();
+                        case MEKHQ -> new PersonnelMarketMekHQ();
+                        case CAMPAIGN_OPERATIONS_REVISED -> new PersonnelMarketCamOpsRevised();
+                        case CAMPAIGN_OPERATIONS_STRICT -> new PersonnelMarketCamOpsStrict();
                     };
+                    personnelMarket.setCampaign(campaign);
```
IllianiBird, line 173: OH! It took me a couple of moments to parse what you were doing here, but this is _so_ much cleaner!

MekHQ/unittests/mekhq/campaign/CampaignTest.java
```diff
@@ -94,22 +95,48 @@
...
+
+        // Add a system to the systems instance; it must exist in the testresources dir
+        config.getSystemsInstance().load(TEST_CANON_SYSTEMS_DIR + "Skye.yml");
+
+        // Instantiate the campaign with the new info
+        Campaign campaign = new Campaign(config);
+
+        // Let's plot a trip from the starting location to Skye!  It should be about 6 days:
+        int travelTime = campaign.getSimplifiedTravelTime(systems.getSystemByName("Skye", config.getDate()));
+        assertEquals(6, travelTime);
```
IllianiBird, line 139: This is super nifty

MekHQ/unittests/mekhq/campaign/mission/TestSystems.java
```diff
@@ -0,0 +1,63 @@
...
+ * MechWarrior Copyright Microsoft Corporation. MekHQ was created under
+ * Microsoft's "Game Content Usage Rules"
+ * <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
+ * affiliated with Microsoft.
+ */
+
+package mekhq.campaign.mission;
+
+import mekhq.campaign.universe.PlanetarySystem;
+import mekhq.campaign.universe.Systems;
```
IllianiBird: If we're matching the location of the original Class, this probably wants to be in campaign/universe/systems and not campaign/mission

MekHQ/unittests/testUtilities/MHQTestUtilities.java
```diff
@@ -35,24 +35,100 @@
...
+     * Then set the heavyweight dependencies here:
+     * 1. Systems (TestSystems for testing purposes)
+     * 2. GameOptions (required for MegaMek, may be candidate for further test class development)
+     * 3. Player instance
+     * 4. LocalDate for Campaign start day
+     * 5. CurrentLocation (created from a PlanetarySystem, which must be retrieved using Systems or TestSystems)
+     * 6. Logistical classes: parts store, new-type personnel market, random death generator, persistent campaign
+     *    summary tracker.
+     */
+    public static CampaignConfiguration buildTestConfigWithSystems(Systems systems) {
```
IllianiBird, line 87: How lightweight is this? By which I mean when should we be using this instead of mockCampaign? Is it a method of last resort, for when we absolutely need campaign to test, or is it something we can use at will without worry of tanking test completion times?

MekHQ/src/mekhq/campaign/CampaignFactory.java
Sleet01: `createCampaign()` gives the caller a `Campaign` instance configured with all the default values we used to set up in the Campaign constructor.  It _calls_ `createCampaignConfiguration()`, which creates a `CampaignConfiguration` object containing those values, and then passes it to `new Campaign()`.

I'll add some comments to clarify the usage:
1. Call `createCampaign()` if you want a Campaign to use in a game.
2. Call `createCampaignConfiguration()` if you want a config object you can adjust before creating a new Campaign, e.g. when writing tests.

In MHQTestUtilities I added two helper functions that do the same things as 1. and 2. above, but using testing-oriented classes and configuration options, to keep things lightweight (TestSystems and TestPartsStore especially).

MekHQ/unittests/mekhq/campaign/mission/TestSystems.java
Sleet01: Hmm, I wonder how I messed that up!

MekHQ/unittests/testUtilities/MHQTestUtilities.java
Sleet01: It should be measurably faster than using `Systems` and `PartsStore`, both of which read a lot of files and instantiate a lot of objects.
Currently TestSystems and TestPartsStore are completely empty so test users will need to populate them for any rigorous testing, but that means they should be fast and lightweight.

IllianiBird: Super. That's the exact answer I was hoping for.

MekHQ/src/mekhq/campaign/CampaignFactory.java
Sleet01: ...also I apparently had the old `createCampaign()` comment above `createCampaignConfiguration()`, because of old brain. -_-;

IllianiBird: That explains it! :D

PR #7569, opened 2025-09-25
Fix: Addressed Rank System Loading Failure

IllianiBird:
This PR addresses the issue of rank systems initialization resulting in an early exit of all processes should ranks fail to load. For example, if loading rank systems in the headless-dateless GitHub test environment.


IllianiBird:
@Sleet01 once this is merged, if you update #7566 you should be good.

I'm handling the failing tests separately as there are a few different reasons behind those tests failing (now that I can see _why_ they're failing) and I'd rather not clutter up this PR unnecessarily.

PR #7589, opened 2025-09-25
Improvement: Updated Faction Standing Notification Text Formatting

IllianiBird:
Originally I added certain quotes into 'blockquote' formatting. However, on second sniff I'm not a fan of how it looks so I went ahead and removed the block quotes.


Sleet01 (approved):
LGTM

PR #7603, opened 2025-09-26
Improvement: Graduated Training Skill from RP Only to Support

IllianiBird:
With the introduction of #7602 'Training' is no longer an RP Only skill and is now classified as a Support skill.


Sleet01 (approved):
LGTM

PR #7614, opened 2025-09-27
Improvement #7468: Updated Prisoner Capacity Contributions for Non-Infantry Units

### LINKED ISSUE #7468: [RFE] Allow crewed units to apply modifier to prisoner capacity
### Enhancement Type *

Right now with regards to prisoner management, your only option to increase capacity is to increase the amount of troopers you're got assigned to security, whether it be battle armour or infantry. Crewed units also increase capacity, but only at a token rate of two capacity for ever one crew member. This was done in order to give them some utility if desired to include as fluff units.

I propose that instead of leaving them as largely symbolic, crewed units can add a multiplier to prisoner capacity instead of a concrete number. This would ideally be based off of the base prisoner capacity to avoid exponential growth, and could be capped a certain amount of capacity provided per unit. Examples of multipliers could be 1.05x or 1.2x depending on the utility of the unit a security.

### Use Cases or Rationale

This would solve the problem of ballooning your unit's personnel count in order to keep prisoners managed. Since infantry are only split up (generally) into squads and platoons, if you take enough prisoners you could have more soldiers as guards than you do all the other personnel in your unit combined.

This would provide an alternate path to improving prisoner capacity while promoting combined arms and lowering the unit footprint/HR needs, as a base number of soldier would still be require to actually *provide* prisoner capacity but then other units could be deployed to add multipliers instead of linearly increasing the amount of manpower you need.

There is precedent for this in-fiction, as there exists already a class of IndustrialMechs called SecurityMechs, which are defined by an in-universe treaty, the "SecurityMech Treaty of 2613". These mechs could benefit from high modifiers/capacity ceilings, while other units that may be less suitable to the task of managing a prison camp and more suitable for actual front line combat could have low or non-existent modifiers. This could mitigate the potential problem of "bigger is better" and having a lance of Atlases being (illogically) assigned to security duty. 

The same logic could be applied to vehicles, although I'm not sure there are such discrete definitions available for security vehicles.

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.06

### Final Verification

IllianiBird:
Close #7468

Rather than offering a token improvement to Prisoner Capacity, non-infantry units increase Prisoner Capacity by 5% to a maximum of +25%.

This means that there is value in adding non-infantry units to Security teams. The prior system was rather underwhelming.


Sleet01 (approved):
LGTM

PR #7621, opened 2025-09-27
Fixed Personnel Market Not Generating Civilian Professions Instead of Generic Dependents

IllianiBird:
The civilian professions were made for the new personnel market, so it was somewhat ironic that it couldn't generate applicants with them. Now it can.


Sleet01 (approved):
LGTM

PR #7631, opened 2025-10-03
Improvement: Displayed Total Skill Level For Those Skills Where Skill Level is Important

IllianiBird:
It was raised on the Discord a little while ago that for some skills (command skills, mainly) you want to be able to skill how many levels a character has in that skill, not just the skill's target number.

This PR updates SkillTypes so that some skills (where appropriate) will now display both the TN and Level of the skill in the character sheet.

image


Sleet01 (approved):
LGTM

PR #7633, opened 2025-10-04
Improvement: Convert Skill Type Storage from Array to Enum

IllianiBird:
This PR relates to the conversion of our current SkillType class from a static String plus hash hybrid and into an enum.

This makes it far easier for developers to add new skills and make changes to existing skills. It makes skill storage more reliable and removes the risk that a skill change will result in damage to a players' save.


MekHQ/src/mekhq/campaign/parts/EnginePart.java
```diff
@@ -438,9 +438,10 @@ public String getDetails(boolean includeRepairDetails) {
     @Override
     public boolean isRightTechType(String skillType) {
         if (getEngine().hasFlag(Engine.TANK_ENGINE)) {
-            return skillType.equals(SkillType.S_TECH_MECHANIC);
+            return skillType.equals(SkillTypeNew.S_TECH_MECHANIC.name());
```
IllianiBird, line 441: You'll see a lot of `SkillTypeNew.SKILL.name()` in this PR. The original SkillType format had all skill references using a static String. That means our codebase is set up to accept Strings and not SkillTypes.

One of my goals with this PR was to make as few changes as possible, to reduce the risk of introducing new bugs with this overhaul. While converting the codebase to use SkillTypes wherever possible would be the correct move, it would balloon the size of this PR beyond reasonable levels. It is better, then, to use legacy architecture that can be replaced ad hoc in future PRs.

PR #7667, opened 2025-10-07
Improvement: Introduced the Abstract Tracking of Infantry Gunnery Skills; Those Skills are No Longer 'RP Only'

IllianiBird:
Earlier in the year we introduced a number of infantry combat skills, such as Support Weapons and Demolitions. Unfortunately, it is not reasonably possible to implement the tracking of what skills are necessary for what platoon loadout. And even if it were, I don't believe it would offer a good play experience.

As a compromise we are introducing the concept of 'Infantry Gunnery Skills'. For Conventional Infantry units each soldier's skill is the highest of the following skills:

- Archery
- Demolitions
- Martial Arts
- Melee Weapons
- Small Arms
- Support Weapons
- Thrown Weapons

That means that if a soldier has Archery 3+, Demolitions 8+, and Small Arms 4+ they will be treated as if they had gunnery 3+. This is handled dynamically by mhq with no intervention required on the part of the player.

Furthermore, any character with one of the above skills is eligible to be assigned to the Soldier profession, even if they don't have Small Arms.

Soldiers generated randomly, such as for the personnel market or for OpFors, will use Small Arms exclusively. This is to allow those players who don't want to engage with this system an easy way to avoid it: simply only ever level up Small Arms and you won't need to worry about this system. Everything will work as it always has.


Sleet01 (approved):
LGTM

PR #7750, opened 2025-10-17
Fix #7746: Overhauled Transportation Costs to Bring Them in Line With CamOps

### LINKED ISSUE #7746: [Issue] Jump costs appear to be too high
### Brief Description *

When transiting through the galaxy, the jump cost seems to be higher than intended. Based on preliminary testing from Illiani it appears to have approximately doubled since 50.05.

### 3. Steps to Reproduce

Any type of interplanetary transportation in the latest builds appears to be high, create a force and check the estimated jump costs. Create an identical force in 50.05 and potentially 50.06 and compare the jump costs.

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

v0.50.10-nightly-2025-10-15

### Operating System *

Windows 11

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.14

### Final Verification

IllianiBird:
Fix #7746

This PR updates our current transportation cost mechanics to bring them in line with CamOps.

## Transport Costs
Previously, we were using home-brew mechanics with some nods given to the official rules. The original logic was flawed which prompted an attempt to fix the logic during the 06 cycle. However, this had the unintended consequences of more than doubling all transport costs.

This PR approaches the issue by using the rules outlined on pg 43 of CamOps. Specifically the abstract (or simplified, if you will) costs. This side-steps the issue of needing to worry about availability or what specific units are being used to transport a force.

Some concessions were made in the name of simplicity. We assume that the players' forces are being transported on a fleet of magical DropShips. Those DropShips will always have exactly the bays the player needs at any given time. Though all magical DropShips are limited to 14 bays.

A similar approach has been given to cargo transportation. With 1874.5 tons of cargo being transported per DropShip. This value was derived from one of the Union types, rather than the mammoth because I am mean and wanted to inflate the transport price a bit. If people want to split hairs this can be logic'd as the player not being important enough to get their own fleet of cargo transports.

For passengers - i.e. civilians and support personnel without units - we use further abstraction. The principle issue is that the only canon DropShip with a reasonable amount of passenger space is the Princess Luxury Liner. And, frankly, hiring one of those dwarfed the cost of most other transport alone. So instead we turn back to our fleet of Magical DropShips. We know travel exists in BattleTech and we can safely assume not everyone is scooting about in a Luxury Liner. Therefore we take the 200 passengers the Princess can transport, divide that by the number of bays in our Magical DropShip and that gives us roughly 15 passengers per bay. A little less than an Infantry Platoon bay. Presumably because they get nicer quarters.

## Paying for Transportation
Previously we handled transportation fees on a per-jump basis. This isn't backed up by CamOps, however. Instead transport fees are handled on a per week (or in some cases per month) basis. Through the power of Java we're able to effortlessly extrapolate these down to the day. This allows us to be very accurate in our transport fees. As well as suitably increasing pay for very long jumps where there is significant time spent recharging.

However, moving away from per-jump fees means that we cannot process fees on new day. Instead, all jump transactions occur when the route is confirmed. A 'are you sure' dialog appears to stop players from accidentally committing to bad routes.

Paying in advance also makes sense from an in-universe perspective. You hire transport for the duration of the journey, not per jump.

Finally, this change removes the really annoying situation where you run out of money mid-route. Allowing us to disable one of the many new day nags.

## Contract Pay
Moving onto contract pay, specifically the transport portion. I introduced a campaign option removing 'two way' transport pay. If two-way pay is disabled the employer will only pay to get the player to the site of the contract. In-lore we know that employers would pay both ways, as the mercenary would use places like Galatea and Outreach as a base of operations.

However, in MekHQ the sheer volume of contract opportunities removes this need. Plus MekHQ doesn't currently allow for splitting a campaign between locations. This resulted in players using reimbursed transport pay as one of their main sources of income. Oftentimes the gains from pocketing half of the transport payment would dwarf the actual contract pay. Especially for larger campaigns.

Finally, I removed 'transit pay'. This cited a lot of CamOps but honestly, I found no basis for this contract payment anywhere in CamOps.


IllianiBird:
This is now complete. Flipping to live. It might look scary but most of the change count is tests

MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -663,36 +663,24 @@ public void calculateContract(Campaign campaign) {
...
-        if (campaign.getCampaignOptions().isUsePeacetimeCost()
-                  && campaign.getCampaignOptions().getUnitRatingMethod().equals(UnitRatingMethod.CAMPAIGN_OPS)) {
-            // contract base * transport period * reputation * employer modifier
-            transitAmount = accountant.getContractBase()
-                                  .multipliedBy(((getJumpPath(campaign).getJumps()) * 2.0) / 4.0)
-                                  .multipliedBy(campaign.getAtBUnitRatingMod() * 0.2 + 0.5)
-                                  .multipliedBy(1.2);
-        } else {
-            transitAmount = Money.zero();
-        }
```
IllianiBird, line 690: I meant to comment on this when I posted this PR. I have no idea what these comments are referring to. Transit pay does not exist in CamOps. And I looked in several places.

HammerGS: I think this is a legacy peace of the original peace time operation costs.

IllianiBird: We got to the bottom of the situation, eventually. And yeah, it's just never explicitly called out. I've restored this code, but once AtBContract is retired I fully desire to just rewrite this whole handling.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
```diff
@@ -1236,6 +1237,7 @@ public CampaignOptions() {
         setContractMarketReportRefresh(true);
         setContractMaxSalvagePercentage(100);
         setDropShipBonusPercentage(0);
+        isUseTwoWayPay = true;
```
psikomonkie: The other values are set via the `set` method, should this also use `setUseTwoWayPay`?

HammerGS (approved):

PR #7788, opened 2025-10-21
Fix #7783: Corrected Large Vessel Ammo Reloading

### LINKED ISSUE #7783: [Issue] 50.07 10-20 nightly.  Dropships cannot reload partially spent ammo bins without using the swap ammo menu to mark them empty.
### Brief Description *

After fights involving dropships, both in air and on the ground, I have not been able to reload their ammo bins, unless they managed to spend more than a full ammo bin.  Looking at them in mekhq, they show that they have reduced ammo, but they do not show up in the repair menu to reload.  If I right click on the dropship in the hangar, then go to swap ammo, a menu comes up, and I can reduce the amount of ammo until it is 0.  After that, I can go to the repair menu and reload the weapon to full.

### 3. Steps to Reproduce

1. Open the save.
2. Go to the overlord in the hangar, confirm that it has some ammo missing from lrm and AC5.
3. Go to the repair bay, verify that the dropship is not available for repairs.
(optional)
4. Go to the hangar again, right click the dropship and select "swap ammo" and reduce the ammo in those bins to 0
5. Go to the repair bay and the dropship should now be available to reload.

### Attach Files

[Dropshipreload.zip](https://github.com/user-attachments/files/23012593/Dropshipreload.zip)

### Severity *

### MekHQ Suite Version *

50.07 Nightly 10-20

### Operating System *

Windows 10

### Java Version *

17.0.14

### Final Verification

IllianiBird:
Fix #7783

We were incorrectly ceiling the return value. This created problems if the ammo used wasn't in increments of whole tons. As the bin would think it was full - as it would round up to the ton.

### DIFF
--- a/MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
+++ b/MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
@@ -167,7 +167,7 @@ public class LargeCraftAmmoBin extends AmmoBin {
      * Gets the unused capacity of the bay, in tons.
      */
     public double getUnusedCapacity() {
-        return getCapacity() - Math.ceil(getCurrentShots() * ammoTonnage / getType().getShots());
+        return getCapacity() - (getCurrentShots() * ammoTonnage / getType().getShots());
     }
 
     @Override


MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
```diff
@@ -167,7 +167,7 @@ public double getCapacity() {
      * Gets the unused capacity of the bay, in tons.
      */
     public double getUnusedCapacity() {
-        return getCapacity() - Math.ceil(getCurrentShots() * ammoTonnage / getType().getShots());
+        return getCapacity() - (getCurrentShots() * ammoTonnage / getType().getShots());
```
psikomonkie, line 170: Was this possibly intentional - where else is this used? 

If a shot is used from a ton of ammo, would that mean the "unused capacity" would be the other x many _whole_ tons?

IllianiBird: > Was this possibly intentional - where else is this used?

So I considered whether this might be intentional, as it's remained unchanged since 2017. However, it's only used in 1 place and that's checking whether a large vessel ammo bin needs to be reloaded. If it were used in a bunch of places I'd say it would probably be intentional, but single use - which is currently bugged? That tells me either it's been broken for years - not impossible, given the save of large vessels in mhq - or we changed something elsewhere, which exposed the bug. I'll comment below why I do think this is a bug.

> If a shot is used from a ton of ammo, would that mean the "unused capacity" would be the other x many whole tons?

I'm not wholly certain what you mean by this. However, I'll explain why I think using `ceil` is a bug and _maybe_ that'll answer your question. If not, feel free to reply with simple words. :D 

Basically, take a theoretical ammo bin. That ammo bin weighs 1 ton for simplicities sake, with each shot weighing just 0.1t (so ten shots total). If you have 2 shots: `Used = 2 * 0.1 = 0.2t`. With `Math.ceil(0.2)` you end up at `1.0t` so the bin thinks its full. Without Math.ceil you get `0.2t` (the actual unused capacity).

psikomonkie: > Basically, take a theoretical ammo bin. That ammo bin weighs 1 ton for simplicities sake, with each shot weighing just 0.1t (so ten shots total). If you have 2 shots: Used = 2 * 0.1 = 0.2t. With Math.ceil(0.2) you end up at 1.0t so the bin thinks its full. Without Math.ceil you get 0.2t (the actual unused capacity).

The method was returning `capacity - Math.ceil(0.2)`, so the method was returning 0, right? Which makes sense for unused capacity - Your 1 ton of ammo _was_ used (just partially, not fully). Should it be returning full tons or partial? IDK, the way the method that utilizes this - `LargeAmmoCraftBin::bayAvailableCapacity` it almost words it like it's looking for full-ton increments:
```    
* to load more ammo into this bin. In the case of an ammo swap some ammo may need to be removed from another bin in
* this bay before more can be loaded.
```

IllianiBird: So, the method is returning unused capacity. If we have `capacity - Math.ceil(0.2)`, ceil will change 0.2 to 1.0 which If capacity is 1 will return 0 - as you say. Essentially telling mhq that there is no unused capacity, therefore the bay doesn't need reloading. Leading us to the bug

psikomonkie: The methods used when `fix()`ing the `LargeCraftAmmoBin` is `loadBinSingleTon`, but it doesn't look like it actually needs to load it in single ton increments... I'm leaving this up to you.

psikomonkie (approved):
Nice! I'm still curious about why it seems like this was geared towards whole ton ammo increments...

MekHQ/src/mekhq/campaign/parts/equipment/LargeCraftAmmoBin.java
IllianiBird: image

PR #7813, opened 2025-10-23
Issue #7080: Bottom-Up TO&E

### LINKED ISSUE #7080: [RFE] Bottom up TO&E parsing
### Enhancement Type *

New Feature

The idea is auto label TO&Es from the bottom up rather than the current top down.

### Use Cases or Rationale

The main reason is many large formations have odd smaller units attached to them at the top level leading to a lot of weirdness in the identity of units in a TO&E.

For example. A clan Cluster will have 3-5 Bi/Trinaries under it with an attached command star. That command star will be labled as a binary/Trinary as well.

A clan Nova should have either 2 stars or 5 points under it. But those become invalid.

Inner sphere regiments can have artillery companies, command lances, Aero wings, and all sorts of things attached outside the three battalions.

At the basic levels if the unit type is IS the bottom of any branch should probably be a lance by default, and a clan should be a star. Though those should be defaults. If the size overrides exist at the bottom we can also do points.

If those lances or stars are contained in a parent unit, they should be Bi/Trinaries or Companies, and so forth.

### Attach Files

_No response_

### MekHQ Suite Version *

.06

### Final Verification

psikomonkie:
Fixes #7080

TO&E will parse from the bottom up instead of top down. 

TO&E will now start with the subforces and work its way up. Parent formations will be at least one level higher than child formations. 

Existing formations that weren't doing anything weird should be unaffected.

For example, this will now allow:
- A (support) company directly attached to a regiment
- Bottom-level forces of 12 units are referred to as "Company"s instead of lances, etc.
- Lances can have a sub-force ("Team")

Some screenshots:
A company now listed as a company: 
image

A "Team" directly underneath of a "Battalion" (not possible before):
image

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/Force.java
+++ b/MekHQ/src/mekhq/campaign/force/Force.java
@@ -249,7 +249,11 @@ public class Force {
     }
 
     public void setOverrideFormationLevel(final FormationLevel overrideFormationLevel) {
-        this.overrideFormationLevel = overrideFormationLevel;
+        if (overrideFormationLevel == FormationLevel.REMOVE_OVERRIDE) {
+            this.overrideFormationLevel = FormationLevel.NONE;
+        } else {
+            this.overrideFormationLevel = overrideFormationLevel;
+        }
     }
 
     public int getScenarioId() {


MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -250,7 +249,11 @@ public FormationLevel getOverrideFormationLevel() {
     }
 
     public void setOverrideFormationLevel(final FormationLevel overrideFormationLevel) {
-        this.overrideFormationLevel = overrideFormationLevel;
+        if (overrideFormationLevel == FormationLevel.REMOVE_OVERRIDE) {
+            this.overrideFormationLevel = FormationLevel.NONE;
+        } else {
+            this.overrideFormationLevel = overrideFormationLevel;
+        }
```
psikomonkie, line 256: This feels weird but IDK how "Remove Override" worked before. During testing it was just setting the formation level to "Remove Override" which looked... Wrong.

IllianiBird: I'll be a liar if I said I remembered. It has been a _very_ long time since I wrote autoAwards - which this was set up to support.

IllianiBird (approved):
This looks fine to my eyes and if it tests ok, I saw let's get it merged.

PR #7816, opened 2025-10-24
Improvement: Rewrote Morale System Based on Player Feedback

IllianiBird:
So a semi-frequent point of user contact over the past half-year has been that they don't feel like the enemy's morale is suitably reacting to their victories. We recently put out a user poll to get a sense of the situation and the results are telling:

image

Taking a look at the code and, frankly, it was a broken mess. I can say that, because I was the one who wrote the code. What I was smoking, however, is probably a question best left unanswered.

Here's some of the noteworthy bugs:
- Clans were always being considered A* rated, no matter their actual quality
- Refusing an Engagement was always being considered a Decisive Defeat, even if the player deployed units to benefit from the Fleet in Being Effect
- The performance modifiers were entirely backwards. That means that the better the player was doing, the harder it would be to break the enemy's morale.

All this combined to create a very frustrating experience for the player.

This PR does the following:
- Fixes all of the above bugs.
- Moves the morale code out of AtBContract.
- Cleans the code up and rewrites it where necessary.
- Makes the morale check roll player-facing.
- Improves the reactivity of morale, so that it changes more often.
- Removes the old documentation, in favor of keeping things in the Glossary, instead. This makes it more accessible for the player.
- Adds a pile of campaign options, so the player can customize their experience. Making their contracts harder or easier as they see fit.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/MHQMorale.java
+++ b/MekHQ/src/mekhq/campaign/mission/MHQMorale.java
@@ -239,7 +239,7 @@ public class MHQMorale {
      * @author Illiani
      * @since 0.50.10
      */
-    private static PerformanceOutcome getOutcome(int decisiveVictoryModifier, int victoryModifier,
+    static PerformanceOutcome getOutcome(int decisiveVictoryModifier, int victoryModifier,
           int decisiveDefeatModifier, int defeatModifier, int performanceModifier) {
         if (performanceModifier == decisiveVictoryModifier) {
             return PerformanceOutcome.DECISIVE_VICTORY;
@@ -264,7 +264,7 @@ public class MHQMorale {
      * @author Illiani
      * @since 0.50.10
      */
-    private static int getReliability(AtBContract contract) {
+    static int getReliability(AtBContract contract) {
         int reliabilityModifier;
 
         int quality = contract.getEnemyQuality();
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -710,7 +604,9 @@ public int getScore() {
         for (Scenario scenario : getCompletedScenarios()) {
             // Special Scenarios get no points for victory and only -1 for defeat.
             if ((scenario instanceof AtBScenario) && ((AtBScenario) scenario).isSpecialScenario()) {
-                if (scenario.getStatus().isOverallDefeat() || scenario.getStatus().isRefusedEngagement()) {
+                if (scenario.getStatus().isOverallDefeat() ||
+                          scenario.getStatus().isRefusedEngagement() ||
+                          scenario.getStatus().isFleetInBeing()) {
```
IllianiBird, line 609: This is old AtB legacy content. So these values are just being added to make sure everything still works for anyone still on AtB

IllianiBird:
> Would you be able to set up unit tests to cover the morale changes? I think it will help with maintaining the morale system.

Done. Thanks for the push :)

HammerGS (approved):

PR #7817, opened 2025-10-24
Fix: Fixed Quick Train Failing on Unknown Skills

IllianiBird:
This fixes a bug with the new Quick Train mechanic where the system will break when trying to train an unknown skill.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
+++ b/MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
@@ -193,7 +193,7 @@ public class QuickTrain {
                     continue;
                 }
 
-                person.spendXP(improvementCost);
+                person.spendXPOnSkills(campaign, improvementCost);
 
                 PerformanceLogger.improvedSkill(
                       isLogSkillGain, person, today, skillName, skill.getLevel());


MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
```diff
@@ -233,7 +241,12 @@ private static void processSkills(Person person, boolean isAdminsHaveNegotiation
 
         // Sort the skills by their total skill level (lowest -> highest)
         targetSkills.sort(
-              Comparator.comparingInt(skillName -> person.getSkill(skillName).getTotalSkillLevel(options, attributes))
+              Comparator.comparingInt(skillName -> {
+                  Skill skill = person.getSkill(skillName);
+                  return (skill == null) // The character doesn't have this skill
+                               ? Integer.MIN_VALUE // Unknown skills should always be trained first
+                               : skill.getTotalSkillLevel(options, attributes);
```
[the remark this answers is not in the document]
IllianiBird: If someone ends up with a skill worse than -2,147,483,648 something has gone terribly wrong

MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
```diff
@@ -185,6 +187,12 @@ private static void handleSkillTraining(int targetLevel, Campaign campaign, bool
 
                 // Refresh skill info after improvement
                 skill = person.getSkill(skillName);
+                if (skill == null) {
+                    // Something went wrong; the skill should have been improved
+                    LOGGER.warn("Failed to improve skill {} for person {}: skill was null after improvement", skillName,
```
psikomonkie: It looks like this warning comes after the XP would've been spent. Is that happening when encountered? Should this be `error` level?

IllianiBird: Good shout. Done.

Also integrated the new `spendXPOnSkills` method from the Veterancy SPA PR

PR #7827, opened 2025-10-24
Fix #7825: Capped Turnover Desirability Modifier to Elite

### LINKED ISSUE #7825: [Issue] [0.50.07] Difficulties with balancing personnel retention
### Brief Description *

Before, difference between retention TN for your best of the best (Elite) and Regular personnel from Desirability modifier would be only 2, which worked well for TN numbers - Elites are more likely to quit than Regulars but not exceedingly so, and Loyalty/officer modifiers could mostly make up for it. However, with 0.50.07 my experience with balancing personnel retention seems to have changed considerably, primarily due to two factors: two new skill levels (Heroic and Legendary) and more AToW skill modifiers from AToW traits and such (meaning more personnel that can suddenly exceed their peers in skill because of good traits/age).

To elaborate on the first factor, the difference between your best people (Legendary) and Regular is no longer +2 TN but rather +4 TN. That means that if you balance the base TN to provide a challenging but reasonable TN to the top people, Regulars almost never quit. And if you balance it for Regulars to still have a chance of quitting, Legendary people will leave you in a year or two (and those are your most renowned veteran personnel of potentially many decades). Example: settings base TN of 2 for Regulars would result in a rather high base TN of 6 for Legendary save other modifiers, and settings with a lower base TN of 4 for Legendary would result in a trivial base TN of 0 for Regulars.

Imho, it would be good to either lower the effect on Desirability from a step in rank (for example +0.5 per rank) or make that a customisable parameter. That way the spread of TNs across the company can be changed and not just shifting it along with the base TN.

### 3. Steps to Reproduce

_No response_

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

0.50.07

### Operating System *

Windows 11

### Java Version *

Eclipse Adoptium 17.0.13

### Final Verification

IllianiBird:
Fix #7825

See original case for a thorough explanation of the issues.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/RetirementDefectionTracker.java
+++ b/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/RetirementDefectionTracker.java
@@ -33,9 +33,11 @@
  */
 package mekhq.campaign.personnel.turnoverAndRetention;
 
+import static java.lang.Math.min;
 import static java.lang.Math.round;
 import static mekhq.campaign.personnel.Person.getLoyaltyName;
 import static mekhq.campaign.personnel.PersonnelOptions.ADMIN_MEDIATOR;
+import static mekhq.campaign.personnel.skills.SkillType.EXP_ELITE;
 import static mekhq.campaign.personnel.turnoverAndRetention.RetirementDefectionTracker.Payout.isBreakingContract;
 
 import java.io.PrintWriter;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/RetirementDefectionTracker.java
```diff
@@ -181,21 +183,19 @@ public Map<UUID, TargetRoll> getTargetNumbers(final @Nullable Mission mission, f
             // Desirability modifier
             if ((campaign.getCampaignOptions().isUseSkillModifiers()) &&
                       (person.getAge(campaign.getLocalDate()) < RETIREMENT_AGE)) {
-                targetNumber.addModifier(person.getExperienceLevel(campaign, false) - 2,
+                targetNumber.addModifier(min(EXP_ELITE - 2, person.getExperienceLevel(campaign, false) - 2),
                       resources.getString("desirability.text"));
             }
 
             // Recent Promotion Modifier
-            if (campaign.getCampaignOptions().isUseSkillModifiers()) {
```
IllianiBird, line 189: I don't know why this was set up to check for this campaign option. I didn't add this, but it seems like a bug.

IllianiBird:
Honestly, I would prefer for Turnover to receive a full update. I've mentioned it before, but fixing Retirement (as it was known, back then) was the first thing I did on the project - long before I was a dev. Plenty of stuff I'd do differently now, not least of which just how many modifiers there are for the player to track. Add to that a pretty lackluster selection of tools to manage Turnover and yeah, not a fan.

I might do a design doc, because I don't think it would take long to redo.

PR #7840, opened 2025-10-26
Improvement: Updated Random Marriage System to Better Reflect Character Sexualities Rather Than Treating All Characters As Either Asexual or Bisexual

IllianiBird:
First, a story.

Whenever I have the time I like to tune into folks streaming mhq campaigns. This has proven time and again to be highly valuable. Whether it's finding bugs, odd interactions, or hearing players come up with ideas for improvements.

During the start of a recent stream - where viewers could submit characters to be in the campaign - one of these viewer submitted characters entered into a relationship. From the players' reaction, the gender the character picked clearly did not match how they viewed their character. This was a poor play experience. It completely cut in the face of who the player imagined their character to be and it was clear that player was not having a good time.

We want players to grow attached to their characters. To do so we need to make sure that we have tools that allow them to express those characters fully.

Prior to this PR all characters fell into two 'pools'. They either have no interest in marriage, or they were willing to marry anyone RNG picked. There was _some_ control. For example, we had a 'same-sex marriage chance' option. This option rolled a 'dice' any time a marriage would occur and if it rolled a certain result that marriage would be same-sex.

This PR adjusts this mechanism. Now, when characters are created we use a bundle of Campaign Options to determine that characters' sexuality. Chiefly, whether they prefer male (inc. non-binary 'male') characters, female (inc. non-binary 'female') characters, both, or neither.

These random sexualities can be easily changed by players using the character right-click menu in the Personnel tab.

Existing characters will have their sexualities 'rolled' the first time the campaign is loaded into 0.50.10. If same-sex marriage was previously disabled no gay characters will be generated. If 'no interest in marriage' was previously disabled no asexual/aromantic characters will be generated.

If the campaign has _not_ previously been saved in 50.10 (i.e. a campaign not from a Nightly) an additional step is taken. Following campaign upgrading we quickly loop through all personnel to ensure that they are attracted to their current spouses' gender. This is not possible on saves from a 0.50.10 nightly due to code limitations.


MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -380,10 +381,12 @@ public class Person {
     private boolean founder; // +1 share if using shares system
     private boolean immortal;
     private boolean quickTrainIgnore;
-    // this is a flag used in determine whether a person is a potential marriage
-    // candidate provided
-    // that they are not married, are old enough, etc.
+    // this is a flag used in determine whether a person is a potential marriage candidate provided that they are not
+    // married, are old enough, etc.
+    @Deprecated(since = "0.50.10", forRemoval = true)
     private boolean marriageable;
```
IllianiBird, line 397: So, a quick note. Right now we're hitchhiking on this variable as a way to confirm whether a character has been updated to the new gender preference system. A future scheme I have in mind is to add an extra step to relationships so that characters start dating _before_ moving to marriage. When that's implemented there will be a big update to how we handle marriage. Including a new 'Interested in Marriage' flag.

HammerGS (approved):

PR #7874, opened 2025-10-29
Improvement: Restored Ability to Deploy from TO&E; Added Ability to Deploy to StratCon from Briefing Room

IllianiBird:
This PR adds the ability to deploy to StratCon scenarios directly from the Briefing Room. Simply right-click on the scenario and off you go.

It also restores the ability to deploy from the TO&E via a handy button. The TO&E deployment button works for all scenario types, not just StratCon scenarios.


Sleet01 (approved):
LGTM.  Very nice!

PR #7877, opened 2025-10-29
Improvement: Added the Ability to Change Combat Roles in the TO&E; Forces Will Now Remember Their Last Assigned Combat Role; Contract Finances Updated to Better Account for Combat Roles

IllianiBird:
This PR does the following:
- Combat Teams (and Forces) will remember their last assigned Combat Role.
- Combat Teams can have their Combat Role assigned in the TO&E
- When arriving at a contract all Combat Teams will be automatically assigned to their last Combat Role.
- Contract pay excludes Combat Teams assigned to non-combat roles. Your employer is paying for your combat forces, not your training battalion*. This is a double-down on the combat forces only considerations made earlier in the year.

*Cadre Duty is going to be updated in a follow-up PR.


Sleet01 (approved):
LGTM

PR #7878, opened 2025-10-29
Improvement: Added 'Cadre' Combat Role to Distinguish Between Forces Assigned to Train Allied Forces & Those Assigned to Train Player Forces

IllianiBird:
This PR adds a new Combat Role 'Cadre'. This role has no mechanical effects. It exists purely for Cadre Duty contracts.

This distinction was made because Cadre Duty contracts ended up becoming a free meal ticket for the player. Not only were they being paid for the contract, but they were also able to use the entire contract to train up their second line forces. This was due to that contract type requiring the Training combat role.

Not only was this a free meal ticket. It was also annoying as, if players had no second line forces to train, they would be spammed weekly with notifications that their forces had nothing to learn.

So this PR was introduced partly to address balance, partly as QoL, and partly for immersion: your employer isn't paying you to train _your_ forces.

A special handler was added. At the end of a Cadre Duty contract any forces assigned to the Cadre role will be automatically reassigned to Frontline. This is to avoid a scenario where #7877 remembers the assignment and consequently stops paying the player for one or more forces.

### DIFF
--- a/MekHQ/unittests/mekhq/campaign/mission/AtBContractTest.java
+++ b/MekHQ/unittests/mekhq/campaign/mission/AtBContractTest.java
@@ -524,6 +524,7 @@ public class AtBContractTest {
             when(mockForce.getFormationLevel()).thenReturn(FormationLevel.INVALID);
             when(mockForce.getAllChildren(mockCampaign)).thenReturn(mockUnits);
             when(mockForce.getAllUnits(anyBoolean())).thenReturn(mockUUIDs);
+            when(mockForce.getCombatRoleInMemory()).thenReturn(CombatRole.FRONTLINE);
 
             forceId = getNextForceId();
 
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/force/Force.java
```diff
@@ -256,6 +258,14 @@ public void setOverrideFormationLevel(final FormationLevel overrideFormationLeve
...
     }
 
+    public CombatRole getCombatRoleInMemory() {
+        return combatRoleInMemory;
+    }
+
+    public void setCombatRoleInMemory(final CombatRole combatRoleInMemory) {
+        this.combatRoleInMemory = combatRoleInMemory;
+    }
+
```
IllianiBird, line 268: The changes in this class are duplicated from #7877. I forgot to remove them before pushing and removing them now would just create conflicts.

Sleet01 (approved):
LGTM

IllianiBird:
> Should auto awards for training be awarded to Cadre duty mentors?

I decided no, at time of pushing; but with the benefit of time I do agree. I'll see it done.

PR #7879, opened 2025-10-30
Improvement: Updated StratCon Scenario Generation Tempo; Introduced Morale-Based Crisis Scenarios

IllianiBird:
A few months ago we updated StratCon's scenario generation tempo. The intention was to create situations where the player was encouraged to deploy half-repaired forces. To create the kinds of desperation moments we read about in the novels.

It worked for a grand total of one release. Afterwhich it became business as usual.

Why did it only have such a short spanned effectiveness? Simple: more scenarios means more salvage. More salvage means bigger player campaigns. Bigger campaigns can more readily handle the increased tempo, leading to more victories, leading to more salvage.

Once the playerbase became used to the updated tempo it acted as a force multiplier to their campaign economics. Essentially worsening the march towards the 'post-scarcity' phase of a campaign. Literally the opposite of what we've been trying to do these part months.

This PR comes in an makes two changes:
- Morale no longer affects scenario spawn changes (except for failed reinforcement attempts). This takes scenario generation back to its' pre-tempo update state.
- Instead Morale now influences the chance a scenario will spawn as a 'Crisis'.

Crisis scenarios were introduced earlier this year. A Crisis scenario acts like a Turning Point, in that losing the scenario reduces victory points by 1. However, unlike a Turning Point, Crisis scenarios do not award a Victory Point on victory. Essentially, this creates scenarios the player _has_ to complete.

Overall this should address balance in a few ways. By reducing scenario tempo we reduce salvage on the field. By creating 'you must do this' scenarios we increase the likelihood the player will need to do a scenario they would have otherwise skipped. Such as no salvage scenarios, or scenarios where they are out-gunned. Finally, it reduces the issue of players destroying entire regiments of stuff every contract. Now, scenarios will be fewer (and due to changes in 50.07, smaller).

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/Scenario.java
+++ b/MekHQ/src/mekhq/campaign/mission/Scenario.java
@@ -95,6 +95,7 @@ public class Scenario implements IPlayerSettings {
     private String desc;
     private String report;
     private ScenarioStatus status;
+    private boolean isCrisis;
     private LocalDate date;
     private List<Integer> subForceIds;
     private List<UUID> unitIds;
@@ -168,6 +169,7 @@ public class Scenario implements IPlayerSettings {
         desc = "";
         report = "";
         setStatus(ScenarioStatus.CURRENT);
+        isCrisis = false;
         date = null;
         subForceIds = new ArrayList<>();
         unitIds = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


Sleet01 (approved):
LGTM

PR #7883, opened 2025-10-31
Feature: Added CamOps Salvage Rules

IllianiBird:
# Requires [Task: Added Salvage Related Methods to Entity & Children](https://github.com/MegaMek/megamek/pull/7552)

This PR addresses a longstanding issue with salvage, where players are able to grab as much salvage as they want without investing any resources into the gathering.

Through this PR we are implementing close to RAW salvage rules, specifically those from CamOps. The differences between the CamOps implementation and ours are minor with one exception. We do not currently support stripping parts off of units in the field.

~~Other differences include us not allowing salvage operations by 'Crippled' units (including those crippled by crew deaths).~~ (removed prior to merging) We have also introduced a couple of non-RAW values into the recovery time table. Specifically we have added a 10 minute recovery time for conventional infantry salvage (as this was completely absent in the salvage table). We have also added a 40 minute 'catch-all' recovery time for any other unit not covered by the table.

The only other difference is that we do not restrict vehicles to towing only other vehicles. This could have been easily done, but I struggled to find a good way to communicate this to the player. While it also introduced a few logic hurdles that I wasn't sad to be able to skip. Plus, if _Grayson "Death" Carlyle_ can do it, so can the players.

The final change to strict RAW is that being assigned to a scenario as a salvage tech depletes that techs' minutes for the day. This isn't directly called out in the manual, but logic dictates a character can't be in two places at once.

All of these rules changes are supported by two campaign options. The first enables-disabled CamOps Salvage. The second introduces 'Risky Salvage'. In CamOps, under the field stripping salvage rules, there is a fumble effect. If the Tech field stripping parts from a salvaged unit rolls snake-eyes they take 1d6 Hits.

With Risky Salvage enabled we extend this to general salvage. Basically, for every unit removed from the field we roll 2d6. If we roll snake eyes a random tech assigned to the scenario takes 1d6 Hits (or Injuries, if Advanced Medical is also enabled).

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
+++ b/MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
@@ -277,7 +277,7 @@ public class CamOpsSalvageUtilities {
         final boolean useInjuryFatigue = campaignOptions.isUseInjuryFatigue();
 
         int injuryEvents = 0;
-        for (int i = 0; i <= numberOfSalvagedUnits; i++) {
+        for (int i = 0; i < numberOfSalvagedUnits; i++) {
             int roll = d6(2);
             if (roll == 2) {
                 injuryEvents++;


MekHQ/src/mekhq/gui/dialog/SalvagePostScenarioPicker.java
```diff
@@ -0,0 +1,1065 @@
...
+
+        return confirmed[0] ? resultHolder.groups : null;
+    }
+
+    /**
+     * This is not currently implemented. The purpose of this method is to allow future developers and easy access point
+     * to implement field stripping, without needing to grok where it should fit in the class (and gui).
+     */
+    private void fieldStrip(SalvageComboBoxGroup group) {
+        TestUnit targetUnit = group.targetUnit;
```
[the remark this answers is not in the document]
IllianiBird: Intentional

MekHQ/src/mekhq/campaign/mission/Scenario.java
```diff
@@ -638,6 +645,22 @@
         return force;
     }
 
+    public List<Integer> getSalvageForces() {
```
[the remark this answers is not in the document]
IllianiBird: Fixed

MekHQ/src/mekhq/campaign/mission/Scenario.java
```diff
@@ -638,6 +645,22 @@
...
+        return salvageForces;
+    }
+
+    public void addSalvageForce(int forceId) {
+        salvageForces.add(forceId);
+    }
+
+    public List<UUID> getSalvageTechs() {
```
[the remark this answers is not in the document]
IllianiBird: Fixed

MekHQ/resources/mekhq/resources/ForceType.properties
```diff
@@ -28,10 +28,13 @@
 # Microsoft's "Game Content Usage Rules"
 # <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
 # affiliated with Microsoft.
+# suppress inspection "UnusedProperty" for whole file
```
Sleet01, line 31: Man, I wish I understood why some property files are detected as used but others aren't.  Probably the way they're loaded...

MekHQ/src/mekhq/campaign/force/ForceType.java
```diff
@@ -47,24 +47,30 @@ public enum ForceType {
...
-    SECURITY(true, true);
+    SECURITY(3, true, true),
+
+    /**
+     * Salvage force type, typically used in post-scenario salvage operations
+     */
+    SALVAGE(4, true, true);
 
     // region Fields
+    private final int key;
```
psikomonkie, line 73: Should we have some kind of comment re: this and Force line +845 letting other developers know to not change the keys until the next major? Otherwise we may break saves if someone else makes a change to the keys for whatever reason.

MekHQ/src/mekhq/campaign/mission/camOpsSalvage/RecoveryTimeCalculations.java
```diff
@@ -0,0 +1,346 @@
...
+    private final static double PITCH_BLACK = 0.5;
+
+    final static double BASE_MULTIPLIER = 1.0;
+    private final static double DEFAULT_MULTIPLIER = 0.0;
+
+    /**
+     * Calculates the total recovery time for an entity based on environmental conditions.
+     *
+     * <p>This method evaluates all environmental factors present in the scenario and applies appropriate multipliers
+     * to the base recovery time. The calculation follows Campaign Operations rules, considering:</p>
```
psikomonkie: Can you include a page reference? (& I think this is SO not CO)

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -7102,4 +7103,26 @@ public List<String> checkForOverCrewing() {
...
+        }
+
+        boolean isMek = entity instanceof Mek;
+        if (!isMek) {
+            boolean hasCargoCapacity = getCargoCapacity() > 0;
+            boolean hasNavalTugAdaptor = isInSpace && CamOpsSalvageUtilities.hasNavalTug(entity);
+            canSalvage = hasCargoCapacity || hasNavalTugAdaptor;
+        }
+
+        return canSalvage && entity.getDamageLevel(true) < Entity.DMG_CRIPPLED;
```
psikomonkie: Why are units unable to salvage if they're considered crippled? 

A mek with internal damage to two torsos counts as crippled. If it still has arms and capacity, should it be able to help?

MekHQ/resources/mekhq/resources/ForceType.properties
IllianiBird: It's direct access vs. dynamic access.

If I reference the resource bundle and directly cite the key it'll be detected as used.

for example, `getTextAt(RESOURCE_BUNDLE, "misterToad.wildRide.getMeOff.text")` will be detected as used. Whereas `getTextAt(RESOURCE_BUNDLE, "disneyOwnsThis." + ownedProperty)` won't, because IDEA can't predict what `ownedProperty` will equal (spoiler alert: everything you love).

MekHQ/src/mekhq/campaign/force/ForceType.java
IllianiBird: So, I'm not sure about this. On one hand, I agree, on the other contributors  _should_ know to never change an access key. If I were to put a disclaimer here I'd need to basically put a disclaimer in every enum I create. It's like serialized enum names, there are just some things you need to know not to fuddle with.

MekHQ/src/mekhq/campaign/mission/camOpsSalvage/RecoveryTimeCalculations.java
IllianiBird: Sure thing, and it is CO, but yeah I'll add a pg

MekHQ/src/mekhq/campaign/unit/Unit.java
IllianiBird: So this is a relic from an earlier iteration and can be removed in favor of just directly checking whether the unit is fully crewed

Sleet01 (approved):
LGTM.
I did have a question about how we'd work Lift Hoists or Salvage Arms into cargo capacity computations, but that's not immediately pertinent.

MekHQ/src/mekhq/campaign/mission/camOpsSalvage/RecoveryTimeCalculations.java
```diff
@@ -0,0 +1,346 @@
...
+     *
+     * @param temperature The ambient temperature in degrees Celsius.
+     *
+     * @return The temperature multiplier to add to the total multiplier.
+     *
+     * @author Illiani
+     * @since 0.50.10
+     */
+    private static double getTemperatureMultiplier(int temperature) {
+        if (temperature > 50 || temperature < 30) {
```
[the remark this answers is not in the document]
IllianiBird: Wups, thanks Copilot!

MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
```diff
@@ -0,0 +1,403 @@
+package mekhq.campaign.mission.camOpsSalvage;
```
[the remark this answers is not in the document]
IllianiBird: Fixed

MekHQ/src/mekhq/gui/dialog/SalvagePostScenarioPicker.java
```diff
@@ -0,0 +1,1140 @@
...
```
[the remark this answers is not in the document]
IllianiBird: This is intentional. It's a placeholder method to make the life of whoever implements field stripping easier.

PR #7885, opened 2025-10-31
Fix: Fixed Contracts Sending You Halfway Around the Inner Sphere Because Your Jump Captain is Scared of Empty Systems

IllianiBird:
This fixes an issue where we were calculating contract market Jump Paths with a mind towards avoiding systems the player is outlawed in and empty systems. This created some very inflated transport clauses in contracts. As well as being a bit odd.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -6252,7 +6252,9 @@ public class Campaign implements ITechManager {
                 String neighborId = neighborSystem.getId();
 
                 // Skip systems without population if avoiding empty systems
-                if (isAvoidingEmptySystems && neighborSystem.getPopulation(currentDay) == 0) {
+                if (!skipEmptySystemCheck
+                          && isAvoidingEmptySystems
+                          && neighborSystem.getPopulation(currentDay) == 0) {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/mission/Contract.java
```diff
@@ -473,7 +473,7 @@ public boolean isActiveOn(LocalDate date, boolean excludeEndDateCheck) {
         if ((getCachedJumpPath() == null) || getCachedJumpPath().isEmpty()
                   || !getCachedJumpPath().getFirstSystem().equals(c.getCurrentSystem())
                   || !getCachedJumpPath().getLastSystem().equals(getSystem())) {
-            setCachedJumpPath(c.calculateJumpPath(c.getCurrentSystem(), getSystem()));
+            setCachedJumpPath(c.calculateJumpPath(c.getCurrentSystem(), getSystem(), true, true));
```
IllianiBird: This just tells the jump path calculator to not avoid empty systems and outlawing.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -6252,7 +6252,9 @@ public JumpPath calculateJumpPath(PlanetarySystem start, PlanetarySystem end, bo
                 String neighborId = neighborSystem.getId();
 
                 // Skip systems without population if avoiding empty systems
-                if (isAvoidingEmptySystems && neighborSystem.getPopulation(currentDay) == 0) {
+                if (!skipEmptySystemCheck
+                          && isAvoidingEmptySystems
+                          && neighborSystem.getPopulation(currentDay) == 0) {
```
IllianiBird, line 6257: This looks weird, so...

- `skipEmptySystemCheck` is a param that tells us to skip checking for empty systems before pathing through them
- `isAvoidingEmptySystems` is a boolean related to the 'avoid empty systems' option in the interstellar tab

So if we're not skipping the empty system check, and we _are_ avoiding empty systems _and_ it's an empty system; early exit. For, for this issue's resolution: if we're skipping the empty system check don't early exit this route.

PR #7901, opened 2025-11-01
Improvement: Reduced User Confusion by Removing Combat Teams Line from Contract Summary

IllianiBird:
As Combat Teams is an internal value only relevant to StratCon and not something users generally need to worry about, I went ahead and removed it from the Contract Market summary panel. Now only elements is displayed.


psikomonkie (approved):
Oh, nice!

PR #7903, opened 2025-11-01
RFE: Simple Cargo (Lift Hoists)

psikomonkie:
Depends on MegaMek/megamek#7556 and  MegaMek/megamek#7579, do not merge without the MegaMek PRs.

The masses yearn for the hoists!

Fixed some tests too. But the main "feature" is that `Unit::getCargoCapacity` will now consider Lift Hoists and Roof Racks. Hooray!

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1566,7 +1566,7 @@ public class Unit implements ITechnology {
     /**
      * @return the total cargo capacity of the entity. Returns 0.0 if the entity is not fully crewed.
      *
-     * @Deprecated create a more specific one like {@link #getCargoCapacityForConvoy} or
+     * @deprecated create a more specific one like {@link #getCargoCapacityForConvoy} or
      *       {@link #getCargoCapacityForSalvage}
      *       <br>
      *       Calculates and returns the cargo capacity of the entity based on its transport bays, mounted equipment, and


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -285,15 +286,13 @@ public void setCampaign(Campaign c) {
 
     /**
      * A convenience function to tell whether the unit can be acted upon e.g. assigned pilots, techs, repaired, etc.
-     *
```
IllianiBird, line 288: Those were my emotional support line breaks! D:

psikomonkie: Right? I finally turned on the automatic reformat file on save, so this is the consequences of that!

IllianiBird (approved):

psikomonkie:
It's failing because it's using the wrong MegaMek branch, that's fun.

MekHQ/src/mekhq/campaign/utilities/CampaignTransportUtilities.java
```diff
@@ -55,59 +57,62 @@ public class CampaignTransportUtilities {
...
+                return calcInfantryBayWeight(transportedInfantry);
             } else if (transporterType == INFANTRY_COMPARTMENT) {
-                return calcInfantryCompartmentWeight(transportedUnit);
+                return calcInfantryCompartmentWeight(transportedInfantry);
             }
         } else if (transporterType == TANK_TRAILER_HITCH) {
-            return transportedUnit.getWeight();
+            return transportedUnit.getTonnage();
+        } else if (transportedUnit instanceof Cargo) {
+            transportedUnit.getTonnage();
```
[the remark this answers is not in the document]
IllianiBird: I think the unit might be correct here, or at least I can't see what we're doing with this information.

IllianiBird (commented):
I think the bot might have picked up on something. Also, do you want to switch over the uses of `getCargoCapacity` to the new methods, or should I? Happy either way, just need to know whether to put in a follow-up PR.

psikomonkie:
> I think the bot might have picked up on something. Also, do you want to switch over the uses of `getCargoCapacity` to the new methods, or should I? Happy either way, just need to know whether to put in a follow-up PR.

Committed both the bot suggestions, they seem right. 

I switched over some of the `getCargoCapacity` - the ones I was confident in the purpose of. The deprecated method will still return the same values it has been returning to preserve functionality. The remaining usages of the deprecated method are because I didn't know which it was, so I just left it for the time being.

IllianiBird:
Cool, I figured that would be your play. I'm familiar enough with the uses that I can easily switch them over.

IllianiBird (approved):
Let's merge the outstanding mm PR and then update this branch (both mm and mhq). If all the tests pass at that point we're good to go.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1556,9 +1555,55 @@ public Money getSellValue() {
...
+    }
+
+    public double getCargoCapacityForConvoy() {
+        return getCargoCapacity(0, ForceType.CONVOY);
+    }
+
+    /**
+     * @return the total cargo capacity of the entity. Returns 0.0 if the entity is not fully crewed.
+     *
+     * @Deprecated create a more specific one like {@link #getCargoCapacityForConvoy} or
```
IllianiBird: This line is what's causing JavaDoc to fail. It needs to be `@deprecated` not `@Deprecated`

PR #7914, opened 2025-11-02
Improvement: Adjusted Healing Times on New Injuries to Make Them Less Oppressive

IllianiBird:
A small adjustment across all injury times. On review I don't think the player base would accept the original healing times.


Sleet01 (approved):
LGTM

PR #7916, opened 2025-11-02
Fix: Allowed Training Forces to Function When Using Mapless Mode

IllianiBird:
Training Forces were not awarding Training progression while using Mapless mode. This PR fixes that.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/education/TrainingCombatTeams.java
+++ b/MekHQ/src/mekhq/campaign/personnel/education/TrainingCombatTeams.java
@@ -129,8 +129,10 @@ public class TrainingCombatTeams {
             StratConCampaignState campaignState = contract.getStratconCampaignState();
             boolean isForceDeployed = campaignState != null &&
                                             campaignState.isForceDeployedHere(combatTeam.getForceId());
-            if (!isUsingStratCon || (!isUsingMaplessMode && !isForceDeployed)) {
-                continue;
+            if (isUsingStratCon) {
+                if (!isUsingMaplessMode && !isForceDeployed) {
+                    continue;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/education/TrainingCombatTeams.java
```diff
@@ -122,8 +123,13 @@ public static void processTrainingCombatTeams(final Campaign campaign) {
...
 
-            if (campaign.getCampaignOptions().isUseStratCon() &&
-                      !contract.getStratconCampaignState().isForceDeployedHere(combatTeam.getForceId())) {
+            CampaignOptions campaignOptions = campaign.getCampaignOptions();
+            boolean isUsingStratCon = campaignOptions.isUseStratCon();
+            boolean isUsingMaplessMode = campaignOptions.isUseStratConMaplessMode();
+            StratConCampaignState campaignState = contract.getStratconCampaignState();
+            boolean isForceDeployed = campaignState != null &&
+                                            campaignState.isForceDeployedHere(combatTeam.getForceId());
+            if (!isUsingStratCon || (!isUsingMaplessMode && !isForceDeployed)) {
```
psikomonkie: Before this change if someone wasn't using Stratcon it would get to the process training below this. 

After this change if someone isn't using stratcon, they will be skipped with the continue. 

`if (isUsingStratCon && (!isUsingMaplessMode && !isForceDeployed)) {` ?

IllianiBird: Ugh, I _knew_ it didn't look right.

Fixed

PR #7898, opened 2025-11-01
Fix #7897: Fixed Legendary MekWarriors Not Counting Towards CamOps Reputation

IllianiBird:
Fix #7897

This came about from a very early misunderstanding of the rules when translating mhq to CamOps. The original interpretation was that we should be comparing total skill level. However, skill levels are an ATOW thing, not TW. CamOps expected us to be comparing skill _target numbers_.

Under the original interpretation a total skill level of `0` would mean the character had no skill whatsoever. As opposed to `0+` meaning the character is the most skilled thing ever.

As we were still checking if the experience score was `>0` any legendary characters were being incorrectly treated as if they were completely unskilled.

This only affected MekWarriors. Proof that 'Meks really aren't the 'kings of the battlefield.'

As a secondary fix I also made sure to include a special handler for Combat Technicians, as they were being incorrectly included in the calculations. It's possible they were being excluded elsewhere - and I think they were - but a little insurance never hurt and adding a single conditional clause isn't going to hurt anyone.

PR #7923, opened 2025-11-02
Improvement: Changed 'Combat Technician' Profession to 'Support'

IllianiBird:
Previously we were classifying 'Combat Technicians' as a 'Combat' profession. This makes sense, given the title of the profession. However, while resolving #7898 I have swiftly come to the conclusion that having what is a special case support profession is going to cause us a lot of pain in the long run. Essentially, any time we'd want to check whether a character was support or combat we'd have to add a special clause to check whether they're actually a Combat Technician.

For the sake of our future sanities, this PR makes it so that Combat Technicians are a support role, like all the other technical professions.


psikomonkie (approved):
Nice!

PR #7929, opened 2025-11-03
Improvement: Updated Scroll Behavior of Immersive Dialogs

IllianiBird:
This PR updates ImmersiveDialogCore so that additional options will scroll with the text. This is useful for when you need to display substantial information beyond the original class' design. Such as in the new salvage system where the player has `n` salvage crews to choose from.


Sleet01 (approved):
LGTM

PR #7961, opened 2025-11-05
Fix: Added Missing Dialog Size Memory for Salvage Pickers

IllianiBird:
The Salvage Force Picker and Salvage Tech Picker dialogs weren't remembering their last opened size, forcing the player to resize them each time they were opened. i.e. before _every_ scenario. This was an awful play-feel. These dialogs now correctly remember their last opened size.


Sleet01 (approved):
LGTM.
I find your patience with our particular, peculiar mix of AWT and Swing quite inspiring.  Inexplicable, but inspiring :)

PR #7968, opened 2025-11-05
Improvement: Adjusted Deployment Date to Equal Battle Date to Improve User Experience

IllianiBird:
StratCon has a mechanic where the date a scenario takes place and the date a player needs to assign forces _to_ that scenario do not always match.

This is largely to represent the time it takes for a players' forces to get into position. However, this is not an intuitive system and we see semi-frequent user contact about it.

This PR adjusts things so that deployment day - by default - will be identical to battle day. I kept the variables separate to allow for developers to overwrite this when creating special scenarios and the like.


Sleet01 (approved):
LGTM

PR #7988, opened 2025-11-07
Fix: Fixed Loading of CamOps Reputation When Campaign is Rated as Ultra-Green

IllianiBird:
We were saving average experience using the enum's label (via an override toString) and not `name()`


Sleet01 (approved):
LGTM

PR #7990, opened 2025-11-07
Improvement: Adjusted Date Range for Random Start Year Button to Remove Pre-Star League Bias

IllianiBird:
Just narrowed the available years so that we don't have an extreme bias towards the Age of War.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -75,7 +75,7 @@ lblGeneralBody.text=<center><p>"War has been described as an art, as a science, 
 lblName.text=Unit Name
 lblName.tooltip=The most important decision you will ever make.
 lblRandomDate.text=<html>Randomize <span style="color:#7FCF43;">\u2606</span></html>
-lblRandomDate.tooltip=Pick a random date between 2850 & 3150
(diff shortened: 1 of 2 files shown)


MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
```diff
@@ -75,7 +75,7 @@ lblGeneralBody.text=<center><p>"War has been described as an art, as a science,
 lblName.text=Unit Name
 lblName.tooltip=The most important decision you will ever make.
 lblRandomDate.text=<html>Randomize <span style="color:#7FCF43;">\u2606</span></html>
-lblRandomDate.tooltip=Pick a random date between 2470 & 3151
+lblRandomDate.tooltip=Pick a random date between 2850 & 3150
```
HammerGS: Want to confirm something with this. 2850 means there will be no First Succession war or Amaris Civil war. Not any of the hidden war of the SL era. Is that the intent?

IllianiBird: It _was_ but I realize that's kinda dumb, so lemme fix it to 2775. That'll allow players to play in the downfall of the Star League, should they wish. We need to be careful how far bad we set the minimum, else we reintroduce an early era bias.

Sleet01 (approved):
LGTM

PR #7998, opened 2025-11-07
Fix: Fixed Edge Not Being Settable from Personnel Tab Right-Click Menu

IllianiBird:
We were incorrectly using the wrong minimum value for Edge.

While I was fixing this I also went ahead and introduced a minimum edge value constant so we're not defining the same value everywhere.


Sleet01 (approved):
LGTM

PR #8005, opened 2025-11-07
Improvement: Increased Recruitment Opportunities for Campaigns with Alternate Advanced Medical Enabled

IllianiBird:
Alt Advanced Medical is built around the idea that players will need to maintain a roster of 'spares', rather than just having their A-Team. When a character is injured they are usually out of action for a realistic, but extended period of time.

Player feedback for AAM has been overall positive, however one friction point is recruitment. Recruitment wasn't factoring in the increased need for bodies.

This PR doubles the number of applicants in the pool each month, if AAM is enabled. This should make it much easier for players to maintain their rosters. This may need further adjustment based on additional feedback, but should be a good starting point.


Sleet01 (approved):
LGTM

PR #8024, opened 2025-11-08
Improvement: Added a Campaign Option to Obstruct Delivery of Parts and Units While in Transit

IllianiBird:
This PR addresses a minor simulation issue: space amazon flying out to meet you mid-jump, even when in unoccupied systems.

When this campaign option is enabled, parts and units will not be delivered unless the campaign is planetside. Delivery time will still go down, while in transit, but will stop at 1 day. This is a gameplay concession, it can be logic'd as the players' Admin/Logistics arranging for the part/unit to be delivered to the destination.


Sleet01 (approved):
LGTM!

PR #8030, opened 2025-11-08
Fix: Fixed MASH & Field Kitchen Capacity Still Not Being Bypassed While in Transit

IllianiBird:
This updates the checks to be more robust, for MASH Kitchens, and to exclude a missed setter for Field Kitchens.


Sleet01 (approved):
LGTM

PR #8031, opened 2025-11-08
Fix: Fixed Engineers Being Able to Be Assigned to Salvage Operations

IllianiBird:
Engineers (techs assigned to self-crewed units) work really weirdly and break salvage. Due to the way they are not individual characters, but a 'face' of an entire team I simply couldn't get them to work. Far better to just exclude the whole lot of them.


Sleet01 (approved):
LGTM

PR #8044, opened 2025-11-09
Fix: Fixed Bug in Vehicle Profession Compatibility Handler

IllianiBird:
This fixes an issue in our vehicle profession compatibility handler. Units are parsed _after_ personnel. This meant that when we fetched the Entity associated with the unit associated with the person it would always return `null`. This caused the checker to use the fallback of Vehicle Crew/Ground.

This PR relocates this check to a point where all vehicles are guaranteed to have been processed.

### DIFF
--- a/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
+++ b/MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
@@ -448,30 +448,6 @@ public record CampaignXmlParser(InputStream is, MekHQ app) {
         LOGGER.info("[Campaign Load] Parts processed in {}ms", System.currentTimeMillis() - timestamp);
         timestamp = System.currentTimeMillis();
 
-        boolean skipAllDeprecationChecks = false;
-        boolean refundAllDeprecatedSkills = false;
-        for (Person person : campaign.getPersonnel()) {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -448,30 +448,6 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
...
-        boolean refundAllDeprecatedSkills = false;
-        for (Person person : campaign.getPersonnel()) {
-            // skill types might need resetting
-            person.resetSkillTypes();
-
-            // Seeing as we're already looping through all personnel, we might as well have the deprecation checks
-            // here, too.
-            if (!DEPRECATED_SKILLS.isEmpty() && !skipAllDeprecationChecks) {
-                // This checks to ensure the character doesn't have any Deprecated skills.
-                SkillDeprecationTool deprecationTool = new SkillDeprecationTool(campaign,
```
IllianiBird, line 461: We've moved this below the vehicle section below it so that we can be absolutely sure units have entities.

This section only concerns the deprecation of skills, so it being relocated isn't going to cause issues and means we don't need to loop through all personnel an extra time.

PR #8050, opened 2025-11-09
Fix: Fixed Filename Errors in Two Resource Bundles

IllianiBird:
`.properties.properties` -> `.properties`

This creates an NPE when enabling certain campaign options.


Sleet01 (approved):
LGTM

PR #8061, opened 2025-11-10
Improvement: Updated Tech Assignment Menus to Include Display of Maintenance Time Available & Required to Reduce Manual Menu Cross-Referencing

IllianiBird:
When a player wants to assign a tech to a unit they need the following information:

- Maintenance time required by unit
- Maintenance time available to tech

Both of these items are available in the gui, but in separate locations and rather hidden behind filters. This PR changes things so that when a player is assigning a tech to a unit this information will be immediately visible in the context menu.


Sleet01 (approved):
LGTM

PR #8062, opened 2025-11-10
Fix: Removed Unit Coloring from Personnel & Hangar Context Menus

IllianiBird:
We were having an issue with some context menus having options using black text. Checking into it we were incorrectly using the unit color method (that determines what color a unit should be in the hangar table) to color text in the context menus. This was a non-issue when everyone was using Light themes, but makes the option very difficult to read in dark themes.

I went ahead and removed the coloring, this should also make it easier on the eyes for players with colorblindness.


Sleet01 (approved):
LGTM

PR #8069, opened 2025-11-11
Improvement: Reworked Advanced Scouting Based on Player Feedback

IllianiBird:
So I've been keeping an eye on feedback surrounding AdScout and I've made a few changes based on it. There will be some players who don't think these changes went far enough. I've seen a couple of suggestions that pretty much amounted to an entirely new mechanic with significant code overhead. That's not really within scope for AdScout. AdScout was intended to be a quick implementation to address a couple of flaws in how StratCon encouraged force composition. At this point I've burned several dev days on it and I need to be done.

With that in mind, this PR changes the following:

- Introduced Ghost (+1 to Stealth), Loud Mouth (-1 to Stealth), and Ranger (+1 to Tracking) SPAs to cover some holes in our SPA coverage of scouting skills
- Sensor Geek now correctly applies its +2 modifier to Sensor Operations
- Enabling AdScout in an in-progress campaign will offer to give 25% of the players' combat roster a random scouting skill at level 1. Originally this was level 3, I ended up knocking it down a few notches based on play feel. Like the other freebies, this is meant to offer minimum viability not solve the players' problems
- If AdScout is enabled when the company generator is run, the above benefit will be applied automatically to the generated forces.
- The target number adjustment for <=35t units has been removed.
- Units <=35t now increase 'scan range' by 1. This allows the force to cover more ground, perfect for filling in the sector. This is explained in more detail in the updated Glossary entry.
- Patrol forces where all units are <=35t now have a daily 1-in-6 chance of completing their patrol early, allowing them to immediately return from the AO.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -3362,8 +3362,16 @@ public class StratConRulesManager {
             }
 
             if (force.getCombatRoleInMemory().isPatrol()) {
+                boolean allLightUnits = true;
+                for (Unit unit : force.getAllUnitsAsUnits(campaign.getHangar(), false)) {
+                    if (unit.getEntity() != null && unit.getEntity().getWeight() > 35) {
+                        allLightUnits = false;
+                        break;
+                    }
+                }
+
                 int roll = d6();
-                if (roll == 6) {
+                if (allLightUnits && roll == 6) {
                     forcesToUndeploy.add(forceID);
                     campaign.addReport(String.format(resources.getString("patrol.undeployed"), force.getName()));
                     continue;


MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -1501,83 +1497,115 @@ private static void scanNeighboringCoords(StratConCoords coords, int forceID, Ca
...
+                        continue;
+                    }
+                    scoutVisited.add(checkCoords);
+
+                    int nextDistance = distance + 1;
+                    if (nextDistance <= individualScanRange) {
+                        // Always enqueue so we can reach further rings,
+                        // even through already-revealed hexes.
+                        scoutQueue.add(new Pair<>(checkCoords, nextDistance));
+                    }
```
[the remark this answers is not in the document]
IllianiBird: I think copilot is high, it's linking to another (completely unrelated) PR

MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -1501,83 +1497,115 @@ private static void scanNeighboringCoords(StratConCoords coords, int forceID, Ca
...
                 Pair<StratConCoords, Integer> current = scoutQueue.poll();
                 StratConCoords currentCoords = current.getKey();
                 int distance = current.getValue();
 
-                // Only process neighbors if they're within the max distance
-                if (distance < scanRangeIncrease) {
-                    for (int direction = 0; direction < 6; direction++) {
-                        StratConCoords checkCoords = currentCoords.translate(direction);
+                // Do not expand beyond this scout's radius
+                if (distance >= individualScanRange) {
```
Sleet01, line 1533: Is this logic actually correct?  The code base in general has some checks that are inclusive of range, and some that exclude hexes _at_ the range...

IllianiBird: Lemme double check what we're indexing at

IllianiBird: Yeah, this is working right. I had to triple check, but this is right

PR #8078, opened 2025-11-11
Improvement: Removed Scenario Odds Reduction when Alt-Advanced Medical is Enabled

IllianiBird:
Previously we reduced scenario odds by 33% when Alt-AM was enabled. This PR removes that based on player feedback following the earlier reduction of AAM healing times.


mmetke-qumulo (approved):
LGTM

PR #8093, opened 2025-11-12
Improvement: Added 'Bolster Contract Skill' Option to Optionally Increase OpFor & Ally Skills

IllianiBird:
I've been keeping an eye on player chatter over the past couple of versions and a theme that I've seen come up a lot is that green allies are not fun to fight alongside. Similarly, neither are green enemies fun to fight against.

With that in mind I've gone ahead and added a campaign option that will increase the experience level of allies and enemies by one level. If previously they would have been green, they are now regular, and so on. This is extended to Cadre Duty.

This will have a few knock-on effects:

- Contract scenarios will, overall, be slightly more challenging on average.
- Prisoners will, on average, be slightly more skilled.
- Clan OpFors will, on average, be slightly smaller (if the Batchall is accepted).
- Allies & OpFors can now be rated as 'Heroic', rather than capping out at Elite. This will allow them to challenge players for longer.

At lot of the points above are marked as 'on average.' This is due to the 'randomize skills' campaign option which can generate crews one experience level above, below, or at the requested skill level. That means 'regular' could generate green, regular, or veteran crews. While 'heroic' could see elite, heroic, or legendary crews.


Sleet01 (approved):
LGTM

PR #8103, opened 2025-11-13
Fix: Fixed HR Strain Penalizing Complete Usage

IllianiBird:
If a campaign had `n`/`n` HR Strain use they would incorrectly get a -1 penalty, when it should have been no penalty. This PR fixes that.


Sleet01 (approved):
LGTM

Issue #8114, opened 2025-11-13
[Issue] Right click personnel breaks hard and persists through save, close, and reload.


trent1098s:
### Brief Description *

I managed to get my personnel list in to an unusable state. I had applied several prosthetics via right click menu. Was looking through personnel, noticed an illiterate staff member w/ 50ish XP available. Went to right click to give them a point in language/any and got this error popup:

<img width="637" height="199" alt="Image" src="https://github.com/user-attachments/assets/7a3cf1cd-4d0f-4f7a-aa9a-f0d025cc23e9" />

Thereafter, ANY staff member I right click on shows the same error; I can no longer interact with staff via right click.

### 3. Steps to Reproduce

as above

### Attach Files

[mekhq.log](https://github.com/user-attachments/files/23533830/mekhq.log)

[The Smoke and Mirrors30690125a.cpnx.gz](https://github.com/user-attachments/files/23533832/The.Smoke.and.Mirrors30690125a.cpnx.gz)

[customs.zip](https://github.com/user-attachments/files/23533834/customs.zip)

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

50.10 Dev environment on main

### Operating System *

Windows 11

### Java Version *

Adoptium 17

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

trent1098s:
Fixed in #8116

PR #8116, opened 2025-11-13
Fix for breaking error on right click of personnel, in Academy skills list

trent1098s:
Fixes issue #8114


IllianiBird (approved):

MekHQ/src/mekhq/campaign/personnel/education/Academy.java
```diff
@@ -1019,6 +1019,8 @@ public static String skillParser(String skill) {
             case "interest/other" -> SkillType.S_INTEREST_OTHER;
             case "interrogation" -> SkillType.S_INTERROGATION;
             case "investigation" -> SkillType.S_INVESTIGATION;
+            //skill string is stored both plural and not plural in some cases
+            case "language/any" -> SkillType.S_LANGUAGES;
             case "languages/any" -> SkillType.S_LANGUAGES;
```
IllianiBird, line 1024: This is unnecessary, it's far better to fix the root cause than to apply a bandaid. I'd like to see this as `language/any` only. I'm fixing the data portion now, which is the root cause.

Issue #8123, opened 2025-11-13
[Issue] Combat Technician available salvage minutes stay at 0 after cycling day


trent1098s:
### Brief Description *

<img width="379" height="61" alt="Image" src="https://github.com/user-attachments/assets/97bba7b0-e5ae-4fbc-8a89-e5b67f0c4982" />

Salvage crew returns from dropship escort mission, one tech injured:

<img width="463" height="142" alt="Image" src="https://github.com/user-attachments/assets/4e8f6e6d-8459-427f-bb93-06c769e708f0" />

Advance day:

<img width="477" height="192" alt="Image" src="https://github.com/user-attachments/assets/90448d6e-b3fb-438f-86f7-0c19fa107f9b" />

Deploy to new mission:

<img width="1073" height="78" alt="Image" src="https://github.com/user-attachments/assets/f45ac6b7-e276-4c85-af8c-736b21403268" />

Combat techs assigned to salvage on prior day have zero available minutes to use:

<img width="1507" height="1188" alt="Image" src="https://github.com/user-attachments/assets/d58e97a0-a38d-45a5-be6a-6b7b744c8556" />

(This is FOR SURE the same red shirts that went out prior as one of them shows the 4 injuries they received)

<img width="1488" height="65" alt="Image" src="https://github.com/user-attachments/assets/0cb7a1ff-0b38-40fd-b46b-71b948132185" />

Thing is, I only sent 4 to that dropship mission. Mission before that, I over-sent techs and had selected 12, or so. SO all of the techs from the prior two missions are at zero minutes.

Eventually I'll run out of techs with all having zero minutes.

### 3. Steps to Reproduce

As above.

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

50.10 dev environment - main branch (current refresh)

### Operating System *

Windows 11

### Java Version *

Adoptium 17

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
I'm going to need a save demonstrating the issue. I'm not able to replicate using a test campaign.

trent1098s:
[...]

PR #8130, opened 2025-11-14
Improvement: Added Current Minutes to Personnel Table

IllianiBird:
While investigating #8123 I realized that we have no easy way for the user to view how many remaining minutes a tech has, on the personnel table. Now we do.


Sleet01 (approved):
LGTM

PR #8134, opened 2025-11-14
Fix: Fixed Combat Technicians Not Counting Towards CamOps Rating

IllianiBird:
We were only checking for Mechanics, now we check for Mechanics and Combat Technicians.

I also renamed `getVehicleCrew` as that was still referencing the now defunct Vehicle Crew profession.


MekHQ/src/mekhq/campaign/rating/CamOpsReputation/SupportRating.java
```diff
@@ -300,6 +300,7 @@ private static Map<String, Integer> calculateTechCounts(Campaign campaign) {
         for (Person person : campaign.getActivePersonnel(false, false)) {
             updateCount(person::isTechMek, "techMekCount", techCounts);
             updateCount(person::isTechMechanic, "techMechanicCount", techCounts);
+            updateCount(person::isCombatTechnician, "techMechanicCount", techCounts); // Counts as Mechanic
             updateCount(person::isTechAero, "techAeroCount", techCounts);
```
IllianiBird, line 304: This is the only change of substance, the rest is the method rename

PR #8145, opened 2025-11-15
Improvement: Updated Crew Required Roles Based on Unit Equipment

IllianiBird:
# Requires [Task: Code Support for Additional Crew Needs](https://github.com/MegaMek/megamek/pull/7633)

Prior to 50.10 we were treating non-gunner, non-driver vehicle crews under the catchall 'Vehicle Crewmember' profession. When we updated vehicle crew, earlier in the 50.10 cycle, we updated 'Vehicle Crewmember' to 'Combat Technician' - to avoid confusion with Vehicle Crew/Ground, Vehicle Crew/Naval, and Vehicle Crew/VTOL. At the time we also made Combat Technicians a tech profession. The original Vehicle Crewmember role used Mechanics, so this made sense.

However, this has had a number of knock-on effects:

- Combat Technician doesn't really have an identity of its own, it's literally just Mechanic by another name.
- Linked to the above, players don't know what a Combat Technician _is_.
- We've received feedback that players are unhappy that this explodes their tech pools.

The first issue is actually a side effect of how phenotypes were programmed. If we don't have a Vehicle Crewmember analog phenotypes break. I'll not going into the gory details, that's just how it is. Legacy code does legacy code things.

Both the second and third issues are legitimate and something we can fix. I'll cover the Combat Technician identity crisis first.

Combat Technicians now use the Mechanics and Tech/Mek skills. This places them as the perfect profession for salvage operations - once we implement field stripping, especially. This skills are also important. With Techs Use Admin enabled (on by default in most presets) Combat Technicians need three skills, compared to the two needed by normal techs. This means they are a good choice for specific things where you need both Mechanics and Tech/Mek - like salvage operations - but they don't outshine normal Mechanics and MekTechs.

Regarding the other changes made by this PR:
- Communications Equipment is crewed by Comms Operators
- MASH Theaters are crewed by Doctors supported by Medics
- Mobile Field Bases are crewed by Combat Technicians supported by Astechs
- Any other additional crew seat (beyond drivers and gunners) can continue to use multiple professions, as per player fiat. However NPC vehicles will crew these seats with Vehicle Crew/x (based on the unit type). This fixes player tech teams exploding in side.

A change not yet implemented is a compatibility handler. That will be coming later, once this PR is ready for review.

### DIFF
--- a/MekHQ/src/mekhq/Utilities.java
+++ b/MekHQ/src/mekhq/Utilities.java
@@ -35,6 +35,7 @@ package mekhq;
 
 import static java.lang.Math.max;
 import static mekhq.MHQConstants.BATTLE_OF_TUKAYYID;
+import static mekhq.campaign.personnel.enums.PersonnelRole.VEHICLE_CREW_EXTENDED_ROLES;
 import static mekhq.campaign.personnel.skills.SkillType.EXP_ELITE;
 import static mekhq.campaign.personnel.skills.SkillType.EXP_GREEN;
 import static mekhq.campaign.personnel.skills.SkillType.EXP_HEROIC;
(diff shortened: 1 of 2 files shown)


IllianiBird:
After long consideration I have decided this would not be a good move at this time, so will be closing this PR.

PR #8154, opened 2025-11-15
Fix for uncrewed field kitchens counting towards capacity

trent1098s:
In the previous code several checks were made for uncrewed, partially crewed, etc. However, if a unit has all personnel stripped it returns Crewstate = unsupported, as technician is checked first, branching out to return unsupported (no tech + no crew) or unmaintained (some or full crew + no tech).

This change simply forces the issue; a full crew complement (including assigned tech) is required for the field kitchen to be fully functional for the unit.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
+++ b/MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
@@ -90,7 +90,6 @@ public class Fatigue {
             if ((unit.getCrewState().isFullyCrewed())) {
                 for (MiscMounted item : unit.getEntity().getMisc()) {
                     if (item.getType().hasFlag(MiscType.F_FIELD_KITCHEN)) {
-                        //final int crewAssignmentState = unit.getCrewState();
                         fieldKitchenCount++;
                     }
                 }


trent1098s:
Aside from removing the comment (which is fine, that was there for testing and I forgot to remove it), the rest is as intended. There's no need to check the other enum types; desired behavior is for .IsFullyCrewed(). 

The enum for state on that works a bit odd, if there are no techs assigned with either partial or no crew, the enum is no longer Uncrewed or PartiallyCrewed, but rather, Unsupported. E.g. there is only one state. Thus, the only state we care about for this IsFullyCrewed; rather than checking for all of the negatives, check for the single positive state we desire.

trent1098s:
Truthfully, that Enum should be refactored entirely; as there can be multiple simultaneous relevant states and the Enum can only reflect one of those at any given time. 

It is possible for a vehicle to have no tech (unmaintained) and also be uncrewed, or partially uncrewed.

I would suggest replacing the enum with a state structure object containing the individual states with appropriate properties and methods, as needed, so that these independent states can be checked individually. Right  now, the state is potentially ambiguous, which is detrimental to developers who expect a certain logic from the properties on unit such as "IsCrewEmpty", etc. Thus, if the unit has no crew but also does not have a tech, it will fail a check for "IsCrewEmpty" - as the enum would change to Unsupported. Or if there is no tech, but full crew, it would still fail IsCrewEmpty as the enum returns "Unmaintained".


Refactoring unit, and the crew state, should be a follow up to this.

MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
```diff
@@ -87,16 +87,11 @@ public static int checkFieldKitchenCapacity(List<Unit> units, int fieldKitchenCa
...
-                      || (unit.isDamaged())
-                      || (unit.getCrewState().isUncrewed())
-                      || (unit.getCrewState().isPartiallyCrewed())) {
-                continue;
-            }
-
-            for (MiscMounted item : unit.getEntity().getMisc()) {
-                if (item.getType().hasFlag(MiscType.F_FIELD_KITCHEN)) {
-                    fieldKitchenCount++;
+            if ((unit.getCrewState().isFullyCrewed())) {
```
IllianiBird, line 90: This should be `!unit.isFullyCrewed()` interacting with crew state is unnecessary and is a mistake in the original code.

Furthermore, early exit rather than nesting in conditionals is the preferred method due to increased code clarity.

MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
```diff
@@ -87,16 +87,11 @@ public static int checkFieldKitchenCapacity(List<Unit> units, int fieldKitchenCa
         int fieldKitchenCount = 0;
 
         for (Unit unit : units) {
-            if ((unit.isDeployed())
-                      || (unit.isDamaged())
-                      || (unit.getCrewState().isUncrewed())
-                      || (unit.getCrewState().isPartiallyCrewed())) {
```
IllianiBird, line 93: `isDamaged` and `isDeployed` are all correct and should not be removed. As per the Turnover & Retention documentation a Field Kitchen is inoperable in any of those states.

MekHQ/src/mekhq/campaign/personnel/turnoverAndRetention/Fatigue.java
trent1098s: I don't believe this will fix 100% of the conditions anyway, after further review.  I do agree the other conditions need to be reapplied - I did so, and further testing with them in place revealed yet another case where a properly maintained, fully crewed vehicle will fail to be counted (closer, but still not complete).

This new case I found - if a tech is assigned to the TOE force, and not a specific tech to the unit, IsFullyCrewed() will always return false. This is because that property checks the Enum for IsFullyCrewed(), and IsFullyCrewed is only true under these conditions:

A tech is assigned to that *specific* unit, 
AND
The vehicle is fully crewed.

If the first condition fails IsFullyCrewed() will not be set on the enum, the IsFullyCrewed() property will return false; even if the unit has a full crew compliment.

This means unit can be properly maintained by a TOE Force technician, not an individual tech; fully crewed; undamaged, and still will not count for kitchen capacity, due to the way that Enum is built. 

TLDR: IsFullyCrewed() only returns true if there is a tech assigned to the _specific_ unit (not a force tech), and it then has a full crew compliment.

trent1098s: (My thought was this enum must predate the TOE force technician capability, and was not updated to reflect TOE force technicians at that time)

Regardless; the enum only allows for one "state" - yet multiple states *can* exist in parallel simultaneously, and not really suitable for an Enum. These really should be individual properties, as damaged/undamaged, partial/full/empty crew, maintained/not maintained are not "variants of the same thing" which an Enum is typically used for - they are separate concepts, entirely.

Issue #8135, opened 2025-11-14
[Issue] Useful Permanent Astechs not adding the correct number of extra Astechs


wildfire142:
### Brief Description *

When using the Useful Permant Astech option you don't get the expected behaviour,
Using one Tech and one Astech to test

Astech skill = 0 temp Astech needed =3
Astech skill = 1 temp Astech needed =3
Astech Skill = 2 temp Astech needed =3
Astech Skill = 3 temp Astech needed =4
Astech Skill = 4 temp Astech needed =4
Astech Skill = 5 temp Astech needed =5
Astech Skill = 6 temp Astech needed =5
Astech Skill = 7 temp Astech needed =5
Astech Skill = 8 temp Astech needed =5
Astech Skill = 9 temp Astech needed =5
Astech Skill = 10 temp Astech needed =5

With the option active as curently implemented your best using people with 0 or 1 skill rank to maximise the benefit for minmal salary costs

### 3. Steps to Reproduce

1) start new campaign
2) choose veteran preset
3) GM add one tech - any type
4) GM add one Astech 
5) change skill level of Astech 
6) remove temp Astech pool
7) advance day and see number change
8) repeat steps 5-7 and watch the pool numbers change

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

Nightly 11-14

### Operating System *

Windows 10 Pro

### Java Version *

Adoptium 17

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

wildfire142:
Closed via https://github.com/MegaMek/mekhq/pull/8155

PR #8155, opened 2025-11-15
Useful Permanent Astech Tweak Fixes issue #8135

wildfire142:
With the change the number of additional Astechs increases as the Skill level increases.


IllianiBird (approved):
I hate this method so much. Thanks for catching this.

We have about five different ways to fetch skill level and all of them work slightly differently.

PR #8184, opened 2025-11-17
Improvement: Added Ability to See Current Salvage % to CamOps Salvage Picker

IllianiBird:
Now it is possible to see current salvage %, not just the maximum in the CamOps Salvage Picker. This should improve usability.


Sleet01 (approved):
LGTM

PR #8210, opened 2025-11-18
Improvement: Optimized Name Fetching

IllianiBird:
Following a discovery by Psi I went through and updated a number of enums that were previously using single-use calls to getFormattedTextAt to instead have the String defined inside the enum entry itself. In this way the getting is run once, used infinitely.


Sleet01 (approved):
LGTM

psikomonkie (approved):
Nice!

PR #8229, opened 2025-11-19
Fix: Added Missing 'Standard' Force Resource Bundle Entry

IllianiBird:
As per title

### DIFF
--- a/MekHQ/src/mekhq/campaign/force/ForceType.java
+++ b/MekHQ/src/mekhq/campaign/force/ForceType.java
@@ -130,16 +130,9 @@ public enum ForceType {
      * a localization resource key in the {@code ForceType} resource bundle, with keys formatted as
      * {@code [enumName].symbol}.</p>
      *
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/force/ForceType.java
```diff
@@ -130,16 +130,9 @@ private String generateDisplayName() {
...
-     * <p>If the current instance is {@code STANDARD}, an empty string is returned as the symbol.</p>
-     *
-     * @return The localized symbol associated with the current instance, or an empty string if the instance is
-     *       {@code STANDARD}.
+     * @return The localized symbol associated with the current instance.
      */
     private String generateSymbol() {
-        if (this == STANDARD) {
-            return "";
-        }
```
IllianiBird, line 141: This stopped working because 'this' cannot equal `STANDARD` because at the time this method is called `STANDARD` doesn't exist.

PR #8250, opened 2025-11-20
Improvement: Added the Ability to Stop Individual Techs from Appearing in the Salvage Op Tech Picker

IllianiBird:
With the removal of Combat Technicians/Engineers players immediately voiced a need for a way to filter personnel out of the salvage tech picker dialog. This flag allows them to do so.

As of this PR only characters with the 'Salvage Supervisor' flag can participate in salvage operations. Whether this flag is enabled or disabled by default is determined by a campaign option (set to enabled by default by default).

While I was adding this flag I also spotted that we were being very inconsistent with how we were handling flag toggling. So I updated that.


Sleet01 (approved):
LGTM

PR #8279, opened 2025-11-21
Improvement: Pyrrhic Victories Now Count as Both Defeat & Victory for Morale; Updated Glossary

IllianiBird:
Based on a discussion between me and Hammer. Pyrrhic victories counting as both victory and defeat is thematically appropriate.


Sleet01 (approved):
LGTM

PR #8282, opened 2025-11-21
Improvement #2321: Added Campaign Option to Prevent Tornadoes From Spawning in Scenarios

### LINKED ISSUE #2321: RFE: alternate weather options in place of tornado
When 'Tornado F1-F3' or 'Tornado F4+' conditions are selected for a scenario, it can be an instant win or instant loss when non-Mech forces are deployed on one side or another (or a double-kill with both sides).  I'd like to see a campaign option that makes the following changes when generating scenario weather:

Limit tornado winds to when all sides are fielding only Mechs.  Otherwise,
Instead of 'Tornado F1-F3', use starting condition of 'Moderate Gale' and variable winds from 'Light Gale' to 'Strong Gale'.
Instead of 'Tornado F4+', use starting conditions of 'Strong Gale' and variable winds from 'Moderate Gale' to 'Tornado F1-F3'.

I've been manually making these changes in campaign play and they seem to work quite nicely when a more conventional force mix is in play.

IllianiBird:
Close #2321

This is one of those perennial requests we see a lot. Basically the issue is that if a tornado spawns on a scenario it's instant death to whichever side has the most vehicles. This isn't a fun experience and creates 'weather gambling' - where you reroll weather against a vehicle force in the hopes of hitting a tornado.

Now, if the campaign option is enabled, F1 -> F3 become Moderate Gales, while F4 becomes Strong Gale.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -1856,6 +1856,9 @@ lblUseLightConditions.text=Enable Light Conditions
 lblUseLightConditions.tooltip=Determine light conditions when generating a scenario.
 lblUsePlanetaryConditions.text=Enable Planetary Conditions
 lblUsePlanetaryConditions.tooltip=Set gravity and atmosphere based on the contract location.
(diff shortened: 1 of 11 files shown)


Sleet01 (approved):
LGTM

Issue #6183, opened 2025-03-06
[RFE] Prompt auto-mothballing on Begin Transit, even off-contract.


Tzahr:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

In the current implementation, when taking a contract, you are helpfully prompted to mothball your units during transit, before they are restored and activated again upon arrival on location. 

This prompt should ideally trigger whenever transit happens, not only on contracts. Ideally, the option should be given each time the Begin Transit button is pressed. 

### Use Cases or Rationale

The existence of this function for contract mission essentially justifies this option as being relevant for any transit period. 
Mothballing the force during transit appears to be normative. 

### MekHQ Suite Version *

v0.50.04-SNAPSHOT

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

PR #8284, opened 2025-11-21
Improvement #2564, #6183: Added AutoMothball Prompt When Beginning Transit - Even if Off Contract

### LINKED ISSUE #2564: RFE: add option to auto-mothball units on long transits
Units can be mothballed to avoid maintenance checks and costs, which can be valuable when making long transits.

A campaign setting to auto-initiate mothballing by unit type in this circumstance would be welcome.  This would be a number indicating the maximum number of jumps without mothballing units, with a check box for each unit type (checked for auto-mothball, unchecked for ignore).  When transit to another system is initiated past that number of jumps, the specified unit types would start to be mothballed by the designated techs.  When the formation jumps to the destination system the units are brought out of mothball while the DropShips are in transit.

IllianiBird:
Close #2564
Close #7611 

Now when entering transit the campaign will be prompted with the opportunity to automatically mothball all units - in the same way as accepting a contract.

I also moved the 'activation' trigger from 'entering contract system' to 'landing on planet' so that it's more reliable and universally applicable.


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1726,13 +1728,31 @@ public void setLocation(CurrentLocation l) {
...
+        setLocation(new CurrentLocation(planetarySystem, 0.0));
         MekHQ.triggerEvent(new LocationChangedEvent(getLocation(), false));
+
+        if (getAutomatedMothballUnits().isEmpty()) {
+            performAutomatedActivation(this);
+        }
+
+        if (campaignOptions.isUseRandomDiseases() && campaignOptions.isUseAlternativeAdvancedMedical()) {
+            Inoculations.triggerInoculationPrompt(this, false);
+        }
```
IllianiBird, line 1755: This method is only used when GM teleporting. We add these calls here to ensure they trigger when teleporting about the universe

Sleet01 (approved):
LGTM

PR #8291, opened 2025-11-22
Improvement: Added Code Support for Skill-Specific Natural Aptitude Traits; Fixed Skill/Ability Check Utilities Burning Edge on Impossible Checks; Added 'Reason' to Skill Checks

IllianiBird:
# Requires [Improvement: Added Utility Method to Compute to Fetch the Highest Two of Any Given List of Integers](https://github.com/MegaMek/megamek/pull/7655)

The PR does the following:

- Adds the behind the scenes code support necessary to implement RAW Natural Aptitude. It's important to note that this PR does _not_ implement RAW NatApt it just puts the necessary code in the necessary places so we can add it later. That won't happen until we more fully move to the Skill Check Utility class. Probably some time during the 50.11 cycle.
- Stops characters burning Edge on impossible skill or attribute checks
- Updated the results text from skill and attribute texts to take up less space
- Added a 'reason' to each skill or attribute check call so players can see _why_ a check is taking place.


Sleet01 (approved):
LGTM

PR #8292, opened 2025-11-22
Improvement: Updated Required Victory Points Calculations; Added Indicator Whether Contract Can End Early to Briefing Room

IllianiBird:
Earlier in the cycle we improved victory points required for contract victory. This PR updates that based on player feedback. We now factor in the following:

- Contract duration (including whether it is Garrison-type)
- scenario chance
- Intensity
- Command Rights

I also added a small indicator to the right of Required Victory Points (in the Briefing Room) to show whether a contract can be concluded early.

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -32,6 +32,7 @@
  */
 package mekhq.campaign.stratCon;
 
+import static java.lang.Math.ceil;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 import static megamek.codeUtilities.ObjectUtility.getRandomItem;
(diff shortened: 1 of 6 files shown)


MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -741,23 +743,21 @@ private static void determineIfTurningPointScenario(AtBContract contract, StratC
             return;
         }
 
-        if (!isCombatChallenge) {
```
IllianiBird, line 744: Unnecessary conditional, so I removed it

MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -741,23 +743,21 @@ private static void determineIfTurningPointScenario(AtBContract contract, StratC
...
+                scenario.setTurningPoint(true);
+                setAttachedUnitsModifier(scenario, contract);
+            }
+            case HOUSE, LIAISON -> {
+                if (randomInt(3) == 0) {
                     scenario.setTurningPoint(true);
                     setAttachedUnitsModifier(scenario, contract);
                 }
-                case HOUSE, LIAISON -> {
-                    if (randomInt(3) == 0 || isObjective) {
```
IllianiBird, line 752: isObjective was unnecessary here

PR #8293, opened 2025-11-22
Improvement: Added Intel Breach Events When Freeing Prisoners

IllianiBird:
We received player feedback - quite rightly - that while there were challenges for keeping prisoners. Chiefly the need to deal with prisoner events and maintain prison guard forces. However, there was nothing stopping players from just freeing all their prisoners. While that isn't really an issue the problem is that without a persistent OpFor there is no mechanical justification for why you're keeping all these PoWs.

With this PR, we add that justification. If you free prisoners there is a chance vital intel will get back to the contract OpFor improving their morale as they utilize this intel to shift the balance of power in their favor. Contracts that are at Overwhelming or Routed cannot be modified in this manner.

The chance of an intel breach is n-in-50 where n is the number of personnel being freed. This means freeing 50+ prisoners guarantees an intel breach.


Sleet01:
In the long run, I'd like us to have a stateful intel state tracking system - partly for modifying scenarios, partly for adjusting deployed munitions.
This seems like a great start!

Sleet01 (approved):
LGTM

PR #8296, opened 2025-11-22
Improvement: Stopped New StratCon Facilities Spawning Mid-Contract

IllianiBird:
While it is fun for new facilities to pop-up mid contract they create an issue. The player will almost always capture them, leading to increasing allied scenario modifiers. What could have been a mid-contract boon for the OpFor instead ends up being another boost to player power.

This PR removes the possibility of new facilities appearing mid-contract.


Sleet01 (approved):
LGTM

PR #8306, opened 2025-11-23
Fix: Fixed Potential Null Error in Scenario Resolution

IllianiBird:
During scenario resolution if a unit's crew have all been vaporized the unit will be treated as uncrewed. However, when fetching pilot we weren't checking that the pilot existed, resulting in the NPE. This PR adds a null safety.


Sleet01 (approved):
LGTM

PR #8310, opened 2025-11-23
Fix: Concurrent Modification in Injury Removal

IllianiBird:
Schoolboy error on my part. Forgot to include defensive copies in the loop.


Sleet01 (approved):
LGTM

PR #8316, opened 2025-11-23
Fix: Fixed Ability to Deploy Salvage Forces Multiple Times

IllianiBird:
Earlier in the cycle we squashed a bug that was allowing players to deploy the same salvage team to salvage ops multiple times prior to the force returning. Unfortunately an oversight meant that you could still do this with dedicated salvage forces, just not non-salvage forces.

This PR fixes that.


MekHQ/src/mekhq/gui/BriefingTab.java
```diff
@@ -1167,29 +1177,21 @@ private List<SalvageForceData> getSalvageForces(Campaign campaign, boolean isSpa
...
-            if (!force.isDeployed() &&
-                      !isDeployedToStratCon &&
-                      force.getSalvageUnitCount(hangar, isSpaceScenario) > 0) {
-                boolean hasDeployedUnit = false;
-                for (Unit unit : force.getAllUnitsAsUnits(getCampaign().getHangar(), false)) {
-                    if (StratConRulesManager.isUnitDeployedToStratCon(unit)) {
-                        hasDeployedUnit = true;
-                        break;
-                    }
-                }
```
IllianiBird, line 1188: I removed this check because it turned out to be unnecessary. `isUnitDeployedToStratCon` doesn't check whether the unit is deployed, but its force. StratCon looks like it was once set up to track unit deployments, as well as force, but that work was never completed.

PR #8330, opened 2025-11-24
Improvement: Extended Bay Rentals to Include Large Craft

IllianiBird:
This PR expands the new bay rental system to incorporate large craft. Not doing so previously has been causing confusion among the player base.


Sleet01 (approved):
LGTM

PR #8331, opened 2025-11-24
Improvement: Added MASH Theater Capacities from Large Craft (inc. Space Stations)

IllianiBird:
CamOps pg 191 states that large craft come equipped with medical facilities (medical teams, but mostly same thing in CamOps -> MekHQ parlance). This PR adds those facilities to the MASH Theater capacity of the campaign.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/medical/MASHCapacity.java
+++ b/MekHQ/src/mekhq/campaign/personnel/medical/MASHCapacity.java
@@ -36,6 +36,7 @@ import java.util.List;
 
 import megamek.common.equipment.MiscMounted;
 import megamek.common.equipment.MiscType;
+import megamek.common.units.Entity;
 import mekhq.campaign.unit.Unit;
 
 public class MASHCapacity {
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/personnel/medical/MASHCapacity.java
```diff
@@ -65,6 +66,18 @@ public static int checkMASHCapacity(List<Unit> units, int capacityPerTheatre) {
                 continue;
             }
 
+            // Certain unit types automatically include medical facilities (CamOps pg 191)
+            Entity entity = unit.getEntity();
+            if (entity != null) {
+                if (entity.isDropShip() || entity.isJumpShip()) {
```
Sleet01, line 72: I don't remember for sure off the top of my head, but these may need to be in reverse order if large craft inherit from JumpShip.

IllianiBird: For `isJumpShip` anything that isn't explicitly a JumpShip will return false. These `isX` booleans aren't foolproof, as they're ostensibly meant for Alpha Strike. But when they _are_ usable they're super useful. Gets rid of all the inheritance landmines.

PR #8347, opened 2025-11-25
Task: Updated 'Learning Ropes' Tutorial Campaign to 50.10

IllianiBird:
As per title. Added a couple of units (in case the player turns on some of the new options), distributed some scouting skills (in case the player turns that option on), and gave out veterancy SPAs.


Sleet01 (approved):
LGTM

PR #8350, opened 2025-11-25
Improvement: Stopped Campaign Upgrader Process From Changing Rank System

IllianiBird:
This was the cause of semi-frequent player contact during release period.

Furthermore, I adopted the previously unused 'ABRIDGED' mode into this so that we reduce the amount of campaign option _stuff_ initialized during the upgrade process. When combined with #8349 this _drastically_ speeds up upgrading.


Sleet01 (approved):
LGTM

PR #8357, opened 2025-11-26
Fix: Stopped Early Contract Finishes via Routing OpFor On Contracts That Don't Allow Early Contract Finishes

IllianiBird:
This removes the capacity for the player to achieve an early contract finish on contracts that are meant to disallow such things.


Sleet01 (approved):
LGTM

PR #8370, opened 2025-11-27
Improvement: Added Ability to Pay Off Bounty Hunters

IllianiBird:
This is a feature I intended to make available in 50.10 but forgot, as I am want to do. All it does is allow the player to extend their protection to characters with bounties. If done any attempts on the characters' life will fail and the characters' bounty (times 2) will be deducted from the campaign finances.

If this doesn't make 50.10 that's fine, but it'd be a nice addition.


Sleet01 (approved):
LGTM

Issue #8393, opened 2025-11-28
[Issue] Error when right click context menu in personnel tab


strategoslevel3:
### Brief Description *

I have just imported my 50.06 campaign to 50.10, but it will not allow me to right click on any personnel

<img width="1537" height="1176" alt="Image" src="https://github.com/user-attachments/assets/6a0c8d6f-bd44-4f29-8067-7069563984c6" />

### 3. Steps to Reproduce

Open file in 50.10
Go to personnel
Right click on personnel

### Attach Files

[logs.zip](https://github.com/user-attachments/files/23829137/logs.zip)

[custom.zip](https://github.com/user-attachments/files/23829143/custom.zip)

[Morning Star Assault Group30320806.cpnx.gz](https://github.com/user-attachments/files/23829144/Morning.Star.Assault.Group30320806.cpnx.gz)

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MekHQ Suite Version *

0.50.10

### Operating System *

MacOS 26.1

### Java Version *

Eclipse 17.0.12

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PR #8395, opened 2025-11-28
Improvement #8393: Added Better Handling When Academy Fails to Parse Skill

IllianiBird:
While investigating #8393 I realized that our current way of reporting skill parse failures in the Education module wasn't great. This PR updates it to trigger an error dialog instead of throwing a true error. This should help users identify the cause in their own custom academies and provide better reporting when we break a stock academy.

### DIFF
--- a/MekHQ/unittests/mekhq/campaign/personnel/education/AcademyTests.java
+++ b/MekHQ/unittests/mekhq/campaign/personnel/education/AcademyTests.java
@@ -34,7 +34,6 @@ package mekhq.campaign.personnel.education;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


MekHQ/unittests/mekhq/campaign/personnel/education/AcademyTests.java
```diff
@@ -183,45 +181,4 @@ void testGetFactionDiscountAdjustedNotPresentInLocationSystems() {
         when(campaign.getFaction()).thenReturn(new Faction("FWL", ""));
         assertEquals(1.0, academy.getFactionDiscountAdjusted(campaign, person));
     }
-
-    @Test
-    public void testSkillParser_ValidSkill() {
-        assertEquals(SkillType.S_PILOT_MEK, Academy.skillParser("piloting/mek"));
```
IllianiBird, line 189: No longer relevant

Sleet01 (approved):
LGTM

PR #8409, opened 2025-11-29
Improvement: Updated Injury Effects to Affect Attribute Score and Not Directly Affect Skills

IllianiBird:
One of the points of feedback we received late in the 50.10 cycle was that AAM's injury penalties were a bit too harsh. I agreed, but there was no time left to change it.

This PR changes injury effects so that the are applied to the characters' Attribute Score and _not_ directly to the skill. This makes them punishing, but not outright crippling. It also matches RAW.

### DIFF
--- a/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
+++ b/MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
@@ -1510,7 +1510,7 @@ atowTab.border="It is only those who have neither fired a shot nor heard the shr
   <br><i>General William Sherman\
   <br>Address at the Michigan Military Academy, June 19th 1879</i>
 lblATOWAttributesPanel.text=Traits and Attributes
-lblUseAttributes.text=Use Attributes \u26A0
+lblUseAttributes.text=New Characters Have Random Starting Attributes \u26A0
 lblUseAttributes.tooltip=If checked, characters will be generated with <i>A Time of War</i> Attribute scores.\
   <br>\
   <br><b>Warning:</b> If enabled (or disabled) mid-campaign, you need to navigate to the personnel tab, and (while in\
--- a/MekHQ/src/mekhq/campaign/personnel/Person.java
+++ b/MekHQ/src/mekhq/campaign/personnel/Person.java
@@ -36,7 +36,6 @@ package mekhq.campaign.personnel;
 import static java.lang.Math.abs;
 import static java.lang.Math.floor;
 import static java.lang.Math.max;
-import static java.lang.Math.min;
 import static java.lang.Math.round;
 import static megamek.codeUtilities.MathUtility.clamp;
 import static megamek.codeUtilities.StringUtility.isNullOrBlank;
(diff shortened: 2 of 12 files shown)


MekHQ/resources/mekhq/resources/CampaignOptionsDialog.properties
```diff
@@ -1510,7 +1510,7 @@ atowTab.border="It is only those who have neither fired a shot nor heard the shr
   <br><i>General William Sherman\
   <br>Address at the Michigan Military Academy, June 19th 1879</i>
 lblATOWAttributesPanel.text=Traits and Attributes
-lblUseAttributes.text=Use Attributes \u26A0
+lblUseAttributes.text=New Characters Have Random Starting Attributes \u26A0
```
IllianiBird, line 1513: Edited to better state what this option does. It is not possible to disable attributes scores no more than it is to disable skills. This is a code limitation.

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -6870,49 +6867,8 @@ public void setAttributeScore(final SkillAttribute attribute, final int newScore
      * @since 0.50.5
      */
     public int getAttributeScore(final SkillAttribute attribute) {
-        if (attribute == null || attribute.isNone()) {
-            LOGGER.error("(getAttributeScore) SkillAttribute is null or NONE.");
-            return DEFAULT_ATTRIBUTE_SCORE;
-        }
```
IllianiBird, line 6876: This logic has been migrated

MekHQ/src/mekhq/campaign/personnel/skills/Attributes.java
```diff
@@ -386,118 +474,6 @@ public int getAttributeCap(Phenotype phenotype, PersonnelOptions options, SkillA
...
 
-    /**
-     * @deprecated use {@link #getAttributeScore(SkillAttribute)}  instead
-     */
-    @Deprecated(since = "0.50.5", forRemoval = true)
-    public int getStrength() {
-        return strength;
-    }
-
-    /**
```
IllianiBird, line 397: There were a lot of deprecated methods that were just cluttering up the class. While I agree with the deprecation process in principle, in particular when we're slowly retiring a system or method. But here, where the getter was deprecated in the same cycle it was introduced, there is nothing to be gained.

MekHQ/src/mekhq/campaign/personnel/skills/Attributes.java
```diff
@@ -719,6 +583,33 @@ public void setCurrentEdge(final int currentEdge) {
...
 
+    public String getTooltip(SkillAttribute attribute, List<InjuryEffect> injuryEffects, PersonnelOptions options) {
+        StringBuilder tooltip = new StringBuilder();
+
+        String flavorText = attribute.getDescription();
+        if (!flavorText.isBlank()) {
+            tooltip.append(flavorText).append("<br><br>");
+        }
+
+        // TODO aging modifiers
```
Sleet01, line 594: Is this still in work, or for a later date?

Sleet01 (approved):
LGTM

MekHQ/src/mekhq/campaign/personnel/skills/Attributes.java
IllianiBird: Later date (I need this PR in place to be able to convert aging modifiers over)

PR #8410, opened 2025-11-30
Fix #8408: Reinforced Handling of Scenario Field Control Dialog by Converting to Immersive Dialog

### LINKED ISSUE #8408: Cannot salvage sometimes after a battle
### Brief Description *

Image

I completed the mission, and after saving the .gif and .mul, I clicked done.  MekHQ very briefly flashed the option to control the battlefield, but before I could choose, it closed that and went onto scenario resolution.  It appeared to resolve just fine until time for salvage and there was none.  It did however pay me for the DropShip kill (so it's strictly a salvage issue I believe).

### 3. Steps to Reproduce

This is one of those it just happens sometimes.  I don't know if you'll even be able to repeat from what I sent, I've never been able to.  This does seem to happen about 1 in 5 games.

### Attach Files

[No Salvage.zip](https://github.com/user-attachments/files/23836316/No.Salvage.zip)

### Severity *

### MekHQ Suite Version *

50.10 milestone

### Operating System *

windows 11

### Java Version *

java 17

### Final Verification

IllianiBird:
Fix #8408

The issue raised in #8408 is one I've seen off-on for several months now. No issues in the log, the dialog just fires and then immediately kills itself. Previously we believed it to have been an OS issue, and that may well still be the case. In particular I've noticed an uptick in this issue since more windows users started migrated to whatever the new Windows version is called.

That said, it may not be a bug at all, but an unfortunate consequence of the basic dialog used here (and elsewhere). In this type of dialog if the user hits ESC at any time the dialog will just close. No option is picked and so we use the fallback of 'no control' (as any boolean is initialized as `false`).

Regardless of the actual cause - user error (though not user fault) or Windows - it's an awful play experience.

This PR addresses this by switching the dialog over to the Immersive Dialog class. This dialog does not accept ESC exits, looks prettier, and avoids many of the issues related to the basic dialog class used previously. In the event this error _does_ return, the default pick for this new dialog is 'control'. So in the worst case the player will get some salvage they weren't entitled to.


Sleet01 (approved):
LGTM

Issue #3191, opened 2022-03-27
[0.49.3] Nimbus Theme Interstellar Map Pop Out Menu Un-clickable


HoneySkull:
### Environment
MekHQ Version 0.49.3
OS: Windows 10
Java 11

### Description
In MekHQ - Change the UI theme to "Nimbus" in the File->Themes->Nimbus menu. 
Navigate to the Interstellar Map tab.
The in-map blue pop-up menu is not painted properly.  It is a solid blue square that is not clickable.
Changing the them to other themes allows this button to function properly.

### Files
![IntersellarMap_Nimbus](https://user-images.githubusercontent.com/83041327/160261846-8811ae40-b37b-4558-bae0-f0321feb024c.png)

HammerGS:
We are currently up to 49.7 have you test this in that version.

HoneySkull:
Yes - I just pulled today's (March 26th, 2022) code and tested with 0.49.8-SNAPSHOT and confirmed the issue still exists with the Nimbus theme.

HoneySkull:
The issue also exists in the stable 0.48.0 - so nothing new introduced with recent changes.

kernhe:
Can approve this bug OS:linux too

Windchild292:
It's because of extra space below the button when using the Nimbus theme.
![image](https://user-images.githubusercontent.com/39067288/160293501-ded4a528-1d05-432d-9f3e-1dda59dcf477.png)

IllianiCBT:
Nimbus theme was removed in 2024 due to numerous issues. Closing.

PR #8432, opened 2025-12-06
Fix MM #3191: Ensure units clear their picked up MW and starting locations on reset

Sleet01:
MHQ side of fix (mainly for double safety).

## NOTE: only merge after https://github.com/MegaMek/megamek/pull/7724 is merged ##

Testing:
- Confirmed proper operation with hand-tweaked units containing fake MW ids in their `entity.pickedUpMekWarriors` vectors.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -35,6 +35,7 @@ package mekhq.campaign.unit;
 
 import static java.lang.Math.ceil;
 import static java.lang.Math.max;
+import static megamek.common.board.Board.START_NONE;
 import static megamek.common.equipment.MiscType.F_CARGO;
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Nice!

IllianiBird:
I wonder if there might also be value in introducing an extended Entity, TransitoryEntity, in mhq. When passing from MegaMek to MekHQ we create a TransitoryEntity from Entity. Basically a stripped down version that only includes what we want to persist. Anything that isn't explicitly included in TransitoryEntity is discarded.

This is future proofed as we add new stuff to Entity. If it isn't included in TransitoryEntity it won't come back to MekHQ.

The downside is that anything new that we do want to persist must be added to TransitoryEntity, but I'll take explicit controlled inclusion over requiring explicit - and clearly prone to failure - exclusion.

PR #8442, opened 2025-12-08
Improvement: Added Support for Partial XP Skill Progress

IllianiBird:
This PR is functionality needed for the upcoming Advanced Character Builder. Currently, MekHQ only tracks what skill level has been obtained. However the ACB allows for partial progress towards a level. A character with enough xp to hit level 3 in a skill, but with 5 xp remainder does not lose that 5 xp, their next improvement in that skill is just reduced by 5.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
+++ b/MekHQ/src/mekhq/campaign/personnel/skills/QuickTrain.java
@@ -204,7 +204,7 @@ public class QuickTrain {
 
                 int improvementCost = person.getCostToImprove(skillName, isUseReasoningMultiplier);
                 improvementCost = (int) Math.round(improvementCost * xpCostMultiplier);
-                int adjustedCost = max(0, skill.getXpProgress());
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/personnel/skills/Skill.java
```diff
@@ -117,26 +118,19 @@ public Skill(String type) {
...
-        this(SkillType.getType(type), level, bonus, hasNaturalAptitude);
+        this(SkillType.getType(type), level, bonus, 0, false);
     }
 
     public Skill(SkillType type, int level, int bonus) {
-        this(type, level, bonus, false);
+        this(type, level, bonus, 0, false);
     }
 
-    public Skill(SkillType type, int level, int bonus, boolean hasNaturalAptitude) {
```
IllianiBird, line 136: I opted to simplify Skill constructors as most of these were either unused or used in only one (or two) instances

Sleet01 (approved):
LGTM

PR #8444, opened 2025-12-09
Improvement: Added Ability to Disable Tabbed Daily Report

IllianiBird:
As per title, this PR adds the ability to disable the new tabbed daily report functionality. It doesn't remove the tabs, as adding & reporting GUI elements is a pain without restarting the MekHQ client. However, it does force all messages to be posted to the General tab. Effectively duplicating the <50.11 experience.


psikomonkie (approved):
Nice!

PR #8451, opened 2025-12-10
Fix #8443: Fixed Advanced Surgeries Not Gifting Appropriate Implant Effects in MegaMek Scenarios

### LINKED ISSUE #8443: [Issue] Implants added through surgery don't show up in megamek
### Brief Description *

The new way of adding Manei Domini implants via surgery doesn't result in units having the corresponding abilities in megamek. The only way to add those implants to personnel in MekHQ in a way that shows up in megamek is the "edit person" option in GM tools, where you can manually add the corresponding abilities. This only affects implants with effects that currently manifest as abilities (VDNI and the like), not prosthetics, as the latter directly modify the pilot stats which then gets communicated to Megamek with no problems.

I am assuming that this is in part because it is a new system, and that this issue is probably known to the developers, but I am submitting an issue report regardless to a. make it easier to track and b. indicate to other players both that it is known and also how to get around it (the GM tools above) to minimise the impact and number of additional reports.

### 3. Steps to Reproduce

1. Use the advanced surgery interface to add an implant with non-stat effects (VDNI or pain shunt for example) to a character
2. Enter a battle with said character
3. Notice they do not have the corresponding ability and that there was no effect from the surgery

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

0.50.10

### Operating System *

Windows 11

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.16  

### Final Verification

IllianiBird:
Fix #8443

I'm not sure when I broke this, because it _was_ working when tested, but regardless this fixes it again.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/AdvancedReplacementLimbDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/AdvancedReplacementLimbDialog.java
@@ -36,6 +36,7 @@ import static java.lang.Math.ceil;
 import static megamek.client.ui.WrapLayout.wordWrap;
 import static megamek.client.ui.util.UIUtil.scaleForGUI;
 import static megamek.common.options.PilotOptions.LVL3_ADVANTAGES;
+import static megamek.common.options.PilotOptions.MD_ADVANTAGES;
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #8453, opened 2025-12-10
Fix: Included Missing Switch Clause for Injury Score Modifiers

IllianiBird:
Removes a single unnecessary log entry


Sleet01 (approved):
LGTM

PR #8467, opened 2025-12-11
Improvement: Updated Scenario Map Size Generation

IllianiBird:
# Requires [Improvement: Updated Scenario Map Size Data & Unified Scenario Spawn Locations](https://github.com/MegaMek/mm-data/pull/211)

With the advent of StratCon Single Drop we've seen a resurgence in talk about scenario map sizes. This is a topic that's been on my mind for a few months, so with the added motivation I updated how scenarios determine map size.

This process replaces the old AtB method. Now we aim to have a map size roughly equal to one map per 4 combatants (ignoring infantry to avoid maps with lots of infantry getting too big). This can be adjusted in the scenario file to account for scenarios where we want the battlefield to be taller or wider. This can also be adjusted in campaign options for players who prefer smaller or larger battlefields.

It should be noted that under normal generation (excluding outside modifiers) maps will not generate deeper than 2 map sheets. Instead, we spread wide. This is to avoid a situation where a large engagement results in a truly massive map where players are forced to spend several turns sprinting just to reach the scrum.


Sleet01 (approved):
LGTM.
mm-data prereq also approved.

PR #8468, opened 2025-12-11
Fix: Fixed Non-Player Infantry Not Swapping to Weather Appropriate Gear

IllianiBird:
This was a frustrating problem to resolve and took more attempts than I care to admit. The root cause was just an order of operations issue: we were setting weather _after_ the infantry were weather equipped.


Sleet01 (approved):
LGTM

PR #8469, opened 2025-12-11
subtype flags merged with flags

exeea:
subtype flags merged with flags
Requires https://github.com/MegaMek/megamek/pull/7758


IllianiBird:
Flipping to draft to prevent merging prior to 50.11 launch

HammerGS (approved):

PR #8502, opened 2025-12-14
Fix #8428: It Is No Longer Possible to Jump Travel With Jump Incapable Units Such as Some Space Stations

### LINKED ISSUE #8428: [Issue] Able to jump with Space Station
### Brief Description *

Space station is behaving like a jumpship or warship by enabling jump.

Possible fix may be quite involved -- close to an RFE.

1. the simplest is simply prohibits jump until the player get rid of the space station.
2. the next possible solution is simply make the space station (and stuff loading on it) either unavailable, invisible, or both until the player returned to the system. Alternatively, make cargo onboard accessible but with latency -- i.e. interstellar shipping.
3. the most difficult would be to make an full instantiation of detached elements so potentially the player may operate on multiple system at the same time. 

### 3. Steps to Reproduce

1. create and crew space station, add to toe
2. travel to another system
3. 

### Attach Files

[customs.zip](https://github.com/user-attachments/files/23951967/customs.zip)

[yThe Reavers27851101.cpnx.gz](https://github.com/user-attachments/files/23951970/yThe.Reavers27851101.cpnx.gz)

### Severity *

### MekHQ Suite Version *

0.50.10

### Operating System *

Windows 10

### Java Version *

Eclipse Adoptium 17.0.17

### Final Verification

IllianiBird:
## See Also [Task: Updated Description of K-F Adaptor & Modular Space Station Options](https://github.com/MegaMek/megameklab/pull/2091)

Fix #8428

If the player attempts to jump travel while the campaign contains non-jump capable units (such as damaged JumpShips or non-jump capable Space Stations) they will receive a blocker dialog. The player can cancel the jump, ignore the restriction (GM), or abandon the affected units.

If the player abandons or sells the affected units they will get a confirmation dialog to make sure they _really_ want to do that.

Mothballing the affected unit will **not** prevent this blocker.

### DIFF
--- a/MekHQ/src/mekhq/campaign/utilities/JumpBlockers.java
+++ b/MekHQ/src/mekhq/campaign/utilities/JumpBlockers.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.StringJoiner;
 
+import megamek.common.units.BuildingEntity;
 import megamek.common.units.Entity;
 import megamek.common.units.Jumpship;
 import megamek.common.units.SpaceStation;
@@ -108,6 +109,12 @@ public class JumpBlockers {
 
         for (Unit unit : campaign.getUnits()) {
             Entity entity = unit.getEntity();
+            // Buildings are, by their nature, not able to leave the planet they're on.
+            if (entity instanceof BuildingEntity) {
+                nonJumpCapableUnits.add(unit);
+                continue;
+            }
+
             if (entity instanceof Jumpship jumpship) {
                 if (jumpship instanceof SpaceStation spaceStation) {
                     if (spaceStation.hasKFAdapter()) {


MekHQ/src/mekhq/campaign/utilities/JumpBlockers.java
```diff
@@ -0,0 +1,223 @@
...
+                if (jumpship.getDriveCoreType() == DRIVE_CORE_NONE) {
+                    nonJumpCapableUnits.add(unit);
+                    continue;
+                }
+
+                if (!jumpship.canJump()) {
+                    nonJumpCapableUnits.add(unit);
+                }
+            }
+        }
```
psikomonkie, line 138: Should we add `BuildingEntity` here - that'd be just "TO:AR Advanced Building" buildings. 

Your other option is `AbstractBuildingEntity` which includes `MobileStructure` entities too. I think those have nuance about if they can be transported, that's more up to you I think.

IllianiBird: I can do that, one moment

IllianiBird: Done

PR #8519, opened 2025-12-14
Fix #8496: Fix MRMS carryover check being applied before work time modifiers

### LINKED ISSUE #8496: [Issue] Instant repair starts jobs that will run until the next day when "Allow tasks to carry over to the next day(s)" is unchecked
### Brief Description *

In the image below, I have a large number of mechs needing repairs, and a very large number of techs available to do those jobs.  "Allow tasks to carry over to the next day(s)" is unchecked, so I expect that all jobs will complete today.  When I run the instant repair

Image

After running the instant repair, I have a large number of jobs that look like this, where an armor job was started that could easily have been finished today by a technician.: 

Image

In a perfect world, the system would prioritize removing any limbs that cannot be repair, then re-attaching all limbs first, because it is the hardest job and a pre-requisite for all other jobs.  Then, it would begin all the other jobs that can be completed today, leaving a large number of technicians with just a few minutes left in their schedule.  

If "allow jobs to run into the next day" was checked, the last step would be to go back over existing jobs and start them with the remaining time technicians have left.

### 3. Steps to Reproduce

1. Load the attached save.
2. Advance 1 day.
3. Go into mass repair, and run a repair with varied settings.

### Attach Files

[Repairstuff.cpnx.zip](https://github.com/user-attachments/files/24143302/Repairstuff.cpnx.zip)

### Severity *

### MekHQ Suite Version *

0.50.11 Nightly 12-12

### Operating System *

Windows 10

### Java Version *

17.0.14 

### Final Verification

psikomonkie:
Fixes #8496 

When doing MRMS, it was checking if a part's work would be more than a tech's time and cancel if the "carryover" is false. This check was done before work time modifiers, so it was not accounting for low skill techs that needed to use extra time, or high skill techs able to rush.

### DIFF
--- a/MekHQ/src/mekhq/service/mrms/MRMSService.java
+++ b/MekHQ/src/mekhq/service/mrms/MRMSService.java
@@ -861,12 +861,12 @@ public class MRMSService {
             // Fallback TN check to account for discrepancies between Techs
             TargetRoll targetRoll = campaign.getTargetFor(partWork, tech);
             if (canChangeWorkTime) {
-                WorkTime wt = techSkillToWorktimeMap.get(skill.getType().getName() + "-" + skill.getLevel());
-                if (null == wt) {
+                WorkTime workTime = techSkillToWorktimeMap.get(skill.getType().getName() + "-" + skill.getLevel());
+                if (null == workTime) {
                     debugLog("[ERROR] Null work-time from techToWorktimeMap for %s", "repairPart", tech.getFullName());
-                    wt = WorkTime.NORMAL;
+                    workTime = WorkTime.NORMAL;
                 }
-                ((Part) partWork).setMode(wt);
+                ((Part) partWork).setMode(workTime);
                 // Get updated TN with worktime in mind
                 targetRoll = campaign.getTargetFor(partWork, tech);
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/service/mrms/MRMSService.java
```diff
@@ -894,7 +894,10 @@ private static MRMSPartAction repairPart(Campaign campaign, IPartWork partWork,
 
             boolean isSameDayTech;
 
-            if ((tech.getMinutesLeft() < partWork.getActualTime())) {
+            WorkTime wt = getWorkTime(canChangeWorkTime, techSkillToWorktimeMap, skill, tech);
```
IllianiBird: We should be moving away from abbreviated variable names. Yes I know it's a pain, but `wt` should be `workTime` or some other descriptive name

IllianiBird (approved):
One tiny change needed. Marking as approved so it doesn't block your merge.

MekHQ/src/mekhq/service/mrms/MRMSService.java
psikomonkie: I'll change it in the other places in the class too.

PR #8526, opened 2025-12-15
Improvement: Minor Changes To Unit Sale Dialogs

IllianiBird:
During the development of Hammer's Obsolete Quick implementation there were a couple of issues I spotted. However, it was one of those situations where explaining exactly what needed to be done would take more time than just making the relatively minor changes myself. I mentioned this to Hammer and they greenlit my changes.

### DIFF
--- a/MekHQ/src/mekhq/campaign/unit/Unit.java
+++ b/MekHQ/src/mekhq/campaign/unit/Unit.java
@@ -1597,7 +1597,7 @@ public class Unit implements ITechnology {
         breakdown.append(getFormattedTextAt(RESOURCE_BUNDLE, "Unit.sellBreakdown.quality",
               qualityName, String.format("%.2f", qualityMultiplier))).append("<br>");
         breakdown.append(getFormattedTextAt(RESOURCE_BUNDLE, "Unit.sellBreakdown.currentWorth",
-              currentWorth.toAmountAndSymbolString())).append("<br>");
+              currentWorth.toAmountString())).append("<br>");
 
         // Check for obsolete quirk
         Money sellFor = currentWorth;
@@ -1613,7 +1613,7 @@ public class Unit implements ITechnology {
 
         breakdown.append("<hr>");
         breakdown.append(getFormattedTextAt(RESOURCE_BUNDLE, "Unit.sellBreakdown.sellFor",
-              sellFor.toAmountAndSymbolString()));
+              sellFor.toAmountString()));
         return breakdown.toString();
     }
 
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -1613,7 +1613,7 @@ public String getSellValueBreakdown() {
 
         breakdown.append("<hr>");
         breakdown.append(getFormattedTextAt(RESOURCE_BUNDLE, "Unit.sellBreakdown.sellFor",
-              sellFor.toAmountAndSymbolString()));
+              sellFor.toAmountString()));
```
IllianiBird, line 1616: `toAmountAndSymbolString` is a weird critter, because it doesn't display very cleanly. 5,000,000 million C-Bills displays as `5000000 CBILL` because `CBILL` is the symbol string. Far easier to just fetch the amount string and add in `C-Bills` inside the user-facing resource.

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -143,6 +143,8 @@
 public class UnitTableMouseAdapter extends JPopupMenuAdapter {
     private static final MMLogger LOGGER = MMLogger.create(UnitTableMouseAdapter.class);
 
+    private static final String RESOURCE_BUNDLE = "mekhq.resources.GUI";
```
IllianiBird, line 146: Resource bundles need to live at the top of the class and should be all caps underscored as they're constants

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -210,6 +212,7 @@ public class UnitTableMouseAdapter extends JPopupMenuAdapter {
     // endregion GM Commands
     // endregion Commands
 
+    @Deprecated(since = "0.50.11", forRemoval = true)
     private final transient ResourceBundle resources = ResourceBundle.getBundle("mekhq.resources.GUI",
```
IllianiBird, line 216: The old `ResourceBundle` approach is being phased out in favor of i18n, so whenever we introduce a `RESOURCE_BUNDLE` constant to a class we need to deprecate the old `resources` bundle so future devs know to stop using it.

HammerGS (approved):

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -287,35 +290,65 @@ public void actionPerformed(ActionEvent action) {
                 }
             }
         } else if (command.equals(COMMAND_SELL)) {
```
IllianiBird, line 292: This is where the bulk of the changes occur.

The long and short of it is that the original implementation would trigger a single dialog for every unit selected. So if the player selected 3,485 individual units and tried selling them at once they would need to go through 3,485 individual dialogs - without any way of breaking that chain.

Now, if one unit is selected we trigger a dialog will all the details (as before), but if multiple units are selected we just show the total sell value for all units.

PR #8527, opened 2025-12-15
Improvement: Adjusted Minimum Map Sizes Based on Player Feedback

IllianiBird:
We received the first batch of player feedback for the new map sizes. The general consensus was that maps were a little too short (and sometimes too narrow), so this PR adds a minimum height and width of 2. This can be overridden in campaign options for players who prefer larger or smaller engagements.


psikomonkie (approved):
Nice!

PR #8536, opened 2025-12-15
Improvement: Deprecated 'No Commander' Nag

IllianiBird:
A few versions ago we added a system to dynamically fetch the campaign commander, in the event none has been explicitly flagged. That removes the need for this nag as it is no longer possible to trigger it through normal play. This PR deprecates it.


Sleet01 (approved):
LGTM

PR #8543, opened 2025-12-16
Fix #8411: Fixed Tactics bonus not applying for individual initiative

### LINKED ISSUE #8411: MM not importing Individual Initiative bonus from MekHQ
### Brief Description *

MegaMek is not importing the Individual Initiative bonus from MekHQ.  Honestly, thinking about it, I haven't seen the Commanders' bonus in a while either, but I haven't been looking either.

### 3. Steps to Reproduce

Set MekHQ and MM to use Individual Initiative.
Give pilot Tactics skill (enough for bonus).
Start MegaMek scenario from MekHQ.
May want to double check Commanders' Initiative.

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.10 DEVELOPMENT RELEASE (see, I CAN learn)

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

psikomonkie:
Fixes #8411 

Individual initiative needs to be set via `initBonus()`. We should only set it if `isUseInitiativeBonus()` is true, and we should only set the commander one if `isUseTactics` is true.

### DIFF
--- a/MekHQ/unittests/mekhq/campaign/mission/AtBDynamicScenarioFactoryTest.java
+++ b/MekHQ/unittests/mekhq/campaign/mission/AtBDynamicScenarioFactoryTest.java
@@ -47,6 +47,7 @@ import megamek.common.game.Game;
 import megamek.common.units.Entity;
 import mekhq.campaign.Campaign;
 import mekhq.campaign.campaignOptions.CampaignOptions;
+import mekhq.campaign.personnel.skills.RandomSkillPreferences;
 import mekhq.campaign.universe.Faction;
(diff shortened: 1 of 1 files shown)


IllianiBird:
One rules implementation question. I tried deriving from the code but my brain is Swiss cheese today so I'll have to ask for your patience:

We have two options: use commander initiative and use individual initiative. Multiple sources of initiative modifiers are do not stack, only the highest applies.

Furthermore, if using individual initiative, only the tactics modifier from that specific unit commander applies, not the overall commander.

Does this change follow those rules?

IllianiBird (approved):

PR #8549, opened 2025-12-17
Fix: Fixed Missing Default Campaign Option Value for Acquisition Type

IllianiBird:
As per title. This would prevent the loading of campaigns if choosing Tabula Rasa as the upgrade option.


Sleet01 (approved):
LGTM

PR #8550, opened 2025-12-17
Fix: Fixed Campaign Upgrader Getting Stuck (For Really Real This Time)

IllianiBird:
This PR is the latest (and greatest) attempt to fix the issue of the campaign upgrader getting stuck. After a _lot_ of digging through old forums and StackOverflow posts I think I've finally found the root cause.

Basically, it all boils down to a threading issue. We were launching the dialogs in the same event thread as the main GUI which caused that to either get stuck or to block the triggering of the various confirmation dialogs. That's my understanding, at least.

The first was to move the confirmation dialogs into their own thread, via 'invokeLater'. I've tested the process multiple times and it _appears_ to be working. Time will tell.

### DIFF
--- a/MekHQ/build.gradle
+++ b/MekHQ/build.gradle
@@ -294,6 +294,10 @@ distributions {
                 exclude 'log4j2.xml'
             }
 
+            from("${mmDir}/megamek/mmconf/munitionLoadoutSettings.yaml") {
+                into 'mmconf'
+            }
+
             from("${mmDir}/megamek/build/launch4j/lib") {
                 into "lib"
             }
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/gui/dialog/CampaignUpgradeDialog.java
```diff
@@ -203,23 +203,24 @@ public static void campaignUpgradeDialog(Campaign campaign, Runnable runnable) {
...
                 }
-                SwingUtilities.invokeLater(runnable);
             }
         }
 
-        new ImmersiveDialogNotification(campaign,
-              getFormattedTextAt(RESOURCE_BUNDLE, "CampaignUpgradeDialog.upgrading", MHQConstants.VERSION.toString()),
-              true);
+        campaign.addReport(DailyReportType.GENERAL, getFormattedTextAt(RESOURCE_BUNDLE,
+              "CampaignUpgradeDialog.upgrading", MHQConstants.VERSION.toString()));
```
IllianiBird: Moved from a dialog to just a campaign report post to reduce the number of dialogs we're triggering (without using another invokeLater.

PR #8553, opened 2025-12-17
Update for new availabilityMap argument in reconfigureEntities

Sleet01:
### NOTE: depends on MegaMek/megamek#7810

Update calls to `reconfigureEntities` for new signature.

At a later date, we may wish to modify which munitions are available to a given force or faction based on Scenario mods.
In that case, we can first generate the availability map explicitly, modify it, and then pass it back into `reconfigureEntities()` where it will replace the default map.

Testing:
- Ran all 3 projects' unit tests.


IllianiBird (approved):
I really don't like how many abbreviated variable names we have here, but I can tell that updating them will cascade across the class so I'm not going to ask for it. Stubbornness compels me to mention it, though, so here we are :D

PR #8554, opened 2025-12-17
Improvement: Added Force Generation, Challenge Skulls, and Generic Battle Value Glossary Entries

IllianiBird:
This addresses a frequent cause of user contact and confusion, namely players not understanding FG3 vs. FG2.5, contract skulls, and GBV.


Sleet01 (approved):
LGTM

PR #8578, opened 2025-12-20
Fix: Allowed Bulk Hiring Skill Override for Civilian Professions

IllianiBird:
This removes an old blocker that was in place to prevent us from attempting to override the skill level of professions without skills. When this guard was put in place civilian professions didn't really exist. Nowadays the only professions without skills are Dependent and None.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/HireBulkPersonnelDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/HireBulkPersonnelDialog.java
@@ -354,7 +354,9 @@ public class HireBulkPersonnelDialog extends JDialog {
         while (number > 0) {
             Person person = campaign.newPerson(selectedItem.getRole());
 
-            if ((useSkill) && (!selectedItem.getRole().isCivilian())) {
+            // Dependents & 'None' don't have skills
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/dialog/HireBulkPersonnelDialog.java
```diff
@@ -354,7 +354,9 @@ private void hire(boolean isGmHire) {
         while (number > 0) {
             Person person = campaign.newPerson(selectedItem.getRole());
 
-            if ((useSkill) && (!selectedItem.getRole().isCivilian())) {
+            // Dependents & 'None' don't have skills
+            PersonnelRole selectedRole = selectedItem.getRole();
+            if (useSkill && !selectedRole.isDependent() && !selectedRole.isNone()) {
```
psikomonkie, line 359: Are there other times where we need to check if Dependents or "None" don't use skills? 

Would it make sense to add a method to `PersonnelRole` like `public boolean usesSkills()` that returns false for none & dependent, so we only need to maintain this in `PersonnelRole`?

PR #8593, opened 2025-12-22
PR: Option to Disable Intel Breach

psikomonkie:
Adds a campaign option to "Disable Intel Breach Chance". This option can only be enabled if the MekHQ prisoner capture style is enabled. If the setting is enabled, intel breach events won't happen when releasing prisoners.


VicenteCartas:
@psikomonkie Is this PR active or could we close it for now?

PR #8633, opened 2026-01-03
Task: Updated Campaign Options Iconography for 50.12

IllianiBird:
Also added the missing iconography from the options added by #8558


psikomonkie (approved):
Thanks!

PR #8643, opened 2026-01-04
PR: Campaign Option Badge Handling

psikomonkie:
Replaces existing campaign option badges with programmatic badges.

# Usage

I've made the conscious decision to not include a `getMetadata` method **without** a Version parameter. Any new campaign option added will use its current version as a parameter. Any "legacy" ones will pass in `LEGACY_RULE_BEFORE_METADATA` (currently null; Using a constant so we can update this more easily, if needed).

## Adding a New Campaign Option
Instead of adding badges manually in the `.properties`, it's added when adding the Checkbox/Label/Dropdown. For a flag, you can do something like: 
```
lblFoo = new CampaignOptionsLabel("Foo", getMetadata(null, CampaignOptionFlag.DOCUMENTED));
```
(Though hopefully any new options won't have null for the version...)

Or you can specify the version it was added in, and the purple/green star will be applied based on MekHQ's current version:
```
lblFoo = new CampaignOptionsLabel("Foo", getMetadata(new Version(0, 50, 12)));
```

Or both!
```
lblFoo = new CampaignOptionsLabel("Foo", getMetadata(new Version(0, 50, 12), CampaignOptionFlag.IMPORTANT));
```

The flag field can take one or more flags, so you can even get wild and do:
```
lblFoo = new CampaignOptionsLabel("Foo", getMetadata(new Version(0, 50, 12), CampaignOptionFlag.CUSTOM_SYSTEM, CampaignOptionFlag.IMPORTANT, CampaignOptionFlag.RECOMMENDED));
```

## Adding a New Campaign Option Badge

1. Add the new badge's symbol (required), description (optional), and color (optional) to `CampaignOptionsDialog.properties`
2. Add an Enum for the new badge to `CampaignOptionFlag.java`

That's all! Now you can use this flag in `getMetadata` and it will apply the badge

## 0.50.11 Milestone
The other reason I want to get this merged is that it'll automatically handle the version badge for the pre-milestone stuff, we don't need to manually update the properties file (beyond what I've already done) - Once `megamek.SuiteConstants.LAST_MILESTONE` is updated (this comes from `mmconf/milestoneReleases.yml`, which hasn't been updated yet!), the features that were added in 0.50.10 or prior won't have the version badge anymore. Wow! The few new options added during the current dev cycle will also maintain the "Added this version" badge until the next dev cycle, and they will then automagically switch to "Added this Milestone".

## Future Enhancements
- Tooltip integration: I defined descriptions for some of the tooltips but they won't be used in the current implementation. My initial idea was that mousing over one of the tooltips would popup what that tooltip means. But I think that'd be a hostile UX: Take "OpFor Equips Special Munitions ✎ ⌘" as an example. Do we want three different messages to pop up depending on if the user mouses over one of the icons or the main text? My current thought is that we should modify the tooltip to include the descriptions of the icons. We've already overloaded & modified the various component classes to do this for the text, and we'd be using the same information here but for the tooltip, so I don't think this would be a massive change. As an example, this is what the tooltip would be like:
```
Use semi-intelligent configuration of allied and enemy munitions, tailored to each scenario.
✎ - Custom system unique to MekHQ
⌘ - Documentation included in MekHQ/docs
```

### DIFF
--- a/MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsSpinner.java
+++ b/MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsSpinner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2025-2026 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *


psikomonkie:
This is finally ready for review.

PR #8721, opened 2026-02-03
Display surgeon skill level instead of target number in prosthetic surgery dialog

HammerGS:
Origin: Player report - confusion between skill level and target number display

  Root Cause: The prosthetic surgery dialog showed the required skill LEVEL (e.g., "5") but displayed the surgeon's
  capability as a TARGET NUMBER (e.g., "5+"). These are different values that happened to look similar, causing players
  to think their surgeon qualified when they didn't.

  Fix: Display the surgeon's skill level instead of target number, matching the requirement display format. This allows
  direct comparison between requirement and capability.


Sleet01 (approved):
LGTM

PR #8730, opened 2026-02-05
Fix warehouse "In Use" column showing incorrect counts (Fixes #8698)

### LINKED ISSUE #8698: [Issue][50.11] Warehouse shows incorrect Parts in use and discrepancy with Parts in use panel
### Brief Description *

In this instance I stripped my only Mek with Endo-steel, sold what was left of the chassis, and proceeded to sell the endo-steel internals. However, they showed as "in use: 1" in the warehouse, eventhough it's gone. The Parts in Use Report, however, correctly showed "0".
I included the save but reloading the save fixed the issue.

### 3. Steps to Reproduce

I guess:
1. strip a mech with unique parts
2. check if they are still "in use" in either displays (warehouse and PiUR)

### Attach Files

[parts_in_use.zip](https://github.com/user-attachments/files/24837941/parts_in_use.zip)

### Severity *

### MekHQ Suite Version *

50.11

### Operating System *

win10

### Java Version *

17.0.11

### Final Verification

HammerGS:
Opening this for Co-Pilot, needs player testing and not 100% convinced yet.

## Root Cause

  The "In Use" column in the Warehouse tab was showing 0 for parts that were actually installed on units. This was caused by three issues:

  1. Stale cache: PartsTableModel.partsUseData was only populated at construction time and never updated when refreshPartsList() was called
  2. Armor name inconsistency: The Armor.changeType() method (used during refits) generated names like "Armor (Standard)" while the constructor generated "Armor (IS Standard)" or "Armor (Clan Standard)". This caused PartInUse.equals() to fail due to description mismatch, tracking the same armor type in separate entries
  3. First-match lookup: The "In Use" column lookup broke on the first matching part found. If parts were incorrectly split into separate PartInUse entries, it might return a zero-count entry instead of the correct one

  ## Changes

  1. PartsTableModel.java - Added updatePartsInUse() method to refresh the cache
  2. WarehouseTab.java - Call updatePartsInUse() in refreshPartsList()
  3. Armor.java - Fixed changeType() to use consistent name format matching the constructor
  4. PartsTableModel.java - Made lookup more robust by finding maximum useCount among all matching parts

  Files Changed

  - MekHQ/src/mekhq/campaign/parts/Armor.java - Name consistency fix
  - MekHQ/src/mekhq/gui/model/PartsTableModel.java - Cache refresh and robust lookup
  - MekHQ/src/mekhq/gui/WarehouseTab.java - Call cache refresh

  Testing

  - Warehouse "In Use" column now correctly shows count of parts installed on units
  - Values update properly when parts are added/removed from units
  - Armor that went through refits displays correct "In Use" counts

  Fixes #8698


SJuliez (commented):
Code is ok. That Copilot issue seems valid, although I don't know what that column should really show

IllianiBird:
@HammerGS was a decision reached for this PR?

HammerGS:
Gonna close this. I'm sure we have issues with Parts in Use and Warehouse. But I think a different approach would be better.

PR #8751, opened 2026-02-18
Fixed Eagle Eyes rolls for Advanced Scouting

heckindoc:
There was a missing negative sign, I added the negative and now eagle eyes does what it is supposed to do

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -1811,7 +1811,7 @@ public class StratConRulesManager {
                 PersonnelOptions options = crewMember.getOptions();
                 int complementaryModifier = !hasSensorEquipment && // Doesn't stack with Sensor Equipment
                                                   options.booleanOption(OptionsConstants.MISC_EAGLE_EYES) ?
-                                                  -1 : 0;
+                                                  1 : 0;
 
                 int scoutSkillLevel = (scoutSkill == null) ? -1 : scoutSkill.getTotalSkillLevel(skillModifierData);
                 scoutSkillLevel += complementaryModifier;


MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -1811,7 +1811,7 @@ private static List<ScoutRecord> buildScoutMap(Force force, Hangar hangar, boole
                 PersonnelOptions options = crewMember.getOptions();
                 int complementaryModifier = !hasSensorEquipment && // Doesn't stack with Sensor Equipment
                                                   options.booleanOption(OptionsConstants.MISC_EAGLE_EYES) ?
-                                                  1 : 0;
+                                                  -1 : 0;
 
                 int scoutSkillLevel = (scoutSkill == null) ? -1 : scoutSkill.getTotalSkillLevel(skillModifierData);
```
psikomonkie, line 1816: For this specific usage, is lower better or worse & should eagle eyes be a bonus or penalty?

The other usage is a TN, lower is better, all good. 

This usage looks like lower is worse - this value will be added to scoutSkillLevel, which defaults to -1 if the skill isn't found. 

I believe the eagle eyes value here should still be positive if it's meant to be a bonus.

heckindoc: Eagle Eyes functions in the same fashion as a beagle probe in the setting of Advanced Scouting, decreasing the TN by 1.

In testing on 50.11, Eagle Eyes increased the TN by 1 and when made negative and retested in the 50.12 environment Eagle Eyes then had the intended effect of reducing the TN by 1.

psikomonkie: Changing "1" to "-1" for the TN is correct, I agree. 

I don't think this specific usage is for a TN though. 

EDIT: This is for determining the best scout. It calculates their effective skill level, a positive value, and returns whoever has the highest value. This change will cause persons with Eagle Eye to be considered worse than they are.

heckindoc: I'll try to take another look at it tonight and test it further and/or see if I can find a different solution. Thank you!

Issue #8691, opened 2026-01-22
[Issue] Advanced Scouting rolls being used on non-existent hexes (50.11)


heckindoc:
### Brief Description *

When a unit is deployed to the very edge of the map, if one of the successful rolls for Advanced Scouting took place on a non-existent hex (off map) within the area of operations, MekHQ is treating the non-existent hex as present for the purposes of successful rolls. In the example below, you can see that there are 3 successful rolls. The next image is of the AoO map immediately following the roll. The base was and already explored hex -- The first successful roll took place to the north hex (off map) then the subsequent two successful rolls moved clockwise as intended. 

I have seen other reports of this within Discord, but have not seen any reports on GitHub. 

<img width="659" height="569" alt="Image" src="https://github.com/user-attachments/assets/f3ee6b4f-ee12-4e03-927b-9fe4f48c96ea" />

<img width="908" height="954" alt="Image" src="https://github.com/user-attachments/assets/ae2233c4-ac0d-4703-9dee-da2d769732e8" />

### 3. Steps to Reproduce

1. From an already explored hex at the edge of the AoO map, deploy a scouting lance
2. Ensure the number of successes coincides with overlapping the off-board portion of the map

### Attach Files

[Cossack's Phalanx30591219_Rubbish-part1.zip](https://github.com/user-attachments/files/24809026/Cossack.s.Phalanx30591219_Rubbish-part1.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

50.11

### Operating System *

Win 11

### Java Version *

JDK 25

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PR #8769, opened 2026-02-27
Stop Advanced Scouting from being able to scout off StraCon board

heckindoc:
Fixes Issue #8691  -- Previously when determining which hexes to scout, advanced scouting would translate clockwise without checking to ensure that the scouted hex existed on the board. We now compare against the size of the board before determining which hexes to reveal.


heckindoc:
> Nice! Can you "Resolve Conversation" on the copilot requested changes once you've made them? Thanks!

I didn't even see that option. All done now. Thank you for taking a look at this!

PR #8850, opened 2026-03-30
Fix NPE in PlanetarySystem.getStar() when star data is missing

HammerGS:
## Root Cause
When a planetary system YAML file fails to parse (e.g., due to invalid enum values), the system can be partially loaded with a null `star` field. `Systems.cleanupSystems()` already has validation to detect and remove systems with missing stars, but `PlanetarySystem.getStar()` calls `getSourcedStar().getValue()` without a null check  so the NPE is thrown before the validation can catch it.

## Changes
  1. `PlanetarySystem.getStar()` - Added null guard on `getSourcedStar()` return value so the method returns `null` instead of throwing an NPE

## Files Changed
  - `MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java` - Null-safe `getStar()` accessor

  ## Testing
  - Verified against log errors caused by mm-data SUCS planetary systems with invalid `Atmosphere` values (`"Low"`, `"Vacuum"`)
  - With this fix, `cleanupSystems()` correctly detects and removes systems with missing star data instead of crashing
  - MegaMek/megamek#8138 separately fixes the Atmosphere deserialization root cause; this fix hardens the cleanup path against any future parse failures

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -666,7 +666,7 @@ public class StratConRulesManager {
         finalizeScenario(backingScenario, contract, campaign);
         swapInPlayerUnits(scenario, campaign, FORMATION_NONE);
 
-        if (!autoAssignLances && !scenario.ignoreForceAutoAssignment()) {
+        if (!autoAssignLances && !scenario.overrideForceAutoAssignment()) {
             for (int forceID : scenario.getPlayerTemplateForceIDs()) {
                 backingScenario.removeFormation(forceID);
             }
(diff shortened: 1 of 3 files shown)


MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
```diff
@@ -473,7 +473,8 @@ public double getTimeToJumpPoint(double acceleration, int sysPos) {
     }
 
     public StarType getStar() {
```
IllianiBird, line 475: If a method can return `null` as part of its normal functionality we wanna make sure we update the signature to tell future developers and the IDE that this method can null.

Just update like 475 to read `public @Nullable StarType getStar() {` and you'll be golden.

HammerGS: Done. Added `@Nullable` annotation to the return type: `public @Nullable StarType getStar()`. Also added the `megamek.common.annotations.Nullable` import.

PR #8864, opened 2026-04-08
fix #8680 Birth announcement missing names

### LINKED ISSUE #8680: [Issue] Birth announcement missing names
### Brief Description *

Image

Image

Image

I'm pretty sure those numbers in the brackets shouldn't be there. (Both children were generated with names, and the mother had one to begin with, but the announcement seems to be just using variable placeholders.

### 3. Steps to Reproduce

_No response_

### Attach Files

[The Alley Cats30520126.cpnx.gz](https://github.com/user-attachments/files/24705692/The.Alley.Cats30520126.cpnx.gz)

### Severity *

### MekHQ Suite Version *

0.50.11

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

AlexOldenburger:
The babyBorn.message strings for twins/triplets/quadrupletsPlus had only a single apostrophe in some places. Since these strings are formatted with MessageFormat.format any apostrophes should be doubled. (So **_I''ve_** instead of **_I've_** for example).

Also with twins/triplets/etc the dialogue was showing up for each new baby, which doesn't make sense as the text mentions all children already in that case. So this was changed to show the dialogue only once.

closes #8680


psikomonkie (approved):
Nice!!

Issue #6606, opened 2025-04-09
Player dropship doesn't spawn despite being assigned automatically to mission


theyoungster21:
### Brief Description *

Deep Raid Defense scenario was spawned through scouting. Player dropship was assigned to it automatically. When playing scenario, dropship does not appear in the scenario either on the map or among the player units. MegaMek appears to indicate that the ship is an illegal design and rejected, despite it being a standard unmodified leopard. Map was completed once and afterwards the player dropship was removed from hanger, despite not being present/damaged in the scenario. During the scenario the dropship is not treated as downed, and it does not default to starting grounded despite the scenario apparently calling for this.

Possible side issue, a allied force dropship player appears to be assigned to the mission with no units, presumed to be a placeholder for if player did not have dropship to assign to scenario. 

### 3. Steps to Reproduce

1. Load save and check hanger/TOE to verify dropship is included among forces
2. Advance 2 days to scenario date
3. Start scenario "Deep Raid Defense"
4. See that player dropship Leopard has not been included among units
5. verify server message that dropship was excluded due to illegal design
6. if scenario is completed in victory (presumably also in defeat) dropship will be removed from hanger afterwards

### Attach Files

[The 16th Freshly Repaired Bugged Dropship Defence.zip](https://github.com/user-attachments/files/19660532/The.16th.Freshly.Repaired.Bugged.Dropship.Defence.zip)

[logs.zip](https://github.com/user-attachments/files/19660534/logs.zip)

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MekHQ Suite Version *

50.04

### Operating System *

Windows 11

### Java Version *

17.0.14

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
<img width="578" alt="Image" src="https://github.com/user-attachments/assets/8a231cf2-02d8-4e58-b355-78615bf78b38" />

I need your custom unit to be able to open your save and investigate this issue :)

theyoungster21:
[customs.zip](https://github.com/user-attachments/files/19675976/customs.zip)

[...]

PR #8866, opened 2026-04-10
Fix: Recover player units rejected by the server instead of treating them as total losses

VicenteCartas:
### Problem

In #6606, a player's standard Leopard dropship was assigned to a Deep Raid Defense scenario, but MegaMek's server rejected it as an illegal design. The unit never appeared in the game. After scenario resolution, MekHQ treated the missing unit as a total loss — removing it from the hangar and losing its crew — even though it never participated in the battle.

### Fix

This PR adds a defensive safety net in `ResolveScenarioTracker` to prevent silent unit loss if any unit is ever rejected by the server in the future. A new `recoverUnfoundDeployedUnits()` method runs after all result categories (active, graveyard, retreated, devastated, wrecked) have been processed in both `processGame()` and `processMulFiles()`. For each deployed unit still marked as a total loss whose ID never appeared in results, the method:

- Calls `assignFoundEntity(entity, false)` to return it to the campaign unharmed
- Adds the crew to the `pilots` map so they aren't marked KIA/MIA
- Logs a warning identifying the affected unit for diagnostics

### Testing

- Added `processGameRecoversDeployedUnitNotInResults` — verifies a deployed unit absent from all results (server-rejected) is recovered
- Added `processGameDoesNotRecoverUnitFoundInResults` — verifies a genuinely destroyed unit in the devastated list remains a total loss
- Validated manually with the save file from the bug report: the Leopard now deploys and resolves correctly


psikomonkie:
Does this fix the entire issue in #6606? If it does close the issue in #6606, you should use a keyword in the PR's description to ensure that when this is merged it'll close the associated issue: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

Reading #6606, I have one question around it:
- Is there a data issue with the Leopard dropship that's causing it to show as illegal when moving between MekHQ and MegaMek?

Would you have time to take a peek into that aspect of the issue - is the Leopard valid in MegaMek if added via MegaMek, but invalid if sent from MekHQ? If so, I think we need to investigate that as a separate bug (and therefore **not** add the keyword to this PR so #6606 remains open). Not asking you to fix it, just wondering if you could take a quick peek so we can determine if we should close #6606 or leave it open.

VicenteCartas:
> id if sent from MekHQ? If so, I think we need to investi

The fix on Megamek side for this issue, sorry I should have pushed that one first instead of this defensive check here first:

https://github.com/MegaMek/megamek/pull/8182

We were using remaining ammo to consider a design legal, and we should be using the design ammo, not what it currently has.

With that fixed, you can see that the Leopard gets to deploy in the scenario (I used the bug save for this repro, updated it manually to 50.06, 50.11, 50.12 and then run on my branch):

Screenshot 2026-04-09 212617

I didn't test playing the whole scenario to see what happens at the end, that's the goal of the unit tests on this PR.

VicenteCartas:
@psikomonkie The check for illegal design is fixed here now in Megamek: https://github.com/MegaMek/megamek/pull/8182

This is just a defensive check in MekHQ, in case there are more illegal situations we don't know about. We could argue we prefer not to have the defensive check so we get the bug report and address the root cause instead for the next report. But I think this defensive check is useful: the illegal unit will not deploy in megamek, the player will realize the unit is missing and report it, and at least they don't lose it in their campaign. Let me know what you think!

Sleet01 (approved):
LGTM

PR #8868, opened 2026-04-10
Fix missing tempCrewDeathBenefitsReport

AlexOldenburger:
After a scenario where a unit with temp crew loses some of the crew, there should be a line in the Personnel tab of the Daily activity log with information about this. But currently the line only shows !tempCrewDeathBenefitsReport! 

The reason for this is that this resource name is not properly referred to, this PR fixes that.

ReportMissing

### DIFF
--- a/MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
+++ b/MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
@@ -862,14 +862,14 @@ public class ResolveScenarioTracker {
     }
 
     /**
-     * Attempts to assign a casualty to temp crew members (blob crew) if any exist.
-     * Casualties are randomly distributed across all temp crew types proportional to their numbers.
-     * If no blob crew exists or the random roll indicates a Person should be hit, the casualty
-     * is assigned to the Person instead (setting dead or wounded status). This can be expanded in the
-     * future to include support for wounding temp crew.
+     * Attempts to assign a casualty to temp crew members (blob crew) if any exist. Casualties are randomly distributed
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
```diff
@@ -1674,28 +1673,29 @@ private void processTempCrewDeathPayouts() {
...
             // Add a report entry
             campaign.addReport(
-                mekhq.campaign.enums.DailyReportType.PERSONNEL,
-                getFormattedTextAt(RESOURCE_BUNDLE, "tempCrewDeathBenefitsReport",
-                    totalPayout.toAmountAndSymbolString(),
-                    totalKilled)
+                  mekhq.campaign.enums.DailyReportType.PERSONNEL,
+                  getFormattedTextAt(RESOURCE_BUNDLE, "ResolveScenarioTracker.tempCrewDeathBenefitsReport",
+                        totalPayout.toAmountAndSymbolString(),
+                        totalKilled)
```
Sleet01, line 1687: It looks like your tabs setting doesn't match the coding standard.

Sleet01 (approved):
Looks fine, the tabs aren't critical but I'd ask that you take a look at your setting and make sure what you have matches the coding standards in the wiki.

MekHQ/src/mekhq/campaign/ResolveScenarioTracker.java
AlexOldenburger: As you mentioned below, I'll have a look again and see if I missed something.

I didn't touch any of those lines with whitespace changes, so I guess they result from applying the editorconfig settings when saving?

PR #8873, opened 2026-04-12
Fix: Exclude unstreamlined DropShips from atmospheric StratCon scenarios

### LINKED ISSUE #6834: [Issue] Invalid Dropship assigned to planetary operations
### Brief Description *

The Behemoth dropship, and theoretically any dropship with Unstreamlined, should not be able to operate in atmosphere. MHQ still selects it for dropship defense type missions.

### 3. Steps to Reproduce

Have a behemoth.
Get a dropship defense mission.
Wonder how the heck you are going to launch

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.05

### Operating System *

Windows 3.2

### Java Version *

No coffee here

### Final Verification

VicenteCartas:
## Summary

Unstreamlined DropShips (e.g., Behemoth) are no longer assigned to ground or low-atmosphere StratCon scenarios when the planet has atmosphere. They remain eligible for space scenarios and for operations on airless (vacuum) worlds.

## Changes

- `isValidUnitForScenario()` now receives `Campaign` and `MapLocation` parameters instead of just a boolean, enabling board-type and planetary condition checks
- Added generic board-type compatibility filtering using `doomedOnGround()` / `doomedInAtmosphere()` / `doomedInSpace()`
- Added specific filtering for the `unstreamlined` negative quirk, with a planet atmosphere lookup to allow operations on vacuum worlds
- Changed method visibility from `private` to package-private for testability
- Added 8 unit tests covering all branches of the new validation logic

Fixes #6834

Companion PR: MegaMek/mm-data#315 (adds missing `unstreamlined` quirk to Behemoth data files)

### DIFF
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -1019,7 +1019,7 @@ public class StratConRulesManager {
      * @return {@code true} if the unit matches the template's requirements and can be included in the scenario,
      *       {@code false} otherwise.
      */
-    private static boolean isValidUnitForScenario(Unit unit, ScenarioForceTemplate scenarioForceTemplate,
+    static boolean isValidUnitForScenario(Unit unit, ScenarioForceTemplate scenarioForceTemplate,
           Campaign campaign, MapLocation mapLocation) {
         // Check if DropShips are allowed and the correct unit type matches
         if (scenarioForceTemplate.getAllowedUnitType() == 11 && !campaign.getCampaignOptions().isUseDropShips()) {
--- a/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
+++ b/MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
@@ -121,6 +121,7 @@ import mekhq.campaign.personnel.turnoverAndRetention.Fatigue;
 import mekhq.campaign.stratCon.StratConContractDefinition.StrategicObjectiveType;
 import mekhq.campaign.stratCon.StratConScenario.ScenarioState;
 import mekhq.campaign.unit.Unit;
+import mekhq.campaign.universe.Planet;
 import mekhq.gui.dialog.nagDialogs.CombatChallengeNagDialog;
 import mekhq.utilities.EntityUtilities;
 import mekhq.utilities.ReportingUtilities;
@@ -901,7 +902,7 @@ public class StratConRulesManager {
             for (Unit unit : potentialUnits) {
                 if (isValidUnitForScenario(unit,
                       scenarioForceTemplate,
-                      campaign.getCampaignOptions().isUseDropShips(),
+                      campaign,
                       scenario.getScenarioTemplate().mapParameters.getMapLocation())) {
                     scenario.addUnit(unit, scenarioForceTemplate.getForceName(), false);
                     AtBDynamicScenarioFactory.benchAllyUnit(unit.getId(),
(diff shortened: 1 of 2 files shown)


VicenteCartas:
@SJuliez This is the fix for the atmospheric operations, but in MekHQ instead, so we can check for vacuum.

binaryspica:
Would it only checks for Un-streamlined design quirk? It is usually said that Mammoth(52,000 tons) is the upper limit of the dropship that could enter atmosphere then go back to the space - although there is no nailed rules for this but it is the well known theory in Battletech universe. And for now Behemoth of MegaMek data lacks Un-streamlined design quirk, so if you only count for the quirk then those dropships are still chosen on the missions.

Thus, I think that it's safe that any dropships with 52,100 tons or more are need to be checked as well. Else is MegaMek already checks that those dropships are unfit for the atmosphere insertion and the addition to the checking for the weight is not required at all?

Edit: Also too slow dropships are must be excluded as well, for if I remember correctly you need at least thrust point of 4 to back to the space, so unless the dropship is bounded to the specific planet then it must not enter the atmosphere if it can't make thrust point of 4 or more.

Still, if make it easier then manually add anything with unfit for the ground duty Un-streamlined design quirk is a solution, though.

MekHQ/src/mekhq/campaign/stratCon/StratConRulesManager.java
```diff
@@ -1006,24 +1009,49 @@ private static Collection<Unit> findPotentialUnits(Campaign campaign, int explic
...
+        }
+
+        // Unstreamlined units (e.g. Behemoth) cannot operate in atmosphere or on the ground,
+        // but they can operate on airless worlds (vacuum)
+        if (isAtmospheric && entity.hasQuirk(OptionsConstants.QUIRK_NEG_UNSTREAMLINED)) {
+            Planet planet = campaign.getLocation().getPlanet();
+            if (planet == null || !planet.getAtmosphere(campaign.getLocalDate()).isNone()) {
+                return false;
+            }
+        }
```
psikomonkie, line 1051: The logic for unstreamlined (and other non-atmospheric capable Aeros) should probably live in MegaMek and not MekHQ. I believe PlanteraryConditions will be what you'll want to look for on the MM side for this. 


EDIT: `PlanetaryConditions::whyDoomed` looks interesting.

VicenteCartas: I tried that first :)

https://github.com/MegaMek/megamek/pull/8183

@SJuliez made a good point about this that checking for atmosphere is not that easy on the Megamek side doomed functions.

I could reopen the original megamek PR and leave the comment there that vacuum is not taken into account, or try to figure out a different way to handle it on the megamek side (I haven't looked a lot at the megamek code).

gsparks3:
> Would it only checks for Un-streamlined design quirk? It is usually said that Mammoth(52,000 tons) is the upper limit of the dropship that could enter atmosphere then go back to the space - although there is no nailed rules for this but it is the well known theory in Battletech universe. And for now Behemoth of MegaMek data lacks Un-streamlined design quirk, so if you only count for the quirk then those dropships are still chosen on the missions.
> 
> Thus, I think that it's safe that any dropships with 52,100 tons or more are need to be checked as well. Else is MegaMek already checks that those dropships are unfit for the atmosphere insertion and the addition to the checking for the weight is not required at all?
> 
> Edit: Also too slow dropships are must be excluded as well, for if I remember correctly you need at least thrust point of 4 to back to the space, so unless the dropship is bounded to the specific planet then it must not enter the atmosphere if it can't make thrust point of 4 or more.
> 
> Still, if make it easier then manually add anything with unfit for the ground duty Un-streamlined design quirk is a solution, though.

There are zero canon DS above 52,000 tons that _don't_ have Un-Streamlined quirk (and very few such DS period; there's only the Castrum and some Behemoth variants). Also, AFAIK that 52,000 ton limit is currently fluff, not a gameplay rule, and therefore potentially subject to change should new DS be introduced. 

Similarly, there are zero canon DS besides the Behemoth that do not have at least 3 Safe Thrust (and therefore 5 Maximum Thrust). The requirement for 4 Thrust is actually to _leave_ the atmospheric interface (see Total Warfare page 79); aerospace units with less than 4 Thrust available are still capable of atmospheric flight, or even atmospheric reentry _from_ space if desired. The requirement discussed on that page is also plausibly meant for "in the heat of combat" situations, given that any craft capable of continuous 1g thrust should be able to reach orbit _eventually_.

As such, nothing mentioned in the quoted post is an issue.

Sleet01 (approved):
This looks fine to me.
It would be ideal if we could rely on MegaMek but it makes sense to also check within MHQ first.

PR #8874, opened 2026-04-13
Fix NPE in Mass Training when person lacks the trained skill

### LINKED ISSUE #8805: [Issue]  Uncaught Exception when using Mass training
### Brief Description *

When using mass training I am sometimes getting uncaught exceptions, and I think its pushed some units to negative xp.

### 3. Steps to Reproduce

Load save.
Go to Mass personnel training
Select doctor, and Administration
there should be 1 doctor to be trained.
Click spend xp.

### Attach Files

[The Expired Licenses30410703_Submarine-part1.zip](https://github.com/user-attachments/files/25886182/The.Expired.Licenses30410703_Submarine-part1.zip)

### Severity *

### MekHQ Suite Version *

0.50.12 Nightly 3-10

### Operating System *

Windows 11

### Java Version *

Java 21

### Final Verification

VicenteCartas:
Fixes #8805

When using Mass Training (BatchXPDialog) to train a skill that a person doesn't yet have, `person.getSkill(skillName)` returns `null`. The skill improvement and XP deduction completed successfully, but the subsequent `PerformanceLogger.improvedSkill()` call dereferenced the stale null `skill` reference, causing an uncaught NPE.

Because the exception prevented `updatePersonnelTable()` from running, the already-trained person could remain in the table and be processed again on retry, potentially leading to negative XP.

### Fix
Re-fetch the skill reference after `improveSkill()` so the logger receives a valid object. `improveSkill()` unconditionally creates or upgrades the skill, so the refreshed reference is always non-null.

### Testing
After fix, the associated campaign from the bug doesn't throw a NPE anymore when using Mass training.

### DIFF
--- a/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
+++ b/MekHQ/src/mekhq/gui/dialog/BatchXPDialog.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016-2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2016-2026 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #8882, opened 2026-04-16
Use automatic alternate table row coloring

SJuliez:
See also MegaMek/megameklab#2174
and megamek/megamek#8197

This makes MHQ use automatic alternate table row coloring. This removes the need for table renderers to do this and it applies to all tables, in principle. There are a bunch of tables in MHQ that use specialized renderers that will hide this, for now, but many tables will use the automatic coloring.


VicenteCartas (approved):

SJuliez:
Ah turns out I should place the UIManager call into a method that is used by all three programs upon startup, then this PR is unnecessary. Therefore closing this.

PR #8899, opened 2026-04-22
Filter out AtBContract / AtBScenario changes

Sleet01:
This PR is a subset of all changes between 84e961b147c919431f7e7182cd52133c979fd900 and  61afd91a11dcacf39703c416c04c895a9d38aed0, but retains as much of the AtBContract and AtBScenario code as seems necessary to maintain Contract Market, Briefing Tab, and StratCon functionality at pre 0.51.00 states.

While cleaning up I did find that the AtBEventProcessor code was basically unused, and the processShipSearch removal could be widened without apparent issue, so I removed those after all.

Since this PR rewinds the Main repo back to 8ae972e29f7be8a08fda21334a1f58c8348111c5 before adding the above Java 21 language changes, I'm not sure what the best method for incorporating its changes in to Main will be.
This is mainly for illustrative / Proof of Concept purposes.


rjhancock:
Is being managed in #8900.

PR #8901, opened 2026-04-23
Fix: Salvage percentage rounding error (#5683)

### LINKED ISSUE #5683: [Issue] Salvage calculation rounding error?
### Brief Description *

Possibly a weird rounding error, possibly a bug. After reaching the Resolve Scenario screen, Salvage tab, taking certain pieces of salvage (in this case, anything under about 200k in value) lowers the percentage from 43% to 42%.
Untitled
Untitled2

### 3. Steps to Reproduce

_No response_

### Attach Files

[autosave.sav.gz](https://github.com/user-attachments/files/18328095/autosave.sav.gz)

### Severity *

### MekHQ Suite Version *

v0.50.02

### Operating System *

Windows 10

### Java Version *

17

### Final Verification

- [X] I have included all necessary information and files to help reproduce this issue

VicenteCartas:
The salvage percentage shown on the Resolve Scenario screen and the Mission view used `BigDecimal.intValue()` to convert the computed share to an integer, which **truncates** rather than rounds. Combined with the intermediate `HALF_EVEN` rounding done by `Money.dividedBy(Money)` at scale 2, this caused values like 42.99% to display as 42%, and small additions to the player's salvage pile could visually drop the percentage by a full point.

### Changes
- Added `Contract.calculateSalvagePercentage(Money, Money)` (static) and `Contract.getCurrentSalvagePct()` (instance convenience), using `BigDecimal.divide(..., 0, HALF_UP)` for proper nearest-integer rounding.
- Replaced the duplicated local helpers in `ResolveScenarioWizardDialog` and `MissionViewPanel` with calls to the shared `Contract` methods.
- Dropped now-unused `Money` import and a confusing/unused parameter in `MissionViewPanel.getCurrentSalvagePct`.

### Behavior
- 42.49% → 42% (was 42)
- 42.50% → 43% (was 42)
- 42.99% → 43% (was 42)  ← the reported bug
- 43.01% → 43% (was 43)

Fixes #5683.


hokvel:
If you look at the original bug report, the first screenshot shows 43% for 42.67% (30142128 and 40498831), which means it was rounded up. And current calculations `unit.multipliedBy(100).dividedBy(unit.plus(employer)).getAmount().intValue()` for this case indeed gives 42. Independently from the rounding method, the result of the current calculation is monotonically increasing. In the bug report it's not. It is 43 for 30142128 / x and 42 for 30326639 / x. My guess is that that calculation has changed since 50.02 and the bug won't be reproducible with the current logic.

hokvel:
I didn't find any changes in the calculation since 50.02, so I must be missing something and there's another explanation for why 30142128 and 40498831 give 43% result in that screenshot.

Independently from that. From the gameplay perspective it should be rounded up. If the contract allows 50% salvage, 50.001% would be a breach.

VicenteCartas:
Thanks for the feedback! I'll take a second pass at this.

VicenteCartas:
> I didn't find any changes in the calculation since 50.02, so I must be missing something and there's another explanation for why 30142128 and 40498831 give 43% result in that screenshot.
> 
> Independently from that. From the gameplay perspective it should be rounded up. If the contract allows 50% salvage, 50.001% would be a breach.

Addressed :)

hokvel (approved):
I'm still not sure how to reproduce the original bug and if this change fixes it, but it definitely fixes the existing rounding behaviour plus makes the implementation cleaner and improves test coverage.

rjhancock (commented):
Sanity Check complete.

rjhancock (approved):

VicenteCartas:
@rjhancock Can you merge? I don't have merge permissions.

rjhancock:
I only did a sanity check on the code. I'll let someone else handle testing and merge.

PR #8911, opened 2026-04-27
Fix: Restore Briefing Room tab content (regression from 61afd91a11)

VicenteCartas:
Commit 61afd91a11 removed the JSplitPane wiring in BriefingTab.initTab() and its add(splitBrief, BorderLayout.CENTER) call without a replacement, leaving the Briefing Room tab empty except for the tutorial hyperlink bar at the bottom. Restore a horizontal JSplitPane between the mission and scenario panels and attach it to BorderLayout.CENTER.


VicenteCartas:
@rjhancock I think this came from #8900

VicenteCartas:
This was fixed in main, closing.

PR #8939, opened 2026-05-05
Give POW awards automatically at mission end

AlexOldenburger:
Currently, POW awards are never given automatically. The reason for this is that this award is based on the POW status of a person, but this status is cleared (just) before the auto awards are processed at mission end.

This PR fixes this issue by building a list of current POWs at mission end, and then using that list to determine which persons should get the award.

Unfortunately, due to the way auto awards are set up, this list needs to be passed on through several methods before it can actually be used to determine if a person should get the award. Overloaded methods help with this but it is still not ideal.


IllianiBird (approved):
autoAwards was my first major project, that it works at all is a miracle. Thanks for fixing this

PR #8964, opened 2026-05-14
Added Ability to Cover Expenses from Compulsion Addiction to Reduce Campaign Impact

IllianiBird:
# Requires #8963

See Also: https://github.com/MegaMek/mm-data/pull/357

Previously there was an issue where certain Flaws had an overwhelming effect on player experience, beyond what was intended. This PR adds the ability for the player to bypass Substance Addiction by paying a weekly stipend.


psikomonkie (approved):
There seems to be a recurring pattern with all these if statements. 

I almost feel like these different conditions should be some kind of new `Enum`. 

Still approved :)

PR #8967, opened 2026-05-14
Fix #8862: Added Notice to Clarify Surgeon Selection for Advanced Surgeries

IllianiBird:
Fix #8862

For Advanced Surgeries, when selecting which doctor will perform the surgery the highest ranked surgeon will be picked. Even if that character is of lower skill than another character. This can create confusion in the event a lower ranked surgeon can perform the surgery but the senior surgeon cannot.

It would be possible to change this so that the highest _skilled_ surgeon performs the surgery however we need to ensure that this remains predictable as certain SPAs can influence the TN.

Therefore we have kept the picker working the same, but instead went ahead and added some text to clarify how surgeon is picked.


psikomonkie (approved):
Nice!

HammerGS (approved):

PR #8976, opened 2026-05-15
Fixed Bloodmark Handling in Education System

IllianiBird:
Previously if a student had a Bloodmark value they could be Bloodhunted and injured. This resulted in injuries that couldn't be healed, because the character wasn't with the campaign. The student would then continue schooling with six broken arms and their bowels hanging out. Not ideal.

This PR addresses this by hooking into the Education injury system. Now Bloodmarked characters have an increased risk of 'accidents' while at school.


psikomonkie:
It's still a little raw, but I hope it'll be ready very soon. This may be of some use to you for an alternate solution, should you be interested: https://github.com/psikomonkie/mekhq/pull/4

IllianiBird:
Mm, potentially. I do think going with this approach is best because otherwise we'd need to refactor education injuries to avoid having in-education injuries handled in two separate ways.

HammerGS (approved):

PR #8997, opened 2026-05-18
Expanded Visibility of Flags in the Personnel View

IllianiBird:
Over time we have added numerous new personnel Flags, but these were not visible in the personnel table. That made it annoying for players to quickly reference who had what flag.

Expanded the single Flags view into three views. Each with up to 7 flags (the most visible before the GUI gets cluttered).


MekHQ/unittests/mekhq/gui/enums/PersonnelTabViewTest.java
```diff
@@ -181,17 +181,6 @@ public void testIsDates() {
...
-    public void testIsFlags() {
-        for (final PersonnelTabView personnelTabView : views) {
-            if (personnelTabView == PersonnelTabView.FLAGS) {
-                assertTrue(personnelTabView.isFlags());
-            } else {
-                assertFalse(personnelTabView.isFlags());
-            }
-        }
-    }
-
```
IllianiBird, line 194: No longer relevant. And all this test does is test whether a change has occurred, _not_ whether something has gone wrong.

HammerGS (approved):

PR #9015, opened 2026-05-20
Added Campaign Option to Improve Relevance of Veterancy Awards

IllianiBird:
When Veterancy Awards were implemented I made the decision to make the SPA award largely random. In my mind this would create memorable situations where a character in one profession might get an amazing SPA, prompting users to cross-train. Or a character might get an otherwise impossible combination.

This approach was also done to reduce the 'power' of these free SPAs.

Over time, based on player feedback, I have softened on this somewhat. Especially with just how many RP SPAs there are.

With all this in mind, I've added a new campaign option. If selected characters will only be given Veterancy Awards that they would be eligible to take had the player purchased the SPA. So no more NatApt on Administrators.


psikomonkie (approved):
Nice!

PR #9018, opened 2026-05-20
Updated Infirmary Nag Text to Better Inform Players of Automation Options

IllianiBird:
We have seen semi-frequent contact from some users unaware that there are automation options for their infirmary.

I went ahead and updated the infirmary nag text to better signpost players to these options.

image


psikomonkie (approved):
Nice!

PR #9020, opened 2026-05-20
[Improvement] Migrate financial tab lines to resource bundle and standardize formatting 

heckindoc:
The goal of this PR is primarily 2 things:
1. Take all consumer facing strings in the Financial Tab summary and move them to Resource Bundles. I opted to use the already created FinancesTab.properties bundle.
2. Create a standardized method for formatting the financial tab summary lines. Previously, each individual line was formatted independently - this created a lot of room for formatting error and increased the amount of work that needed to be put into adding any new lines to the summary. I've added a `formattingFinancialReport` method which does a few things 
 - standardizes the SB length to 44, this way Java isn't trying to create unnecessarily long strings for each line
 - I increased the total length of each line by 1 character to allow for larger number values to be displayed.
 - provides standardized formatting based on indentation for each line in the finances tab
 - makes it significantly easier to modify the formatting on ALL lines of the financial tab summary in the future through modifying one method, as opposed to modifying approximately 15 individual lines.

Before:
image

After:
image


MekHQ/src/mekhq/gui/FinancesTab.java
```diff
@@ -514,53 +542,56 @@ public String getFormattedFinancialReport() {
...
-              .append('\n');
+
+        sb.append(formattingFinancialReport(resourceMap.getString("netWorth.text"), 1,
+              String.format(formatted, netWorth.toAmountAndSymbolString())));
+        sb.append('\n');
+
+        sb.append(formattingFinancialReport(resourceMap.getString("assets.text"), 2,
+              String.format(formatted, assets.toAmountAndSymbolString())));
+
+        sb.append(formattingFinancialReport(resourceMap.getString("cash.text"), 3,
```
IllianiBird, line 553: This isn't cause for a 'request changes' but an FYI. We're trying to avoid using single letter and abbreviated variable names nowadays, to improve code clarity.

MekHQ/src/mekhq/gui/FinancesTab.java
```diff
@@ -501,6 +501,34 @@ public void refreshFinancialReport() {
...
+      the switch will default to no indentation.
+
+      Note: if the title for a line (ie. under Monthly Expenses) ever exceeds 25 characters, this
+            method will need to be modified to account for additional characters.
+    */
+    public String formattingFinancialReport(String financialLine, int indentLevel, String printedValue) {
+        //Total length of all characters in report line is 44
+        StringBuilder sb = new StringBuilder(44);
+        switch (indentLevel) {
+            case 1 -> sb.append("");
```
IllianiBird, line 517: Another FYI, no change needed. You can use the escape key `&nbsp;` in a resource bundle to have it include a leading whitespace. i.e.

`someResourceKeyTitle=&nbsp;&nbsp;&nbsp;BlahMcBlahBlah`

Would display as...

`   BlahMcBlahBlah`

IllianiBird (approved):

MekHQ/src/mekhq/gui/FinancesTab.java
heckindoc: Also I wanted to say thank you for this! I purposely didn't do this in case the terms for the resource bundle were used elsewhere and the leading white space wasn't desired in that location.

PR #9024, opened 2026-05-21
Improved Fetching of Character Information for Personnel Tables

IllianiBird:
The method we used to build strings for display in the personnel tables was gross, with lots of repetitive code. I needed to make changes for another PR, so I took the opportunity to update everything.

### DIFF
--- a/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
+++ b/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
@@ -37,6 +37,8 @@ import static mekhq.campaign.personnel.turnoverAndRetention.Fatigue.getEffective
 import java.time.LocalDate;
 import java.util.Comparator;
 import java.util.ResourceBundle;
+import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.swing.JTable;
 import javax.swing.SortOrder;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
```diff
@@ -644,525 +649,301 @@ public String getCellValue(final Campaign campaign, final PersonnelMarket person
...
-            case WILLPOWER:
-                return getAttributeScoreDisplay(person, SkillAttribute.WILLPOWER);
-            case CHARISMA:
-                return getAttributeScoreDisplay(person, SkillAttribute.CHARISMA);
-            case EDGE:
-                int currentAttributeValue = person.getAttributeScore(SkillAttribute.EDGE);
-                int attributeCap = person.getAttributeCap(SkillAttribute.EDGE);
-                return currentAttributeValue + " / " + attributeCap;
-            default:
-                return "UNIMPLEMENTED";
```
psikomonkie, line 1179: Do we need a `default` case? If we don't include one, that'll have the IDE preventing it from compiling unless the case is explicitly handled, which I think is a good thing instead of letting accidentally missed cases render as "UNIMPLEMENTED". If we want to add something that is unimplemented, we can explicitly add it to this list and have it return "UNIMPLEMENTED" or whatever

psikomonkie: That's what it did before, and I even responded to the "removed" version not the added version... The merge conflicts are from a fix to the personnel crew roles, I just want to make sure that's properly merged into this PR and I'll approve this.

psikomonkie (approved):
Nice! Excited to fix my conflicts... ;)

Issue #1099, opened 2019-02-21
Feature Request: Planetary Base Hangar


Morkath:
### Environment
0.44.0

### Description
It would be useful if there was a way to store equipment in a location that did not need to be brought with you on contracts.  This would allow smaller merc companies to accrue some material without necessarily needing a bigger dropship, and adjust their forces based on the type of work they are currently doing.

While having a stockpile of 50 Large Lasers can be useful so you always have them available for replacements, I don't really need to bring them all with me every contract.

As an example layout flow; add a button on the interstellar map tab, to set a planet as your home base.  Then add a tab in between the Warehouse and Repair Bay tabs for the Base UI.

* Only one planet could be active as a Home Base at a time.  Attempting to choose a new one should prompt the user that they will be abandoning their old base and its contents.
* It would have a numerical field for user to choose the available storage tonnage for the base.
* Have an option for having a Field Workshop, Transport Bay, Maintenance Facility, or Factory on site.
* Stretch Goal: Allow the user to set units to a Training role while remaining at the planet.
* Increase monthly expenses based on these options.

Then be able to right click an item in the Personnel/Hangar/Warehouse and Send it to the Base, and vice versa, when on that planet.

jbhuddleston:
**Environment**

0.46.0rc1

**PartialSolution**

I have the same issues and am trying the following to meet some of them.

Build a minimal dropship in MegaMekLab that has just the features you are discussing then save it in customs and add it to your campaign hangar. Add the entire crew complement then tweak them to have real base admin roles other than maintaining a dropship. Now you have mech bays and cargo space. When you leave the planet, you will have to split your campaign then merge it when you return unless you take your base with you :-) Not perfect but usable so far.

Perhaps Ground Bases could be added to the features for MegaMekLab?

Here is my 16 mech, 6 vehicle Base requiring 1 pilot and 3 crew as staff.
[LandBase Underground-16m6v.zip](https://github.com/MegaMek/mekhq/files/3501952/LandBase.Underground-16m6v.zip)

NickAragua:
There's an "export campaign subset" feature coming in the next dev version which should take some of the sting out of splitting up a campaign.

PR #9025, opened 2026-05-21
PR: Refactor `CurrentLocation` to better support objects being in different locations

psikomonkie:
Refactors `CurrentLocation` into:
- `AbstractLocation`: A specific location/position, currently only implemented in `CurrentLocation` but I intend to implement `Location` in the next PR which'll be used for things that will never move - Academies. 
- `LocationNode`: A simple node for a tree structure for storing where things are.
- `ILocation`: An interface for classes with a `LocationNode` that supports various basic location operations

Also adds unit tests.

This is the first step towards implementing features like MegaMek/mekhq#1099. 

This PR should not have any impact on the user or gameplay. This PR updates `Campaign` to implement `ILocation` and to utilize a `LocationNode` instead of directly working with a `CurrentLocation`. 

Additional design notes can be found on Discord.


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -801,7 +808,7 @@
     }
 
     public PlanetarySystem getCurrentSystem() {
-        return location.getCurrentSystem();
+        return getCurrentLocation().getCurrentSystem();
```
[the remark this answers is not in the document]
psikomonkie: fixed

MekHQ/src/mekhq/campaign/events/LocationChangedEvent.java
```diff
@@ -34,7 +34,7 @@
 package mekhq.campaign.events;
 
 import megamek.common.event.MMEvent;
-import mekhq.campaign.CurrentLocation;
+import mekhq.campaign.AbstractLocation;
```
IllianiBird, line 37: Hrm, I'm a bit concerned that `AbstractLocation` is going to be confusing. Especially as we have several things 'location' could refer to. This isn't a major thing, but maybe change it to `AbstractGalacticLocation` or something similar.

MekHQ/src/mekhq/campaign/location/ILocation.java
```diff
@@ -0,0 +1,339 @@
...
+
+import mekhq.campaign.AbstractLocation;
+import mekhq.campaign.JumpPath;
+import mekhq.campaign.parts.Part;
+import mekhq.campaign.personnel.Person;
+import mekhq.campaign.unit.Unit;
+import mekhq.campaign.universe.Planet;
+import mekhq.campaign.universe.PlanetarySystem;
+
+public interface ILocation {
```
IllianiBird, line 65: I thought you were going with IPlace to avoid confusion and conflicts with unit part locations?

MekHQ/src/mekhq/campaign/location/ILocation.java
```diff
@@ -0,0 +1,339 @@
...
+
+    LocationNode getLocationNode();
+
+    default boolean hasLocationNode() {
+        return getLocationNode() != null;
```
IllianiBird, line 76: We're using `null` as shorthand for 'nothing is assigned'. What might be a good move is to create a class-level static that equals `null` and then propagate that, instead of putting `null` each time.

It's an additional level of abstraction, but means that if you ever change what the no assigned location signpost is, you only need to change it in one place.

It will also make the code itself a bit easier to parse, compare...

`return getLocationNode() != null`

vs.

`return getLocationNode() != NO_ASSIGNED_LOCATION`

MekHQ/src/mekhq/campaign/location/LocationNode.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public LocationNode(ILocation locatable) {
+        this.locatable = locatable;
+    }
+
+    public ILocation getLocatable() {
+        return locatable;
+    }
+
+    public AbstractLocation getCurrentLocation() {
```
IllianiBird: As this can return `null` as part of its normal operation it should have `@Nullable` in its method signature

There might be other instances of this in the PR, this is just the first one I spotted.

MekHQ/src/mekhq/campaign/market/personnelMarket/markets/PersonnelMarketMekHQ.java
```diff
@@ -199,7 +198,7 @@ public ArrayList<Faction> getApplicantOriginFactions() {
      */
     @Override
     public String getAvailabilityMessage() {
-        CurrentLocation location = getCampaign().getCurrentLocation();
+        mekhq.campaign.AbstractLocation location = getCampaign().getCurrentLocation();
```
IllianiBird: Do we need the full class address here, or can we just import it into the class?

I'm seeing this pattern across a few classes, so won't flag them all.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -532,7 +535,11 @@ public Campaign(Game game,
         game.setOptions(gameOptions);
         this.techFaction = techFaction;
         this.systemsInstance = systemsInstance;
-        location = startLocation;
+        this.locationNode = new LocationNode(this);
+        if (startLocation != null) {
+            locations.add(startLocation);
+        }
+        this.setParent(startLocation);
```
IllianiBird, line 540: Is there value in breaking this out into a helper method?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1716,8 +1719,22 @@ public List<Scenario> getActiveScenarios() {
         return scenarios.values().stream().filter(s -> s.getStatus().isCurrent()).toList();
     }
 
-    public void setLocation(CurrentLocation l) {
-        location = l;
+    public void setLocation(AbstractLocation l) {
```
IllianiBird: I know it's a small thing, but I'd really rather we avoided single-letter and abbreviated variable names.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -8341,7 +8364,7 @@ public void shutdownAtB() {
      *       "Employee Turnover", 1 if user selected "Advance Day Regardless", 2 if user selected "Cancel Advance Day"
      */
     public int checkTurnoverPrompt() {
-        if (!location.isOnPlanet()) {
+        if (!isOnPlanet()) {
             return -1;
```
IllianiBird, line 8367: I know this isn't your code, but it'd be good to define `-1` somewhere, so we're not returning a magic number

MekHQ/src/mekhq/campaign/AbstractLocation.java
```diff
@@ -0,0 +1,413 @@
...
+          TransportCostCalculations transportCostCalculations) {
+        double currentRechargeTime = currentSystem.getRechargeTime(date, isUseCommandCircuit);
+
+        StringBuilder report = new StringBuilder();
+        report.append("<html>")
+              // First Line
+              .append("In ").append(currentSystem.getPrintableName(date)).append(' ');
+
+        if (isOnPlanet()) {
+            report.append("on planet ").append(getPlanet().getPrintableName(date));
```
IllianiBird: We shouldn't use player-facing hardcoded text any more. I know this wasn't your code originally, but as we're moving it across can we get a RESOURCE_BUNDLE set up and use that, instead?

MekHQ/src/mekhq/gui/dialog/CampaignExportWizard.java
```diff
@@ -40,16 +40,7 @@
...
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
-import java.util.ResourceBundle;
-import java.util.UUID;
+import java.util.*;
```
IllianiBird, line 43: Doesn't look like anything was done in this class, probably should revert this dirty edit. Or don't, it's not really a problem. I just think we should always avoid dirty edits where possible, but that's probably just an ingrained thing from my modding days. Your call, either way is good.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -64,8 +64,8 @@
 import javax.xml.parsers.DocumentBuilder;
 
 import megamek.Version;
-import megamek.client.ui.CopySystemDataAction;
 import megamek.client.generator.RandomUnitGenerator;
+import megamek.client.ui.CopySystemDataAction;
```
IllianiBird, line 68: See prior dirty edit comment

MekHQ/unittests/mekhq/campaign/stratCon/StratConRulesManagerTest.java
```diff
@@ -59,12 +59,11 @@
 import java.util.UUID;
 import java.util.Vector;
 
-import org.junit.jupiter.api.Test;
-import org.mockito.MockedStatic;
-import org.mockito.Mockito;
-
-import mekhq.campaign.CurrentLocation;
+import megamek.common.options.OptionsConstants;
```
IllianiBird, line 62: Looks like more dirty edits

IllianiBird (changes requested):
Just some relatively minor feedback. Not everything needs actioning, but there are a couple of things I'd like to see tweaked.

Otherwise, great work as always! :)

MekHQ/src/mekhq/campaign/events/LocationChangedEvent.java
psikomonkie: I added Javadocs to the class, if you think that helps enough. I'll keep thinking of a better name. I did consider something like that at one point but I felt it was a mouthful.

MekHQ/src/mekhq/campaign/location/ILocation.java
psikomonkie: Yesn't. That idea needs baked a little more. But regardless `ILocation` will still remain, for anything that "has a location". `IPlace` will be a specific case of it that for some special kinds of `ILocation`.

MekHQ/src/mekhq/campaign/location/ILocation.java
psikomonkie: This isn't the absence of a location, this is the absence of the physical tether to reality. This shouldn't happen. An `ILocation` should always have an associated `LocationNode`. 

However, I do agree that there should be a static `NO_ASSIGNED_LOCATION` but used one level up - If `getCurrentLocation` etc. etc. returns null then that'd be a good case to use `NO_ASSIGNED_LOCATION`.

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: It's used in two places, and the other spot it's used is just magic numbers... It needs refactored to an enum probably. I'll submit an issue for it so it isn't lost, but I'd rather not include it here just so we don't have to hit more classes in this PR.

psikomonkie: New issue for tracking this is #9065.

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: Good catch! Already got one, replaced 539-542 with `Campaign.setLocation(AbstractLocation)`.

MekHQ/src/mekhq/campaign/market/personnelMarket/markets/PersonnelMarketMekHQ.java
psikomonkie: The usages in this PR were when there were 3 or less usages of the class. I switched them all to imports.

But I do have a question now - do we have a policy on this ?

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: Great time to ask for it! Done.

MekHQ/src/mekhq/gui/dialog/CampaignExportWizard.java
psikomonkie: These dirty edits are from the auto format not hitting files that were changed in #9051 but the auto format didn't hit until I merged that into this branch. They'll be the same changes that'll happen next time someone edits the file, and they should've been changed in #9051. 

Therefore, I want to leave them.

MekHQ/src/mekhq/campaign/market/personnelMarket/markets/PersonnelMarketMekHQ.java
IllianiBird: I don't actually know, if we do it is not remotely enforced

IllianiBird (approved):

MekHQ/src/mekhq/campaign/location/LocationNode.java
psikomonkie: Done and did other spots too

MekHQ/src/mekhq/campaign/AbstractLocation.java
psikomonkie: Done!

MekHQ/src/mekhq/campaign/location/ILocation.java
psikomonkie: Done

PR #9030, opened 2026-05-21
Fix getting multiple injuries for twins/triplets/etc and remove all injuries from generated personnel

AlexOldenburger:
If a person gives birth to multiple children, a post-partum injury (or a hit) would be added for each baby. This is unintended, it should be a single injury (or hit).

Also, when using the option to run a simulation for a few years in the Company Generator, the generated personnel would also give birth. The post-partum injuries (or hits) from that would still be present on the person, even if the birth occurred years ago. This PR fixes that by removing all injuries (or hits) from the generated personnel.


IllianiBird (approved):
Great work :)

Issue #7849, opened 2025-10-26
[RFE] Chancing column title under Personnel > Flags to match the flag chosen


Insert25c:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

Currently a Column Title when looking at personnel under the Flags view reads "Interested in Children". Besides the creepy name, this doesn't match what the flag is called when selecting it from the Right Click menu. 

I propose changing it the "Trying to Conceive" or similar.

This also matches with the fact that the Yes/No data only applies to Female unit members (Males are listed as N/A)

### Use Cases or Rationale

1: its creepy

2: It keeps verbiage the same throughout the program 

3: title fits better with the data presented below 
Currently Males are listed as N/A because they cant conceive. Females are Yes/No. But from the standpoint of wanting children, a male could definitely want to have kids (its a common thing sought in a potential mate)

### Attach Files

Right Click Menu > Flags options
<img width="147" height="157" alt="Image" src="https://github.com/user-attachments/assets/cb77e330-865f-48ad-a073-bfd5d85860ac" />

Column Appearance currently
<img width="877" height="803" alt="Image" src="https://github.com/user-attachments/assets/a4acc17f-f87b-40c5-af29-743c7e400885" />

### MekHQ Suite Version *

.50.07

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

AlexOldenburger:
@Insert25c 

PR #9040 changes the column title to 'Wants children', would this satisfy this request? Maybe not since you mention that for males that could also be true.

So maybe the display should also be changed for them from N/A to Yes/No? In which case, for females a Yes would mean they want children and are trying to conceive. For males a Yes would mean they want children as well (but obviously not the trying to conceive part).

IllianiBird:
We shouldn't expose this for male characters _right now_ unless we're also going to edit the underlying relationship code to allow male interest in procreation is taken into consideration. We'd also need to change character generation so we roll for procreation interest at character creation. Neither should be too hard, but it would need to be done. We'd also likely need additional campaign options, one set for male procreation interest and another for female.

PR #9040, opened 2026-05-22
Some tweaks to the Flags pages in Personnel

AlexOldenburger:
Recently the Flags page in the Personnel tab was expanded. However, not all new flags were included in the new pages, some flags were duplicated and there were other issues.

This PR fixes this by doing the following:

-- Remove the (duplicate) columns for MARRIAGEABLE as the flag is deprecated and cannot be accessed any longer
-- Remove duplicate entry for TRYING_TO_CONCEIVE, and also change the text to 'Wants Children' to make it less creepy
-- Add new column definitions for the new COVER_MEDICAL_EXPENSES and BLOCK_MATERNITY_LEAVE flags
-- Assign colums for these two new flags and also for HIDE_PERSONALITY since it was missing
-- Rearrange the flags to fit together better thematically
-- Change the distribution to 6-6-4 to have more space for a column on each page 

-- Added a helper function to convert a boolean into a Yes or No text

-- For children, changed the display to N/A for LIKES_MEN, LIKES_WOMEN and TRYING_TO_CONCEIVE
    (Wait until they are of age before we concern ourselves with this.)


MekHQ/resources/mekhq/resources/GUI.properties
```diff
@@ -1354,7 +1354,7 @@ PersonnelTableModelColumn.FOUNDER.text=Founder
 PersonnelTableModelColumn.CLAN_PERSONNEL.text=Clan Personnel
 PersonnelTableModelColumn.MARRIAGEABLE.text=Interested in Marriage
 PersonnelTableModelColumn.DIVORCEABLE.text=Divorceable
-PersonnelTableModelColumn.TRYING_TO_CONCEIVE.text=Interested in Children
+PersonnelTableModelColumn.TRYING_TO_CONCEIVE.text=Wants Children
```
IllianiBird, line 1357: I wholeheartedly support this.

It used to be _way_ worse: "Trying to Conceive" was the original version of this flag

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
```diff
@@ -866,6 +868,9 @@ private String getDisplayString(Campaign campaign, PersonnelMarket personnelMark
                                                                   "No.text") :
                                                                  "NA.text");
             case UNDER_PROTECTION -> resources.getString(person.isUnderProtection() ? "Yes.text" : "No.text");
+            case COVER_MEDICAL_EXPENSES ->
+                  resources.getString(person.isCoverIllicitMedicalExpenses() ? "Yes.text" : "No.text");
```
IllianiBird: Not necessary, but if you wanna be clever and fancy, adding a helper method that takes in a bool and spits out yes.text or no.text wouldn't be amiss. There are _loads_ of examples of that pattern in this switch

IllianiBird (approved):
Approved, but a single suggestion. It's non-essential so if you don't feel like incorporating it - or want to do it in a follow-up PR - just lemme know :)

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
AlexOldenburger: I may have to revisit this anyway. Issue #7849 has the notion that male personnel may also want children. So maybe it would be better to just make this column a yes/no for everyone? Thoughts on that?

IllianiBird: I left some thoughts in the cited issue (I saw that before I saw this). I have zero conceptual issues with the move, my thoughts in the cited issue cover the immediate steps needed to handle that.

AlexOldenburger: I've added a helper function now. Other changes are best done with another PR.

PR #9041, opened 2026-05-22
Fix IllegalArgumentException with Financial Report Changes 

heckindoc:
Thanks to @AlexOldenburger for finding this problem

The changes I had previously made improved the formatting of the financial report, however when any lines financial amount exceeded 9,999,999,999 C-bills and illegal argument exception was thrown. This change increases the total length allowable to go up to 9 Trillion C-Bills natively, and a try-catch now exists to allow it to flex further up to 99 trillion. 

I don't like having this try catch block in place, but it allows for very large campaigns to continue. I will make another PR after I create a new method to grab the financial values of all lines and pull the maximum amount which can then be used to determine the length of the printed value line. 

Since I was in the area, I also improved the readability of the code by changing the StringBuilder `sb.*` to `financialLineFormatter.*`


IllianiBird (approved):
Campaigns are too damn rich xD

MekHQ/src/mekhq/gui/FinancesTab.java
```diff
@@ -511,22 +511,29 @@ public void refreshFinancialReport() {
...
-        //Total length of all characters in report line is 44
-        StringBuilder sb = new StringBuilder(44);
+        //Total length of all characters in report line is 50
+        StringBuilder financialLineFormatter = new StringBuilder(50);
         switch (indentLevel) {
-            case 1 -> sb.append("");
-            case 2 -> sb.append("    ");
-            case 3 -> sb.append("       ");
-            default -> sb.append("");
+            case 1 -> financialLineFormatter.append("");
```
IllianiBird, line 517: I appreciate you nixing the abbreviated variable name

PR #9050, opened 2026-05-23
fix AmmunitionRestock (fixes #8971)

### LINKED ISSUE #8971: [Issue] Ammunition Restock is Broken
### Brief Description *

Automatic ammunition restocking no longer works on my save, and I'm not sure why. I started setting my ammo percentages to be automatically set to 300%, but no ammunition doesn't restock at all. Changing the percentages on ammunition back to 100% doesn't fix it. I have the weekly restock on, but neither that nor the "Order Parts to Minimum Stock Levels" actually restock ammunition. The restocking of other parts and equipment seems not to be effected.

After some quick testing, it only restocks ammunition when I have no ammunition of that type.

I've had this issue for a while, it seems like setting them to higher percentages might cause the issue. It started affecting my ammunition once I started setting certain types to higher percentages for restocking, but as mentioned before lowering them again doesn't fix it.

### 3. Steps to Reproduce

1. Load save
2. "Open Parts In Use Report"
3. Observe ammunition below desired restock levels, in this case Gauss Rifle Ammo [IS]
4. Press "Order Parts to Minimum Stock Levels" button

### Attach Files

[Wallace's Warriors30520408.cpnx.gz](https://github.com/user-attachments/files/27779173/Wallace.s.Warriors30520408.cpnx.gz)

[customs.zip](https://github.com/user-attachments/files/27779178/customs.zip)

[logs.zip](https://github.com/user-attachments/files/27779212/logs.zip)

Note, the logs file was too large with the game summaries, so I had to remove them. They theoretically shouldn't be relevant for this issue, but let me know if you need them.

### Severity *

### MekHQ Suite Version *

0.50.11

### Operating System *

### Java Version *

Java Vendor: Eclipse Adoptium   Java Version: 17.0.18

### Final Verification

AlexOldenburger:
Currently, the automatic restock of ammunition does not work as intended. To determine how many should be ordered, the current inventory (+ # ordered and # in transit) is tallied up and this is compared to the requested tonnage. However, for ammunition the inventory is not the tonnage, but the total number of shots, so the automatic restock fails as it appears there is enough in stock. 

This PR fixes this issue by converting the inventory of shots in stock to their tonnage before calculating the number needed to fill the stock to the requested amount.

Note: the partInUse parameter name was changed to conform to the coding standards (should not start with a capital letter).

Fixes #8971

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/PartsInUseManager.java
+++ b/MekHQ/src/mekhq/campaign/market/PartsInUseManager.java
@@ -454,21 +454,26 @@ public class PartsInUseManager {
      * <ul>
      *     <li>Current inventory: parts in storage, in transfer, and planned acquisitions</li>
      *     <li>Required stock: the requested stock percentage applied to the part's usage count</li>
+     *     <li>Ammo storage adjustment: for ammo the amount in the inventory is the number of shots, so this needs to
+     *     be converted to its tonnage to determine the amount to buy.</li>
      *     <li>Bundle adjustment: for bundled items (e.g., armor sold in 5-ton blocks), the quantity is adjusted
      *         based on tonnage per item</li>
      * </ul>
      *
-     * @param PartInUse the {@link PartInUse} instance to calculate stock requirements for
+     * @param partInUse the {@link PartInUse} instance to calculate stock requirements for
      *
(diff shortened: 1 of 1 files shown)


IllianiBird (approved):
How many times do we need to fix this bug!? T_T Please let this be the last.

PR #9056, opened 2026-05-23
Updated Company Generator to Automatically Assign Ranks to Additional Support Unit Personnel

IllianiBird:
The existing company generator was never designed with anything but Meks in mind. When we added the new supporting units (such as convoys and soldiers) this was an absolute hack. Not in the sense that it was bad code, rather than we were bending the company generator in ways it wasn't intended.

Ultimately, this meant that the crew manning supporting units were generated with no ranks. This forced the player to manually go through and assign ranks.

Now, I have added a bit of code that will assign relatively appropriate ranks to all personnel that previously lacked them. It is a pretty primitive system and is not foolproof, but will work for 99% of player campaigns. Including those campaigns using custom rank sets.

Rather than going into the ins and outs of the actual assignment process I'll include some example pictures. First of conventional infantry, and then vehicles. These examples are using the default rank system used by the SLDF:

## Conventional Infantry
image

For conventional infantry the overall unit commander has the top rank, then their squad leaders are one step below. With everyone else getting a step below that.

## Vehicles
image

For vehicles, the overall unit commander has the top rank. One character is picked as their sub-commander (a following the rules for conventional infantry squad leaders). Everyone else gets a rank below that.

Ranks are given based on experience level, with better people getting a better rank. This is true for both conventional infantry and vehicles.

## Addendum
Again, this is a very primitive system. I would love to see someone take this and run with it, but I opted for a simple system we could implement now and improve later. Rather than just not implementing anything.

Special handlers are in place for ComStar/WoB and the Clans.


IllianiBird:
Popping this into draft while I adjust the approach

IllianiBird:
And back into live it goes

IllianiBird:
And back into draft it goes, weeee

IllianiBird:
And back to live again. In out, in out, shake it all about...

HammerGS (approved):

PR #9062, opened 2026-05-24
Fixed Prisoner-Defectors Being Incorrectly Recruited as Camp Followers

IllianiBird:
The prisoner system predates the concept of Camp Followers. There was a small oversight where, when recruiting defectors, they would be recruited as Camp Followers. Now they are employed fully.


MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -985,6 +985,10 @@ public void actionPerformed(ActionEvent action) {
                 for (Person person : people) {
                     if (person.getPrisonerStatus().isPrisonerDefector()) {
                         person.setPrisonerStatus(getCampaign(), PrisonerStatus.FREE, true);
+
+                        // You need to set the prisoner-defector to ACTIVE, otherwise they will be recruited as a
+                        // Camp Follower.
+                        person.changeStatus(getCampaign(), getCampaign().getLocalDate(), PersonnelStatus.ACTIVE);
```
AlexOldenburger, line 991: If you do a base assault, you may end up with defectors that only have civilian skills. (Because they are numerous, in my experience these occur the most often). It could be that you want to add those as camp followers, but maybe not as people on your payroll.

Is it possible to detect the type of defector and change their new status accordingly?

IllianiBird: Yes, super easy to do so, too.

My concern is that we're then creating two types of behavior for a single option. I'd be worried the player might get frustrated with the inconsistency.

Thoughts? :)

AlexOldenburger: Add another menu option 'Recruit and Employ'?

Or make a new campaign option to either just recruit or to recruit and employ.

I don't mind it much that I have to do two actions (in some cases). Other people may want a different behaviour.

If this gets merged as-is it's also not a big deal as after recruiting you can always change the status back to camp follower.

HammerGS (approved):

PR #9079, opened 2026-05-24
Fix #8801: Stopped Characters From Being Given Gunnery/ProtoMek as a Random Skill

### LINKED ISSUE #8801: [RFE]Prevent personnel to be generated with Gunnery/ProtoMek unless they are protomech pilot.
### Enhancement Type *

Currently there is the chance to gain the random skills in addition to their own. But it is weird that there would be protomech gunnery skill. By kerensky and blake, is it even makes sense?

I said prevent non-protomech pilot to gain gunnery/protomech correctly. It is intended for EI is needed to even train the protomech skills in Battletech universe so not even clanners are able to access for this unless they also has EI. So the skill must be exclusive for the protomech pilots. Maybe, Aerospace Pilot clanner with EI by default on them could have it, but it would be edge case.

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.11

### Final Verification

- [ ] I've discussed or asked about this enhancement on MegaMek Discord

IllianiBird:
Fix #8801

This created lore inconsistencies, blocking the skill was the easiest implementation.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/generator/DefaultSkillGenerator.java
+++ b/MekHQ/src/mekhq/campaign/personnel/generator/DefaultSkillGenerator.java
@@ -156,6 +156,14 @@ public class DefaultSkillGenerator extends AbstractSkillGenerator {
             List<String> possibleSkills = new ArrayList<>();
             for (String skillType : SkillType.skillList) {
                 SkillType type = SkillType.getType(skillType);
+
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

I wonder if we should have a skill : first available date map, similar to tech progressions?

IllianiBird:
The idea has merit, leave it to me. I'll do a follow-up PR after dinner

PR #9083, opened 2026-05-24
Cleaned Up Risky Salvage Code

IllianiBird:
I was asked to check the risky salvage code, to see if it was possibly inflicting more accidents than intended.

I did a thorough review and walked through the code with breakpoints and I didn't see any issues.

I did however go ahead and clean the code up a bit, there _was_ a minor bug but it wouldn't have actually resulted in bugged behavior, rather we just wouldn't have early exited saving a few parses.

### DIFF
--- a/MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
+++ b/MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
@@ -86,6 +86,7 @@ import mekhq.campaign.stratCon.StratConTrackState;
 import mekhq.campaign.unit.TestUnit;
 import mekhq.campaign.unit.Unit;
 import mekhq.campaign.unit.enums.TransporterType;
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/campaign/mission/camOpsSalvage/CamOpsSalvageUtilities.java
```diff
@@ -359,61 +360,27 @@ public static void performRiskySalvageChecks(Campaign campaign, List<UUID> techU
...
+        int injuryEvents = getInjuryEventsCount(numberOfSalvagedUnits);
 
-            techs.add(tech);
-        }
+        List<Person> techs = getValidTechs(campaign, techUUIDs);
 
         boolean didAccidentOccur = false;
         for (int i = 0; i < injuryEvents; i++) {
-            if (techUUIDs.isEmpty()) {
+            if (techs.isEmpty()) {
```
IllianiBird, line 369: This is the only functional change

Sleet01 (approved):
LGTM!

PR #9086, opened 2026-05-25
Close #8904: Added Field Control Statement to Salvage Formation & Tech Pickers

### LINKED ISSUE #8904: [RFE] Display field control on salvage force allocation panel on scenario start
### Enhancement Type *

Display scenario field control conditions on the salvage force allocation panel in the pre-scenario wizard; optionally display deployment time, and number of enemy units deployed in the scenario. Adding better percentage-per-unit in the post resolution dialogue would also be useful. See rationale for full breakdown. 

This is the relevant panel:

Image

### Use Cases or Rationale

When deploying to a scenario, you are given the option to allocate a salvage force. This pre-mission dialogue consists of three panels.
The first is a copy of the mission briefing dialogue and the objective. Thanks to consistent writing, this will usually include a paragraph about who controls the field. 
The second one is the salvage force selection. 
The third one is the tech allocation. 

My user experience so far has been that I will often click past the first panel to get the scenario started, and then remember I need to set a salvage force. Because I have ADHD-I, the note of who controls the field and therefore determines whether or not I need to bring a salvage force is something I have often already forgotten, or missed reading initially. This forces me to cancel the dialogue boxes, re-read the briefing to see who will hold the field, and then re-start the scenario. 

This could be improved by displaying field control on the second panel; this is the point where the player actually wants to know if it is worth deploying a salvage force; if the enemy controls the field, you almost never want to deploy a force. 
Optionally: displaying the number of enemy units deployed in the scenario might also prove to be useful, as would be the expected return time for the salvage force. This is all information that is pertinent to the allocation of salvage teams. 

Finally, displaying a better breakdown of what salvage percent a unit in the post-engagement salvage dialogue breaks down to would also be useful; specifically, if a unit has been marked for recovery, display the % of the total salvage value this unit represents. The intended user experience is therefore as follows: allocate recovery units, and then use the % value to determine what picks the player wishes to salvage / cash in, and which ones go to the employer. This is may be more intuitive than trying to balance salvage value by C-Bills, especially since the player/employer split is shown in percentages. 

### Attach Files

_No response_

### MekHQ Suite Version *

v0.50.13-nightly-2026-04-07

### Final Verification

IllianiBird:
Close #8904

Now the salvage formation and tech pickers will show who will control the field below the instructions.

For non-AtBDynamicScenarios (i.e. scenarios without FieldControl) we display 'Unknown'.

### DIFF
--- a/MekHQ/resources/mekhq/resources/SalvageFormationPicker.properties
+++ b/MekHQ/resources/mekhq/resources/SalvageFormationPicker.properties
@@ -29,15 +29,16 @@
 # <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
 # affiliated with Microsoft.
 # suppress inspection "UnusedProperty" for whole file
-SalvageFormationPicker.instructions=Under CamOps salvage rules a salvage team cannot partake in the fighting. They sit\
-  \ on\
-  \ the sidelines until the OpFor has been routed.\n\n\
+SalvageFormationPicker.instructions=Under CamOps salvage rules a salvage team cannot partake in the fighting. They \
+  sit on the sidelines until the OpFor has been routed.\n\n\
   Only 'Meks and vehicles can participate in ground salvage. Only DropShips and WarShips can participate in space \
   salvage. Units that are not fully crewed cannot salvage.\n\n\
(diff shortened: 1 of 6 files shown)


Sleet01 (approved):
LGTM

PR #9089, opened 2026-05-25
Close #8813: Arriving at Your Destination or Completing a Contract With Now Populate the Personnel Market Instantly

### LINKED ISSUE #8813: [Issue] [50.11] Personnel market not refreshing when a contract ends / contract market not refreshing when reaching HH
### Brief Description *

It's neither really a bug nor a RFE.

Markets refresh at the start of the month, but the personnel market doesn't when in space or on operations.
However, it's not that it shouldn't refresh, but rather that you can't make use of it. Wouldn't it be reasonable to refresh it (i.e. populate it) right when a contract ends? It's a bit strange that you have to skip days up until the new month.

Similarly, when reaching a new destination system, shouldn't the contract market refresh according to the new parameters (i.e, Hiring Hall situation)? Again, otherwise you have to oddly skip days, eventhough it would seem that the contracts were already there - just not you!

### 3. Steps to Reproduce

_No response_

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.11

### Operating System *

win10

### Java Version *

17.0.11

### Final Verification

IllianiBird:
Close #8813

Previously the market would only populate at the start of the campaign, or on the first of each new month.

Has no effect on the old, deprecated, personnel markets.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -4337,13 +4337,14 @@ public class Campaign implements ITechManager, ILocation {
     /**
      * Refreshes the personnel markets based on the current market style and the current date.
      *
-     * @param isCampaignStart {@code true} if campaign method is being called at the start of the campaign
+     * @param bypassDateRestrictions {@code true} if we want the market to refresh at an unusual time, such as campaign
+     *                               start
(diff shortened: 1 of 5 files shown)


Sleet01 (approved):
LGTM

Issue #8889, opened 2026-04-21
[v0.50.11] Monthly Expenses in Finances are Wildly Inaccurate


WallaceLordofOwls:
### Brief Description *

Books Referenced: 
Campaign Operations with Errata v5.0
Strategic Operations with Errata 3.02
Field Manual: Mercenaries Revised, Unknown Version

Currently, MekHQ has maintenance costs for units implemented both using the rules in Campaign Operations page 24, a system calculating maintenance costs directly using a percentage of unit costs, and a system calculating maintenance values using the table in the Field Manual Mercenaries pg. 148. Both the CamOps system and the percentage unit cost systems are on by default under the "Veteran" preset, and the "Maintenance" category showing the percentage unit costs and the FM:Merc are always present, even when these options are turned off.

To clarify, under the Monthly Expenses section in the Finances section of the MekHQ interface, there are six costs listed: Salaries, Maintenance, Overhead, Spare Parts, Training Munitions, and Fuel.

Under Campaign Operations, only the Salaries, Maintenance, Spare Parts, Training Munitions and Fuel costs are calculated.
Here's an example to demonstrate:
<img width="337" height="106" alt="Image" src="https://github.com/user-attachments/assets/6403b863-1bdf-4186-9488-0b890d74db8f" />
This is a mercenary company with a single Leopard (2537) to confirm maintenance costs.
The values for spare parts, training munitions, fuel all match calculated values. More specifically:
- Spare Parts: 1900 Tons * 0.001 (0.1%) * 15000 C-Bills = 28500 C-Bills
- Training Munitions: 27000 C-Bills (Cost per Ton of LRM Ammunition) * 3 Tons (Of Ammunition) * 1/4 = 45000 C-Bills
- Fuel: 0 C-Bills. This is a bit odd, but CamOps p. 24 references Strategic Operations p. 179, which states that units with Fusion Engines (which includes the Leopard), can generate 10 Tons of Hydrogen, the fuel Dropships use, per Maintence/Repair Cycle (One day, per SO p. 166). While per CamOps p. 24 the Leopard would consume 15 burn-days of fuel, which for the Leopard at 1.84 tons per burn day would be 27.6 tons, it would be able to produce 300 tons of hydrogen over the month. I do not know if MekHQ tracks this or simply assumes peacetime fuel costs for dropships are zero, but that is outside the scope of this issue.

[...]

PR #9090, opened 2026-05-25
[improvement] Remove zeroed out lines from Finance Tab Summary

heckindoc:
Partially addresses #8889 

Depending on the settings that a player has, they may not have to pay any overhead or maintenance costs. When this happens they have two lines in their monthly costs that will persistently be zero. Additionally, a large number of player will not have loans, and they will have lines on their finance tab which will always be zero. In order to decrease clutter and improve readability, I just added `if` statements for each line that can be zeroed out in settings that I'm aware of.

### DIFF
--- a/MekHQ/src/mekhq/gui/FinancesTab.java
+++ b/MekHQ/src/mekhq/gui/FinancesTab.java
@@ -616,12 +616,16 @@ public final class FinancesTab extends CampaignGuiTab {
                       .append('\n');
             }
         }
-        sb.append(formattingFinancialReport(resourceMap.getString("liabilities.text"), 2,
-              String.format(formatted, liabilities.toAmountAndSymbolString())));
+        if (!liabilities.isZero()) {
+            sb.append(formattingFinancialReport(resourceMap.getString("liabilities.text"), 2,
+                  String.format(formatted, liabilities.toAmountAndSymbolString())));
+        }
 
-        sb.append(formattingFinancialReport(resourceMap.getString("loans.text"), 3,
-              String.format(formatted, r.getLoans().toAmountAndSymbolString())));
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/FinancesTab.java
```diff
@@ -616,12 +616,16 @@ public String getFormattedFinancialReport() {
...
-              String.format(formatted, liabilities.toAmountAndSymbolString())));
+        if (!liabilities.isZero()) {
+            sb.append(formattingFinancialReport(resourceMap.getString("liabilities.text"), 2,
+                  String.format(formatted, liabilities.toAmountAndSymbolString())));
+        }
 
-        sb.append(formattingFinancialReport(resourceMap.getString("loans.text"), 3,
-              String.format(formatted, r.getLoans().toAmountAndSymbolString())));
-        sb.append("\n\n");
+        if (!r.getLoans().isZero()) {
```
IllianiBird: I know it's not your code, but would you be willing to rename `r` to a proper variable name? `sb` is fine, because it's pretty standard for StringBuilder, but `r` really isn't clear without digging through the code

IllianiBird (changes requested):
Just a small change, I know it isn't your code but little improvements over time lead to a tidy codebase.

MekHQ/src/mekhq/gui/FinancesTab.java
heckindoc: I was actually thinking about how, with as much as I've been playing in the area of the code, I really should rename those variables. 

Changes have been made and pushed. Let me know if there's anything else you'd like done.

IllianiBird (approved):
Thanks for getting this done

PR #9091, opened 2026-05-25
Fix #8737: Added Safeguard for Missing Game Options When Saving Older Campaigns

### LINKED ISSUE #8737: [Issue] Can't save after moving from MHQ 0.50.06 to MHQ 0.50.11
### Brief Description *

Image
I get this every time I try to save the campaign. I've tried saving it in different file locations as well. No luck.

### 3. Steps to Reproduce

1. Open MHQ with old campaign
2. awesome built in updates apply automatically.
3. Try to save.
4. Fails with message.

Image

### Attach Files

[Customs.zip](https://github.com/user-attachments/files/25223057/Customs.zip)

[mekhq.log](https://github.com/user-attachments/files/25223062/mekhq.log)

[megamek.log](https://github.com/user-attachments/files/25223063/megamek.log)

[Plot Armor BrigadeIlClan.cpnx.gz](https://github.com/user-attachments/files/25223016/Plot.Armor.BrigadeIlClan.cpnx.gz)

### Severity *

### MekHQ Suite Version *

0.50.11

### Operating System *

Windows 11 Home 25H2 26200.7705 Windows Feature Experience Pack 1000.26100.291.0

### Java Version *

Eclipse Adoptium Temurin JRE with Hotspot 21.0.1+7 (x64)

### Final Verification

IllianiBird:
Fix #8737

Under rare circumstances it seems that a campaign can end up with malformed MegaMek Game Options. `null` to be precise. This then prevents saving, because the saving process assumes `GameOptions` has been substantiated.

I've explored the campaign creation process and I cannot find any explanation why campaigns are instantiating with `null GameOptions`. It _shouldn't_ be possible.

Clearly that isn't the case, so what I've gone ahead and done is added a catch that will self-correct in the event a `null GameOption` is detected during the save process. The player will be notified when this occurs.

### DIFF
--- a/MekHQ/resources/mekhq/resources/Campaign.properties
+++ b/MekHQ/resources/mekhq/resources/Campaign.properties
@@ -238,6 +238,5 @@ stratCon.earlyContractEnd.objectives=You have resolved all outstanding objective
   <p>The contract will end tomorrow. If the contract was successful, any outstanding payments will be rendered once \
   the contract has concluded.</p>
 gameOptions.save.failure.title=Broken Game Options
-gameOptions.save.failure.body=MegaMek's game options failed to save.n\n\
-  All options have been reset to their default values.n\n\
-  If this keeps happening, please report to the MekHQ team.
+gameOptions.save.failure.body=MegaMek's game options failed to save.n\n\All options have been reset to their default \
+  values.n\n\If this keeps happening, please report to the MekHQ team.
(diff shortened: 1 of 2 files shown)


Sleet01:
From what I recall, we might have seen this when a campaign contained game options that were no longer supported, or when the campaign's options did _not_ contain a value that was now required (due to a recent change in the nightly builds, say).
We'd throw an exception that wouldn't be handled until higher up the call stack than where we set the options instance, and would end up with the default null value.
But I never found the smoking gun.

Sleet01 (approved):
LGTM

PR #9092, opened 2026-05-25
Introduce CurrentLocationPanel and TransitStatusChangedEvent

hokvel:
This change introduces a brand new `CurrentLocationPanel` that displays a location image, planetary conditions, socio-industrial data, travel progress, jump costs, and the hiring hall button. Additionally, it takes a stab at a cleaner separation of event processing and domain logic versus the UI layer, simplifies `AbstractLocation` and `CampaignGUI` logic.

Detailed changes:
- `AbstractLocation.getReport` separated into more specific methods (`getTitle`, `getPlanetaryConditionsInfo`, `getCourseInfo`, `getJumpCostInfo`, `getSocioIndustrialInfo`) and moved to `CurrentLocationPanel`
- Hiring Hall button moved from the Marketplace panel to the `CurrentLocationPanel`, conditionally enabling based on its availability
- Planetary conditions surfaced in the UI including highlighting based on modifiers they provide
- Introduced `TransitStatusChangedEvent` to broadcast in-system transit ticks, jump path alterations, and jumpship charging progress
- Removed manual UI updates, simplifying `CampaignGUI` and other map views by relying on event subscriptions
- Renamed the GM charging action from `setRecharged()` to `chargeFully()`
- Improved test coverage for `CurrentLocation`, notably for `newDay`
- Switched to `campaign.isUseCommandCircuit()` calls where possible


MekHQ/src/mekhq/campaign/events/TransitStatusChangedEvent.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final CurrentLocation location;
+
+    public TransitStatusChangedEvent(CurrentLocation location) {
+        this.location = location;
+    }
+
+    public CurrentLocation getLocation() {
+        return location;
+    }
```
psikomonkie: Could this be an `AbstractLocation` so if any future `AbstractLocations` were to have a transit status but not extend `CurrentLocation` this would still work?

MekHQ/unittests/mekhq/campaign/CurrentLocationTest.java
```diff
@@ -229,11 +261,7 @@ void fullyChargedReportAddedWhenRechargeComplete() {
             // Two reports: spending hours + fully charged
             verify(campaign, times(2)).addReport(eq(GENERAL), anyString());
         }
-    }
 
-    /** Tests for {@link CurrentLocation#writeToXML(java.io.PrintWriter, int)}. */
-    @Nested
-    class WriteToXml {
```
psikomonkie, line 236: Nesting here shouldn't be removed. This has the various `writeToXML` test methods under the `RechargeAccumulation` nested test suite.

psikomonkie (changes requested):
Nice! I love the new `CurrentLocationPanel`! I'm glad that inside `CurrentLocationPanel` it uses `AbstractLocation`. 

Two requested changes, but otherwise I think this looks great.

Thanks!

MekHQ/src/mekhq/campaign/events/TransitStatusChangedEvent.java
hokvel: Yes, good catch. Missed that when was rebasing this PR, which I started working on before AbstractLocation was introduced. Will double check if type requirements can be relaxed in other places.

MekHQ/unittests/mekhq/campaign/CurrentLocationTest.java
hokvel: Absolutely, will fix.

psikomonkie (approved):
Excellent, thanks for getting those changes in quick, hokvel! This is approved!

PR #9094, opened 2026-05-25
fix children still having portraits

AlexOldenburger:
The campaign option to have a default portrait for children that was introduced in PR #9007 does not work properly.

The underlying reason is that a call to isChild() is used to determine if a New Person is a child. However, at birth the Person is created first (including a portrait) and only then is the birthdate set for the new person.

This PR fixes this issue by checking if the new option is active after the birthdate is set. There is still a check for isChild(), since with historic births the baby may have been born many years earlier than the current date in the campaign and could now be at the age where a portrait should be added.

If the campaign option is used and the person is still a child, a default Portrait will be set.


IllianiBird (approved):
Whoopsie doodles

PR #9103, opened 2026-05-26
Fix #8735: Added Instant MRMS Button Back to Command Center

### LINKED ISSUE #8735: [Issue][50.11] Show Instant repair button in Command Center doesn't seem to work
### Brief Description *

It seems that this option doesn't actually show said button (I think it used to work in earlier versions). Or maybe the button is placed in a location where some unusual layouts don't show it, e.g. I'm using a 1280x1024 and 1024x1280 screen, which tends to mess up a lot of things in the Command Center display, which can't be repositionned.

Either way it is a trivial issue as the daily auto-repair does a similar thing, but...

### 3. Steps to Reproduce

_No response_

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.11

### Operating System *

win10

### Java Version *

17.0.11

### Final Verification

IllianiBird:
Fix #8735

We used to have an option that would cause an MRMS button to be available in the command center. At some point, I have no idea when, it got deleted. Not sure why, not super important. The weird thing is that the MekHQ client option was left in place, as well as all its hookups. Super weird.

Anyhoo, I think it's important we don't hide GUI elements behind client options most players won't be aware of. Rather than restoring the original option I went ahead and ripped out the option code and now the button will always be visible.

The area where it lives was previously labeled 'procurement information', as that was no longer accurate I went ahead and removed the label. This has the added benefit of allowing us to place non-procurement stuff there, moving forward.

image


Sleet01 (approved):
LGTM

PR #9107, opened 2026-05-26
Added MekHQ Client Option to Allow Players to Set Whether Weekly AutoLogistics Runs Automatically

IllianiBird:
Received player feedback that it was unintuitive that autologistics didn't run weekly, by default. We didn't want to force this change on players, so added a client option. If enabled (default) new campaigns will running weekly autologistics automatically.

For existing campaigns, the Client Option will have no effect. This is noted in the client option tooltip. This was done to avoid overwriting existing player choices.


Sleet01 (approved):
LGTM

PR #9109, opened 2026-05-26
Fixed Error When Replacing Fire Control Systems for Aero Units Such as DropShips

IllianiBird:
I double-checked all other instances of this method (there are around 106 overrides) and this was the only instance I found where we weren't checking for a `null` unit before attempting action.


Sleet01 (approved):
LGTM

PR #9110, opened 2026-05-26
Add Blob Crews to debiting

heckindoc:
The individual payouts for blob crews were being tracked and logged, however the actual costs of blob crews was not being deducted or added to the total which was calculated within the financial tab. Additionally the calculation for payment for blob crews included infantry, regardless of what settings were selected. 

The calculation for blob crews was added to the `getTheoreticalPayroll` method, wish a conditional statement to allow the deduction of infantry from contract calculations. In the previous commit I was doing this in the summary method which would change the calculation for paying infantry a salary at all, and the option in the settings is for including infantry pay in contract calculations.

If the point behind that setting is to not pay your infantry then we should probably consider renaming the option and I will revert this back to the previous commit.


IllianiBird:
Leaving this for Psi to review, as it touches blob crews

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -106,13 +106,29 @@ private Money getTheoreticalPayroll(boolean noInfantry) {
...
+                                        PersonnelRole.VESSEL_CREW };
+
+        if (getCampaignOptions().isInfantryDontCount()) {
+            for (PersonnelRole personnelRole : notInfantry) {
+                salaries.plus(getTempCrewPay(personnelRole, campaign().getTempCrewPool(personnelRole)));
+            }
+            for (PersonnelRole personnelRole : campaign().getTempCrewRoleKeys()) {
+                salaries.plus(getTempCrewPay(personnelRole, campaign().getTempCrewPool(personnelRole)));
+            }
+        }
```
psikomonkie: I think this looks very similar to `Accountant.sumTempCrewPay()`. 

Could `sumTempCrewPay()` be overloaded like `sumTempCrewPay(boolean isInfantryDontCount)`(with the old `sumTempCrewPay()` now just calling `sumTempCrewPay(false)`)?

Then we can use `sumTempCrewPay(getCampaignOptions().isInfantryDontCount())` in this method.

That would let us have one method that is responsible for summing and returning temp crew pays.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -106,13 +106,29 @@ private Money getTheoreticalPayroll(boolean noInfantry) {
...
+        salaries.plus(getTempCrewPay(PersonnelRole.ASTECH, campaign().getTemporaryAsTechPool()));
+        salaries.plus(getTempCrewPay(PersonnelRole.MEDIC, campaign.getTemporaryMedicPool()));
+
+        PersonnelRole[] notInfantry = { PersonnelRole.BATTLE_ARMOUR,
+                                        PersonnelRole.VEHICLE_CREW_GROUND,
+                                        PersonnelRole.VEHICLE_CREW_VTOL,
+                                        PersonnelRole.VEHICLE_CREW_NAVAL,
+                                        PersonnelRole.VESSEL_PILOT,
+                                        PersonnelRole.VESSEL_GUNNER,
+                                        PersonnelRole.VESSEL_CREW };
```
psikomonkie: If we add another `PersonnelRole` in the future as a type of temp personnel, should we need to seek out and update this, or can we assume that anything new is **not** infantry? 

If we want to assume any new `PersonnelRole` is some kind of `Infantry` that should be excluded, then this should stay as is. If we want to assume any new `PersonnelRole` is NOT some kind of `Infantry`, we should switch this to a list of `PersonnelRole[] isInfantry`.

MekHQ/src/mekhq/campaign/finances/FinancialReport.java
psikomonkie, line 1: Were there any changes to this other than renaming `r` to `financialReport`? I appreciate making `r` more readable, it's just a little distracting mixed in with a normal PR. If you ever come across one of these that you want to refactor, especially if it's part of a larger PR you're doing, consider making a separate branch + pr just for the rename so it can get review and merged immediately, and then it won't show on the diff mixed in with actual changes. 

Please don't remove this from this PR, I just want to mention this for next time you're improving the legibility to ensure it's merged as smoothly and quickly as possible. I don't always do this myself but I still feel better for the reviewer when I can separate it from code changes.

psikomonkie (changes requested):
Hey Doc, thanks for the PR!

One change I need, one change I'd like, and one comment for next time.

Thanks for fixing the silly formatting error I left in `Accountant`, and for refactoring `FinancialReport`'s `r` to `financialReport`!

MekHQ/src/mekhq/campaign/finances/FinancialReport.java
heckindoc: No there weren't. I normally do these kind of refactorings in separate PRs and I don't know why I included it here. Sorry about that, and thank you for the recommendation!

MekHQ/src/mekhq/campaign/finances/Accountant.java
heckindoc: I actually tried that originally. This was all in the `sumTempCrewPay()` method, but I ran into a problem that I think is more indicative of a larger problem that's outside the scope of this PR. 

When everything is in `sumTempCrewPay()` and is called by `getTheoreticalPayroll()`, the pay displayed when you change the "Infantry Not Counted Toward Contract Pay" option changes as well -- which is incorrect. Based on how that option is worded this should have no affect on what you're actually paying your infantry. Below are photo examples:

Before any changes were made:
Screenshot 2026-05-26 154348

After changes with option enabled:
image

After changes with option disabled:
Screenshot 2026-05-27 140035


When everything was moved up to `getTheoreticalPayroll()` we no longer saw any of those changes in the financial report, but we did see appropriate appearing changes noted to the contract. 

I opted not to have a boolean passed into the method as we were already calling CampaignOptions at a class level, so instead of having to pass into a boolean we are just pulling whether that option is enabled directly with invoking `campaignOptions.isInfantryDontCount()`. But if it would be better to have the boolean passed in, I'm happy to do that

I did notice, looking through everything for this reply, that I had neglected to get rid of the addition from the summed amount from `getTheoreticalPayroll()` and so I was actually charge the player double for their temp crews. I'll fix this in the next commit

MekHQ/src/mekhq/campaign/finances/Accountant.java
heckindoc: I'm happy to do it whichever way you would prefer. The _right_ way to do it would probably to make this a method within the `HumanResources.java` class that way it can be pulled from there, and if we use it anywhere else it only has to be modified in one place.

psikomonkie: The way I'd prefer? In general, prefer to be "generous" - I would rather be too lenient than too strict. 

So in this case, I think explicitly naming the "Infantry" personnel roles is the lenient option, as any new temp personnel crew roles will Just Work. We only need to update the code here if it's supposed to count as "Infantry".

If we explicitly name the "not-infantry" roles we're being "strict" and having very narrowly scoped reach. This has any newly added temp personnel role treated as if it's "Infantry" until the code here is updated.

In addition, I think it is more likely that a newly added temp personnel role would be treated like non-infantry than like infantry.

MekHQ/src/mekhq/campaign/finances/Accountant.java
psikomonkie: The key thing I need is that I do not want the same logic being implemented in two different locations. Any solution that has a common method being used for summing temp crew pay in `getTheoreticalPayroll` and `getPayRollSummary`/`sumTempCrewPay` will make me very happy. 


> I opted not to have a boolean passed into the method as we were already calling CampaignOptions at a class level, so instead of having to pass into a boolean we are just pulling whether that option is enabled directly with invoking campaignOptions.isInfantryDontCount(). But if it would be better to have the boolean passed in, I'm happy to do that

I think we do need to pass a Boolean. The issue isn't necessarily being able to get to `CampaignOptions` but if we should even care about that option. At an isolated level, when summing temp personnel pay, does that option alone indicate that we should or should not include Infantry? I don't think it does, because when that option is enabled we still need to pay our temp infantry. 

So I believe at some point we need a `sumTempCrewPay` that takes a Boolean. _You do have a choice on what that Boolean is named_. Is it `sumTempCrewPay(boolean excludeInfantry)` or `sumTempCrewPay(Boolean forContractValue)` or whatever. I think either of those would work.  

But in any case I would rather we have one method responsible for summing all temp personnel pays instead of having bespoke implementations like it currently is. 

> I did notice, looking through everything for this reply, that I had neglected to get rid of the addition from the summed amount from getTheoreticalPayroll() and so I was actually charge the player double for their temp crews. I'll fix this in the next commit

Unit tests!! Proper unit test coverage would've caught this. `AccountantTest` already exists but it only covers food and housing. 

I don't want to delay this PR longer with getting the tests written, but I think writing them would be a good exercise. Let's see if we can get these changes made and get this merged, then let's connect on Discord and we can work on getting tests set up together on a call or something.

MekHQ/src/mekhq/campaign/finances/Accountant.java
heckindoc: Not a problem, I'll make this happen after work today!

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -106,13 +106,8 @@ private Money getTheoreticalPayroll(boolean noInfantry) {
...
 
-        // And pay our pool
-        salaries = salaries.plus(campaign().getCampaignOptions()
-                                       .getRoleBaseSalaries()[PersonnelRole.ASTECH.ordinal()].getAmount()
-                                       .doubleValue() * campaign().getTemporaryAsTechPool());
-        salaries = salaries.plus(campaign().getCampaignOptions()
-                                       .getRoleBaseSalaries()[PersonnelRole.MEDIC.ordinal()].getAmount().doubleValue() *
-                                       campaign().getTemporaryMedicPool());
+        // Add all temporary personnel (medics, astechs, temp crew)
+        salaries = salaries.plus(sumTempCrewPay());
```
psikomonkie: When summing the temp crew pay at line 110, should we be considering the `boolean noInfantry` parameter? Should 110 be:
```
        // Add relevant temporary personnel (medics, astechs, temp crew)
        salaries = salaries.plus(sumTempCrewPay(noInfantry));
```
instead of
```
        // Add all temporary personnel (medics, astechs, temp crew)
        salaries = salaries.plus(sumTempCrewPay());
```
?

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -626,24 +621,30 @@ public Map<Person, Money> getPayRollSummary() {
             payRollSummary.put(person, person.getSalary(campaign()));
         }
         // And pay our pool
-        payRollSummary.put(null, Money.of(sumTempCrewPay()));
```
psikomonkie, line 629: (EDIT: This was supposed to include your change not just your removal, sorry)

`getPayRollSummary()` is the method used when collecting everyone's salaries for new day payroll processing. Should we be **paying** our temp soldiers when `getCampaignOptions().isInfantryDontCount())` is true?
Should this be
```
  payRollSummary.put(null, Money.of(sumTempCrewPay()));
```
instead of 
```
  payRollSummary.put(null, Money.of(sumTempCrewPay(getCampaignOptions().isInfantryDontCount())));
```
?

psikomonkie (changes requested):
This is so close! This is looking great so far.

I think `isInfantryDontCount` functionality got swapped between "we don't consider infantry for contract pay" and "we don't pay temp infantry".

MekHQ/src/mekhq/campaign/finances/Accountant.java
heckindoc: yup, that was the intention and then I just forgot to add it. My bad!

MekHQ/src/mekhq/campaign/finances/Accountant.java
heckindoc: done!

psikomonkie (approved):
Fantastic, amazing, thank you so much!

PR #9118, opened 2026-05-27
Close #9113: Updated Automatic Bot Behavior Configuration to More Dynamically Assign Pirate, Convoy/VIP Flags

### LINKED ISSUE #9113: [RFE] Automatically Apply Convoy Bot Behavior as needed
### Enhancement Type *

A "Convoy" behavior flag exists that causes the bot to disregard outgoing damage potential while planning movement. This is useful for ensuring that the flimsy civilian units you are trying to escort don't charge at the enemy so they can use their single machine gun (getting themselves killed in the process). However, this flag is not actually applied to such units in MekHQ escort missions unless the player does so manually. It would be useful if this was applied automatically for all scenarios where such behavior would be appropriate. 

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.12

### Final Verification

IllianiBird:
Close #9113

Previously, if the OpFor faction is marked as a pirate, the 'isPirate' flag is turned on for _all_ Bot Forces. This kinda makes sense, if you're fighting pirates you're probably not taking prisoners. However, we can handle this with a bit more nuance.

Now, only the specific bot force will be marked as pirate if their faction is a pirate. This means that non-pirate factions will act 'honorably' until such a time that the opposing faction acts dishonorably. So no Lyrans shooting dismounted pilots by default if the OpFor is pirate.

Furthermore, we now correctly set the 'ignore outgoing damage' flag for convoy forces. So your convoy drivers won't suddenly decide that Mad Max was a life aspiration and try killing enemy assault meks with their hire hydrants or whatever.


AlexOldenburger:
Would it be useful to make the VIP opponents in the scenarios where they appear more cautious? I think applying the 'ignore outgoing damage' flag would do that? Right now, they have a tendency to charge headlong at the player forces, making their job much easier.

Caveat: I do not know what kind of behaviour this may lead to. Will they go hide in a corner or go flee off the field? The first would be good, the second not at all.

IllianiBird:
I think that idea might have merit, lemme dig through the code one more time. My concern is that we hardcode "you have no guns" values into the AI so it's possible it might not shoot at all. I don't _think_ we'd do that, but I also wasn't the developer behind this function and she's not with the project any more so can't ask.

IllianiBird:
Confirming that all `ignoreDamageOutput` is set outbound damage potential to `0` for path ranking (movement phase) only, so should work well for VIPs

IllianiBird:
Done, great suggestion @AlexOldenburger

Sleet01 (approved):
LGTM

PR #9123, opened 2026-05-27
Revamp the Top Bar in the CampaignGUI

hokvel:
- Improved the top bar's adaptability to different window sizes
- Enforced a minimum window size for the `CampaignGUI`
- The "Company Generator" button is now dynamically hidden and will only appear when both the hangar and personnel lists are empty
- Modularized the top bar by introducing the `AdvanceTimePanel` (displaying the current date and Advance Day buttons) and the `CommandSummaryPanel` (displaying funds, unit rating, and combat force)
- Moved the "Change Faction" button under the player faction icon
- Moved the "Purchase Parts" button from the top bar to the Procurement panel within the CommandCenterTab, as well as to the Warehouse tab
- Moved the "Mass Personnel Training" button to the Personnel Tab
- Moved the "Allow Overtime" toggle down into the status bar
- Simplified UI initialization and update logic within `CampaignGUI`
- Introduced shared layout gap size constants in `CampaignGUI`

Important: this PR uses two icons that will be placed in the data project, but it can be safely merged without them because missing icon case is gracefully handles.

Before:
Before

After:
After

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -42,7 +42,6 @@ import static mekhq.campaign.universe.Faction.MERCENARY_FACTION_CODE;
 import static mekhq.gui.dialog.nagDialogs.NagController.triggerDailyNags;
 import static mekhq.gui.enums.MHQTabType.COMMAND_CENTER;
 import static mekhq.utilities.MHQInternationalization.getFormattedTextAt;
-import static mekhq.utilities.MHQInternationalization.getText;
 
 import java.awt.*;
 import java.awt.event.ActionEvent;
@@ -99,7 +98,6 @@ import mekhq.campaign.campaignOptions.AcquisitionsType;
 import mekhq.campaign.campaignOptions.CampaignOptions;
 import mekhq.campaign.enums.DailyReportType;
 import mekhq.campaign.events.*;
-import mekhq.campaign.events.assets.AssetEvent;
 import mekhq.campaign.events.loans.LoanEvent;
 import mekhq.campaign.events.missions.MissionEvent;
 import mekhq.campaign.events.persons.PersonEvent;
@@ -137,8 +135,8 @@ import mekhq.campaign.unit.Unit;
 import mekhq.campaign.universe.Faction;
 import mekhq.campaign.universe.NewsItem;
 import mekhq.campaign.universe.Systems;
-import mekhq.campaign.universe.factionStanding.GoingRogue;
 import mekhq.campaign.utilities.AutomatedTechAssignments;
+import mekhq.gui.baseComponents.HorizontallyConstrainedPanel;
 import mekhq.gui.baseComponents.immersiveDialogs.ImmersiveDialogSimple;
 import mekhq.gui.baseComponents.roundedComponents.RoundedJButton;
 import mekhq.gui.baseComponents.roundedComponents.RoundedLineBorder;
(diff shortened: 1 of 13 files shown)


MekHQ/src/mekhq/campaign/finances/Finances.java
```diff
@@ -212,7 +212,7 @@ public Money getLoanBalance() {
         return balance.plus(loans.stream().map(Loan::determineRemainingValue).collect(Collectors.toList()));
     }
 
-    public boolean isInDebt() {
+    public boolean hasLoan() {
```
IllianiBird: I would have preferred this in a separate, follow-up PR, as it doesn't directly relate to the task of revamping the top bar. Means this PR hits a lot more classes than it otherwise would.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -3293,7 +3135,11 @@ private void refreshGMMenuItems() {
         }
     }
 
-    private final ActionScheduler fundsScheduler = new ActionScheduler(this::refreshFunds);
+    private void refreshCampaignControlButtons() {
+        boolean emptyHangar = getCampaign().getHangar().getUnits().isEmpty();
```
IllianiBird: You can use `getCampaign().getUnits().isEmpty()` here. Functionally the same, but reads a little cleaner.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -3593,8 +3437,8 @@ private boolean checkForInvalidFaction(DayEndingEvent dayEndingEvent) {
      */
     @Subscribe
     public void handleNewDay(NewDayEvent newDayEvent) {
-        refreshCalendar();
-        refreshFunds();
+        refreshWindowTitle();
```
IllianiBird, line 3440: This is a positive rename

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -3684,25 +3526,9 @@ public void handle(TransactionEvent transactionEvent) {
...
-     * <p><b>Important:</b> This method is not directly evoked, so IDEA will tell you it has no uses. IDEA is
-     * wrong.</p>
-     *
-     * @param assetEvent the event indicating a change in assets
-     */
-    @Subscribe
-    public void handle(AssetEvent assetEvent) {
-        fundsScheduler.schedule();
-    }
-
```
IllianiBird, line 3705: I'm struggling to tell why this subscriber was removed, could you clarify for me? :)

IllianiBird (commented):
For the most part, the code checks out. I do think this PR _is_ trying to do too much, which makes it difficult to review.

In future, it'd probably be better off making a bunch of smaller PRs. MekHQ PRs get reviewed and merged pretty quickly, so it wouldn't delay for long. But it would make it easier to get an 'at glance' idea of where things have been moved to and what's been functionally changed instead of just renamed.

Case in point, as an example, it took me a good moment to figure out that the calendar refresh method had just been renamed.

MekHQ/src/mekhq/gui/CampaignGUI.java
hokvel: It was needed to update the funds label. The label was moved to CommandSummaryPanel, which now subscribes to this event instead.

hokvel:
Pushing to retrigger the build, seems like it failed due to an issue with CI infra.

MekHQ/src/mekhq/gui/CampaignGUI.java
IllianiBird: Perfect, I kinda gathered it had been moved, but I wasn't able to spot where to. Not going to lie this was a tough PR to review

IllianiBird (approved):

PR #9124, opened 2026-05-28
Improved How We Evaluate Personnel for Automated Personnel Clean Up

IllianiBird:
While working on a bug I noticed some oversights with the automated personnel clean up that might result in some personnel being removed that shouldn't have been. It is important to note that this wouldn't have 'cleaned up' characters still active in the campaign, only characters that had already left and **weren't** dead or retired.

The limitation of the earlier method wasn't that it didn't work. It actually did. The issue is that Genealogy was changed at some point during its development, many years ago, and despite being set up to hold a characters' entire genealogical data it actually only holds parents, children, spouses, and ex-spouses. And the method `getFamily` only returns this limited pool of people.

The intent was that would wouldn't remove anyone with an active genealogy, no matter how distant that relationship might be. To ensure a characters' legacy is kept intact.

With this PR, that intention is realized. I also added two new tests as I wanted to confirm how Background Characters were being handled.


MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
```diff
@@ -456,26 +459,147 @@ public void clearGenealogyLinks() {
...
-        for (List<Person> relationshipGroup : family.values()) {
-            for (Person relation : relationshipGroup) {
-                if (!relation.getStatus().isDepartedUnit()) {
+    public boolean isActive(HumanResources humanResources) {
+        HashSet<Person> allFamilyMembers = new HashSet<>();
+        return collectRelatives(humanResources, origin, allFamilyMembers, true);
+    }
+
+    /**
+     * Recursively traverses the genealogical graph from {@code currentPerson}, collecting all reachable relatives and
```
IllianiBird, line 484: I was nervous about using recursion, as it can be expensive, but benchmarking this method showed that (due to the early exit clause) it's not particularly expensive. As it's only run once a month I don't anticipate any performance issues, even for large campaigns.

MekHQ/unittests/mekhq/campaign/utilities/AutomatedPersonnelCleanUpTest.java
```diff
@@ -33,50 +33,41 @@
...
+    private Campaign testCampaign;
+    private final LocalDate today = LocalDate.of(3151, 1, 1);
 
     @BeforeEach
     void setUp() {
-        mockCampaign = mock(Campaign.class);
-        when(mockCampaign.getFaction()).thenReturn(Factions.getInstance().getDefaultFaction());
-
-        today = LocalDate.of(3151, 1, 1);
+        testCampaign = MHQTestUtilities.getTestCampaign();
```
IllianiBird, line 55: Figured I might as well take the opportunity to move this test to use a real campaign object

MekHQ/unittests/mekhq/campaign/utilities/AutomatedPersonnelCleanUpTest.java
```diff
@@ -33,50 +33,41 @@
...
         // Setup
-        Collection<Person> personnelForConsideration = new ArrayList<>();
+        HumanResources humanResources = testCampaign.getHumanResources();
         for (int i = 0; i < 10; i++) {
-            Person ineligiblePerson = new Person(mockCampaign);
+            Person ineligiblePerson = new Person(testCampaign);
             ineligiblePerson.setStatus(PersonnelStatus.ACTIVE);
 
-            personnelForConsideration.add(ineligiblePerson);
+            humanResources.importPerson(ineligiblePerson);
```
IllianiBird, line 66: `importPerson` adds them to the 'personnel' array, but without going through the heap of additional steps normally required when recruiting a character, making this method fantastic for unit testing.

MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
Sleet01: Recursion is no better or worse than other methods... so long as one remembers their termination cases :)

MekHQ/src/mekhq/campaign/utilities/AutomatedPersonnelCleanUp.java
```diff
@@ -62,14 +64,16 @@ public class AutomatedPersonnelCleanUp {
      * Constructs an automated personnel cleanup process.
      *
      * @param today                      the current date used as a reference for evaluating time-based conditions
-     * @param personnelToProcess         the list of personnel to be examined for potential cleanup
      * @param isUseRemovalExemptRetirees whether retired personnel should be exempt from removal
      * @param isUseRemovalExemptCemetery whether deceased personnel should be exempt from removal
+     * @param humanResources             the container for all personnel, used to check whether a character is
+     *                                   genealogically
```
Sleet01, line 70: We may get style complaints about the parameter order not matching the doc string, so this should probably get moved  up (and the `today` parameter removed from the `@param` list.

Sleet01 (approved):
There's a method with some out-of-date / out-of-order `@param` entries in the javadoc block, but functionally it looks fine.

PR #9128, opened 2026-05-28
Implement dynamic GUI scaling for the Campaign Window and Top Panel

hokvel:
Includes the following changes:
- Add GUI scaling support in `HorizontallyConstrainedPanel` and rename it to `ScalingWidthConstrainedPanel`
- Add GUI scaling support in `VerticalFillImage` and rename it to `ScalingVerticalFillImage`
- Update the `CurrentLocationPanel` to accept `iconMaxHeight` instead of using a hardcoded value
- Introduce `refreshGuiScale()` method in `CampaignGUI` that applies GUI scaling to the Campaign Window and Top Bar
- Improve the documentation

I recommend opening the following screenshots in a separate tab and setting 100% scale before comparing.

Before (70% scale):
Before_70

After (70% scale):
After_70

Before (160% scale):
Before_160

After (160% scale):
After_160

### DIFF
--- a/MekHQ/src/mekhq/gui/CampaignGUI.java
+++ b/MekHQ/src/mekhq/gui/CampaignGUI.java
@@ -136,7 +136,7 @@ import mekhq.campaign.universe.Faction;
 import mekhq.campaign.universe.NewsItem;
 import mekhq.campaign.universe.Systems;
 import mekhq.campaign.utilities.AutomatedTechAssignments;
-import mekhq.gui.baseComponents.HorizontallyConstrainedPanel;
+import mekhq.gui.baseComponents.ScalingWidthConstrainedPanel;
 import mekhq.gui.baseComponents.immersiveDialogs.ImmersiveDialogSimple;
 import mekhq.gui.baseComponents.roundedComponents.RoundedJButton;
 import mekhq.gui.baseComponents.roundedComponents.RoundedLineBorder;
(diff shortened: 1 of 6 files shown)


MekHQ/src/mekhq/gui/baseComponents/ScalingVerticalFillImage.java
```diff
@@ -165,7 +168,12 @@ public Dimension getMinimumSize() {
 
     @Override
     public Dimension getMaximumSize() {
-        return new Dimension((int) (maxHeight * aspectRatio), maxHeight);
+        int h = getScaledMaxHeight();
```
IllianiBird: I'd prefer if we didn't use abbreviated or single letter names. I'm aware here it's a tiny instance, but we used to have a massive problem with this and it took months to slowly unravel.

MekHQ/src/mekhq/gui/baseComponents/ScalingVerticalFillImage.java
```diff
@@ -165,7 +168,12 @@ public Dimension getMinimumSize() {
...
+        return new Dimension((int) (h * aspectRatio), h);
+    }
+
+    private int getScaledMaxHeight() {
+        return Math.max(MIN_HEIGHT, UIUtil.scaleForGUI(maxHeight));
```
IllianiBird, line 176: Good safely check

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1579,6 +1578,24 @@ private void refreshThemeChoices() {
...
+        pnlTop.setMaximumSize(new Dimension(Integer.MAX_VALUE, UIUtil.scaleForGUI(TOP_PANEL_HEIGHT)));
+    }
+
+    private static Dimension getUsableScreenSize() {
+        GraphicsDevice gd = GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice();
+        Dimension screenSize = Toolkit.getDefaultToolkit().getScreenSize();
+        Insets screenInsets = Toolkit.getDefaultToolkit().getScreenInsets(gd.getDefaultConfiguration());
+        int usableScreenWidth = screenSize.width - screenInsets.left - screenInsets.right;
+        int usableScreenHeight = screenSize.height - screenInsets.top - screenInsets.bottom;
+        return new Dimension(usableScreenWidth, usableScreenHeight);
```
IllianiBird, line 1596: To confirm, there is no risk of a 0 or negative value here? I don't believe there is, but I need your confirmation.

IllianiBird (changes requested):
A small needed change and a quick question, otherwise good.

MekHQ/src/mekhq/gui/baseComponents/ScalingVerticalFillImage.java
hokvel: I don't do this outside of small local scopes, as it can totally result in maintainability issues. Will update this and one more instance in the same class.

IllianiBird: Yeah, it's one thing that I'm particularly hard line on. I appreciate that I'm probably _too_ hard line on it, but just outright avoiding them is better than needing to make a judgment call every time.

MekHQ/src/mekhq/gui/CampaignGUI.java
hokvel: Yes, these insets correspond to OS toolbars and should not overflow.

IllianiBird (approved):

PR #9135, opened 2026-05-28
Extract menu logic from CampaignGUI into MekHQMenuBar

hokvel:
Currently, menu-related logic accounts for approximately 48% of the CampaignGUI class. Extracting this code is the first step toward the future goal of creating a clearer separation between the UI and control layers, as well as addressing existing issues with localization.

This PR extracts all menu-related logic into a dedicated MekHQMenuBar class. To minimize the risk of regressions given the large footprint of this change, every method has been moved exactly as it is. The only modifications made were to rewire access to the MekHQ, Campaign, and CampaignGUI objects using dedicated getters.

### DIFF
--- a/MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
+++ b/MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
@@ -1,0 +1,1827 @@
+/*
+ * Copyright (c) 2009 Jay Lawson (jaylawson39 at yahoo.com). All rights reserved.
+ * Copyright (C) 2013-2026 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MekHQ.
+ *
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -0,0 +1,1826 @@
+/*
+ * Copyright (C) 2026 The MegaMek Team. All Rights Reserved.
```
IllianiBird: This should be

```
 * Copyright (c) 2009 Jay Lawson (jaylawson39 at yahoo.com). All rights reserved.
 * Copyright (C) 2013-2026 The MegaMek Team. All Rights Reserved.
```

Migrating code doesn't remove the original copyrights

IllianiBird (changes requested):

IllianiBird (approved):

PR #9139, opened 2026-05-29
Added Tests for Payroll Summary to AccountantTest

heckindoc:
As noted in title, several tests were added to test `GetPayRollSummary()` within `AccountantTest`


IllianiBird:
Leaving this one for @psikomonkie to review, as it was them that requested the additional test cases

psikomonkie (approved):
> Leaving this one for @psikomonkie to review, as it was them that requested the additional test cases

TBH I probably shouldn't review this as we wrote these on a call together! But it's just unit tests, and I do think they're good :)

PR #9143, opened 2026-05-29
Close #9142: Rolled Back Circa 2025 Change to Contract Clause Generation, Contract Clauses Will Now Become Worse (When Re-Negotiated) Less Often

### LINKED ISSUE #9142: Legendary Negotiaters always making things worse.
“Illiani told me to make this RFE”

So, when negotiating contracts, all my guys are 1+ or 2+. Yet, on contracts, they make things worse just as much if not more than they make it better. I Suspect a coding issue is causing this.

I can't see the rolls, so I can't speculate, but perhaps their skills aren't being read right?

IllianiBird:
Close #9142

In modern MekHQ we still use AtB's old contract generation. Back in 2025 we made several changes to how clauses were determined (especially for re-negotiations).

This had the unfortunate consequence of making clause re-negotiations significantly harder than was intended.

This PR reverses those changes back to their AtB original state. Additional steps were taken to clean up the code, as the original code was a little crusty.

## Addendum 1
Only being able to renegotiate each clause once is still in effect. No plans to change that exist at this time.

## Addendum 2
I had wanted to convert these negotiation checks to use the SkillCheckUtility but that isn't possible. Negotiation checks aren't true skill checks, instead the characters experience milestone (green, regular, etc) applies a modifier to a 2d6 roll.

### DIFF
--- a/MekHQ/src/mekhq/campaign/market/contractMarket/AtbMonthlyContractMarket.java
+++ b/MekHQ/src/mekhq/campaign/market/contractMarket/AtbMonthlyContractMarket.java
@@ -35,10 +35,8 @@ package mekhq.campaign.market.contractMarket;
 
 import static java.lang.Math.max;
 import static megamek.common.compute.Compute.d6;
-import static megamek.common.enums.SkillLevel.ELITE;
 import static megamek.common.enums.SkillLevel.GREEN;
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #9144, opened 2026-05-29
Updated StratCon Scenario Generation So the Clans Will Stop Cheating (Yet Again)

IllianiBird:
Added a campaign option, if enabled (enabled by default) and the player has agreed to the terms of a Batchall, Clan forces will no longer reinforce with additional units from facility or random scenario modifiers.

If the player breaks the terms of the Batchall all bets are off.


Sleet01 (approved):
LGTM

PR #9145, opened 2026-05-29
Added New Aggregate Daily Report Tab Client Option That Will Add a New Tab That Displays All Reports

IllianiBird:
This PR adds a new daily report tab, the 'aggregate' tab. This requires a Client Option to enable. Due to how the daily reports work it is not possible to hide the tab if the option is disabled, it just won't be populated with anything.

If the option is enabled, all daily reports will appear in two places: their individual tabs, _and_ the aggregate tab.

This will increase load, save times. It will also increase how long it takes to advance to a new day. Smaller campaigns won't likely notice any different unless they have a ton of things happening on a single day. For larger campaigns this will likely result in a noticeable decline in performance.

The option includes a warning to this effect.

## Addendum
This is a feature that has been requested ever since we implemented the tabbed daily reports. Which themselves were persistently requested. However, I couldn't find an RFE so none is attached.


Sleet01 (approved):
LGTM

Issue #1979, opened 2020-09-02
Unload, partial load, and GM manage ammunition counts


SuperStucco:
Please add additional options in the MekHQ Repair Bay tab for ammunition bins:

- empty a bin back into the 'warehouse', the same tech action as 'load ammunition' but in reverse
- do a partial load for a specific bin, the same tech action as 'load ammunition'
- add ammunition bin counts to the GM "Edit damage..." dialog so counts can be manually adjusted

PR #9146, opened 2026-05-29
Add a menu action to unload ammo

AlexOldenburger:
Currently, to remove ammo from a unit you must switch the unit to Strip and remove the entire ammo bin.

This PR adds a new option to the Swap Ammo menu in the Hangar tab to unload ammo from a specific bin on a unit. The ammo will then go to the warehouse. Since the unit is now missing ammo, it will show up in the Repair Bay as well.

Making the unloading of ammo easier is just something to make it easier to do what is essentially a very simple task.

The action does not need a technician, and it also takes no time. This makes it something alike applying a new paint scheme on a unit. The reasoning behind this is that it is a very simple task where no special skill is needed (so it could be done by AsTechs).

Another (more technical) reason is that the curent repair system is not really geared towards removing something from a unit, so it would be difficult to add a new repair action for this.

This implements the first suggestion from issue #1979


MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -960,7 +961,7 @@ protected Optional<JPopupMenu> createPopupMenu() {
             // swap ammo
             if (oneSelected) {
                 if (unit.getEntity().usesWeaponBays()) {
-                    menuItem = new JMenuItem("Swap ammo...");
+                    menuItem = new JMenuItem(resources.getString("swapAmmo.text") + "...");
```
IllianiBird: This should be `getText("swapAmmo.text")`, we're moving away from `[RESOURCE MAP].getString()`. If you search for `RESOURCE_BUNDLE` (all caps) in MekHQ you'll see loads of examples of the new system

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -973,15 +974,21 @@ protected Optional<JPopupMenu> createPopupMenu() {
                               .stream()
                               .anyMatch(m -> (m.getType() instanceof InfantryWeapon) &&
                                                    ((InfantryWeapon) m.getType()).hasInfernoAmmo())) {
-                        menuItem = new JMenuItem("Swap ammo...");
+                        menuItem = new JMenuItem(resources.getString("swapAmmo.text") + "...");
```
IllianiBird: I'm not going to mark every instance, just assume I commented about the resource bundle thing for all.

No need to change existing resource bundle calls, unless you want to or change them

IllianiBird (changes requested):
I absolutely would have loved to have this earlier in the week. 10/10

I do have some feedback, but it's minor stuff.

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
AlexOldenburger: I've updated all the instances I could find in this file from resources.getString to getText

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
AlexOldenburger: I've updated all the instances I could find in this file from resources.getString to getText

IllianiBird (approved):

PR #9152, opened 2026-05-30
Display Edge in Salvage Tech Picker Dialogue

AlexOldenburger:
With risky salvage operations, when assigning techs as salvage supervisors it is important to know if a tech has any edge left, as this would enable a reroll if anything were to go wrong.

This PR addresses this by adding a new column showing the current edge of all techs, allowing the player to make a better informed decision on who to select.

The new column is not shown if either edge or edge for support personnel is not enabled in the campaign.

Using edge during risky salvage is now also dependent on both edge and edge for support personnel being enabled. Previously, only edge needed to be enabled.


psikomonkie (approved):
Nice!

PR #9182, opened 2026-06-01
Fixed Background Marriages Occurring Within the Company Generator Failing to Add Spouses to Campaign Under Certain Circumstances Resulting in the Marriage Being Erased On Campaign Load

IllianiBird:
This bug came about due to use of the `isBackground` flag. That flag prevents characters from being added to a campaign, generally due to the addition being from the Simulate Relationship Histories campaign option. However, this flag had been incorrectly set to `true` for the company generator.

If, during the company generator's processes, a character got married. And the character they married was external to the unit (an NPC). Then (and only then) that character would **not** to added to the campaign roster, resulting in the marriage being removed the next time the campaign loaded (to protect the user from NPEs).

### DIFF
--- a/MekHQ/src/mekhq/campaign/universe/generators/companyGenerators/AbstractCompanyGenerator.java
+++ b/MekHQ/src/mekhq/campaign/universe/generators/companyGenerators/AbstractCompanyGenerator.java
@@ -704,11 +704,7 @@ public abstract class AbstractCompanyGenerator {
 
                 for (final CompanyGenerationPersonTracker tracker : trackers) {
                     if (getOptions().isSimulateRandomMarriages()) {
-                        // While this is happening to generate a character's background, it does not use the
-                        // simulated relationship history system. Therefore, 'isBackground' needs to == false,
-                        // otherwise any marriages will be voided the next time the campaign is loaded as the spouse
-                        // will not be added to the campaign
-                        campaign.getMarriage().processNewWeek(campaign, date, tracker.getPerson(), false);
+                        campaign.getMarriage().processNewWeek(campaign, date, tracker.getPerson());
                     }
 
                     if (getOptions().isSimulateRandomProcreation()) {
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/campaign/universe/generators/companyGenerators/AbstractCompanyGenerator.java
```diff
@@ -704,7 +704,11 @@ private void finalizePersonnel(final Campaign campaign,
 
                 for (final CompanyGenerationPersonTracker tracker : trackers) {
                     if (getOptions().isSimulateRandomMarriages()) {
-                        campaign.getMarriage().processNewWeek(campaign, date, tracker.getPerson(), true);
+                        // While this is happening to generate a character's background, it does not use the
+                        // simulated relationship history system. Therefore, 'isBackground' needs to == false,
+                        // otherwise any marriages will be voided the next time the campaign is loaded as the spouse
+                        // will not be added to the campaign
+                        campaign.getMarriage().processNewWeek(campaign, date, tracker.getPerson(), false);
```
hokvel: As an observation, now that both `processNewWeek` usages pass isBackground == false, it makes sense to consider removing it from arguments. Then `AbstractMarriage::marryRandomSpouse` will be called from two places: `processNewWeek` (isBackground always being false because the naming implies campaign time advancement) and `processBackgroundMarriageRolls` (always passing true).

hokvel (approved):

MekHQ/src/mekhq/campaign/universe/generators/companyGenerators/AbstractCompanyGenerator.java
IllianiBird: Good shout.

Done

hokvel (approved):

HammerGS (approved):

PR #9183, opened 2026-06-01
Fixed Attribute Cost Campaign Option Not Being Correctly Hooked Up to GUI

IllianiBird:
Unfortunately it looks like, when the ability to change Attribute Costs was added, it was not properly affixed to the UI. Meaning the cost was in place, but there was no way for the players to access it. Now there is.

Marked it as a new-to-51.0 option, as to the player base it effective is.


VicenteCartas (approved):
I'll make sure I merge this to my branch.

PR #9185, opened 2026-06-01
Lowered Accessibility of `CurrentLocation` Text Helper Methods

IllianiBird:
Spotted a couple of help methods in `CurrentLocation` that had been set up as global statics. Methods like this should never be made global as they should not be accessed outside of the relevant class/package.

Reduced accessibility to local static and moved the helper into the class that uses them.


psikomonkie (approved):
Nice!

hokvel:
That was my change. The reasoning is that we want to share the resource bundle between `CurrentLocationPanel` and `AbstractLocation`. Agreed that access scope is too broad. Maybe the middle ground is to copy the methods, but reuse `RESOURCE_BUNDLE` from `AbstractLocation`.

hokvel:
Also, the same public static is likely still present in `AbstractLocation`.

IllianiBird:
Yeah that should probably be package-private

IllianiBird:
Confirming that the static in AbstractLocation is package-private, but won't be accessible to CurrentLocationPanel as it doesn't extend AbstractLocation. The change in this PR matches current practice elsewhere.

PR #9186, opened 2026-06-01
Added Ability to Pay XP to Replenish Edge at Any Time

IllianiBird:
Designed to work alongside an earlier change that reduced the rate Edge is replenished, this PR allows players to pay 20 XP (campaign option) to replenish all spent edge. Perfect if there is a hard repair task or scenario you want to be fully loaded for.

### DIFF
--- a/MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
+++ b/MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
@@ -579,6 +579,7 @@ public class CampaignOptions {
 
     private int attributeCost;
     private int edgeCost;
+    private int edgeRefreshCost;
     // endregion Experience Tab
 
     // region Skills Tab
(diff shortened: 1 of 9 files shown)


MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
```diff
@@ -1250,6 +1251,7 @@ public CampaignOptions() {
         missionXpSuccess = 3;
         missionXpOutstandingSuccess = 5;
         edgeCost = 10;
+        edgeRefreshCost = 20; // Taken from ATOW
```
hokvel: ATOW 3rd printing, p.43. Also it is called "edge recovery" in the rulebook, so it might be a better name to avoid confusion in the future.

IllianiBird: Good call

HammerGS (approved):

PR #9187, opened 2026-06-01
Fixed `getFormattedText` Incorrectly Double Formatting Text Resulting in Display Errors

IllianiBird:
`getFormattedText` calls `getFormattedTextAt`, both parse the text through `MessageFormat.format` which causes anything with an apostrophe to get mangled. Normally you need to double an apostrophe, to escape it, however due to this bug developers would have actually needed to escape apostrophe _twice_ with `'''` appearing as `'`, whereas missing the third apostrophe would cause html to break formatting.

### DIFF
--- a/MekHQ/src/mekhq/utilities/MHQInternationalization.java
+++ b/MekHQ/src/mekhq/utilities/MHQInternationalization.java
@@ -124,7 +124,7 @@ public class MHQInternationalization {
      * @return the localized string
      */
     public static String getFormattedText(String key, Object... args) {
-        return MessageFormat.format(getTextAt(getInstance().defaultBundle, key), args);
+        return getFormattedTextAt(getInstance().defaultBundle, key, args);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/utilities/MHQInternationalization.java
```diff
@@ -124,7 +124,7 @@ public static String getText(String key) {
      * @return the localized string
      */
     public static String getFormattedText(String key, Object... args) {
-        return MessageFormat.format(getFormattedTextAt(getInstance().defaultBundle, key), args);
+        return MessageFormat.format(getTextAt(getInstance().defaultBundle, key), args);
```
psikomonkie: Would using just `return getFormattedTextAt(getInstance().defaultBundle, key, args);` work - removing the extra MessageFormat.format wrapping it, instead of replacing the inner method?

IllianiBird: For sure! :)

 That was my original change but then second-guessed myself. Easily corrected.

NickAragua (approved):

HammerGS (approved):

PR #9190, opened 2026-06-01
Show 'Allow Overtime' button only in the Repair Bay and Warehouse tabs

hokvel:
The button is now present only in places where we show AsTech pool instead of being in the status bar. Wiring that synchronizes button state across the tabs was also added.

Before:
Screenshot 2026-06-01 at 19 03 16

After:
Screenshot 2026-06-01 at 19 04 51
Screenshot 2026-06-01 at 19 05 08

### DIFF
--- a/MekHQ/src/mekhq/gui/WarehouseTab.java
+++ b/MekHQ/src/mekhq/gui/WarehouseTab.java
@@ -401,7 +401,9 @@ public final class WarehouseTab extends CampaignGuiTab implements ITechWorkPanel
         btnOvertime.addActionListener(evt -> {
             getCampaign().setOvertime(btnOvertime.isSelected());
             RepairTab repairBayTab = getCampaignGui().getRepairBayTab();
-            repairBayTab.refreshOvertimeStatus();
+            if (repairBayTab != null) {
+                repairBayTab.refreshOvertimeStatus();
+            }
         });
         gridBagConstraints = new GridBagConstraints();
         gridBagConstraints.gridx = 0;
--- a/MekHQ/src/mekhq/gui/WarehouseTab.java
+++ b/MekHQ/src/mekhq/gui/WarehouseTab.java
@@ -139,6 +139,7 @@ public final class WarehouseTab extends CampaignGuiTab implements ITechWorkPanel
     private RoundedMMToggleButton btnShowAllTechsWarehouse;
     private JLabel lblTargetNumWarehouse;
     private JTextArea textTargetWarehouse;
+    private RoundedMMToggleButton btnOvertime;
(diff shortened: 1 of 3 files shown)


IllianiBird:
It might be better to have the overtime options to be next to the 'show all techs' buttons

hokvel:
> It might be better to have the overtime options to be next to the 'show all techs' buttons

I tried that and it seemed a bit disconnected from the AsTech pool time statistic that it affects, especially with refresh logic in place:

https://github.com/user-attachments/assets/ac8bd21c-1b6e-4fd7-bc16-77aad91edee4

IllianiBird:
Mm, I see your point. Works for me, if we get any feedback we can easily move it.

MekHQ/src/mekhq/gui/WarehouseTab.java
```diff
@@ -394,6 +395,21 @@ public void initTab() {
...
         panelDoTask.add(scrollTechTable, gridBagConstraints);
 
+        btnOvertime = new RoundedMMToggleButton(resourceMap.getString("btnOvertime.text"));
+        btnOvertime.setToolTipText(resourceMap.getString("btnOvertime.toolTipText"));
+        btnOvertime.setSelected(getCampaign().isOvertimeAllowed());
+        btnOvertime.addActionListener(evt -> {
+            getCampaign().setOvertime(btnOvertime.isSelected());
+            RepairTab repairBayTab = getCampaignGui().getRepairBayTab();
+            repairBayTab.refreshOvertimeStatus();
+        });
```
IllianiBird, line 407: `getRepairBayTab` is flagged as Nullable, if `repairBayTab == null` it will trigger an NPE on line 404 (appropriate, no? :D)

IllianiBird (changes requested):
One small change, otherwise happy for this to be merged.

MekHQ/src/mekhq/gui/WarehouseTab.java
hokvel: Good point. Fixed this and the case with nullable warehouse tab too. It makes me want to investigate how to clean that up, because these tabs should always be present, but `getTab` is nullable because the StratCon tab is optional.

IllianiBird (approved):

MekHQ/src/mekhq/gui/WarehouseTab.java
IllianiBird: The root cause is JTabbedPane. It doesn't natively support hiding tabs. Which means our only recourse is to not add the tab, which then leads to NPEs... it's all a bit of a mess, honestly.

PR #9194, opened 2026-06-02
Marks Complete and Delete Mission as GM only buttons in Briefing room

VicenteCartas:
image

Marks Complete Mission and Delete Mission as GM only buttons.


IllianiBird:
I am so sorry, I should have read the feedback closer, these should absolutely **not** be GM only. Else players will never be able to complete contracts or clean up old ones

PR #9197, opened 2026-06-02
Always add free units to the default formation, even without ranks

AlexOldenburger:
When generating a company with the company generator, it is possible to receive free units (for supply, salvage and so on). These free units should be automatically added to default formations in the TOE.

However, if generating the company without automatically assigning ranks to the new personnel, the free units were added to the roster, but they were not added to the default formations in the TOE.

This PR fixes this by always adding the free units to the default formations in the TOE, even when no ranks are generated.


IllianiBird (approved):
Whoopsie doodles

PR #9203, opened 2026-06-02
Update AtTech pool minutes info on overtime mode changes

hokvel:
Changes include:
- Add wiring for 'Allow Overtime' buttons in the Repair Bay and Warehouse tabs and corresponding AsTech pool info labels
- Make `refreshTechsList` private in both tabs
- Extract AsTech pool info update logic into a separate method and reuse it during tab initialization to remove duplicate code

PR in action:

https://github.com/user-attachments/assets/6fafdbb5-f735-41b2-9785-a42577eadee7

### DIFF
--- a/MekHQ/src/mekhq/gui/RepairTab.java
+++ b/MekHQ/src/mekhq/gui/RepairTab.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2017-2026 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MekHQ.
  *
--- a/MekHQ/src/mekhq/gui/RepairTab.java
+++ b/MekHQ/src/mekhq/gui/RepairTab.java
@@ -984,7 +984,7 @@ public final class RepairTab extends CampaignGuiTab implements ITechWorkPanel {
      * Updates the AsTech pool statistics (minutes, overtime availability, and AsTech count) in the UI label.
      */
     public void refreshAsTechPool() {
-        String astechString = "<html><b>AsTech Pool Minutes:</> " + getCampaign().getAsTechPoolMinutes();
+        String astechString = "<html><b>AsTech Pool Minutes:</b> " + getCampaign().getAsTechPoolMinutes();
         if (getCampaign().isOvertimeAllowed()) {
             astechString += " [" + getCampaign().getAsTechPoolOvertime() + " overtime]";
         }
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/RepairTab.java
```diff
@@ -986,6 +980,18 @@ public void refreshTechsList() {
         }
     }
 
+    /**
+     * Updates the AsTech pool statistics (minutes, overtime availability, and AsTech count) in the UI label.
+     */
+    public void refreshAsTechPool() {
+        String astechString = "<html><b>AsTech Pool Minutes:</> " + getCampaign().getAsTechPoolMinutes();
```
IllianiBird: These should be in a Resource Bundle using getTextAt pr getFormattedTextAt. We shouldn't be adding new hardcoded Strings.

MekHQ/src/mekhq/gui/WarehouseTab.java
```diff
@@ -681,6 +675,18 @@ public void refreshTechsList() {
         }
     }
 
+    /**
+     * Updates the AsTech pool statistics (minutes, overtime availability, and AsTech count) in the UI label.
+     */
+    public void refreshAsTechPool() {
+        String astechString = "<html><b>AsTech Pool Minutes:</> " + getCampaign().getAsTechPoolMinutes();
```
IllianiBird: And here. Potentially we could combine the two refreshAstechPool methods, as they seem pretty similar

IllianiBird (changes requested):
A couple minor changes needed

MekHQ/src/mekhq/gui/RepairTab.java
hokvel: This is old code

MekHQ/src/mekhq/gui/WarehouseTab.java
hokvel: Same here, old code

hokvel: I also want to refactor it, but as a separate change, otherwise it will extend the scope considerably.

IllianiBird: Sounds good. Generally speaking, even for old code, if we're touching the area where the hardcode strings live it's best to break out those strings into a resource bundle using the new getters. Don't need to do the entire class, just the areas you're touching.

IllianiBird (approved):

MekHQ/src/mekhq/gui/RepairTab.java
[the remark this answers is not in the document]
IllianiBird: This looks like a typo.

MekHQ/src/mekhq/gui/WarehouseTab.java
hokvel: Of course, it's just this particular case that I think deserves a separate PR because it will be about removing code duplication and improving localization at the very least. Potentially even more, because there's a lot of repetition around the techs panels in the Warehouse and Repair Bay tabs.

PR #9212, opened 2026-06-02
Unsubscribe CommandGUI from the event bus on campaign creation and loading

hokvel:
A regression was introduced in #9135 because `MekHQ.unregisterHandler(this)` was moved as is and it did not break the build. This change ensures that `CampaignGUI` is correctly unsubscribed.


hokvel:
Additionally, all other cases of `this` usage were checked, this was the only regression of such type.

IllianiBird (approved):

IllianiBird:
Good catch

PR #9218, opened 2026-06-03
Reduce code duplication in MekHQMenu CSV Export logic

hokvel:
- Move file selection logic into `save*CSV` methods in `FileDialogs`
- Introduce `FileDialogs.saveWithBackup` helper that handles output file backups and default directory updates
- Remove the try-catch blocks from the `miExport` item listeners because export methods are already safe
- Update export methods in `MekHQMenu` to use file selectors from `FileDialogs`, add early returns to prevent actions when tables are empty, and simplify their signatures
- Rename `checkFileEnding` to `enforceFileExtension`, make it static, make it accept `FileType` instead of a raw string format, and move to `FileDialogs`
- Move hardcoded filename string suffixes (e.g. _ExportedPersonnel) into MekHQMenuBar.properties
- Move default filename formatting logic to `FileDialogs`
- Fix an inconsistency where exporting personnel omitted campaign name prefix, all three exports now uniformly use [Campaign Name][Date]_[Suffix].csv format
- Update `Utilities.exportTableToCSV` docs to accurately reflect its return behavior
- Date time formatter for file names is now static and reused across `FileDialogs` methods
- Reduce method access to private where applicable
- Add a `CampaignGUI.getHangarTab` helper method
- Create `FileDialogs.properties` file to house localized dialog titles and file suffixes


MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
+            JTable table = ((PersonnelTab) getGui().getTab(MHQTabType.PERSONNEL)).getPersonnelTable();
+            if (table.getRowCount() == 0) {
+                JOptionPane.showMessageDialog(getFrame(), resourceMap.getString("dlgNoPersonnel.text"));
+            } else {
+                saveToUserSelectedCSV(file -> Utilities.exportTableToCSV(table, file),
+                      MekHQ.getPersonnelDirectory(), resourceMap.getString("dlgSavePersonnelCSV.suffix"),
+                      resourceMap.getString("dlgSavePersonnelCSV.text"));
+            }
+        } catch (Exception ex) {
+            logger.error("Error exporting personnel to CSV", ex);
```
IllianiBird: Do we want to add a player-facing notification to make them aware export failed?

Not sure what the answer is, so mostly flagging this for consideration

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
-                }
-                JOptionPane.showMessageDialog(getFrame(), report);
-            });
-        } else {
-            JOptionPane.showMessageDialog(getFrame(), resourceMap.getString("dlgNoUnits.text"));
+    private void exportUnits() {
+        try {
+            JTable table = ((HangarTab) getGui().getTab(MHQTabType.HANGAR)).getUnitTable();
+            if (table.getRowCount() == 0) {
+                JOptionPane.showMessageDialog(getFrame(), resourceMap.getString("dlgNoUnits.text"));
```
IllianiBird: This should be using `getText` or `getTextAt`

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
+            } else {
+                saveToUserSelectedCSV(file -> Utilities.exportTableToCSV(table, file),
+                      MekHQ.getUnitsDirectory(), resourceMap.getString("dlgSaveUnitsCSV.suffix"),
+                      resourceMap.getString("dlgSaveUnitsCSV.text"));
+            }
+        } catch (Exception ex) {
+            logger.error("Error exporting units to CSV", ex);
```
IllianiBird: Similar comment to earlier's try-catch.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
-                }
-                JOptionPane.showMessageDialog(getFrame(), report);
-            });
-        } else {
-            JOptionPane.showMessageDialog(getFrame(), resourceMap.getString("dlgNoFinances.text"));
+    private void exportFinances() {
+        try {
+            Finances finances = getCampaign().getFinances();
+            if (finances.getTransactions().isEmpty()) {
+                JOptionPane.showMessageDialog(getFrame(), resourceMap.getString("dlgNoFinances.text"));
```
IllianiBird: getText/getTextAt

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
+            } else {
+                saveToUserSelectedCSV(file -> finances.exportFinancesToCSV(file.getPath(),
+                            FileType.CSV.getRecommendedExtension()),
+                      MekHQ.getFinancesDirectory(), resourceMap.getString("dlgSaveFinancesCSV.suffix"),
```
IllianiBird: I'm going to stop flagging these moving forward, but we really need to be moving away from `resourceMap.getString`.

The reason, chiefly, is that the `getText` family of methods (make sure to use the MHQ one) is our modern approach. The other (and this is the bigger reason imo) is that it includes protections against missing or malformed keys. Whereas `getString` will straight up just _brick_. This has caused a lot of problems in the past, including breaking releases.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
-
     /**
-     * Checks if a file already exists, if so it makes a backup copy.
+     * Asks the user to select output CSV file, enforces its extension, and runs {@code contentGenerator} for it.
+     * If the output file already existed, makes its backup.
      *
-     * @param file to determine if there is an existing file with that name
-     * @param path path to the file
+     * <p>
+     * To streamline UX, pre-populates output file name to [Campaign Name][Date]_[Suffix].csv and updates preferred
```
IllianiBird: Let's use `{@code [Campaign Name]` and `{@code [Date]_[Suffix].csv}`, helps it stand out in the JavaDoc and makes the doc easier to parse at a glance.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
-        File backupFile = new File(path2);
-        if (file.exists()) {
-            Utilities.copyfile(file, backupFile);
-        }
+    private void saveToUserSelectedCSV(Function<File, String> contentGenerator,
+          ObservableString dir, String filenameSuffix, String dialogTitle) {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(MHQConstants.FILENAME_DATE_FORMAT)
+                                            .withLocale(MekHQ.getMHQOptions().getDateLocale());
+        String recommendedFilename = getCampaign().getName() + getCampaign().getLocalDate().format(formatter) +
+                                           "_" + filenameSuffix + '.' + FileType.CSV.getRecommendedExtension();
```
IllianiBird: This is a big block of chained calls. Can we break it up at all?

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
+        GUI.fileDialogSave(getFrame(), dialogTitle, FileType.CSV, dir.getValue(), recommendedFilename).ifPresent(f -> {
+            File outputFile = enforceFileExtension(f, FileType.CSV);
+            // if the file already exists, make a backup copy.
+            if (outputFile.exists()) {
+                String backupPath = outputFile.getPath() + "_backup";
+                Utilities.copyfile(outputFile, new File(backupPath));
+            }
+            JOptionPane.showMessageDialog(getFrame(), contentGenerator.apply(outputFile));
+            dir.setValue(f.getParent());
```
IllianiBird: My _preference_ (emphasis important) is that we avoid code blocks in lambdas. I'd prefer for you to create a utility method and then call that method from within the lambda. Makes for more readable code.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
      */
-    private void checkToBackupFile(File file, String path) {
-        // check for existing file and make a back-up if found
-        String path2 = path + "_backup";
-        File backupFile = new File(path2);
-        if (file.exists()) {
-            Utilities.copyfile(file, backupFile);
-        }
+    private void saveToUserSelectedCSV(Function<File, String> contentGenerator,
+          ObservableString dir, String filenameSuffix, String dialogTitle) {
```
IllianiBird: `dir` I would prefer a more descriptive param name. I know `dir` is fairly self-explanatory but we're trying to avoid abstracted variable names. It caused a lot of problems in the codebase previously.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
```diff
@@ -1054,129 +1024,104 @@ protected void exportPlanets(FileType format, String dialogTitle, String filenam
...
-     * @param file to determine if there is an existing file with that name
-     * @param path path to the file
+     * <p>
+     * To streamline UX, pre-populates output file name to [Campaign Name][Date]_[Suffix].csv and updates preferred
+     * output directory to the user-selected selected one.
+     * </p>
+     *
+     * @param contentGenerator a function that writes to user-selected file
+     * @param dir              preferred output directory
+     * @param filenameSuffix . recommended file suffix
```
IllianiBird: Errant `.`

IllianiBird (changes requested):
A couple of minor points of feedback. I know a lot of these code isn't yours originally, but whenever we're updating existing code we should be attempting to modernize it as much as possible. With a project like this, it's slow, but it allows us to improve maintainability over a long enough period.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Good point, will add that

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Everything in that file uses `resourceMap.getString`. It will be another PR where all usages will be replaced, otherwise we will have a mix of two ways of getting the strings

IllianiBird: Mix actually isn't an issue. The current approach is to deprecate the old resourceMap, create a RESOURCE_BUNDLE and then migrate to that as we go.

If you're going to do that in a follow-up PR, though, non-issue. Whichever works, we just need to stop using `getString`.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Yes, I also dislike it. And there are so many things to improve. I just want to be methodical with the changes, focusing on one thing at a time

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Will do

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Will look into that

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Will look into that too

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Haven't considered that, it was the old structure of the code. I'll check a few options here

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Thought it was something on my screen lol

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
IllianiBird: Yeah, it's an older style. You'll see it used a lot in our older classes. But for new stuff, and refactors, I prefer for us to break the code out. Also has the added benefit of making it more testable.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Yes, I know what you meant. To clarify: there are 112 `resourceMap.getString` calls and replacing a few of them in this PR will barely make a dent. There will be a separate PR to address it.

IllianiBird: Sounds perfect :)

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: I just found `FileDialogs`. Seems like it is the way to go instead of keeping it in `MekHQMenu`. Thoughts?

IllianiBird: I have no objections

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Turns out, there's nothing to catch there. Methods that do the exporting are already wrapped into try-catch.

MekHQ/src/mekhq/gui/menus/MekHQMenuBar.java
hokvel: Reusing patterns from `FileDialogs` allowed to improve quite a few things. Now export calls are chains of `Optional(selectFile(UI)).map(export(dataModel)).ifExist(showConfirmation(UI))" with clear separation of UI and data model, while file selection logic is completely static. Plus dialog strings now have a new home in `FileDialogs.properties`.

hokvel: More details in the commit message here https://github.com/MegaMek/mekhq/pull/9218/changes/7af54fb0ff13b5c0711e5ecf6a7757cffc340692

IllianiBird (approved):

PR #9226, opened 2026-06-04
Fixed NPE in AutoRank System

IllianiBird:
Added `null` protections to autoranks ranks system. While a `null` person _shouldn't_ be passed in it's entirely possible that it could happen (and has) so let's protect ourselves.

Marked as critical as this can cause several systems to brick, it turns out.

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/ranks/AutoAssignRankForCompanyGenerator.java
+++ b/MekHQ/src/mekhq/campaign/personnel/ranks/AutoAssignRankForCompanyGenerator.java
@@ -127,6 +127,10 @@ public final class AutoAssignRankForCompanyGenerator {
         }
 
         for (Person person : crew) {
+            if (person == null) {
+                continue;
+            }
+
             // Skip anyone who already has a rank
             if (!hasNoRank(person)) {
                 continue;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/campaign/personnel/ranks/AutoAssignRankForCompanyGenerator.java
```diff
@@ -127,6 +127,10 @@ public static void assignRanks(Campaign campaign, Unit unit, Faction faction) {
         }
 
         for (Person person : crew) {
+            if (person == null) {
```
hokvel: My concern is that `crew` here is assigned from `Unit.getCrew()`, which has 37 usages. If we suspect that it can have nulls, we should fix it at the source, otherwise this block will signal that we can't trust it. I also took a look at how crew-related nulls are handled in `Unit` and it seems to be safe.

IllianiBird: Sounds good, I was just being paranoid after I hit an NPE earlier

IllianiBird:
Took another look, and actually found the root cause. I opted to leave two of the `null` checks in place regardless (though one was removed). I left those in as they were instances where we were passing in `Person` objects and so couldn't vet them prior to processing.

VicenteCartas (approved):

PR #9228, opened 2026-06-04
PR: Refactor usages of campaign.getWarehouse() with getWarehouse()

psikomonkie:
While working on #1099 I came across this refactor that hits a ton of files. I don't want that to clutter an already large pull request. 

This PR does not change any functionality. It adds a new method to `IWork` - `getWarehouse()` - that is implemented by the few base classes to do `campaign.getWarehouse()`. I then replaced all the other usages of `campaign.getWarehouse()` in the parts/work/refit to instead use the new `getWarehouse()` method. 

This refactor will allow #1099 to update the few actual implementations of `getWarehouse()` instead of all the files that were using `campaign.getWarehouse()`. 

I suspect I'll need to do this for `campaign.getHangar` and refits too, but I'll leave that for another PR, this one's already hitting 60 files.


IllianiBird (approved):
I know this isn't ready for prime time, just making it as approved so you can merge post-51.0's launch at your convenience

PR #9231, opened 2026-06-04
Removed 'Mercenary Campaign Size Affects Contract Pay' Campaign Option Due to Not Doing What It Said It Did and Not Working in a Way That is Conducive to a Fun Experience

IllianiBird:
As per title, the way this option worked was incredibly gross and impossible for the player to track. Basically each experience rating was given a break point. For every 12 units above that break point a random contract clause would be a (stacking) -1 penalty. Have a battalion over this breakpoint and you'd be looking at three (stacking) -1 penalties hitting 1-3 clauses.

The inability of the player to track this makes it nothing more than a "gotcha" that punishes success. Like it's sister option, that limited the size of combat teams, this is an old AtB campaign option that needs to go.

It's a newbie trap, nothing more.


hokvel (approved):
I was looking for documentation on why this logic existed in the first place and tracked it back to [this commit](https://github.com/MegaMek/mekhq/commit/9e27614e83d0cab5508cf0e5aaacdabce9ae6956#diff-1d46e7c8881873bcd6bbec0e5d335f007c45c7777e4238ed36a3758a5e6e176a), which is impressively old. I agree with the reasoning you provided for its removal.

VicenteCartas (approved):

PR #9232, opened 2026-06-05
Fixed Faction Standing Accolades Incorrectly Factoring in Climate Regard Resulting in Faction-Locked Rewards Being Given to Non-Faction Campaigns

IllianiBird:
Certain Accolade awards are meant to be locked behind the need to actually be a member of that faction. Due to an incorrect boolean polarity we were allowing non-faction campaigns to benefit from faction only awards. Now we don't.

## Addendum
Censures are still affected by climate regard, this is intentional.


Sleet01 (approved):
LGTM

PR #9233, opened 2026-06-05
Added Faction Standing Annual Decay Notice to Try and Reduce Player Confusion

IllianiBird:
I noticed many players are getting confused by the annual degradation of faction standing points (regard). I opted to include a little notice to clarify that this is annual decay in an attempt to mitigate this confusion.


Sleet01 (approved):
LGTM

PR #9235, opened 2026-06-05
Fix #8639: Fixed Faction Standing Incorrectly Reporting Positive Regard Gain as Negative When Regard is Already at Maximum Value

IllianiBird:
Fix #8639

I do love a single-character fix!

If the player received Regard for a faction and their Regard is already maxed out they wouldn't see any improvement. This would cause the polarity of the report to incorrectly call it a loss. Now it doesn't

This was a visual bug only.

PR #9236, opened 2026-06-05
Minor Renaming of a Handful of Faction-Related Variables to Improve Clarity

IllianiBird:
While investigating #9235 I found myself tripping over a few variable names and so opted to rename them for improved clarity.

Going to mark this PR for new dev cycle, not because it's remotely risky, but because there is 0 need for it to be included in 51.0


MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -535,20 +535,19 @@ public void updateEnemy(Campaign campaign, LocalDate today, @Nullable String ene
 
         // Update the Batchall information
         batchallAccepted = true;
-        Faction faction = getEnemy();
```
IllianiBird, line 539: Removed this line because it was wholly unnecessary, we've already fetched the enemy faction earlier in the method. No reason to fetch it again.

VicenteCartas (approved):

PR #9238, opened 2026-06-05
Fixes resolve scenario wizard from extending off the bottom of the screen most of the time

### LINKED ISSUE #8568: [RFE] After action report window should remember window size adjustments
### Enhancement Type *

Currently the after action report screen (the one where you see your injuries, prisoners, scenario results, etc) is a little too tall on my screen and I need to resize it every time before I can hit the finish button.  It would be helpful if it would remember the new window position.

### Use Cases or Rationale

1. It would remove a minor irritation that happens after every scenario as I could close the window without resizing and dragging it every time.

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.10

### Final Verification

### LINKED ISSUE #8660: [RFE] Ensure popup windows fit within the screen without clipping into the toolbar
### Enhancement Type *

Since neither megamek nor MekHQ can be made fullscreen, both have occasional issues with popups clipping out of screen and needing to be resized constantly. Making these popups adapt to the size of the game would improve the experience.

Image

### Use Cases or Rationale

Having to resize certain windows every time they appear is a minor annoyance and so fixing it would improve user experience.

### Attach Files

_No response_

### MekHQ Suite Version *

50.11

### Final Verification

heckindoc:
Should (mostly) fix #8568 and fix #8660 

Currently the preferred height for the scenario summary wizard is set to 1000 pixels, which since it starts approximately 300 pixels from the top of the screen always has the bottom of the window extending off the bottom of the screen where it then has to be moved or resized after every scenario. We could A. move the window up B. decrease the "preferred" window size or C. do both. I opted with option B because I couldn't immediately find option A and figured having a window that could potentially be too small, but has all buttons immediately available and could be resized was preferrable to a window that was too large and always had to be resized. So now the window defaults to being 700 pixels tall.


IllianiBird (approved):
If you want a goal for a follow-up PR, take a look at `mekhq/gui/dialog/FamilyTreeDialog.java` specifically lines `137` and `156` -> `164`. If you can implement preferences in this class it will cause MekHQ to remember the last size and (potentially, I forget) position. This will mean users won't need to keep resizing it.

heckindoc:
ooo yeah, I'll take a look at that. Thank you! Hopefully this is an acceptable quick fix now and that can be a goal for the next dev cycle?

IllianiBird:
Absolutely

rjhancock (approved):

PR #9242, opened 2026-06-05
Fixed Null Protection in `getSpecialAbilities`

IllianiBird:
This PR is in response to MegaMek/megamek#8311 but does not directly fix the root cause of that PR.

The goal of this PR is to fix the underlying failure that led to MegaMek/megamek#8311, so that we don't have a broken Nightly, tonight. I can then, tomorrow, go through and find the root cause.


MekHQ/src/mekhq/campaign/personnel/generator/SingleSpecialAbilityGenerator.java
```diff
@@ -304,14 +304,21 @@ private List<SpecialAbility> getEligibleSPAs(Person person, boolean isVeterancyA
...
+                if (spa == null) {
+                    LOGGER.error("Missing ability: {}", ability.getName());
+                    continue;
+                }
+
                 if (isVeterancyAward && spa.getOriginOnly()) {
                     continue;
                 }
 
-                if ((spa == null) || (spa.getWeight() <= 0)
```
IllianiBird, line 311: This null protection was too late in the function to protect against NPEs. So I moved it higher as well as adding better logging.

VicenteCartas (approved):

PR #9253, opened 2026-06-06
Fixed Null Protection Reporting False-Positive Events

IllianiBird:
My earlier fix to a faulty `null` guard was a little too paranoid miss-reporting missing SPA data, when this is in fact working as intended. The bug was the faulty guard and not an issue with how we were handling data.


Sleet01 (approved):
LGTM

PR #9254, opened 2026-06-06
Fixed Inverted Polarity in Veterancy Award Eligibility

IllianiBird:
This conditional was incorrectly flagging characters as ineligible to receive Veterancy Awards. Now it isn't.

Marking this as critical, not because it's game breaking, but because if we don't fix this before release we will never hear the end of it :D


Sleet01 (approved):
LGTM

PR #9260, opened 2026-06-06
PR: Refactor certain ILocations to be IPlace

psikomonkie:
**Incudes #9228 and #9261**

Yet another PR for supporting #1099.

This PR implements the `IPlace` interface. `IPlace` extends `ILocation` and is used for special locations that are expected to have certain properties. 

Currently that's just changing the default `ILocation`'s `getHangar()`, `getWarehouse()`, and `getPersonnel()` to return null - we don't want to keep going up the tree once we reach a place, we want the place to return its hangar/warehouse/personnel, if it has one.


psikomonkie:
Please use "Update Branch" after merging #9261 but before reviewing this. Thanks!

IllianiBird (approved):

PR #9261, opened 2026-06-06
PR: Stub implementations for `ILocation`

psikomonkie:
**Includes #9228**

Another refactor for #1099. 

This PR should be strict refactor with absolutely no gameplay impact.

This is a stub implementation of `ILocation` for the classes that will be made into `ILocation`s for #1099. Each of these classes now have a `LocationNode`, and is added into the campaign's `LocationNode` tree.

There will only be one tree with this implementation - the main force's (`Campaign`).

PR #9262, opened 2026-06-06
PR: Replace campaign.* usages from ILocation implementations

psikomonkie:
**Includes #9228 and #9260 and #9261**

Another prerequisite on the road to #1099. 

This PR should be strict refactor with absolutely no gameplay impact.

Using `campaign.getWarehouse()` and `campaign.getHangar()` is incredibly common in the code. 

When moving to multiple locations we cannot get the warehouse or hangar this way because the campaign will have multiple warehouses and hangars - we need to get the _relevant_ warehouse or hangar.

The classes inherit the `getPersonnel()`, `getWarehouse()`, and `getHangar()` methods. When any of these relevant objects are added, we set up the `LocationNode` tree. These are not necessarily the final implementations of these methods. The important part was extracting all of the `campaign.` calls so the PR won't be so cluttered when I do finish implementing it the location-aware logic. 

With those new methods, I then tried to replace every usage of an `ILocation`'s `campaign.*` with its new getter. 

There are some spots where we want to get all units/people/parts, like for accounting. There are new methods `getAllPersonnel`, `getAllWarehouse`, etc. that just return the same thing as the normal getter. It's the same deal - I just wanted to get the rename done so the full #1099 PR can be as small as possible.


psikomonkie:
Please use "Update Branch" after merging #9260 but before reviewing this. Thanks!

MekHQ/src/mekhq/campaign/CampaignSummary.java
```diff
@@ -150,7 +150,7 @@ public void updateInformation() {
         aeroCount = 0;
         infantryCount = 0;
         int squadCount = 0;
-        for (Unit u : campaign.getHangar().getUnits()) {
+        for (Unit u : campaign.getAllHangar().getUnits()) {
```
IllianiBird: `unit`

IllianiBird (approved):

MekHQ/src/mekhq/campaign/CampaignSummary.java
psikomonkie: Threw in a bonus `e` -> `entity`, enjoy!

PR #9266, opened 2026-06-06
Fix assisted healing using free edge rolls

hokvel:
In the case of assisted healing rolls with edge were not correctly expending it nor checked if the doctor had edge to spend.


IllianiBird (changes requested):
We actually handle checking for available edge inside SkillCheckUtility (and AbilityCheckUtility), for the former it's the method `performInitialRoll`.

See also `rollWithEdge`

hokvel:
> We actually handle checking for available edge inside SkillCheckUtility (and AbilityCheckUtility), for the former it's the method `performInitialRoll`.

In that method both rolls are done with `useEdge == false` most likely because rerolling needs to be done under stricter rules.

IllianiBird (approved):
You're absolutely right. I think we can both guess how this bug came about in the first place ;)

Issue #3, opened 2016-01-21
Ships engineer not assigned properly when Hire Full Compliment used on Space Stations


pheonixstorm:
0.3.23

After buying a space station, using hire full compliment will not assign a tech to some space stations. After several tests it seems that this only happens with stations that do not have gunners. Normal crew assignments work fine.

This issue will go away if you un-assign one crew member and then reassign them to the station.
[Station test.zip](https://github.com/MegaMek/mekhq/files/98617/Station.test.zip)

pheonixstorm:
Small update. The unassign/assign does not last. When you save/reload the campaign the assigned tech is blank again.

pheonixstorm:
Update 2.
This bug also seems to cause repairs to the affected station to be shorthanded.

Dylan-M:
Space stations are not fully supported at this time.

pheonixstorm:
Yeah, just checked with 0.3.18 as I thought it was related to the commit I posted about in the first ticket but happens prior to those changes as well.

pheonixstorm:
Question. Is there any reason a space station does not have a minimum of one pilot? They do have station keeping drives so someone competent should be at the helm.

mkerensky:
Still an issue as of 0.43.4.   It would seem that the unarmed station doesn't count as a unit or its personnel don't count as people -  Unit.resetEngineer never triggers for the unarmed station, though it does for all other units - mechs, other spacecraft, etc. I think it actually triggers on a per-person basis.  Digging continues...

mkerensky:
Fixed in my current PR dealing with edge for non-combat rolls. Had to do some other work on the ship's engineer anyway, so it made sense to add this in too.  Unit.calcCompositeCrew() was returning empty because the unarmed stations lacked both pilots and gunners, thus triggering getCrew().isMissing() in Entity.  I added a check to ignore space stations for this line, but adding at least 1 pilot to the station crew would be another possible fix, and I agree with your point that space stations should have pilots if JumpShips do.  I would be inclined to add 3 just like all other large craft have.

arlith:
Fixed with merge of MM PR #701

Issue #8336, opened 2025-11-24
[Issue] Pacar finishes, but mekhq does not "notice".


Daishi-5:
### Brief Description *

Sometimes, when running pacar for a scenario, the scenario will finish, but nothing happens in mekhq when it should ask if I control the battlefield.  

I can run pacar again, but usually a second try to run pacar does nothing.  It acts like it will start, but does nothing after clicking ok.

Trying a third time usually works to start pacar, but usually it ends the same way, where mekhq acts like nothing happened.  Once, while testing it, the third time did work though, so it is not completely consistent.

The attached save has a scenario queued up where the issue has just happened, and all logs are included.

This issue has occurred several times when I clicked "request victory" so it might be possible to replicate by quickly claiming victory on several scenarios in a row.

### 3. Steps to Reproduce

Unfortunately, loading a save often fixes the problem, so it is hard to reproduce, but usually after 2-4 successful pacar simulations, the issue will occur again.

### Attach Files

[Pacarissue.zip](https://github.com/user-attachments/files/23727542/Pacarissue.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

0.50.10 Nightly 11-24

### Operating System *

Windows 11

### Java Version *

17.0.16

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

smrogers91:
On Nightly 11-24 on JDK17 I just had this same issue happen. It occurred on an instance of mekhq which I had running for a long period of time (6-8 hours) and after I had done about 6-7 scenarios. Some of those resolved with PACAR and some of those by playing it out. 

At the conclusion of the PACAR run through, I hit complete and it closed the window and went to MekHQ as though the scenario had not been run. 

When I get some time shortly I'll upload the log and save

IllianiBird:
Adopting this from @Sleet01 as I'm planning a big PACAR QA pass next week

PR #9274, opened 2026-06-07
New Campaign Options dialog

VicenteCartas:
## Summary

This PR rebuilds the **Campaign Options dialog** around a clean separation between *navigation*, *form/view*, and *domain model*. The old dialog was a deeply nested stack of `JTabbedPane`s where each tab class both built Swing widgets **and** read/wrote `CampaignOptions` inline. The new design introduces a searchable navigation tree, a per-area "options model" seam, and a library of reusable form components — and the view itself is decomposed into thin per-area **coordinators** that delegate to small **single-screen page** classes.

It's large, but the churn is concentrated and repetitive: once you understand **one** vertical slice, every other slice has the same shape. See **"How to review this efficiently"** below.

---

## Why

- **Slow startup.** Opening Campaign Options meant building all the UI up front, even for screens the user may never visit.
- **Navigation didn't scale.** Options lived in nested tabs-within-tabs; finding a setting meant clicking through several tab levels.
- **No seam between form and domain.** Tabs poked `CampaignOptions` getters/setters directly, so there was no unit-testable boundary and no single place to see "which options does this screen own."
- **Monolithic screens.** Each area was one giant tab class (often 1,000–1,800 lines) mixing several unrelated screens together.
- **Layout was ad hoc.** Each tab hand-rolled its `GridBagLayout`, so spacing/width/section behavior drifted between screens.

## What changed (architecture)

The dialog is now four layers: a **navigation shell**, an **options-model layer**, a **view layer** (per-area page coordinators + single-screen pages) built from **reusable components**, plus **tests** that the new seam makes possible.

### 1. Navigation shell — *new* (`gui/campaignOptions/`)
| File | Role |
|------|------|
| `CampaignOptionsPane` | Orchestrator. Owns the list of routes, **lazily** builds pages via factory suppliers, caches them, and maintains a search index. |
| `CampaignOptionsNavigationPanel` | Searchable **`JTree`** navigation rail + filter field that replaces the nested tabs. |
| `CampaignOptionsContentHost` | Central scrollable content area + sticky **contextual help** surface. |
| `CampaignOptionsRoute` | Value object for a navigable destination (id, hierarchical path, title resource keys, searchable text). |
| `CampaignOptionsHelpPanel` | Contextual help/tip panel, fed through `CampaignOptionsUtilities.setTipTextConsumer`. |

### 2. Options-model layer — *new* (`contents/*OptionsModel.java`) — **the heart of the refactor**
Each domain area gets a plain data-holder model that mirrors its slice of `CampaignOptions` and knows how to **load from** / **apply to** it: Personnel, Biography, AwardsAndRandomization, Finances, Markets, Relationships, Salaries, Skills, Systems, Rulesets, TurnoverAndRetention, EquipmentAndSupplies, RepairAndMaintenance. This is the seam that decouples the Swing form from the domain object. (The **General** and **Abilities** screens apply directly to `CampaignOptions` and have no model.)

### 3. View layer — per-area coordinators + single-screen pages, built from reusable components
The former monolithic `*Tab` classes are gone. Each area is now a thin **coordinator** (`contents/*Pages.java`) that:
- owns its `*OptionsModel` and the **load-from / apply-to** lifecycle, and
- registers the area's screens, delegating each screen's Swing construction to a small **leaf page** (`contents/*Page.java`).

Each leaf `*Page` builds exactly one screen from the shared `components/` primitives and syncs to/from the model via `readFromModel` / `writeToModel`; the coordinator never hand-builds widgets. For example, **Personnel**: `PersonnelPages` (coordinator) delegates to `PersonnelGeneralPage`, `AwardsPage`, `MedicalPage`, `PersonnelInformationPage`, and `PrisonersAndDependentsPage`. This took the largest areas from ~1,000–1,800-line classes down to ~150–270-line coordinators plus focused page files.

A few areas stay a single class because they are one screen (or one screen rendered per category): `GeneralPage` (single screen, no model), and the category-parametrized `SkillsPages`, `SalariesPages`, and `AbilitiesPages`.

- `components/` (shared building blocks): `CampaignOptionsPagePanel` (fluent **Builder** that assembles a standard page: header → intro → collapsible sections → quote footer at a uniform width), `CampaignOptionsFormPanel`, `CampaignOptionsPairedFieldGridPanel`, `CampaignOptionsModifierTablePanel`, `CampaignOptionsIntroPanel`, `CampaignOptionsHeaderPanel`, `SectionHeaderControlProvider`, plus typed widgets (`CampaignOptionsButton/CheckBox/Label/Spinner/TextField`).
- Editors: `AbilitySelectorDialog`, `SkillPrerequisitesDialog`, `SkillAdvancedEditorDialog` + `SkillConfiguration` (reworked skills/abilities editing).

### 4. Tests — *new* (`unittests/.../campaignOptions/`)
- `contents/*OptionsModelTest` + `OptionsModelTestSupport`: **round-trip** tests (load → mutate every field → apply → assert) — only possible because of the model seam. These are the best guard against field-mapping regressions.
- `components/*Test`: behavior of the shared form/page panels.

### 5. Preset save/load UX
- **Save Preset** now auto-saves into the user campaign-preset directory (the folder Load Preset scans) with filename sanitization and an overwrite confirmation, instead of a free-form file chooser.
- The success dialog shows a **selectable path** with an inline **open-folder** button.
- On a **declined overwrite**, the builder re-opens with all selections preserved — `CreateCampaignPreset` now repopulates name/description/toggles/combos from a passed-in preset (which also makes the dormant "Edit Preset" path functional).

### Peripheral changes
`RankSystemsPane`, `RankTableModel`, `EditSkillPreRequisiteDialog`; small touches to `CampaignSummary`, `CampaignNewDayManager`, `EntityUtilities`, `Fatigue`, `TrainingCombatTeams`; resource bundles (`CampaignOptionsDialog.properties`, `GUI.properties`, `GlossaryEntry.properties`); `FileDialogs` drops the now-unused `saveCampaignPreset`.

---

## How to review this efficiently

1. **Start with the shell** to understand the runtime structure: `CampaignOptionsPane` → `CampaignOptionsRoute` → `CampaignOptionsNavigationPanel` → `CampaignOptionsContentHost`.
2. **Read one vertical slice end-to-end**, e.g. Personnel: `PersonnelOptionsModel` (model) → `PersonnelPages` (coordinator) → its leaf pages (`PersonnelGeneralPage`, `AwardsPage`, …) → `PersonnelOptionsModelTest` (round-trip). **Every other area is the same shape** — once this clicks, the rest is pattern-matching.
3. **Skim `components/`** to learn the shared page/form primitives the pages now use.
4. **The `*Pages`/`*Page` diffs look enormous but are largely mechanical**: inline `CampaignOptions` get/set → model fields, bespoke `GridBagLayout` → shared builders, and one big tab → a coordinator plus per-screen pages. Review **one** area carefully (coordinator + one leaf), then skim the others for deviations.
5. **Behavior changes live in the models.** If you're hunting for regressions, compare each `*OptionsModel`'s load/apply mappings against the old tab's get/set calls; the model tests assert these.

## Screenshots

image
image
image
image
image
image
image


MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsDialog.java
```diff
@@ -247,24 +274,129 @@
...
+                openFolderButton.setIcon(folderIcon);
+            } else {
+                openFolderButton.setText(getTextAt(bundle, "savePresetSuccess.openFolder"));
+            }
+            openFolderButton.setToolTipText(getTextAt(bundle, "savePresetSuccess.openFolder"));
+            openFolderButton.addActionListener(evt -> {
+                try {
+                    Desktop.getDesktop().open(presetDirectory);
+                } catch (Exception ex) {
+                    LOGGER.error(ex, "Failed to open campaign preset folder: {}", presetDirectory);
```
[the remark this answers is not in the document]
VicenteCartas: You seem to get this wrong constantly.

HammerGS (commented):
## Code-quality review (draft PR)

Reviewed the whole diff: the new dialog shell (route model, navigation panel, content host, page panels, form helpers), the migrated pages, the `*OptionsModel` classes and their tests, the new dialogs, and the rank-table changes.

Overall this is a strong refactor. Routing values through per-section `OptionsModel` classes made load/apply symmetric on almost every page - I checked them field-by-field and found very few mismatches. The model unit tests use a reflection round-trip harness that would catch most cross-wiring, which is a good call.

Findings are grouped by theme, with `file:line` and a proposed fix for each. Nothing here blocks the direction of the PR. (I couldn't get the inline-comment API to resolve lines on this merged branch, so they're listed here instead.)

---

### Must-fix

**1. Unicode literals in source** - `contents/SkillsPages.java:624, 645, 648`
Raw `·` and `—` glyphs in UI strings (`String.format("G %d · R %d ...")` and the summary builder). These are real bytes, not escapes, so they break the no-Unicode-in-source rule (Windows encoding/build hazard).
Fix: use `\u00B7` / `\u2014`, or move the separators into the resource bundle.

**2. Camo icon not scaled after selection** - `contents/GeneralPage.java:456`
`btnCamo.setIcon(camouflage.getImageIcon());` sets the icon at native size. The initial icon (line 243) uses `getImageIcon(UIUtil.scaleForGUI(75))`, so the button jumps to the wrong size once the user picks a camo.
Fix: `btnCamo.setIcon(camouflage.getImageIcon(UIUtil.scaleForGUI(75)));`

**3. Shared column width not GUI-scaled** - `components/CampaignOptionsFormPanel.java:431`
`setMinimumWidth` uses the raw constants `DEFAULT_LABEL_WIDTH` (300) / `DEFAULT_CONTROL_WIDTH` (220). At higher GUI scales the column floor stays at 1x while fonts grow, so columns misalign.
Fix: `int width = Math.max(preferredSize.width, UIUtil.scaleForGUI(minimumWidth));` - this also covers pages that pass raw pixel widths through this panel (e.g. PriceMultipliersPage, StratConPage).

---

### Architecture: global static help-text consumer

`CampaignOptionsUtilities` holds a `static Consumer<String> tipTextConsumer` (line 99). `CampaignOptionsContentHost` sets it in its constructor and nulls it in `removeNotify()`. Two problems:

1. Global mutable UI state on a class whose own Javadoc says it is stateless. Two dialogs (or a stale one) would fight over the same field, so hover help from one drives the other.
2. `removeNotify()` nulls it unconditionally. `removeNotify` runs on any reparent, not just teardown, and it doesn't check whether `this` is still the owner - so a surviving host can lose its help wiring.

Fix: give the host an instance reference and pass it down the page-construction path instead of a static. If that's too large for this PR, at least make the clear a compare-and-clear (only null when `this` registered it).

---

### Correctness edge cases

**4. Null spinner value on load** - `contents/PersonnelMarketPage.java:310`
`entry.getValue().setValue(model.personnelMarketRandomRemovalTargets.get(entry.getKey()));` - the spinner map is keyed by `SkillLevel.values()` (all 8), but the model map only holds what the save stored. An older save missing a level (e.g. HEROIC/LEGENDARY) returns null, and `JSpinner.setValue(null)` throws, breaking page load.
Fix: `Integer target = model.personnelMarketRandomRemovalTargets.get(entry.getKey()); if (target != null) { entry.getValue().setValue(target); }`

**5. AIOOBE on combo with no selection** - `contents/AcquisitionPage.java:516`
`ProcurementPersonnelPick.values()[cboProcurementPersonnelPick.getSelectedIndex()]` throws if the selected index is -1, and it addresses the value differently from the read side (line 481 selects by `toString()`). It works now only because the combo is built in enum order.
Fix: make it a typed `MMComboBox<ProcurementPersonnelPick>` and use `getSelectedItem()` on both sides.

**6. Portrait read/write asymmetry** - `contents/NameAndPortraitGenerationPage.java:314` (vs 342-347)
The Civilian checkbox is written out to every civilian role on apply, but read back from `usePortraitForRole[i]` by array index - a single role. It only round-trips today because civilian roles are always written together.
Fix: read it the way you write it, from `PersonnelRole.getCivilianRoles()`, so the two sides can't drift. (Minor - fragility, not a live bug.)

---

### GUI scaling (same root cause as must-fix #3)

Raw pixels not run through `UIUtil.scaleForGUI()`:

- `CampaignOptionsNavigationPanel` `NAVIGATION_WIDTH = 240` (preferred size / split divider) is unscaled, and the same constant is duplicated in `CampaignOptionsPane`. Scale it and share one constant.
- `CampaignOptionsHeaderPanel` passes `DEFAULT_IMAGE_SIZE = 80` to `scaleImageIcon` unscaled (the body-text width on the next line *is* scaled, so the file is inconsistent). It also reloads/retints the header image from disk on every build - worth caching by (address, size, tint).
- `CampaignOptionsPairedFieldGridPanel` and `CampaignOptionsModifierTablePanel` feed raw width constants into `setPreferredSize`/`setMinimumSize` too.
- `RankSystemsPane:565` adds a flat `+ 6` px column padding; wrap in `scaleForGUI`.

---

### Naming (reviewer rule - PR #8336 / #8350)

New code with abbreviated identifiers reviewers will flag:

- `RankSystemsPane`: `rh` (row-header table, 430-460), `cm` (column model, 418/448/468/497), lambda param `evt` (459/526/530) -> `rowHeaderTable`, `columnModel`, `event`.
- `RandomizationPage`: `lblCombatSA` / `spnCombatSA` / `lblSupportSA` / `lblArtyProb` / `spnArtyProb` / `lblSecondProb` (~125-140). The resource keys already use full words; only the Java identifiers need it.
- `AbilitySelectorDialog`: `DocumentListener` params named `e` (105/110/115) -> `event`.

---

### Minor / nice-to-have

- `components/AbilitySelectorDialog.java:135` - a new `NaturalOrderComparator` is constructed inside the sort lambda, so one is allocated per comparison. Hoist a single instance out and reuse it (and rename lambda params `a`/`b`).
- `CampaignOptionsDialog:325` - `presetDirectory.mkdirs()` return value ignored; if it fails the later write throws an opaque error. Check it and log/abort.
- `ContractMarketPage` - field `useInfantryDoseNotCountBox` / `model.infantryDontCount`: misspelled ("Dose"->"Does") and breaks the `chk` prefix convention; the i18n key inherits the typo.
- `CreateCampaignPreset:409` - sets `JTextField.placeholderText` on a `JTextArea`. FlatLaf only renders that placeholder for `JTextField`, so the description box shows no hint (small regression vs the old seeded text).
- `CampaignOptionsContentHost` / `CampaignOptionsModifierTablePanel` use old-style `instanceof`+cast on touched lines, while the sibling `CampaignOptionsPairedFieldGridPanel` already uses pattern-matching `instanceof`. Easy to align.

---

### Performance (already on your plan.md, step 10 - flagging, not asking to fix here)

`AbilitiesPages` builds every ability's full panel up front though sections are collapsed, and rebuilds all created categories on each refresh. `SkillsPages` builds the detailed cost/milestone controls eagerly. These are the known heavy-construction spots.

---

### Test gaps (for the separate test PR)

The model tests are good, with one real hole: `SkillsOptionsModelTest` only round-trips `edgeCost`/`attributeCost` and one `targetNumber`. The per-skill `costs[]` array and the six milestone level fields - the model's core payload - are never asserted to survive `applyTo`, which is exactly where a load/apply bug would hide. `SkillConfigurationTest` covers the DTO in isolation, but not the model wiring.

Also: `RepairAndMaintenanceOptionsModelTest` is hand-written rather than using the reflection harness, so it'll silently miss any field added later.

I can put these (plus regression tests for the correctness edge cases above) into a test-only branch off `vc/options` as a separate PR you can review and merge, per the plan. Say the word.

*(Reviewed and verified against the source before posting; posted via Claude Code.)*

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsCheckBox.java
```diff
@@ -120,11 +121,12 @@ public CampaignOptionsCheckBox(String name, @Nullable CampaignOptionsMetadata me
...
      */
-    public CampaignOptionsCheckBox(String name, @Nullable Integer customWrapSize,
+    public CampaignOptionsCheckBox(@Nonnull String name, @Nullable Integer customWrapSize,
                                    @Nullable CampaignOptionsMetadata metadata) {
         // Fetch base text and append badges
-        // Sets the checkbox's text from the resource bundle, wrapped in HTML tags
-        super(String.format("<html>%s%s</html>",
+        // Sets the checkbox's text from the resource bundle, wrapped in HTML tags. The <nobr> keeps the label and its
+        // trailing badges on a single line so the badge symbols do not wrap onto a second line in tight grid columns.
+        super(String.format("<html><nobr>%s%s</nobr></html>",
```
IllianiBird, line 129: Do we really want to use String.format here, or would getFormattedTextAt be better?

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHeaderPanel.java
```diff
@@ -59,21 +59,16 @@
...
-    }
-
-    public static String getTipPanelName() {
-        return TIP_PANEL_NAME;
-    }
+    // Header images are reused across every page build; cache the loaded/scaled/tinted icon by (address, size, tint)
+    // so we don't reload from disk and re-tint each time. Accessed only on the EDT, so a plain HashMap is safe.
+    private static final Map<String, ImageIcon> HEADER_IMAGE_CACHE = new HashMap<>();
 
     @Deprecated(since = "0.50.06", forRemoval = true)
```
IllianiBird: This is a substantive enough rework we should just remove any deprecated methods instead of updating them.

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsPagePanel.java
```diff
@@ -0,0 +1,540 @@
...
+
+    @Override
+    public void doLayout() {
+        Dimension preferredSize = pageBody.getPreferredSize();
+        // Give the page body its full content width, even when that exceeds the viewport. Squeezing the inner
+        // GridBagLayout below its content width makes the section stack collapse to its zero minimum width and the
+        // header text wrap vertically, which wipes the whole page. Letting any overflow clip on the right keeps the
+        // header on one line and the sections visible, which is a far better failure mode than a blank page.
+        int contentWidth = preferredSize.width;
+        int x = Math.max(0, (getWidth() - contentWidth) / 2);
```
IllianiBird, line 195: Single letter variable name

(this is a joke)
(I am very funny)

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsPagePanel.java
```diff
@@ -0,0 +1,540 @@
...
+        controls.setOpaque(false);
+        controls.add(expandAllButton);
+        controls.add(collapseAllButton);
+
+        return controls;
+    }
+
+    private JButton createSectionActionButton(String resourceKey) {
+        JButton button = new JButton(getTextAt(getCampaignOptionsResourceBundle(), resourceKey));
+        button.putClientProperty("JComponent.sizeVariant", "small");
```
IllianiBird: If we find ourselves using this a lot it would be better to define `"small"` as a static const so we only need to change the one place if we ever want to change size. Non-essential change, entirely your call.

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
```diff
@@ -0,0 +1,798 @@
...
+              .header(headerPanel)
+              .showDetailsPanel(false)
+              .component(createAbilityButtonBar(abilityCategory));
+
+        // Every special ability is selectable as a prerequisite/incompatible/removed entry, so build the lookup map
+        // once and share it with each ability control's selector popups.
+        Map<String, SpecialAbility> allSPAs = new HashMap<>();
+        for (CampaignOptionsAbilityInfo info : allAbilityInfo.values()) {
+            allSPAs.put(info.getAbility().getName(), info.getAbility());
+        }
```
IllianiBird: Let's break this out into a utility method

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
```diff
@@ -0,0 +1,798 @@
...
+        for (String abilityName : sortedAbilityNames) {
+            CampaignOptionsAbilityInfo abilityInfo = allAbilityInfo.get(abilityName);
+
+            if (abilityInfo.getCategory() == abilityCategory) {
+                AbilityOptionPanel abilityPanel = new AbilityOptionPanel(abilityInfo, allSPAs);
+                builder.literalSection(abilityInfo.getAbility().getDisplayName(),
+                      abilityPanel.getSummaryText(),
+                      abilityPanel);
+            }
+        }
```
IllianiBird, line 302: And this

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
```diff
@@ -0,0 +1,798 @@
...
+            setName("pnl" + ability.getName() + "Ability");
+
+            buildHeaderControl();
+
+            // The cost resource is a formatted "XP Cost: %s" string; format with an empty value to reuse it as the
+            // spinner's prefix label without introducing a new resource key.
+            String costPrefix = String.format(getTextAt(getCampaignOptionsResourceBundle(), "abilityCost.text"), "")
+                                      .trim();
+            JLabel lblCost = new JLabel(costPrefix);
+            JSpinner spnCost = new JSpinner(new SpinnerNumberModel(ability.getCost(), -100000, 100000, 1));
```
IllianiBird, line 449: Something I've been wanting for a while is a utility JSpinner inheritor that adds 'min, max, and default' to the bottom of any tooltips. We'd then replace every JSpinner across the codebase to use this new inheritor.

Not something that should be done in this PR, but if you've an interest it would directly benefit IIC-2

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
```diff
@@ -0,0 +1,798 @@
...
+         * @param skillPrerequisite the prerequisite to format
+         *
+         * @return an HTML snippet sized to the column width
+         */
+        private static String formatSkillPrerequisite(SkillPrerequisite skillPrerequisite) {
+            String body = skillPrerequisite.toString();
+            if (body.startsWith("{") && body.endsWith("}")) {
+                body = body.substring(1, body.length() - 1);
+            }
+            body = body.replace("<br>OR ", "<br><span style='color:gray;'>or </span>");
```
IllianiBird: If we're going to use String comparison (god I hate String comparison) let's pass in a const rather than a magic string. It's also worth noting that, due to using String comparison, this will break if we ever change the underlying language used by SPA displays. Please make a note wherever that code lives that here will need to be updated, too.

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
```diff
@@ -0,0 +1,798 @@
...
+
+            List<String> sorted = new ArrayList<>(current);
+            sorted.sort(Comparator.comparing(SpecialAbility::getDisplayName, String.CASE_INSENSITIVE_ORDER));
+            for (String abilityName : sorted) {
+                column.add(makeChipRow(abilityName, () -> {
+                    Vector<String> updated = new Vector<>(current);
+                    updated.remove(abilityName);
+                    setter.accept(updated);
+                    rebuild.run();
+                }));
```
IllianiBird: If a lambda ends up more than a couple lines long it should be broken out into a utility method. Elsewise code readability can take a nosedive.

MekHQ/src/mekhq/gui/campaignOptions/contents/BiographyGeneralPage.java
```diff
@@ -0,0 +1,327 @@
...
+    /**
+     * Builds the General page, populates its controls from the supplied model, and returns the assembled panel.
+     *
+     * @param model the shared biography options model to populate the freshly built controls from
+     *
+     * @return a {@link JPanel} representing the General Page
+     */
+    @Nonnull JPanel createPanel(@Nullable BiographyOptionsModel model) {
+        // Header
+        String imageAddress = getImageDirectory() + "logo_clan_blood_spirit.png";
```
IllianiBird, line 112: I can't help but wonder if we should create an enum with these references in so we can just grab the names from there (as well as seeing if a logo has already been used). Not a needed change, just a wild thought as I'm reviewing.

MekHQ/src/mekhq/gui/campaignOptions/contents/DeathPage.java
```diff
@@ -0,0 +1,202 @@
...
+        if (!created || model == null) {
+            return;
+        }
+
+        chkUseRandomDeathSuicideCause.setSelected(model.useRandomDeathSuicideCause);
+        spnRandomDeathMultiplier.setValue(model.randomDeathMultiplier);
+        for (final AgeGroup ageGroup : AgeGroup.values()) {
+            JCheckBox ageGroupCheckBox = chkEnabledRandomDeathAgeGroups.get(ageGroup);
+            if (ageGroupCheckBox != null) {
+                ageGroupCheckBox.setSelected(Boolean.TRUE.equals(model.enabledRandomDeathAgeGroups.get(ageGroup)));
```
IllianiBird: Can this be the primitive? Or does it need to be whatever the opposite of a primitive is called?

MekHQ/src/mekhq/gui/campaignOptions/contents/NameAndPortraitGenerationPage.java
```diff
@@ -0,0 +1,367 @@
...
+                        civilianSelected = model.usePortraitForRole[role.ordinal()];
+                        break;
+                    }
+                }
+                chkUsePortrait[i].setSelected(civilianSelected);
+                continue;
+            }
+            chkUsePortrait[i].setSelected(model.usePortraitForRole[i]);
+        }
+    }
```
IllianiBird: Is there any way we could make this not o(n)? I'm fine if we can't, but for-loops nested inside for-loops is a warning sign.

MekHQ/src/mekhq/gui/campaignOptions/contents/NameAndPortraitGenerationPage.java
```diff
@@ -0,0 +1,367 @@
...
+            if (chkUsePortrait[i] == null) {
+                continue;
+            }
+            if (i == chkUsePortrait.length - 1) {
+                for (PersonnelRole role : PersonnelRole.getCivilianRoles()) {
+                    if (role.ordinal() < model.usePortraitForRole.length) {
+                        model.usePortraitForRole[role.ordinal()] = chkUsePortrait[i].isSelected();
+                    }
+                }
+                continue;
```
IllianiBird: Same here

MekHQ/src/mekhq/gui/campaignOptions/contents/PersonnelMarketPage.java
```diff
@@ -0,0 +1,338 @@
...
+    @Deprecated(since = "0.50.06")
+    private JPanel pnlRemovalTargets;
+    @Deprecated(since = "0.50.06")
+    private JLabel lblPersonnelMarketDylansWeight;
+    @Deprecated(since = "0.50.06")
+    private JSpinner spnPersonnelMarketDylansWeight;
+    @Deprecated(since = "0.50.06")
+    private Map<SkillLevel, JLabel> lblPersonnelMarketRandomRemovalTargets;
+    @Deprecated(since = "0.50.06")
+    private Map<SkillLevel, JSpinner> spnPersonnelMarketRandomRemovalTargets;
```
IllianiBird: As mentioned before, let's just get rid of these deprecated values. I won't comment on this again, just assume I did for each instance.

MekHQ/src/mekhq/gui/campaignOptions/contents/SkillsPages.java
```diff
@@ -0,0 +1,889 @@
...
+    /**
+     * Builds a short, human-readable summary of a skill's experience milestones (the level at which a character is
+     * considered Green, Regular, Veteran, Elite, Heroic, and Legendary).
+     *
+     * @param configuration the configuration to summarize
+     *
+     * @return a compact milestone summary string
+     */
+    private static String buildProgressionSummary(SkillConfiguration configuration) {
+        return String.format("G %d  \u00B7  R %d  \u00B7  V %d  \u00B7  E %d  \u00B7  H %d  \u00B7  L %d",
```
IllianiBird: This looks like a hardcoded player-facing String. If so, it should be in a resource bundle.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsContentHost.java
```diff
@@ -0,0 +1,228 @@
...
+        setName("campaignOptionsContentHost");
+
+        helpPanel = new CampaignOptionsHelpPanel();
+        tipConsumer = helpPanel::setHelpText;
+        CampaignOptionsUtilities.setTipTextConsumer(tipConsumer);
+
+        contentPanel = new CampaignOptionsContentPanel();
+        contentPanel.setName("campaignOptionsContentPanel");
+
+        contentScrollPane = new JScrollPane(contentPanel,
```
IllianiBird: This should probably be FastJScrollPane. Generally speaking there are no reasons to use a vanilla JScrollPane over FastJScrollPane. I imagine there are other instances of this through the PR, this was just the one I noticed.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsContentHost.java
```diff
@@ -0,0 +1,228 @@
...
+     * box are left untouched so their tooltips remain the only hover help.
+     *
+     * @param component the content subtree to process
+     */
+    private static void clearRedundantOptionToolTips(Component component) {
+        if (component instanceof JComponent jComponent
+                  && (component instanceof CampaignOptionsCheckBox
+                  || component instanceof CampaignOptionsSpinner
+                  || component instanceof CampaignOptionsLabel
+                  || component instanceof CampaignOptionsTextField)) {
```
IllianiBird: I generally prefer for chained conditions to be broken out into booleans and a single boolean passed in. This can also be handled by a utility method if you wanna be really clean.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
```diff
@@ -0,0 +1,370 @@
...
+        setName("campaignOptionsNavigationPanel");
+        // Match the content panel's scroll-pane border instead of a TitledBorder. A TitledBorder reserves vertical
+        // space for its caption and draws its top line at the caption's mid-height, which left the navigation frame
+        // sitting lower than the content frame on the right. Using the same border keeps both top edges aligned.
+        Border frameBorder = UIManager.getBorder("ScrollPane.border");
+        if (frameBorder == null) {
+            frameBorder = BorderFactory.createLineBorder(UIManager.getColor("Component.borderColor"));
+        }
+        setBorder(BorderFactory.createCompoundBorder(frameBorder,
+              BorderFactory.createEmptyBorder(6, 6, 6, 6)));
```
IllianiBird: Should these be scaled for gui?

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
```diff
@@ -0,0 +1,370 @@
...
+        add(navigationScrollPane, BorderLayout.CENTER);
+
+        buildNavigationTree("");
+    }
+
+    private void showLegend(JComponent anchor) {
+        JLabel legend = new JLabel("<html><body>"
+              + getTextAt(getCampaignOptionsResourceBundle(), "lblGeneralIconLegend.text")
+              + "</body></html>");
+        legend.setBorder(BorderFactory.createEmptyBorder(8, 8, 8, 8));
```
IllianiBird: And here, I'll not comment any more on this, so make sure to do a quick sweep if the feedback is valid.

MekHQ/src/mekhq/gui/dialog/EditSkillPreRequisiteDialog.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2009 Jay Lawson (jaylawson39 at yahoo.com). All rights reserved.
- * Copyright (C) 2014-2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2014-2026 The MegaMek Team. All Rights Reserved.
```
IllianiBird, line 3: I haven't been checking legal years, so please do one last double-check you caught them all

MekHQ/src/mekhq/gui/dialog/EditSkillPreRequisiteDialog.java
```diff
@@ -109,8 +109,8 @@ private void initComponents() {
             panMain.add(choiceLvl);
         }
 
-        JPanel panButtons = new JPanel(new GridLayout(0, 2));
-        btnOK.setText("Done");
+        JPanel panButtons = new JPanel();
+        btnOK.setText("OK");
```
IllianiBird: player-facing hardcoded String. I believe we can use getText("ok.text")

IllianiBird (changes requested):
With a PR of this magnitude it's not going to be possible for any one reviewer to find every issue. However, I did go through every class and change and here's the handful of problems I found.

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
VicenteCartas: I can tackle that after this PR.

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHeaderPanel.java
VicenteCartas: @IllianiBird 

Should I remove also this?

/**
     * Creates the UI panel for the Legacy AtB configuration.
     * <p>
     * This section configures opponent force generation, scenario generation
     * probabilities, and customization of battle
     * intensities for "Against the Bot" campaigns.
     * </p>
     *
     * @return a {@link JPanel} containing all Legacy AtB settings.
     */
    @Deprecated(since = "0.51.0", forRemoval = true)
    public @Nonnull JPanel createLegacyPage() {

VicenteCartas: I'm going to remove it, if you disagree let me know :)

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsPagePanel.java
VicenteCartas: I'll centralize this.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsContentHost.java
VicenteCartas: @IllianiBird I'm sorry I don't understand this, like do this:

boolean a = condition1;
boolean b = condition2;
boolean 3 = condition1 && condition2;
if (boolean3) {
   ...
}

?

MekHQ/src/mekhq/gui/campaignOptions/contents/PersonnelMarketPage.java
VicenteCartas: On it.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsContentHost.java
VicenteCartas: There's a few more on the PR, I'll update all.

MekHQ/src/mekhq/gui/dialog/EditSkillPreRequisiteDialog.java
VicenteCartas: Double checked, all good.

MekHQ/src/mekhq/gui/dialog/EditSkillPreRequisiteDialog.java
VicenteCartas: Double checked, all good.

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsCheckBox.java
VicenteCartas: Mmm, I can use getFormattedTextAt, but I don't think the strings we are formatting are something related to localization. The only localizable thing is the label name, and that's localized, the rest is formatting that would be language independent.

Happy to change it if you disagree.

MekHQ/src/mekhq/gui/campaignOptions/contents/DeathPage.java
VicenteCartas: Boxed type :) I'll change this.

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
VicenteCartas: Moved to constants in SkillPrerequisite.java, so both pieces of code stay in sync and can't break.

MekHQ/src/mekhq/gui/campaignOptions/contents/AbilitiesPages.java
VicenteCartas: Lambdas as private functions now.

MekHQ/src/mekhq/gui/campaignOptions/contents/SkillsPages.java
VicenteCartas: Yes they are, moved to a bundle.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
VicenteCartas: Fixes

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
VicenteCartas: Fixed

MekHQ/src/mekhq/gui/campaignOptions/contents/BiographyGeneralPage.java
VicenteCartas: This should be a follow up, so we can touch all the code that is actually using logos (like also ImmersiveDialog)

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsContentHost.java
IllianiBird: Correct, yeah

IllianiBird: It takes up slightly more room but makes it much easier to follow the logic and vastly less likely for a logical error to enter the conditional

MekHQ/src/mekhq/gui/campaignOptions/contents/BiographyGeneralPage.java
IllianiBird: Oh for sure, not something here, was just a ponderance

MekHQ/src/mekhq/gui/campaignOptions/contents/NameAndPortraitGenerationPage.java
VicenteCartas: This work is O(n), but removed the nested loops.

IllianiBird (approved):

PR #9276, opened 2026-06-07
Fixed Parts in Use Failing to Correctly Set Minimum Stock Percentage for Missing Parts

IllianiBird:
If a player salvaged a number any missing parts would be placed in the Parts in Use dialog with 0% minimum stock. This would only happen if that part had not been present on another unit earlier in the campaign.

The issue was caused by an inheritance issue which this PR fixes.


AlexOldenburger:
OK, I see what you did here. Yes, that should fix it.

So, LGTM

HammerGS (approved):

PR #9309, opened 2026-06-10
Disposable infantry weapons in campaigns: parts, acquisition, refit, reload

HammerGS:
## Summary
  Campaign support for conventional infantry / BattleArmor disposable weapons (TO:AR p.106): buy, sell, value, refit, and reload them as parts, modeled one per trooper.

  ## Changes
  1. `InfantryDisposableWeaponPart` / `MissingInfantryDisposableWeaponPart`: one part per trooper (like primary/secondary weapons), so a platoon is valued / refit / bought-and-sold as that many individual weapons. Spent state is driven read-only by the platoon's single shared fireable mount (no entity mutation during refit diffing); reload is a platoon-level action that draws one spare per trooper; warehouse matching is by weapon type so existing stock is consumed first.
  2. `Unit.initializeParts`: one disposable part per trooper for CI and BA.
  3. **Bulk acquisition for infantry weapons and ammo only**: a single successful acquisition roll fills the whole platoon's order (`IAcquisitionWork.isBulkAcquisition`, `MissingEquipmentPart`, `InfantryAmmoStorage`, `Campaign.acquireEquipment`) instead of one roll per weapon. All other parts are unchanged.
  4. Refit: preserve a user/MegaMekLab-set Model name through the refit; refresh the Begin Refit button on every change across all MekLab panels.
  5. `Campaign`: automatic-success reloads for disposables and infantry field-gun ammo.

  ## Files Changed
  - `campaign/parts/equipment/InfantryDisposableWeaponPart.java` (new)
  - `campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java` (new)
  - `campaign/parts/equipment/MissingEquipmentPart.java`
  - `campaign/parts/InfantryAmmoStorage.java`
  - `campaign/work/IAcquisitionWork.java`
  - `campaign/Campaign.java`
  - `campaign/unit/Unit.java`
  - `campaign/parts/Refit.java`
  - `gui/MekLabTab.java`
  - `unittests/.../InfantryDisposableWeaponPartTest.java` (new)

  ## Testing
  New `InfantryDisposableWeaponPartTest` (6 tests) plus the equipment-parts package, all green.
  Manual: buy/refit/reload a disposable platoon; confirm bulk-order behaviour, the Begin Refit button
  updating on changes, and the typed Model name surviving the refit.

  ## Dependencies
  Requires the MegaMek PR (`Fixes MegaMek/megamek#148`) — uses the `ConvInfantry` disposable API and
  `WeaponMounted` disposable/fired state. **Merge after megamek.** Independent of the MegaMekLab PR.

  Part of [MegaMek/megamek#148](https://github.com/MegaMek/megamek/pull/8326)


MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * A single one-shot Disposable Weapon (TO:AR p.106) carried by one trooper. A platoon/squad gets one of these per
+ * trooper (like the primary/secondary {@link InfantryWeaponPart}), so the loadout is valued, refit, and bought/sold as
+ * that many individual weapons. The whole platoon fires its disposables together in one volley; this part tracks
+ * whether that has happened (and not yet been reloaded) via the {@code spent} flag, which is driven by the platoon's
+ * single fireable Disposable Weapon mount ({@link WeaponMounted#isFired()}).
+ */
+public class InfantryDisposableWeaponPart extends InfantryWeaponPart {
+    private static final MMLogger LOGGER = MMLogger.create(InfantryDisposableWeaponPart.class);
```
IllianiBird, line 57: This is the wrong logger method. It should be...

`private static final String LOGGER = [file address]`

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+    private static final String DISPOSABLE_SUFFIX = " (Disposable)";
+
+    private boolean spent;
+
+    public InfantryDisposableWeaponPart() {
+        this(0, null, -1, null);
+    }
+
+    public InfantryDisposableWeaponPart(int tonnage, EquipmentType et, int equipNum, Campaign c) {
```
IllianiBird: No single letter, abbreviated, or truncated param or variable names please. I'm not going to go through and mark everyone, as there's a lot. Could you give the PR a pass and update wherever you find them? :)

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+        clone.spent = spent;
+        return clone;
+    }
+
+    /** @return the platoon's single fireable Disposable Weapon mount of this weapon type, or null */
+    private @Nullable WeaponMounted findDisposableMount() {
+        if ((unit == null) || (unit.getEntity() == null) || (type == null)) {
+            return null;
+        }
+        return unit.getEntity().getWeaponList().stream()
```
IllianiBird: This is a long stream to be having as the return value. Let's use a traditional loop and then return first first valid value (if appropriate)

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+              .filter(WeaponMounted::isDisposableWeapon)
+              .filter(weaponMounted -> (weaponMounted.getType() != null)
+                    && type.getInternalName().equals(weaponMounted.getType().getInternalName()))
+              .findFirst()
```
IllianiBird: I mentioned this in my mm review, but is it possible a unit might have multiple disposable weapons? If so, this will only return the first found.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+              .orElse(null);
+    }
+
+    @Override
+    public void updateConditionFromEntity(boolean checkForDestruction) {
+        super.updateConditionFromEntity(checkForDestruction);
+        // Read-only: the whole platoon shares one fireable mount, so every per-trooper disposable part is spent iff
```
IllianiBird: `iff` -> `if`

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+    public boolean needsFixing() {
+        return spent || super.needsFixing();
+    }
+
+    /** @return the number of troopers carrying this disposable weapon (i.e. matching parts on the unit). */
+    private int trooperCount() {
+        if (unit == null) {
+            return 1;
+        }
+        return (int) Math.max(1, unit.getParts().stream()
```
IllianiBird: A stream is really unnecessary here. Streams are powerful, but they can also be expensive. We should be using traditional loops when working with sample sizes of less than a thousand objects.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+              .count());
+    }
+
+    /** @return total spare weapons of this disposable type available in the warehouse. */
+    private int availableReloadStock() {
+        final String weaponInternalName = (type == null) ? null : type.getInternalName();
+        if (weaponInternalName == null) {
+            return 0;
+        }
+        return getWarehouse().getSpareParts().stream()
```
IllianiBird, line 176: A stream is good here, as the warehouse can contain very large volumes of objects.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+    private @Nullable Part findSpareReload() {
+        final String weaponInternalName = (type == null) ? null : type.getInternalName();
+        if (weaponInternalName == null) {
+            return null;
+        }
+        return getWarehouse().findSparePart(spare -> (spare instanceof EquipmentPart equipmentSpare)
+              && (spare.getId() != getId())
+              && (spare.getQuantity() > 0)
+              && (equipmentSpare.getType() != null)
+              && weaponInternalName.equals(equipmentSpare.getType().getInternalName()));
```
IllianiBird: I'm not a fan of chaining conditionals like this. It makes it hard to read the return value at a glance. Please break these out into their own booleans:

boolean condition1 = [conditional]
boolean condition2= [conditional]
boolean condition3 = [conditional]

return condition1 && condition2 && condition3

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+            mount.setFired(false);
+        }
+        spent = false;
+    }
+
+    @Override
+    public @Nullable String checkFixable() {
+        WeaponMounted mount = findDisposableMount();
+        if (spent && (mount != null) && mount.isFired() && (availableReloadStock() < trooperCount())) {
+            return getName() + ": not enough spares in stock to reload the platoon";
```
IllianiBird: Player-facing Strings should be in a resource bundle, not hardcoded. Not going to mark all the instances, just assume this comment applies to them all.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+        return super.checkFixable();
+    }
+
+    @Override
+    public int getBaseTime() {
+        return spent ? 15 : super.getBaseTime();
```
IllianiBird: Can we define this magic number as a class-level static? I know we're only using it once, but it improved code readability and maintainability

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+        writeToXMLEnd(pw, indent);
+    }
+
+    @Override
+    protected void loadFieldsFromXmlNode(Node wn) {
+        NodeList nl = wn.getChildNodes();
+        for (int x = 0; x < nl.getLength(); x++) {
+            Node wn2 = nl.item(x);
+            try {
+                if (wn2.getNodeName().equalsIgnoreCase("equipmentNum")) {
```
IllianiBird: We're reusing `wn2.getNodeName()` a lot, let's cache the value and reuse that reference instead.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,253 @@
...
+                    equipmentNum = Integer.parseInt(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("typeName")) {
+                    typeName = wn2.getTextContent();
+                } else if (wn2.getNodeName().equalsIgnoreCase("equipTonnage")) {
+                    equipTonnage = Double.parseDouble(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("spent")) {
+                    spent = Boolean.parseBoolean(wn2.getTextContent().trim());
+                }
+            } catch (Exception e) {
+                LOGGER.error("", e);
```
IllianiBird: If you cache the value of `wn2.getNodeName()` you can actually pass it in here, too:

`LOGGER.error(nodeName, e);`

This means when we get an error we know _exactly_ what value caused the problem

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,117 @@
...
+import org.w3c.dom.NodeList;
+
+/**
+ * The "missing" / needs-acquisition counterpart of {@link InfantryDisposableWeaponPart}. It is the acquisition work for
+ * buying a spare Disposable Weapon (TO:AR p.106). One is created per missing trooper's disposable, so a refit shops for
+ * one per trooper. It accepts a plain {@link EquipmentPart} of the same weapon type as a replacement, because the parts
+ * store stocks a disposable weapon as a generic EquipmentPart - so existing warehouse stock is consumed first.
+ */
+public class MissingInfantryDisposableWeaponPart extends MissingEquipmentPart {
+    private static final MMLogger LOGGER = MMLogger.create(MissingInfantryDisposableWeaponPart.class);
```
IllianiBird, line 53: This isn't the correct way to create a resource bundle in mhq, see prior comment.

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final String DISPOSABLE_SUFFIX = " (Disposable)";
```
IllianiBird: Public facing Strings should not be hardcoded

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public MissingInfantryDisposableWeaponPart() {
+        this(0, null, -1, null);
+    }
+
+    public MissingInfantryDisposableWeaponPart(int tonnage, EquipmentType et, int equipNum, Campaign c) {
```
IllianiBird: single letter and abbreviated param names

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public boolean isAcceptableReplacement(Part part, boolean refit) {
+        // Match any spare weapon part of the same weapon type - the parts store stocks a disposable weapon as a plain
+        // EquipmentPart, so existing warehouse stock (and freshly bought spares) satisfy the refit/replacement.
+        return (part instanceof EquipmentPart equipmentPart)
+              && (type != null)
+              && (equipmentPart.getType() != null)
+              && type.getInternalName().equals(equipmentPart.getType().getInternalName());
```
IllianiBird: While this is less egregious that the chained conditionals I cited earlier, the code would still be more readable if we break out each condition into its own boolean value

IllianiBird: As a more serious issue, String comparison may cause problems. I'm not wholly sure internal name is unique across different types of equipment. We should test Clan equipment and IS equipment and make sure this conditional still returns true.

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
```diff
@@ -0,0 +1,117 @@
...
+            try {
+                if (wn2.getNodeName().equalsIgnoreCase("equipmentNum")) {
+                    equipmentNum = Integer.parseInt(wn2.getTextContent());
+                } else if (wn2.getNodeName().equalsIgnoreCase("typeName")) {
+                    typeName = wn2.getTextContent();
+                } else if (wn2.getNodeName().equalsIgnoreCase("equipTonnage")) {
+                    equipTonnage = Double.parseDouble(wn2.getTextContent());
+                }
+            } catch (Exception e) {
+                LOGGER.error("", e);
```
IllianiBird: Similar comment to earlier about caching `wn2.getNodeName()`

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -3818,6 +3825,19 @@ private boolean acquireEquipment(IAcquisitionWork acquisition, Person person, Pl
...
+            // platoon's quantity too (while it can be paid for), counting as a single acquisition for the daily limit.
+            if (acquisition.isBulkAcquisition()) {
+                int bulkDelivered = 1;
+                while ((acquisition.getQuantity() > 1) && canPayFor(acquisition)) {
+                    acquisition.find(transitDays, valueChange);
+                    acquisition.decrementQuantity();
+                    bulkDelivered++;
+                }
+                if (bulkDelivered > 1) {
+                    report += " (bulk order of " + bulkDelivered + ")";
```
IllianiBird: Hardcoded String. Not going to mark them all, some in this class are pre-existing issues others are not. I'm going to cede to your best judgment.

IllianiBird (changes requested):
Some relatively minor feedback. I am very nervous about this PR (and the MegaMek portion). We're touching on a lot of classes and there is a reasonably high chance of something not working, especially with mhq's parts code.

Finally, this PR is going to conflict terribly with Psi's bases PR. Please reach out to them and negotiate whose gets merged first.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: The MMLogger.create(Class) line matches every other part in mekhq.campaign.parts (Armor, AmmoStorage, BattleArmorSuit, InfantryAmmoStorage, ...), so I kept it. Reading this together with your resource-bundle comment on the missing part, I took the underlying ask to be the RESOURCE_BUNDLE pattern for the hardcoded strings - that is in place as of 1b47ff0082 (RESOURCE_BUNDLE = mekhq.resources.Parts + getFormattedTextAt, same as Part.java). Happy to adjust if you meant something else here.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Swept the PR in 1b47ff0082: et -> equipmentType, c -> campaign, pw -> printWriter, wn/wn2/nl/x -> node/childNode/childNodes/index, e -> exception, and disposableW -> disposableWeaponPart in Unit.java. I left the pre-existing primaryW/secondaryW in Unit.java alone to keep the diff scoped - can rename those too if you want them gone.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Converted to a traditional loop in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Same answer as the mm thread: a CI platoon carries at most one Disposable Weapon (equipDisposableWeapon replaces any existing disposable mount), and this lookup additionally filters by this part's weapon type, so the first match is the single mount the whole platoon shares. Documented in the method javadoc in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Converted to a traditional loop in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Kept the stream here per this comment.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Broken out into named booleans in 1b47ff0082 - here (now an isUsableReloadSpare helper), in checkFixable, and in the missing part's isAcceptableReplacement.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Moved to Parts.properties in 1b47ff0082: the (Disposable) name suffix, the not-enough-spares message and the spent details string all come from the bundle now.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Extracted as RELOAD_BASE_TIME_MINUTES in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Cached in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: Done in 1b47ff0082 - the cached nodeName is passed to LOGGER.error in both XML loaders.

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
HammerGS: Switched to the standard pattern in 1b47ff0082: InfantryDisposableWeaponPart declares RESOURCE_BUNDLE = mekhq.resources.Parts and exposes a shared buildDisposableName() helper that this class reuses, so the missing part has no hardcoded display strings left.

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
HammerGS: Moved to Parts.properties in 1b47ff0082 (shared with the main part via buildDisposableName).

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
HammerGS: Renamed in 1b47ff0082 (equipmentType, campaign).

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
HammerGS: Broken out into named booleans in 1b47ff0082.

HammerGS: Internal name is the unique key equipment registers under - EquipmentType.get() resolves by it - and IS and Clan variants register under distinct internal names (e.g. ISERMediumLaser vs CLERMediumLaser), so they cannot cross-match. To pin that down I added internalNameMatchingSeparatesClanAndInnerSphere to InfantryDisposableWeaponPartTest in 1b47ff0082: the exact IS type is accepted as a replacement, the Clan variant is rejected. Matching by internal name (rather than object identity or strict isSamePartType) is deliberate - it survives save/load and the equipment-singleton restore, and it is the same approach Quartermaster.findSpareAmmo takes.

MekHQ/src/mekhq/campaign/parts/equipment/MissingInfantryDisposableWeaponPart.java
HammerGS: Cached here too in 1b47ff0082.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
HammerGS: That one was intentional (if-and-only-if), but you are right that it reads as a typo - reworded to "exactly when" in 1b47ff0082.

MekHQ/src/mekhq/campaign/Campaign.java
HammerGS: The new bulk-order report string moved to Campaign.properties (acquireEquipment.bulkOrder.format) in 1b47ff0082. I left the two TargetRoll reason strings ("disposable weapon reload" / "infantry field gun reload") hardcoded since every neighbouring reason string in getTargetFor is - happy to move those too if you would prefer.

MekHQ/src/mekhq/campaign/parts/equipment/InfantryDisposableWeaponPart.java
IllianiBird: It would be nice to remove them all, but non-essential

MekHQ/src/mekhq/campaign/Campaign.java
IllianiBird: Would be best to move, but non-essential

IllianiBird (approved):

PR #9310, opened 2026-06-11
RFE: #1099 - Planetary Bases

### LINKED ISSUE #1099: Feature Request: Planetary Base Hangar

psikomonkie:
Closes #1099

The player is now able to send people, units, and parts to “Bases”. These objects can then be sent back to the main force or to a different base. Things traveling to bases use the same logic as the main force does. A future enhancement to this is to charge the player for transportation to their bases, but also allowing the player to provide their own transport.   Anyway, this PR forces us handle multiple `Warehouse` and `Hangar` instances. We already had some support for `Personnel` from the education changes. Different bases and the main force have different warehouses or hangars. Fortunately, it should be easy to get - any `ILocation` should have methods `getWarehouse`, `getHangar`, or `getPersonnel` that will return the appropriate warehouse, hangar, or personnel for that part, unit, or person. 

## Features
- Player Bases
        - Has a `Warehouse`, `Hangar`, and `Personnel` - can “contain” `Part`s, `Unit`s, and `Person`s. 
        - Has two free-text fields - Display Name, and Display Type
            - Display Name is the base’s name that’ll be shown in the UI
            - Display Type (optional) will “nest” bases under submenus / let the player categorize their bases
- Personnel, Hangar, and Warehouse Tab UI Updates
    - Location View
        - Personnel and Hangar menus now have a “View” option for the table of “Location” that’ll show the current system, planet, and location name, and the destination system, planet, and location name.
            - Warehouse doesn’t support alternate views - currently you can only change the active location to filter the list
    - “Send to…” Menu Option for `Person`, `Part`, and `Unit`
        - Dispatches `Person`, `Part`, or `Unit` to the target
        - Includes “New Base” which opens a prompt for creating a new base
            - Display Name, Display Type, System (searchable), Planet (searchable)
- Repair Tab
    - Only shows units at the set location
    - Techs and parts are filtered to only show parts or techs relevant to the unit’s location
- Active Location
    - Visible and changeable via a dropdown in the bottom-left corner
    - Options for Campaign (main force) and bases
        - Defaults to “All (Main Force)” - Shows units, people, and parts from all locations, but will still default any new purchases to go to the main force
        - Additional location option for “Main Force” (hides things at other locations) and then each Player Base
        - Specific options will filter the Personnel, Hangar, Repair Bay, and Warehouse tabs to exclude things that are not at that location

## Out of Scope
- Base Configuration / Management
    - I know I really should include this but I think this PR is big enough, and another GUI to support this will just make reviewing it harder. I promise this'll be one of the first enhancements pushed, I have Big Plans for bases once the basis are in.
- TOE / Formation
    - This was a consideration in the design, but not implemented as part of the MVP. I intend to expand the “Send To…” functionality to support the TOE screen & formations too, but not as part of this PR. It’s complicated enough!
- $$$ and/or Providing DS/JS
    - We don’t charge the player for dispatching things. We should be doing that, but only if the player isn’t providing their own transport. This becomes a bigger lift and I think will be a PR on its own

## TODO
- [x] Final Review


MekHQ/resources/mekhq/resources/PersonnelStatus.properties
```diff
@@ -106,6 +106,10 @@ CAMP_FOLLOWER.label=Camp Follower
 CAMP_FOLLOWER.tooltip=Is just along for the ride.
 CAMP_FOLLOWER.report=%s is no longer {0}<b>Employed</b>{1}.
 CAMP_FOLLOWER.log=Was marked as no longer employed by the unit.
+AWAY_FROM_MAIN_FORCE.label=Away from Main Force
+AWAY_FROM_MAIN_FORCE.tooltip=They are stationed at a player base or traveling to one.
+AWAY_FROM_MAIN_FORCE.report=%s is {0}<b>Away from the Main Force</b>{1}.
```
IllianiBird: Looks like we're mixing placeholder keys here, should `%s` be `{0}`? Will need us to increment to increment `0` and `1`, too.

MekHQ/src/mekhq/campaign/base/AbstractBase.java
```diff
@@ -0,0 +1,228 @@
...
+    private final Warehouse baseWarehouse = new Warehouse();
+    private final Hangar baseHangar = new Hangar();
+
+    /**
+     * Creates a new base anchored under {@code parentLocation}.
+     *
+     * @param parentLocation the location this base lives under; must not be {@code null}
+     */
+    protected AbstractBase(ILocation parentLocation) {
+        Objects.requireNonNull(parentLocation, "parentLocation must not be null");
```
IllianiBird, line 77: If `parentLocation` is `null` we're going to hit an error here. Is there value in adding some kind of handler, or is this a case where we want it to fail loudly?

MekHQ/src/mekhq/campaign/base/AbstractBase.java
```diff
@@ -0,0 +1,228 @@
...
+    public @Nullable String getPlanetId() {
+        return planetId;
+    }
+
+    public void setPlanetId(@Nullable String planetId) {
+        this.planetId = planetId;
+    }
+
+    /**
+     * Returns the immediate parent of this base in the location tree, or {@code null} if unparented.
```
IllianiBird, line 162: I foresee players wanting to be able to see this tree if we end up more than one step deep. Might be worth hijacking some code from the family tree class. Your call, but if you think it's necessary I should be able to knock something out for you - I know you're not a huge fan of gui stuff so lemme know on Discord if this is needed.

MekHQ/src/mekhq/campaign/base/AbstractBase.java
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Reads a single child node into the common base fields. Subclasses call this inside their
+     * XML parsing loop.
+     *
+     * @return {@code true} if the node was consumed
+     */
+    protected static boolean readBaseFieldFromXML(AbstractBase base, Node wn2) {
+        switch (wn2.getNodeName().toLowerCase()) {
+            case "id" -> {
+                base.id = UUID.fromString(wn2.getTextContent().trim());
```
IllianiBird: This can error out if the text content isn't a valid UUID.

I'd prefer if we enclosed this switch in a try-catch so we don't end up bricking saves if something goes horribly wrong.

MekHQ/src/mekhq/campaign/base/AbstractBase.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Dispatches XML deserialization to the correct {@link AbstractBase} subclass based on the
+     * element name of {@code wn}.
+     *
+     * @return the deserialized base, or {@code null} if the node name is unrecognized
+     */
+    public static @Nullable AbstractBase generateInstanceFromXML(Node wn, Campaign campaign,
+          megamek.Version version) {
+        return switch (wn.getNodeName().toLowerCase()) {
```
IllianiBird, line 229: Right now this isn't needed by the code we have here. So measure this note accordingly. However, it may be worth wrapping this in a try-catch, too. Again, the code right now doesn't warrant it, I'm just thinking it would be good to add it now so it's future proofed.

MekHQ/src/mekhq/campaign/CampaignNewDayManager.java
```diff
@@ -1358,6 +1370,17 @@ public void processNewDayUnits() {
             }
 
             if (null != tech) {
+                // If the tech has moved to a different location since the assignment was made,
+                // cancel it and notify the player rather than silently failing.
+                ILocation repairTarget = (part.getUnit() != null) ? part.getUnit() : part;
+                if (!LocationUtils.areSameEffectiveLocation(tech, repairTarget)) {
+                    campaign.addReport(TECHNICAL, String.format(
```
psikomonkie: Don't use string.format

MekHQ/src/mekhq/campaign/base/PlayerBase.java
```diff
@@ -0,0 +1,214 @@
...
+        return result;
+    }
+
+    /**
+     * Returns the {@link FixedLocation} that anchors this base, navigating up the location tree
+     * from this base's node.
+     */
+    public @Nullable FixedLocation getFixedLocation() {
+        ILocation parent = getParent();
+        return parent instanceof FixedLocation fl ? fl : null;
```
IllianiBird: `fl` -> `fixedLocation`. I am trying to be better about letting sensible abbreviations through, but seeing this bothers me.

I'd even be ok if it were just `FixedLocation instance ? instance : null`, because that takes fewer mental parses than `fl` where I need to reread the code to confirm what `fl` means.

MekHQ/src/mekhq/campaign/base/PlayerBase.java
```diff
@@ -0,0 +1,214 @@
...
+    public void writeToXML(PrintWriter pw, int indent) {
+        MHQXMLUtility.writeSimpleXMLOpenTag(pw, indent++, "playerBase");
+        writeBaseFieldsToXML(pw, indent);
+        if (getCurrentLocation() != null && getCurrentLocation().getCurrentSystem() != null) {
+            MHQXMLUtility.writeSimpleXMLTag(pw, indent, "systemId",
+                  getCurrentLocation().getCurrentSystem().getId());
+        }
+        // Persons who have arrived live under basePersonnel.
+        for (LocationNode child : getBasePersonnel().getLocationNode().getChildren()) {
+            if (child.getLocatable() instanceof Person p) {
```
IllianiBird: `person` please

MekHQ/src/mekhq/campaign/base/PlayerBase.java
```diff
@@ -0,0 +1,214 @@
...
+        MHQXMLUtility.writeSimpleXMLCloseTag(pw, --indent, "playerBase");
+    }
+
+    public static @Nullable PlayerBase generateInstanceFromXML(Node wn, Campaign campaign,
+          Version version) {
+        try {
+            String systemId = null;
+            PlayerBase base = new PlayerBase();
+
+            NodeList nl = wn.getChildNodes();
```
IllianiBird: `nodeList`

MekHQ/src/mekhq/campaign/base/PlayerBase.java
```diff
@@ -0,0 +1,214 @@
...
+            for (int x = 0; x < nl.getLength(); x++) {
+                Node wn2 = nl.item(x);
+                if (wn2.getNodeType() != Node.ELEMENT_NODE) {
+                    continue;
+                }
+                if (wn2.getNodeName().equalsIgnoreCase("systemId")) {
```
IllianiBird: Can we cache the return of `wn2.getNodeName()` as we're using that value a lot.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -179,7 +179,7 @@ public Money getMonthlyFoodAndHousingExpenses() {
         int officerFoodUsage = 0;
 
         // Determine housing and food requirements
-        List<Person> personnel = new ArrayList<>(campaign().getPersonnel());
+        List<Person> personnel = new ArrayList<>(campaign().getAllPersonnel());
```
IllianiBird, line 182: `getPersonnel` gets the personnel at the local location?
`getAllPersonnel` gets everybody everywhere?

If so, may I recommend a follow-up PR that does one thing and one thing only:

rename `getPersonnel` to `getLocalPersonnel` and `getAllPersonnel` to `getGlobalPersonnel`. I think that change will make it clearer which `getPersonnel` method to use in any give use-case.

Renaming a core method like `getPersonnel` is going to hit a bazillion classes, which is why I think leave both methods as they are in this PR. We merge it, and then wack up a follow-up to rename them.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -526,6 +529,37 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
...
+
+                if ((unit.getFormationId() > 0) && (campaign.getFormation(unit.getFormationId()) == null)) {
+                    unit.setFormationId(FORMATION_NONE);
+                }
+
+                final EquipmentUnscrambler unscrambler = EquipmentUnscrambler.create(unit);
+                final EquipmentUnscramblerResult result = unscrambler.unscramble();
+                if (!result.succeeded()) {
+                    LOGGER.warn(result.getMessage());
+                }
```
IllianiBird: Let's move the contents of this lambda into a utility method, makes the code more readable. My general preference is if a lambda is more than a couple lines it should be calling a utility class.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -753,14 +787,17 @@ public Campaign parse() throws CampaignXmlParseException, NullEntityException {
             correctSexualPreferencesForCurrentSpouse(campaign.getAllPersonnel());
         }
 
-        // Reconnect all persons to the main-force personnel node. Persons whose location was
-        // serialized inside a travel or campus node will be re-parented during reconnectChildren.
+        // Reconnect persons to the main-force personnel node. Skip persons already placed by
+        // processPlayerBaseNodes or reconnectChildren (base / travel / campus persons).
         for (Person person : campaign.getAllPersonnel()) {
-            person.setParent(campaign.getMainForcePersonnel());
+            if (person.getLocationNode().getParent() == null) {
```
IllianiBird, line 785: Is there a risk of `getLocationNode` returning `null`? I'm particularly concerned about personnel predating our new location code.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1433,6 +1470,34 @@ private static void processLocations(Campaign campaign, Node wn) {
         }
     }
 
+    private static void processPlayerBaseNodes(Campaign campaign, Node wn, Version version) {
+        NodeList nl = wn.getChildNodes();
```
IllianiBird: `nodeList`

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1433,6 +1470,34 @@ private static void processLocations(Campaign campaign, Node wn) {
...
+                PlayerBase base = PlayerBase.generateInstanceFromXML(wn2, campaign, version);
+                if (base != null) {
+                    campaign.addPlayerBase(base);
+                    for (UUID personId : base.drainPendingPersonIds()) {
+                        Person person = campaign.getPerson(personId);
+                        if (person != null) {
+                            person.setParent(base.getBasePersonnel());
+                        }
+                    }
+                    for (mekhq.campaign.parts.Part part : base.drainPendingBaseWarehouseParts()) {
```
IllianiBird: Do we need to include the full reference here or can we just put `Part part`?

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1433,6 +1470,34 @@ private static void processLocations(Campaign campaign, Node wn) {
...
+                        LocationNode.LocationManager.setLocation(part, base.getBaseWarehouse());
+                    }
+                    for (mekhq.campaign.unit.Unit unit : base.drainPendingBaseHangarUnits()) {
```
IllianiBird: Same as last message, but this time `Unit unit`

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1623,16 +1688,63 @@ private static void correctSexualPreferencesForCurrentSpouse(Collection<Person>
     }
 
     /**
-     * Reconnects persons to their location nodes after a save/load.
+     * Reconnects persons, units, and parts to their travel {@link CurrentLocation} nodes after a
+     * save/load.
+     *
+     * <p>During save, each {@link CurrentLocation} writes the UUIDs/IDs of its direct children
```
IllianiBird: Are they `UUID` or `ID`? The latter suggests something like a String, whereas the former is clear. I'd prefer we avoid ambiguity. You might think I'm splitting hairs, I am, but only because your code today is going to become a core function in MekHQ from this point on.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1623,16 +1688,63 @@ private static void correctSexualPreferencesForCurrentSpouse(Collection<Person>
...
+     * ({@link Person}, {@link Unit}, {@link Part}). On load those parent links are lost because
+     * the {@link mekhq.campaign.location.LocationNode} tree is not serialized directly. This pass
```
IllianiBird: We _probably_ don't need the full reference

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1623,16 +1688,63 @@ private static void correctSexualPreferencesForCurrentSpouse(Collection<Person>
...
+                        Person person = campaign.getPerson(personId);
+                        if (person != null && person.getLocationNode().getParent() == null) {
+                            person.setParent(campaign.getMainForcePersonnel());
+                            LOGGER.warn("reconnectPersonsToTravelLocations: person {} had no parent "
+                                  + "(orphaned arrived node); re-homed to main force", personId);
+                        }
+                    }
+                    for (UUID unitId : currentLocation.drainPendingUnitIds()) {
+                        Unit unit = findUnitAnywhere(campaign, unitId);
+                        if (unit != null && unit.getLocationNode().getParent() == null) {
```
IllianiBird, line 1745: Similar to an earlier comment, is there any risk whatsoever of `unit.getLocationNode()` returning `null`?

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1623,16 +1688,63 @@ private static void correctSexualPreferencesForCurrentSpouse(Collection<Person>
...
+                            LocationNode.LocationManager.setLocation(unit, campaign.getHangar());
+                            LOGGER.warn("reconnectPersonsToTravelLocations: unit {} had no parent "
+                                  + "(orphaned arrived node); re-homed to main hangar", unitId);
+                        }
+                    }
+                    for (int partId : currentLocation.drainPendingPartIds()) {
+                        Part part = findPartAnywhere(campaign, partId);
+                        if (part != null && part.getLocationNode().getParent() == null) {
```
IllianiBird, line 1753: Ditto here

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1662,6 +1791,36 @@ private static void reconnectPersonsToTravelLocations(Campaign campaign) {
         }
     }
 
+    /** Searches campaign hangar then all base hangars for a unit by UUID. */
+    private static Unit findUnitAnywhere(Campaign campaign, UUID unitId) {
```
IllianiBird: Needs `@Nullable`

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1662,6 +1791,36 @@ private static void reconnectPersonsToTravelLocations(Campaign campaign) {
...
+            unit = base.getBaseHangar().getUnit(unitId);
+            if (unit != null) {
+                return unit;
+            }
+        }
+        return null;
+    }
+
+    /** Searches campaign warehouse then all base warehouses for a part by ID. */
+    private static Part findPartAnywhere(Campaign campaign, int partId) {
```
IllianiBird: `@Nullable` needed

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1678,7 +1837,7 @@ private static void migrateLegacyEducationTravel(Campaign campaign) {
         int travelMigrated = 0;
         int skipped = 0;
 
-        for (Person person : campaign.getPersonnel()) {
+        for (Person person : List.copyOf(campaign.getPersonnel().values())) {
```
IllianiBird, line 1855: Not seen `List.copyOf()` before, I am absolutely using this in future

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -831,6 +835,11 @@
         this.campaignStartDate = campaignStartDate;
     }
 
+    public PlanetarySystem getCurrentSystem() {
```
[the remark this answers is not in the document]
psikomonkie: fixed

MekHQ/src/mekhq/campaign/location/IPlace.java
```diff
@@ -88,7 +94,74 @@ default Warehouse getWarehouse() {
...
+     * zero by default — {@link mekhq.campaign.Campaign} overrides this to add shopping-list orders.</p>
+     */
+    default PartInventory getPartInventory(Part part) {
+        PartInventory inventory = new PartInventory();
+        Warehouse warehouse = getWarehouse();
+        if (warehouse == null) {
+            return inventory;
+        }
+        int nSupply = 0;
+        int nTransit = 0;
```
IllianiBird: Not sure how I feel about these two variables, I think we can make the names clearer. Right now it's not super clear what they mean, whereas something like `localSupplyCount` and `countInTransit` might work better

MekHQ/src/mekhq/campaign/location/IPlace.java
```diff
@@ -88,7 +94,74 @@ default Warehouse getWarehouse() {
...
+        for (Part p : warehouse.getParts()) {
```
IllianiBird: `Part part` please

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -164,8 +171,27 @@ private static void landFromTravelNodeImpl(CurrentLocation travelNode,
...
-     * {@code destination}, and registers it with the campaign.
+     * Builds a zero-transit "arrived" node parented under {@code destination} and registers it
+     * with the campaign. The node has {@code transitTime = 0}, so {@code isOnPlanet()} is
+     * immediately {@code true} and {@link IPlace#processArrivals} will land all carried items on
+     * its next call.
+     */
+    private static CurrentLocation buildArrivedNode(PlanetarySystem system,
+          ILocation destination,
+          Campaign campaign,
+          String logContext) {
```
IllianiBird: Seems like we have a bit auto formatter gore here

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -203,7 +229,7 @@ public static void dispatchToLocation(Collection<Person> people,
           ILocation destination,
           Campaign campaign) {
 
-        PlanetarySystem destinationSystem = destination.getCurrentSystem();
+        PlanetarySystem destSystem = destination.getCurrentSystem();
```
IllianiBird: `PlanetarySystem destinationSystem` please

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
             PlanetarySystem fromSystem = entry.getKey();
             List<Person> group = entry.getValue();
 
-            if (destinationSystem == null || fromSystem.equals(destinationSystem)) {
-                group.forEach(p -> p.setParent(destination));
+            if (destSystem == null || fromSystem.equals(destSystem)) {
+                if (destination instanceof AbstractBase) {
+                    PlanetarySystem system = destSystem != null ? destSystem : fromSystem;
+                    CurrentLocation arrivedLocation = buildArrivedNode(system, destination, campaign, "dispatchToLocation");
```
IllianiBird: Can we have this magic String converted into a const. I'm _really_ concerned it's going to break if we ever edit the string parsing for `buildArrivedNode`

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
...
+                    group.forEach(p -> p.setParent(arrivedLocation));
+                } else {
+                    group.forEach(p -> p.setParent(destination));
+                }
                 continue;
             }
 
             Optional<CurrentLocation> maybeTravelLocation = buildTravelNode(
-                  fromSystem, destinationSystem, destination, campaign, "dispatchToLocation");
+                  fromSystem, destSystem, destination, campaign, "dispatchToLocation");
```
IllianiBird: I'm not going to comment all of these magic string instances, you get the idea.

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
...
+     * Same-system dispatches (or dispatches with no calculable path) skip the travel node and land the unit
+     * immediately.</p>
+     *
+     * @param units       the units to dispatch; must not be {@code null}
+     * @param destination the target {@link ILocation}; must not be {@code null}
+     * @param campaign    the active campaign; must not be {@code null}
+     */
+    public static void dispatchUnitsToLocation(Collection<Unit> units,
+          ILocation destination,
+          Campaign campaign) {
```
IllianiBird, line 249: gore

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
...
+
+        Hangar arrivalHangar = (destination instanceof AbstractBase base)
+                                     ? base.getBaseHangar()
+                                     : campaign.getHangar();
+
+        PlanetarySystem destSystem = destination.getCurrentSystem();
+
+        Map<PlanetarySystem, List<Unit>> bySystem = units.stream()
+                                                          .collect(Collectors.groupingBy(u -> {
+                                                              PlanetarySystem sys = u.getCurrentSystem();
```
IllianiBird: `planetarySystem` or even better `unitLocationSystem`

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
...
+     * {@code destination}) so that location columns show in-transit status. Same-system dispatches skip the travel
+     * node.</p>
+     *
+     * @param parts       the parts to dispatch; should all be spare parts; must not be {@code null}
+     * @param destination the target {@link ILocation}; must not be {@code null}
+     * @param campaign    the active campaign; must not be {@code null}
+     */
+    public static void dispatchPartsToLocation(Collection<Part> parts,
+          ILocation destination,
+          Campaign campaign) {
```
IllianiBird: gore

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
```diff
@@ -215,19 +241,165 @@ public static void dispatchToLocation(Collection<Person> people,
...
+
+        Warehouse arrivalWarehouse = (destination instanceof AbstractBase base)
+                                           ? base.getBaseWarehouse()
+                                           : campaign.getWarehouse();
+
+        PlanetarySystem destSystem = destination.getCurrentSystem();
+
+        Map<PlanetarySystem, List<Part>> bySystem = parts.stream()
+                                                          .collect(Collectors.groupingBy(p -> {
+                                                              PlanetarySystem sys = p.getCurrentSystem();
```
IllianiBird: `partLocationSystem` or something similar, just not `sys`

MekHQ/src/mekhq/campaign/location/LocationUtils.java
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * <p>Both items must be non-null, neither may currently be in transit, and they must share
+     * the same nearest {@link AbstractBase} ancestor (or both have none, meaning main force).</p>
+     *
+     * @param a first location item
+     * @param b second location item
+     *
+     * @return {@code true} if the two items are co-located
+     */
+    public static boolean areSameEffectiveLocation(@Nullable ILocation a, @Nullable ILocation b) {
```
IllianiBird: `firstLocation` and `secondLocation` please

MekHQ/src/mekhq/campaign/parts/missing/MissingPart.java
```diff
@@ -479,7 +479,7 @@ public void reservePart() {
             if (replacement.getQuantity() > 1) {
                 Part actualReplacement = replacement.clone();
                 actualReplacement.setReservedBy(getTech());
-                campaign.getQuartermaster().addPart(actualReplacement, 0, false);
+                replacement.getWarehouse().addPart(actualReplacement, true);
```
IllianiBird, line 482: This was previously `false`, is the change to `true` intentional?

MekHQ/src/mekhq/campaign/parts/missing/MissingPart.java
```diff
@@ -496,9 +496,8 @@ public void cancelReservation() {
             if (replacement != null) {
                 replacement.setReservedBy(null);
 
-                // Only return the replacement part to the campaign if we have one
                 if (replacement.getQuantity() > 0) {
-                    campaign.getQuartermaster().addPart(replacement, 0, false);
+                    replacement.getWarehouse().addPart(replacement, true);
```
IllianiBird, line 500: Same as last comment

MekHQ/src/mekhq/campaign/parts/PodSpace.java
```diff
@@ -304,6 +305,11 @@ public Person getTech() {
         return tech;
     }
 
+    @Override
+    public LocationNode getLocationNode() {
```
IllianiBird: `@Nullable`

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -9339,4 +9340,21 @@ public void invalidateAdvancedAsTechContribution() {
     public void setAdvancedAsTechContribution(int contribution) {
         advancedAsTechContribution = contribution;
     }
+
+    @Override
+    public boolean setParent(ILocation parent) {
+        ILocation oldParent = getLocationNode().getParent() != null ?
```
IllianiBird: If we're calling `getLocationNode().getParent()` multiple times let's cache it and reuse the return instead of reparising

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -70,7 +70,7 @@ public String getCombatPersonnelDetails() {
         int countRetired = 0;
         Money salary = Money.zero();
 
-        for (Person p : getCampaign().getPersonnel()) {
+        for (Person p : getCampaign().getPersonnel().values()) {
```
IllianiBird: `person`, I know it's not your code, but it's an easy fix so we might as well

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -5954,6 +5993,10 @@ public void setTechOfficer(Person person, boolean useTransfers) {
     public void setTech(Person p) {
```
IllianiBird: While you're here, would you object to renaming this to `person`?

MekHQ/src/mekhq/campaign/work/IWork.java
IllianiBird, line 1: This was the hardest class of all to review. I really struggled :D

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1797,6 +1809,24 @@ public List<AbstractLocation> getLocations() {
         return Collections.unmodifiableList(locations);
     }
 
+    public void addPlayerBase(PlayerBase base) {
```
IllianiBird: Do we wanna put a `@Nullable` tag in front of `PlayerBase`?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1797,6 +1809,24 @@ public List<AbstractLocation> getLocations() {
...
+        if (base == null) {
+            return;
+        }
+        playerBases.add(base);
+    }
+
+    public void removePlayerBase(PlayerBase base) {
```
IllianiBird: And here

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4134,11 +4183,23 @@ public void refit(Refit theRefit) {
...
         if (!(repairable instanceof OmniPod)) {
-            getQuartermaster().addPart(repairable, 0, false);
+            if (targetWarehouse == getWarehouse()) {
+                // Main-force spare: use the Quartermaster for full processing.
+                getQuartermaster().addPart(repairable, 0, false);
+            } else {
+                // Base spare: add directly to the base warehouse and wire the locationNode.
+                repairable.setDaysToArrival(0);
+                repairable.postProcessCampaignAddition();
+                targetWarehouse.addPart(repairable, true);
```
IllianiBird, line 4202: `getQuartermaster().addPart()` also adds the part to the warehouse, will use adding the part to `targetWarehouse` duplicate the part?

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4160,6 +4221,15 @@ public Part fixWarehousePart(Part part, Person tech) {
      * @return a <code>String</code> of the report that summarizes the outcome of the attempt to fix the part
      */
     public String fixPart(IPartWork partWork, Person tech) {
+        // Enforce location constraint: tech must be at the same location as the repair target.
+        ILocation repairTarget = (partWork instanceof Part p && p.getUnit() != null)
+                                       ? p.getUnit() : (ILocation) partWork;
+        if (!LocationUtils.areSameEffectiveLocation(tech, repairTarget)) {
+            String report = tech.getFullName() + " cannot repair " + partWork.getPartName()
+                                  + " — not at the same location.";
```
IllianiBird: Hardcoded String, this should be in a resource bundle

MekHQ/src/mekhq/campaign/CampaignNewDayManager.java
```diff
@@ -439,6 +442,15 @@ public boolean newDay() {
...
 
         processNewDayPersonnel();
+
+        for (AbstractLocation location : new ArrayList<>(campaign.getLocations())) {
+            location.processArrivals(campaign);
+        }
+        for (PlayerBase base : campaign.getPlayerBases()) {
+            base.processArrivals(campaign);
+        }
+        campaign.processArrivals(campaign);
```
IllianiBird: Let's break this out into its own method. `newDay` is already super busy

MekHQ/src/mekhq/campaign/CurrentLocation.java
```diff
@@ -342,6 +360,10 @@ public void writeToXML(final PrintWriter pw, int indent) {
         for (LocationNode child : locationNode.getChildren()) {
             if (child.getLocatable() instanceof mekhq.campaign.personnel.Person person) {
                 MHQXMLUtility.writeSimpleXMLTag(pw, indent, "personId", person.getId().toString());
+            } else if (child.getLocatable() instanceof Unit u) {
```
IllianiBird: `Unit unit`

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -171,17 +171,51 @@ public void addPart(Part part, int transitDays, boolean isBrandNew) {
...
+     * Adds infantry ammo to a specific warehouse.
+     *
+     * @param warehouse      The warehouse to add ammo to.
+     * @param ammoType       The type of ammo to add.
+     * @param infantryWeapon The type of infantry weapon using the ammo.
+     * @param shots          The number of rounds of ammo to add.
+     */
+    public void addAmmo(Warehouse warehouse, AmmoType ammoType, InfantryWeapon infantryWeapon, int shots) {
+        Objects.requireNonNull(ammoType);
+        Objects.requireNonNull(infantryWeapon);
```
IllianiBird, line 217: I'm not sure about how we're handling these. I'd prefer that if we anticipate we're going to hit a `null` value that we include some management. Otherwise we're just going to be throwing errors at the player.

MekHQ/src/mekhq/campaign/Quartermaster.java
```diff
@@ -286,40 +319,54 @@ public int removeCompatibleAmmo(AmmoType ammoType, int shotsNeeded) {
...
+
+    /**
+     * Finds spare ammo of a given type in a specific warehouse, if any.
+     */
+    private @Nullable AmmoStorage findSpareAmmo(Warehouse warehouse, AmmoType ammoType) {
+        return (AmmoStorage) warehouse.findSparePart(part -> isAvailableAsSpareAmmo(part)
+                                                                  &&
+                                                                  ((AmmoStorage) part).isSameAmmoType(ammoType));
+    }
+
```
IllianiBird, line 374: I'm concerned that we're going to get a class cast exception if we ever pass a non-AmmoStorage into here

MekHQ/src/mekhq/gui/adapter/PartsTableMouseAdapter.java
```diff
@@ -522,6 +528,14 @@ protected Optional<JPopupMenu> createPopupMenu() {
             popup.add(menu);
         }
 
+        List<Part> spares = Arrays.stream(parts)
+                                  .filter(Part::isSpare)
+                                  .collect(Collectors.toList());
+        JMenuHelpers.addMenuIfNonEmpty(popup, new SendToLocationMenu(
+              gui.getCampaign(), gui.getFrame(), spares,
+              dest -> LocationDispatch.dispatchPartsToLocation(
```
IllianiBird: `destination`

MekHQ/src/mekhq/gui/adapter/PersonnelTableMouseAdapter.java
```diff
@@ -2516,6 +2518,10 @@ protected Optional<JPopupMenu> createPopupMenu() {
         }
 
         JMenuHelpers.addMenuIfNonEmpty(popup, new AssignPersonToUnitMenu(getCampaign(), selected));
+        List<mekhq.campaign.personnel.Person> selectedPeople = Arrays.asList(selected);
+        JMenuHelpers.addMenuIfNonEmpty(popup, new SendToLocationMenu(getCampaign(), getFrame(),
+              selectedPeople,
+              dest -> LocationDispatch.dispatchToLocation(selectedPeople, dest, getCampaign())));
```
IllianiBird: `destination`

MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java
```diff
@@ -1056,6 +1058,12 @@ protected Optional<JPopupMenu> createPopupMenu() {
 
             JMenuHelpers.addMenuIfNonEmpty(popup, new AssignUnitToForceMenu(gui.getCampaign(), units));
 
+            List<mekhq.campaign.unit.Unit> unitList = List.of(units);
+            JMenuHelpers.addMenuIfNonEmpty(popup, new SendToLocationMenu(
+                  gui.getCampaign(), gui.getFrame(), unitList,
+                  dest -> LocationDispatch.dispatchUnitsToLocation(
+                        unitList, dest, gui.getCampaign())));
```
IllianiBird: `destination` not `dest`

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        super(parent, true);
+        this.campaign = campaign;
+        this.existingBase = existingBase;
+        initComponents(defaultSystem, defaultPlanet);
+        setLocationRelativeTo(parent);
+    }
+
+    private void initComponents(@Nullable PlanetarySystem defaultSystem,
+          @Nullable Planet defaultPlanet) {
+        setTitle(existingBase == null ? "New Base" : "Edit Base");
```
IllianiBird: We need to avoid these hardcoded strings

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
+
+        LocalDate today = campaign.getLocalDate();
+
+        txtDisplayName = new JTextField(20);
+        txtDisplayType = new JTextField(20);
```
IllianiBird: I'm concerns 20 characters might be a bit small.

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        if (existingBase != null) {
+            txtDisplayName.setText(existingBase.getDisplayName());
+            if (existingBase.getDisplayType() != null) {
+                txtDisplayType.setText(existingBase.getDisplayType());
+            }
+            PlanetarySystem system = existingBase.getCurrentSystem();
+            if (system != null) {
+                cboSystem.setSelectedItem(system);
+                if (existingBase.getPlanetId() != null) {
+                    Planet p = system.getPlanetById(existingBase.getPlanetId());
```
IllianiBird: `Planet planet`

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+            if (defaultPlanet != null) {
+                cboPlanet.setSelectedItem(defaultPlanet);
+            }
+        } else {
+            cboSystem.setSelectedIndex(-1);
+        }
+
+        JButton btnOk = new JButton("OK");
+        btnOk.addActionListener(e -> onOk());
+        JButton btnCancel = new JButton("Cancel");
```
IllianiBird: More hardcoded Strings

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        btnCancel.addActionListener(e -> dispose());
+
+        JPanel content = new JPanel(new GridBagLayout());
+        GridBagConstraints gbc = new GridBagConstraints();
+        gbc.insets = new Insets(4, 4, 4, 4);
```
IllianiBird: These inserts need to be scaled to gui, `UIUtil.scaleForGUI()` will set you right

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        gbc.anchor = GridBagConstraints.WEST;
+        gbc.fill = GridBagConstraints.HORIZONTAL;
+
+        int row = 0;
+        gbc.gridx = 0;
+        gbc.gridy = row;
+        gbc.weightx = 0;
+        content.add(new JLabel("Display Name:"), gbc);
```
IllianiBird: Hardcoded String

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        gbc.gridx = 1;
+        gbc.weightx = 1.0;
+        content.add(txtDisplayName, gbc);
+        row++;
+
+        gbc.gridx = 0;
+        gbc.gridy = row;
+        gbc.weightx = 0;
+        content.add(new JLabel("Display Type:"), gbc);
```
IllianiBird: I'm not going to comment on all of these, you get the gist. But we really shouldn't be including hardcoded Strings in new content.

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+                // throw "Attempt to mutate in notification". The user's typed text is already
+                // correct, so there is nothing to update.
+                if (updating[0]) {
+                    return;
+                }
+                updating[0] = true;
+                try {
+                    if (anObject == null) {
+                        editorField.setText("");
+                    } else if (anObject instanceof String s) {
```
IllianiBird: `String string`

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+                return text;
+            }
+
+            @Override
+            public void selectAll() {
+                editorField.selectAll();
+            }
+
+            @Override
+            public void addActionListener(java.awt.event.ActionListener l) {
```
IllianiBird: We don't need full class addresses here. AI loves them, but they're really not needed over just importing and they make no discernible performance impact.

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+                editorField.addActionListener(l);
+            }
+
+            @Override
+            public void removeActionListener(java.awt.event.ActionListener l) {
```
IllianiBird: `ActionListener listener`

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        });
+
+        // When the editor loses focus, restore the text to the selected item's display name
+        // (if one is selected). This handles the case where the user typed a partial search,
+        // did not pick from the popup, and clicked away — the editor snaps back to the real
+        // selection so the combo always looks consistent.
+        editorField.addFocusListener(new FocusAdapter() {
+            @Override
+            @SuppressWarnings("unchecked")
+            public void focusLost(FocusEvent e) {
```
IllianiBird: I prefer `evt` instead of just `e` for events. Personal preference, but I think it makes it more obvious what it is, whereas `e` has more room for uncertainty. This is especially important if you're actually using the event, as you are here

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+            @Override
+            public void changedUpdate(DocumentEvent e) {
+            }
+        });
+    }
+
+    /** Clears the editor text field of an editable combo without firing a selection event. */
+    private static void clearEditorText(JComboBox<?> combo) {
+        Component editor = combo.getEditor().getEditorComponent();
+        if (editor instanceof JTextField tf) {
```
IllianiBird: `JTextField textField`

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    private void onOk() {
+        String displayName = txtDisplayName.getText().trim();
+        if (displayName.isEmpty()) {
+            txtDisplayName.requestFocusInWindow();
+            return;
+        }
+
+        Object systemSel = cboSystem.getSelectedItem();
```
IllianiBird: `selectedSystem` or similar, please

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
```diff
@@ -0,0 +1,471 @@
...
+        if (!(systemSel instanceof PlanetarySystem system)) {
+            cboSystem.requestFocusInWindow();
+            return;
+        }
+
+        Object planetSel = cboPlanet.getSelectedItem();
+        Planet planet = planetSel instanceof Planet p ? p : null;
```
IllianiBird: `selectedPlanet`

MekHQ/src/mekhq/gui/menus/AssignUnitToPersonMenu.java
```diff
@@ -168,13 +169,16 @@ private void createPersonAssignmentMenus(final Campaign campaign, final Unit...
...
+        final IPlace unitPlace = units[0].getPlace();
+        List<Person> personnel = campaign.getPersonnel().values()
                                        .stream()
                                        .filter(person -> person.getStatus().isActive())
                                        .filter(person -> !person.getPrisonerStatus().isCurrentPrisoner())
                                        .filter(Person::isEmployed)
                                        .filter(person -> person.getUnit() == null)
+                                       .filter(person -> unitPlace == null || person.getPlace() == unitPlace)
                                        .sorted(new PersonTitleSorter().reversed())
                                        .collect(Collectors.toList());
```
IllianiBird, line 183: Man these streams are gross. Not your code, definitely not your problem, just needed to comment.

MekHQ/src/mekhq/gui/menus/SendToLocationMenu.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private void initialize(Campaign campaign, JFrame frame,
+          List<? extends ILocation> items,
+          Consumer<ILocation> dispatcher) {
+        if (items.isEmpty()) {
+            return;
+        }
+
+        setText("Send To...");
```
IllianiBird: Another hardcoded String

MekHQ/src/mekhq/gui/menus/SendToLocationMenu.java
```diff
@@ -0,0 +1,185 @@
...
+
+        // Determine the ILocation parent all selected items share (if any), to exclude it.
+        // ILocation subtypes like Personnel extend LinkedHashMap, so Objects.equals() would
+        // compare map contents rather than identity — two empty Personnel objects would falsely
+        // match, hiding every base from the menu. Use an identity-keyed set instead.
+        Set<ILocation> currentLocations = Collections.newSetFromMap(new IdentityHashMap<>());
+        for (ILocation item : items) {
+            var node = item.getLocationNode();
```
IllianiBird: I prefer strong typecasting over `var`, makes for clearer code

MekHQ/src/mekhq/gui/menus/SendToLocationMenu.java
```diff
@@ -0,0 +1,185 @@
...
+
+        // Main Force entry (= the campaign itself).
+        // Items at main force are parented to mainForcePersonnel, hangar, or warehouse — not
+        // to the campaign object directly — so check all three sub-resources.
+        boolean alreadyAtMainForce = sharedCurrent == campaign
+              || sharedCurrent == campaign.getMainForcePersonnel()
+              || sharedCurrent == campaign.getHangar()
+              || sharedCurrent == campaign.getWarehouse();
+        if (!alreadyAtMainForce) {
+            JMenuItem mainForce = new JMenuItem("Main Force");
```
IllianiBird: Hardcoded String. Not going to mark any more in this class, so please make sure to do a sweet and make sure they're all grabbed into a resource bundle

MekHQ/src/mekhq/gui/menus/SendToLocationMenu.java
```diff
@@ -0,0 +1,185 @@
...
+                continue;
+            }
+            addBaseItem(this, base, dispatcher, pathToMenu);
+        }
+
+        addSeparator();
+
+        JMenuItem newBase = new JMenuItem("New Base...");
+        newBase.addActionListener(e -> {
+            PlanetarySystem contextSystem = items.get(0).getCurrentSystem();
```
IllianiBird, line 144: This will throw an exception if `items` is empty

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+                }
+                cursor = cursor.getParent();
+            }
+        }
+
+        if (!isTraveling && node != null) {
+            AbstractBase base = LocationUtils.findEffectiveBase(item);
+            if (base != null) {
+                String name = base.getDisplayName();
+                return (name != null && !name.isBlank()) ? name : "Unnamed Base";
```
IllianiBird: You assign a resource bundle to this class, but are using a hard coded String here

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+        if (isTraveling) {
+            CurrentLocation currentLocation = (CurrentLocation) location;
+            JumpPath path = currentLocation.getJumpPath();
+            PlanetarySystem system = currentLocation.getCurrentSystem();
+
+            if (path.size() > 1 && currentLocation.isAtJumpPoint()) {
+                double neededHours = system.getRechargeTime(today, currentLocation.computeIsUseCommandCircuit(campaign));
+                double remainingHours = neededHours - currentLocation.getRechargeTime();
+                if (remainingHours > 0) {
+                    int days = (int) Math.ceil(remainingHours / 24.0);
```
IllianiBird: Let's make a constant for this magic number (yes, I'm aware we're not likely to change how many hours there are in a day, but it makes for cleaner code).

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+     * Returns {@code "-"} when the item is not traveling.</p>
+     */
+    public static String getDestinationName(ILocation item, Campaign campaign, LocalDate today) {
+        AbstractLocation location = getNearestAbstractLocation(item);
+        if (location instanceof CurrentLocation currentLocation) {
+            JumpPath path = currentLocation.getJumpPath();
+            if (path != null && !path.isEmpty()) {
+                PlanetarySystem dest = path.getLastSystem();
+                LocationNode currentLocationNode = currentLocation.getLocationNode();
+                if (currentLocationNode != null) {
```
IllianiBird: There is a lot of conditional nesting here, it might be worth seeing if we can break some of the code that's not feeding conditionals into a utility method to make it easier to see the flow of logic at a glance.

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+                    LocationNode parent = currentLocationNode.getParent();
+                    while (parent != null) {
+                        if (parent.getLocatable() instanceof AbstractBase base) {
+                            String name = base.getDisplayName();
+                            return (name != null && !name.isBlank()) ? name : "Unnamed Base";
```
IllianiBird: Hardcoded String

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+                            return campaign.getName();
+                        }
+                        parent = parent.getParent();
+                    }
+                }
+                if (dest != null) {
+                    for (PlayerBase base : campaign.getPlayerBases()) {
+                        if (dest.equals(base.getCurrentSystem())) {
+                            String name = base.getDisplayName();
+                            return (name != null && !name.isBlank()) ? name : "Unnamed Base";
```
IllianiBird: You get the gist

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+     * <p>Walks the {@link CurrentLocation}'s parent chain for an {@link AbstractBase}; if found
+     * returns the base's display name, otherwise the last system in the {@link JumpPath}.
+     * Returns {@code "-"} when the item is not traveling.</p>
+     */
+    public static String getDestinationName(ILocation item, Campaign campaign, LocalDate today) {
+        AbstractLocation location = getNearestAbstractLocation(item);
+        if (location instanceof CurrentLocation currentLocation) {
+            JumpPath path = currentLocation.getJumpPath();
+            if (path != null && !path.isEmpty()) {
+                PlanetarySystem dest = path.getLastSystem();
```
IllianiBird: `PlanetarySystem destination`

MekHQ/src/mekhq/gui/model/LocationDisplay.java
```diff
@@ -0,0 +1,275 @@
...
+
+    /**
+     * The printable name of the destination system when in transit, or {@code "-"} otherwise.
+     */
+    public static String getDestinationSystem(ILocation item, LocalDate today) {
+        AbstractLocation location = getNearestAbstractLocation(item);
+        if (location instanceof CurrentLocation currentLocation) {
+            JumpPath path = currentLocation.getJumpPath();
+            if (path != null && !path.isEmpty()) {
+                PlanetarySystem dest = path.getLastSystem();
```
IllianiBird: dest again

MekHQ/src/mekhq/gui/model/LocationFilterItem.java
```diff
@@ -0,0 +1,96 @@
...
+ * <ul>
+ *   <li>{@link #ALL} — show items from every location.</li>
+ *   <li>{@link #MAIN_FORCE} — show items belonging to the campaign's main force.</li>
+ *   <li>A base entry created via {@link #forBase(AbstractBase)} — show items at that base only.</li>
+ * </ul>
+ */
+public final class LocationFilterItem {
+
+    /** Sentinel: show items from every location, including main force and all bases. */
+    public static final LocationFilterItem ALL = new LocationFilterItem(null, "All (Main Force)");
```
IllianiBird: If this String, and the others in this class, are player-facing they need to be moved into a resource bundle

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -137,11 +138,10 @@ private Campaign getCampaign() {
 
     public void refreshData() {
         if (!isGroupByUnit()) {
-            setData(new ArrayList<>(getCampaign().getPersonnel()));
+            setData(new ArrayList<>(getCampaign().getAllPersonnel()));
         } else {
-            Campaign c = getCampaign();
             List<Person> commanders = new ArrayList<>();
-            for (Person p : c.getPersonnel()) {
+            for (Person p : getCampaign().getAllPersonnel()) {
```
IllianiBird: `Person person` if you don't mind. Not your code so not going to be a hard ass about this

MekHQ/src/mekhq/gui/model/TaskTableModel.java
```diff
@@ -139,8 +139,8 @@ public Component getTableCellRendererComponent(JTable table, Object value,
             } else {
                 if (part instanceof MissingPart) {
```
IllianiBird: Let's change this to `instanceof MissingPart missingPart` here, then we can just use `missingPart` in the rest of this method instead of recasting.

MekHQ/src/mekhq/gui/model/UnitTableModel.java
```diff
@@ -136,6 +142,12 @@ public String getColumnName(int column) {
             case COL_MODE -> "Mode";
             case COL_SHIP_TRANSPORT -> "Ship Transport";
             case COL_TAC_TRANSPORT -> "Tactical Transport";
+            case COL_LOCATION_NAME -> "Location";
+            case COL_LOCATION_SYSTEM -> "System";
+            case COL_LOCATION_PLANET -> "Planet";
+            case COL_DESTINATION_NAME -> "Destination";
+            case COL_DESTINATION_SYSTEM -> "Dest. System";
+            case COL_DESTINATION_PLANET -> "Dest. Planet";
```
IllianiBird: I'd _really_ like for these hardcoded Strings to find their way into a resource bundle.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -399,6 +404,10 @@ private void initStatusBar() {
         lblTempVesselCrew = new JLabel();
         lblPartsAvailabilityRating = new JLabel();
 
+        statusPanel.add(new JLabel(resourceMap.getString("lblActiveLocation.text")));
```
IllianiBird: Should be `getText` or `getTextAt` we shouldn't be using `getString` for updated or new resource bundle calls.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1414,6 +1427,81 @@ private String statusBarLabel(String key, Object... args) {
...
+        model.addElement(LocationFilterItem.MAIN_FORCE);
+        for (PlayerBase base : getCampaign().getPlayerBases()) {
+            model.addElement(LocationFilterItem.forBase(base));
+        }
+        return model;
+    }
+
+    /** Refreshes the three location-filtered tabs when the Active Location selection changes. */
+    private void refreshLocationFilteredTabs() {
+        HangarTab ht = getHangarTab();
```
IllianiBird: `HangarTab hangarTab`

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -1414,6 +1427,81 @@ private String statusBarLabel(String key, Object... args) {
...
+        if (ht != null) {
+            ht.refreshUnitList();
+        }
+        PersonnelTab pt = getPersonnelTab();
```
IllianiBird: `PersonnelTab personnelTab` same for the rest of this chain of calls.

MekHQ/src/mekhq/gui/PersonnelTab.java
```diff
@@ -504,11 +505,22 @@ public void refreshPersonnelList() {
...
+            people = new ArrayList<>(getCampaign().getMainForcePersonnel().fetchPersonnelAtLocation());
+        } else {
+            people = new ArrayList<>(locationFilter.getBase().fetchPersonnelAtLocation());
+        }
+        personModel.setData(people);
+
         for (int row = 0; row < personnelTable.getRowCount(); row++) {
             Person p = personModel.getPerson(personnelTable.convertRowIndexToModel(row));
-            if (p.getId().equals(selectedUUID)) {
+            if (p != null && p.getId().equals(selectedUUID)) {
```
IllianiBird: `person`

MekHQ/src/mekhq/gui/RepairTab.java
```diff
@@ -793,7 +798,19 @@ public boolean include(Entry<? extends TechTableModel, ? extends Integer> entry)
                 }
                 TechTableModel techModel = entry.getModel();
                 Person tech = techModel.getTechAt(entry.getIdentifier());
-                if (unit != null && unit.isSelfCrewed() && tech != unit.getEngineer()) {
+                // Tech must be at the same location as the unit being repaired
+                ILocation repairTarget = (unit != null) ? unit
+                      : (part instanceof Part p && p.getUnit() != null) ? p.getUnit() : (ILocation) part;
```
IllianiBird: `Part part`

IllianiBird (changes requested):
This is really impressive work, but there's definitely some areas that could do with a second pass.

I mentioned this on Discord, but I completely ignored all copilot feedback. I leave it to you to action any copilot stuff that seems valid.

The biggest areas of concern:
- hard coded strings
- abbreviated, single letter, and truncated variable names. Especially `dest` which appears a lot.

Programmatically everything _largely_ seems solid. This PR will conflict with Hammers new disposable infantry weapon PR. I recommend discussing with him which of your PRs should be merged first.

MekHQ/src/mekhq/campaign/base/AbstractBase.java
psikomonkie: Not here. A Base is just under a `FixedLocation` which'll just be the system & planet. Bases won't be parented to other stuff. Or at least I don't think normally would be unless we're considering a base on an ArkShip a base but uh that's stretching definitions a bit.

MekHQ/src/mekhq/campaign/base/AbstractBase.java
psikomonkie: I think this is a case where we'll want to fail loudly. Without a parent location we just have a base that is floating with no location. Very bad!

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: Fun fact! We don't actually save the explicit `LocationNode`. When anything with an `ILocation` is created, it should instantiate a new `LocationNode`. When saving, we don't save the `LocationNode`s, just what `ILocation`'s are parented/childed to another. This is then rebuilt on load. 

If a `LocationNode` is null, nothing will work, and we should throw an error immediately. 

Regarding your specific concern - even with these changes there is no `LocationNode` saved to `Person` during serialization, so I don't believe there's a risk.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: Hahaha. But it is ambiguous! `Person` and `Unit` use `UUID`, but `Part`s use `int` for ID. So yes, it's a string. I'd love to refactor `Part` sometime to use a `UUID` but I don't think I want to push the scope of this even more.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: Same as `Person`

psikomonkie: Though that's a good point, why'd I mark it `@Nullable`? I should change that.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: Returning list copies is very handy!

MekHQ/src/mekhq/campaign/location/LocationDispatch.java
psikomonkie: The string is just used for logging. If `buildArrivedNode` fails, I want to know where it was called from.

MekHQ/src/mekhq/campaign/Quartermaster.java
psikomonkie: This is how it currently works - we do not anticipate to hit a null. The purpose of those is to not need to do `!= null` constantly by asserting that the variable is definitely not null before continuing.

MekHQ/src/mekhq/campaign/Quartermaster.java
psikomonkie: `isAvailableAsSpareAmmo` only returns true if the `part` is an `AmmoStorage`, so it shouldn't be possible, and explicitly checking it here would be flagged as redundant by the IDE.

MekHQ/src/mekhq/gui/dialog/BaseSettingsDialog.java
psikomonkie: fwiw this whole dialog is placeholder, barebones. it'll need a better one with more options for its final form.

MekHQ/src/mekhq/gui/menus/SendToLocationMenu.java
psikomonkie: Lines 87-89, return early if `items.isEmpty()` - items will not be empty.

MekHQ/src/mekhq/campaign/location/IPlace.java
psikomonkie: This was ripped straight from `Campaign` iirc

MekHQ/src/mekhq/campaign/parts/missing/MissingPart.java
psikomonkie: Yes, different method that just happens to have the same name. The quartermaster one's boolean is for brand new, the warehouse's is for merge with existing.

MekHQ/src/mekhq/campaign/parts/missing/MissingPart.java
psikomonkie: Yes, different method that just happens to have the same name. The quartermaster one's boolean is for brand new, the warehouse's is for merge with existing.

MekHQ/resources/mekhq/resources/PersonnelStatus.properties
psikomonkie: Done - I'm not gonna comment this on every fix, I'll let you know when it's all resolved.

MekHQ/src/mekhq/campaign/base/PlayerBase.java
psikomonkie: Sure

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
IllianiBird: Method, not class

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: I've made `getLocationNode` Nonnull, btw.

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: No, I don't think so.

IllianiBird (approved):

PR #9311, opened 2026-06-11
Docs refactoring - Add AM documentation to Glossary

heckindoc:
This adds the previously standalone documentation about Advanced Medical (AM) to the glossary and moves the previously stand alone document into a folder marked for removal. 

While AM is really from around version 0.3.0, the last update to it was made in 0.50.11, where the TW hits was fixed, so that's where I marked its version for glossary purposes.


IllianiBird:
Glossary entry version should be the version the *entry* was added or last meaningfully updated. Not necessarily when the underlying system was last updated. So this should be marked as 51.01

MekHQ/resources/mekhq/resources/GlossaryEntry.properties
```diff
@@ -81,6 +81,21 @@ ADVANCED_SCOUTING.definition=If 'Advanced Scouting' is enabled in Campaign Optio
   Each day there is a 1-in-6 chance the formation will finish their patrol and return to base. This allows you to explore\
   \ the <a href='GLOSSARY:AREA_OF_OPERATIONS'>Area of Operations (AO)</a> much quicker than were you to use heavier \
   units.
+ADVANCED_MEDICAL.title=Advanced Medical (AM)
```
IllianiBird, line 84: We should probably include a link in this entry to the Alt AM entry, and if Alt AM mentions vanilla AM, that reference should be updated to link to your AM entry

IllianiBird (changes requested):

MekHQ/resources/mekhq/resources/GlossaryEntry.properties
heckindoc: These are all super reasonable and I should have done this up front, I'll get working on it!

IllianiBird (approved):

PR #9324, opened 2026-06-12
PR: Nested Map inside Navigation Tab and added Locations subtab

psikomonkie:
Now that there are multiple locations that the player's stuff can be, and bases, we need some UI to display that.

This PR renames the top level tab to "Navigation" and has two nested tabs below it: The existing Interstellar Map, and a new Locations tab. Both of these tabs, like normal campaign tabs, can be dragged out and made into windows. 

This also exposes the base settings dialog for existing bases - this will make it possible to edit existing bases.

image
image


MekHQ/src/mekhq/gui/view/LocationPlacePanel.java
```diff
@@ -0,0 +1,256 @@
...
+ * Detail panel for a single {@link IPlace}. Shows the place's name, type, current location status, system, planet, and
+ * (when in transit) its destination.
+ *
+ * <p>Follows MVC: {@link LocationPlacePanelModel} computes display strings, {@link
+ * LocationPlacePanelView} owns the widgets, and {@link LocationPlacePanelController} drives the single
+ * populate-on-construction update.</p>
+ */
+public class LocationPlacePanel extends JScrollablePanel {
+
+    private static final String RESOURCE_BUNDLE = "mekhq.resources.GUI";
```
IllianiBird: You don't need to assign a RESOURCE_BUNDLE when using GUI. Instead of `getTextAt` and `getFormattedTextAt` you just use `getText` and `getFormattedText`

MekHQ/src/mekhq/gui/LocationsTab.java
```diff
@@ -0,0 +1,458 @@
...
+    private static final class LocationsTabModel extends AbstractTableModel {
+
+        private static final int COL_NAME = 0;
+        private static final int COL_TYPE = 1;
+        private static final int COL_MODE_A = 2;  // System (General) or At Location (others)
+        private static final int COL_MODE_B = 3;  // Planet (General) or In Transit (others)
+        private static final int COL_EXTRA_START = 4;
+
+        // Unit type breakdown columns (displayed in UNITS mode, col >= COL_EXTRA_START)
+        private static final int[] UNIT_TYPES = {
```
IllianiBird, line 138: What about Space Stations?

MekHQ/src/mekhq/gui/LocationsTab.java
```diff
@@ -0,0 +1,458 @@
...
+            toolbar.add(new JLabel(getTextAt(RESOURCE_BUNDLE, "LocationsTab.toolbar.view")));
+            toolbar.add(viewDropdown);
+
+            table = new JTable(model);
+            JScrollPane tableScrollPane = new FastJScrollPane(table);
+
+            detailScrollPane = new FastJScrollPane();
+            detailScrollPane.setBorder(RoundedLineBorder.createRoundedLineBorder());
+            detailScrollPane.setHorizontalScrollBarPolicy(ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
+            detailScrollPane.setMinimumSize(new Dimension((int) (PLACE_DETAIL_WIDTH * 0.9), 600));
```
IllianiBird: It's highly unlikely to ever be an issue, but casting like this can result in a minimum size of 0.

MekHQ/src/mekhq/gui/NavigationTab.java
```diff
@@ -0,0 +1,99 @@
...
+    public NavigationTab(CampaignGUI gui, String tabName) {
+        super(gui, tabName);
+    }
+
+    @Override
+    public void initTab() {
+        setLayout(new BorderLayout());
+
+        ResourceBundle resources = ResourceBundle.getBundle("mekhq.resources.GUI",
+              MekHQ.getMHQOptions().getLocale());
```
IllianiBird: We shouldn't be using this resource method any more.

IllianiBird (changes requested):
Functionally everything _looks_ good. Just a couple minor points of feedback regarding String bundles. In both cases I only commented the first issue I encountered, rather than every instance.

MekHQ/src/mekhq/gui/view/LocationPlacePanel.java
psikomonkie: Done, not going to comment on each I'll resolve it if it's done.

MekHQ/src/mekhq/gui/LocationsTab.java
psikomonkie: Fair, I'll round up instead so it's at least 1 px.

MekHQ/src/mekhq/gui/LocationsTab.java
psikomonkie: Sure, added them and buildings.

IllianiBird (approved):

PR #9344, opened 2026-06-14
Consolidate and improve readability of Mission descriptions in Glossary

heckindoc:
This PR attempts to consolidate all of the mission type descriptions so that they are grouped together within the glossary. While doing this, I discovered that the terms "Contract" and "Mission" are used nearly interchangeably throughout MekHQ -- this creates space for confusion. As the we that we complete contracts is by selecting a button called "Complete Mission" I made the decision here to migrate the majority of the times where the term "contract" is used to "mission." I did, however, add a "Contracts Information" entry which links to the "Mission Types" entry. As I migrated to making everything "Mission," I also slightly modified the mission completion entry to ensure that it was placed next to all of the other Mission entries.  Finally, the `Missions and Contracts` entry was modified to `Contracts and Missions` so that players would have an easily accessible entry when they look for "Contract" that will then link them to the rest of the information on Missions. 

Overall, I'm hoping that this makes it easier for players to navigate through the glossary and find the information that the are wanting. If folks would prefer that everything that was changed to "Mission" is instead changed to "Contract" then I'm happy to do so.

image


IllianiBird (approved):
Good idea

Issue #8934, opened 2026-05-03
Customize Person origin-system picker should respect birth-date ownership and exclude connector systems


HammerGS:
## Problem

The origin-system dropdown in ``CustomizePersonDialog`` (Edit Person via GM tools) has two related shortcomings:

1. **Date inconsistency.** The default unfiltered list sorts and names systems using ``campaign.getLocalDate()`` (today's date), while the "Only Our Faction" filtered list uses ``person.getDateOfBirth()``. A character born in 3047 being edited in 3071 sees today's system names in one mode and birth-era names in the other, and the unfiltered list will happily include systems whose ownership history makes them implausible for that birth date.

2. **Connector systems leak in.** Synthetic "connector" systems used for jump-path routing (DPR0101, HWY-Coreward-Spinward-001, LTR-Coreward-001, FDR-Lokoja-001, etc.) appear in the dropdown alongside real canon systems. These have no inhabitants, no faction history, no real lore — they're noise in a birthworld picker.

The cleaner mental model is: the picker is choosing **where the character was born**, so the entire list should be filtered and named per the character's date of birth, and connector systems should never appear.

## Example

In a 3071 campaign editing a character born 3047:

- Unfiltered dropdown shows hundreds of entries including ``DPR 0101``, ``HWY-Coreward-001``, ``LTR-Coreward-001`` etc. (synthetic connector systems)
- ``Tharkad`` shows under its 3071 name in the unfiltered list, but under its 3047 name in the filtered list
- A planet that was uninhabited or unaffiliated at 3047 but settled later still appears as a valid birthworld option in the unfiltered list

## Proposed fix

In ``MekHQ/src/mekhq/gui/dialog/CustomizePersonDialog.java``:

- ``getPlanetarySystemsComboBoxModel()`` (line 1403, the unfiltered variant) — change the sort key from ``campaign.getLocalDate()`` to ``person.getDateOfBirth()`` so naming is consistent across both modes. Add a filter that excludes connector systems.
- ``getPlanetarySystemsComboBoxModel(Faction faction)`` (line 1416, the faction-filtered variant) — already uses ``person.getDateOfBirth()``; add the same connector-systems filter.
- Both — also consider filtering systems whose ``getFactionSet(person.getDateOfBirth())`` is empty (uninhabited at birth — not a plausible birthworld either way).

Connector-system detection requires a small change in ``MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java`` and ``Systems.java``:

[...]

PR #9350, opened 2026-06-14
Fix connector-system detection (zipped + unzipped) and quiet close-systems log spam

HammerGS:
## Summary

The "Extremely close systems detected" check floods the MekHQ log with hundreds of warnings on every startup. Root cause: connector systems were never being flagged as connectors in **either** data layout, so nothing could filter them out. This fixes connector detection for both layouts (zipped and unzipped) and skips connectors in the proximity check.

## Bug Fixed

`PlanetarySystem.isConnector()` was never `true` for any connector system, because `Systems.isConnectorPath()` only matched the substring `/connector_systems/` (slashes on both sides) - which matched neither real layout:

- **Unzipped**: loose files are flagged via `isConnectorPath(dirName)` with the `connector_systems` *directory* path (no trailing slash) -> not matched.
- **Zipped** (default runtime): connectors ship as `connector_systems.zip` with *flat* entry names (e.g. `HWY-Spinward-040.yml`); neither the parent dir, the entry names, nor the zip filename matched.

Effects:
- **Log spam**: `logVeryCloseSystems()` warns for any two systems within 1 ly; the ~1,240 connector systems (synthetic jump-path waypoints) sit dense along routes on purpose and tripped it by the hundreds, with no way to exclude them.
- **Broader**: every other `isConnector()` consumer was affected too, e.g. the origin-system picker filter (issue #8934).

## Changes (all in `Systems.java`)

1. **`isConnectorPath()`** - also match the connector_systems *directory* form (`endsWith "/connector_systems"`, `equals "connector_systems"`), keeping the full-path-segment guard so names like `my_connector_systems_notes.yml` are not flagged. Handles the **unzipped** layout.
2. **zip loading** - derive connector-ness from the zip's own filename (`connector_systems.zip`), since its entries are flat. Handles the **zipped** layout.
3. **`logVeryCloseSystems()`** - skip connector systems on both sides of the proximity check.

## Files Changed

- `MekHQ/src/mekhq/campaign/universe/Systems.java`

## Testing

- `isConnectorPath` now returns true for the connector_systems directory and its files, and for the `connector_systems.zip` filename; false for `canon_systems` (dir + zip) and look-alikes (`my_connector_systems_notes.yml`).
- Live data: 86 close (<1 ly) pairs - 85 connector-involved, 1 canon. After the fix only canon-vs-canon pairs can log.
- Reviewer check, both layouts (shipped zip, and after unzipping `connector_systems.zip` into a `connector_systems/` folder): launch MekHQ and confirm the startup log no longer floods with "Extremely close systems detected"; genuine close canon pairs still warn.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MekHQ/src/mekhq/campaign/universe/Systems.java
+++ b/MekHQ/src/mekhq/campaign/universe/Systems.java
@@ -482,11 +482,21 @@ public class Systems {
     }
 
     private void logVeryCloseSystems() {
-        // Planetary sanity check time!
+        // Planetary sanity check: this catches genuine canon coordinate data errors (two real
+        // systems at nearly the same spot). Connector systems are synthetic jump-path waypoints
+        // (see PlanetarySystem#isConnector / issue #8934) that are intentionally placed dense
+        // along routes, so they trip this check en masse with no data-error meaning. Skip them on
+        // both sides to keep the warning useful instead of flooding the log every startup.
         for (PlanetarySystem system : systemList.values()) {
+            if (system.isConnector()) {
+                continue;
+            }
             List<PlanetarySystem> veryCloseSystems = getNearbySystems(system, 1);
             if (veryCloseSystems.size() > 1) {
                 for (PlanetarySystem closeSystem : veryCloseSystems) {
+                    if (closeSystem.isConnector()) {
+                        continue;
+                    }
                     if (!system.getId().equals(closeSystem.getId())) {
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Nice! That explains why the issue seemed to have "appeared" all of a sudden. Thanks!

PR #9357, opened 2026-06-16
Add contract gauges to the briefing room and reorganise the contract panel

VicenteCartas:
image

## Summary

Replaces several plain text rows in the briefing-room contract panel with compact, colour-coded gauges, and regroups the panel so the most useful "at a glance" information sits together near the top.

The panel now shows a **dashboard cluster** of gauges — enemy morale, victory points, support points, salvage, and a contract timeline — followed by the static reference details. Rows that became redundant with the new gauges are removed to keep the panel compact.

## New components

- **`GradientMarkerBar`** (`gui/baseComponents`) — a reusable, value-agnostic gauge that draws a continuous gradient track with one or more labelled markers. The track stretches to keep out-of-range markers in view and proportionally placed. Designed to be reusable (a future PR can adopt it to replace `FactionStandingSlider`). Covered by unit tests.
- **`ContractMeterBar`** (`gui/view`) — a contract-specific wrapper around `GradientMarkerBar` exposing four factory variants:
  - **Victory points** — red→green value meter; reaching the target's reference is muted to a "(full term)" title cue for contracts that cannot be ended early.
  - **Support points** — red→green value meter against the negotiable reserve.
  - **Salvage** — red→green value meter of current vs. negotiated maximum salvage %.
  - **Timeline** — a deliberately **neutral** progress gauge (no red→green, since time has no good/bad direction) marking where "today" sits between the contract's start and end.

## Behaviour notes

- **Two visual languages:** "value meters" use a warm red→gold→green gradient ("more is better"); the timeline uses a single neutral track ("progress, not performance").
- **Timeline only when meaningful:** shown while the contract is genuinely under way at its destination system (dates fixed, today within the window). While travelling — when the engine slides the start/end dates each day to track the ETA — or after the contract ends, the dates are shown as a single compact `Dates: start – end` text row instead.
- **Salvage fallbacks:** the salvage-exchange and no-salvage cases keep their existing text presentation.
- Each gauge degrades to a plain label/value text row when its reference value is not meaningful (e.g. no positive victory-point target).

## Panel reorganisation

- The mission **status** moved into the section's title border (`"<name> - <status>"`), freeing a row.
- **Employer** and **Enemy** rows removed; the faction names are now tooltips on their belligerent logos (employer left, enemy right). The logos were also reduced in size to make room.
- **Mission Type** row removed (it is already part of the contract name in the title).
- The repetitive per-row `GridBagConstraints` boilerplate in `fillStatsAtBContract` was refactored into small `addStatRow` / `addHeaderRow` / `addGaugeRow` helpers.

## Conventions

- Localised strings via `getTextAt` / `getFormattedTextAt`; all new strings keyed in `ContractViewPanel.properties`.
- All pixel dimensions scaled through `UIUtil.scaleForGUI`.
- `jakarta` `@Nonnull` / `@Nullable` on the new components.

## Testing

- `GradientMarkerBarTest` (new) plus the existing `SegmentedBarTest` pass (39 tests).
- Main and test sources compile cleanly.
- Verified in-app across active, in-transit, and completed contracts; StratCon and non-StratCon; salvage-exchange / no-salvage; and negative / overshooting gauge values.

## Notes for reviewers

- `MoraleBar` now renders a "Morale" title, which also appears above the bar in the
  immersive **Morale Update** dialog (via `createDialogPanel`). This is intentional but is
  a visible change outside the briefing panel — happy to make the title opt-in if preferred.
- Employer/enemy faction names are now discoverable only on hover (the logo tooltips).


MekHQ/src/mekhq/gui/view/MissionViewPanel.java
```diff
@@ -137,7 +139,8 @@ public void updateDeploymentCoverage() {
     private void initComponents() {
         GridBagConstraints gridBagConstraints;
 
-        JPanel statsSection = BriefingStyle.createSectionPanel(mission.getName());
+        // The mission status is shown in the section's title border ("<name> - <status>") rather than as a row.
+        JPanel statsSection = BriefingStyle.createSectionPanel(mission.getName() + " - " + mission.getStatus());
         pnlStats = new JPanel();
         txtDesc = new JTextPane();
```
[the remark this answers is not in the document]
VicenteCartas: I tried that, but it looked terrible. I think losing the tooltip is better than adding it to the panel.

MekHQ/src/mekhq/gui/view/MoraleBar.java
```diff
@@ -92,8 +94,11 @@ public class MoraleBar extends JPanel {
...
      */
     public MoraleBar(@Nonnull final AtBMoraleLevel moraleLevel, @Nonnull final String labelText) {
+        // The segments sit at the top of the wrapped bar, so the title is placed directly above them with no extra gap.
         super(new BorderLayout());
         setOpaque(false);
+        add(new JLabel(getTextAt(RESOURCE_BUNDLE, "contractMoraleBar.title.text"), SwingConstants.CENTER),
+              BorderLayout.NORTH);
         add(bar, BorderLayout.CENTER);
 
         final AtBMoraleLevel[] levels = AtBMoraleLevel.values();
```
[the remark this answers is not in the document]
VicenteCartas: Good catch

MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -708,6 +708,17 @@ public int getCurrentSupportPoints() {
         return stratconCampaignState.getSupportPoints();
     }
 
+    /**
+     * Returns the support-point reserve this contract can be negotiated up to, used as the "full reserves" reference
+     * when displaying support points. This mirrors the cap applied during initial support-point negotiation (see
+     * {@code SupportPointNegotiation}): three per required combat team.
```
IllianiBird: This should probably be `@link` instead of `@code`

MekHQ/src/mekhq/campaign/mission/AtBContract.java
```diff
@@ -2084,19 +2095,23 @@ public JPanel getContractDifficultySkulls() {
...
      *
      * @return a {@link JPanel} with the employer and enemy faction logos, with a divider in between
      *
      * @author Illiani
      * @since 0.50.06
      */
-    public JPanel getBelligerentsPanel(int gameYear) {
-        final int SIZE = 100;
+    public JPanel getBelligerentsPanel(int gameYear, @Nullable String employerTooltip, @Nullable String enemyTooltip) {
+        final int SIZE = 64;
```
IllianiBird, line 2108: Should this be scaled to gui?

MekHQ/src/mekhq/gui/baseComponents/GradientMarkerBar.java
```diff
@@ -0,0 +1,481 @@
...
+    private static final int MARKER_OVERHANG = 3;
+    private static final int SOLID_MARKER_WIDTH = 4;
+    /** Width of a bold (emphasised) {@link MarkerStyle#SOLID} handle. */
+    private static final int BOLD_MARKER_WIDTH = 6;
+    private static final int TICK_MARKER_WIDTH = 2;
+    private static final int TRACK_ARC = 6;
+    /** Vertical gap between a marker label and the track. */
+    private static final int LABEL_GAP = 2;
+    /** Alpha applied to the contrasting outline drawn around markers. */
+    private static final int OUTLINE_ALPHA = 160;
```
IllianiBird, line 140: Looks like we're initializing a bunch of size values, they should probably be scaled for the users' gui.

Update: it looks like we do scale to gui, but later in the code. It'd probably be better to scale on initialization rather than needing to rescale each time one of these statics is called.

IllianiBird (changes requested):
Two very tiny points of feedback, otherwise this is excellent

MekHQ/src/mekhq/campaign/mission/AtBContract.java
VicenteCartas: This gets passed to scaleImageIcon, which already calls UIUtil.scaleForGUI.

VicenteCartas: We should probably at some point annotate our params to know if they expect scaled or unscaled values.

MekHQ/src/mekhq/gui/baseComponents/GradientMarkerBar.java
VicenteCartas: One issue if we scale on the statics, is that if the user were to change the scale mid-session, the statics are already set until they close mekhq. I'm ok either way, but whatever you prefer.

IllianiBird: Hrm, a very valid point

IllianiBird (approved):

PR #9361, opened 2026-06-16
Use more intuitive entry points for report generation

hokvel:
This change removes reports panel from the Command Center tab. 8 report types that it featured are triggered as follows:
- "Reputation Report" - by clicking `lblRating`
- "Personnel Breakdown" - by clicking `lblPersonnel`
- "Hangar Breakdown" - by clicking `lblComposition`
- "Transport Capacity" - by clicking `lblTransportCapacity`
- "Cargo Capacity" - by clicking `lblCargoSummary`
- "Faction Standing Report" and "Diplomacy Report" moved to the Faction panel as is
- "Jump Fee Summary Report" moved to the Map tab

All hyperlinked reports are still available via menu.

This streamlines UX by removing the step of finding corresponding report buttons when user wants to see details related to the stats featured in the Command Center.

Additionally:
- Simplify stat label initialization
- Clean up logic in cargo report generation
- Remove html tag wrapping from cargo report to make it compatible with `ClickableLabel`

Before:
Before

After:
Screenshot 2026-06-17 at 15 59 09


hokvel:
Relates to https://github.com/MegaMek/megamek/pull/8349

IllianiBird (changes requested):
Code-wise everything checks out, but I'm a little unsure about this PR. I worry that this level of obfuscation will result in users - especially new users - being unable to find these reports. If a player is unaware a report exists they're not likely to look for it.

Marked as 'request for change' till further discussion can be had.

hokvel:
> I worry that this level of obfuscation will result in users - especially new users - being unable to find these reports. If a player is unaware a report exists they're not likely to look for it.

Valid point. It’s definitely a significant UI change, and the last thing we want is to obscure report access. That said, the current 8-button panel is a major UX bottleneck. This initial PR is just one approach. Here is the breakdown of the alternatives and their limitations:
- The ⓘ symbol is a workaround for how `ClickableLink` currently behaves (it only underlines on hover); making links always underlined would be more intuitive, but it breaks consistency
- Changing the font color (like in the glossary links) is also problematic since it conflicts with existing color-coded labels (see cargo summary above)
- A safer fallback is row-level icons, but it introduces a brand-new UI pattern for contextual info, which I'd prefer to avoid if possible

TL;DR: The 8-button clutter needs a fix, and making reports contextual seems like the right path. However, the visual cue must be intuitive. I'm open to ideas on how we can achieve that while keeping the UI consistent. Here are some alternatives (they will require changes in `ClickableLink`)

Underscore:
Screenshot 2026-06-17 at 13 49 02

Underscore + color:
Screenshot 2026-06-17 at 13 51 54

IllianiBird (approved):

PR #9366, opened 2026-06-17
Update deprecated skill check in assited healing

hokvel:
Additionally, remove private static `Campaign` state from `AdvancedMedicalAlternateHealing` and make skill check methods private.


IllianiBird (approved):
Yeah! One less Campaign import!

PR #9368, opened 2026-06-17
PR: Keep LocationNode logic within location package

psikomonkie:
This is a refactor with no changes to functionality. 

I made a mistake by directly accessing the `LocationNode` trees from wherever. This makes the methods for getting a `LocationNode`'s `LocationNode` children or parents package-private instead of public. There should be helper methods available to do whatever a contributor needs regarding locations without working directly with the `LocationNode`s tree.


IllianiBird (approved):
Failing tests, but otherwise seeing no issues

PR #9372, opened 2026-06-17
Remove deprecated skill check calls

hokvel:
As a side effect it makes assisted healing, embezzlement, and surgery checks respect aging effects and other campaign settings affecting skill checks.

Additionally, do minor renaming in `AdvancedMedicalAlternateHealing`.


IllianiBird (approved):
LGTM

PR #9387, opened 2026-06-18
Added a Family Filter to Personnel Tab so Players Can More Easily See Relationships

IllianiBird:
image


MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
```diff
@@ -257,20 +264,21 @@ public boolean hasParents() {
...
-     * @return {@code true} if the person at least one living parent, otherwise {@code false}. Will also return false if
-     *       the person has no parents
+     * @return {@code true} if the person has at least one living parent, otherwise {@code false}. Will also return
+     *       false if the person has no parents.
      */
     public boolean hasLivingParents() {
-        if (hasParents()) {
-            return getFamily().get(FamilialRelationshipType.PARENT).stream()
-                         .anyMatch(parent -> !parent.getStatus().isDead());
+        // Avoid stream — early-exit loop once we find a living parent.
```
psikomonkie, line 274: `.anyMatch()` will short-circuit if it finds a match - it should function the same.

MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
```diff
@@ -326,10 +352,29 @@ private Set<Person> getAncestors(int depth) {
...
+        Set<Person> seen = new HashSet<>(8);
+        List<Person> grandparents = new ArrayList<>(4);
+        for (Person parent : parents) {
+            for (Person grandparent : parent.getGenealogy().getParents()) {
+                if (seen.add(grandparent)) {
+                    grandparents.add(grandparent);
+                }
+            }
+        }
+        return grandparents;
```
psikomonkie: Is `seen` just to avoid duplicates in `grandparents`? I believe you could achieve this by making `grandparents` a `Set<Person>` and then `return grandparents.stream().toList();` (or however you want to convert a Set to a List for the return)

Same deal in the other places where we count up grandparents.

psikomonkie (approved):
Nice! 

Two comments, please take a look at them before merging this. One of them applies repeatedly and I think would simplify the code a bit.

MekHQ/src/mekhq/campaign/personnel/familyTree/Genealogy.java
IllianiBird: I'm specifically avoiding streams. The reason this optimization pass was done was because we use these methods a lot in GUI components and the streams were surprisingly expensive.

psikomonkie: Fair enough, there's a dozen other ways to convert a set to a list. Using seen is definitely unnecessary though, even if you're removing streams.

IllianiBird: Mm, I see your point. Gimme two ticks, going to move to this pattern:

image

psikomonkie: Sets cannot have duplicates, do we need to be setting the initial capacity to 8 still?
EDIT: Otherwise yes, I think this's much cleaner!

IllianiBird: Initial capacity is a micro-optimization. It means that we're not always needing to expand the capacity of the array every time we add a new entry

psikomonkie: Right but are we expecting the average person to have 8 unique grandparents? The comment implies to me it should be 4.

IllianiBird: Ack, you're right

PR #9392, opened 2026-06-19
Improved Contract XML IO By Using a Map-Lookup Pattern

IllianiBird:
Also made some minor improvements to how individual fields were parsed.


exeea (approved):

MekHQ/src/mekhq/campaign/mission/AbstractMission.java
```diff
@@ -1547,197 +1554,225 @@ protected void writeToXMLEnd(final PrintWriter pw, int indent) {
...
+            Scenario scenario = Scenario.generateInstanceFromXML(scenarioNode, campaign, version);
+
+            if (scenario != null) {
+                addScenario(scenario);
+            }
+        }
+    }
+
     public void loadFieldsFromXmlNode(Campaign campaign, Version version, Node node) throws ParseException {
+        initializeNodeSetters(version);
```
exeea: why re-initialize the map every time you call loadFieldsFromXmlNode?
`version` doesn't alter the mapping.
I think is safe to move initializeNodeSetters in the constructor (with the other init) and pass version in another way.

options:
- saving it into a property of the class (but then you have a prop polluting your class).
- pass the version as a `context` to the called methods (then each consumer can decide to use it), maybe this is the cleanest way.

MekHQ/src/mekhq/campaign/mission/AbstractMissionTransition.java
```diff
@@ -193,6 +195,9 @@ public class AbstractMission {
 
     private final List<Scenario> scenarios = new ArrayList<>();
 
+    protected final Map<String, BiConsumer<String, Campaign>> fieldSetters = new HashMap<>();
```
exeea, line 199: these maps could become static and their initialization too.
The result would be that you don't have anymore a map for each instance but a single one shared between all missions.

(this means that loadFieldsFromXmlNode should pass as context the instance itself)

IllianiBird: This is an interesting idea. I'll take a look later today.

MekHQ/src/mekhq/campaign/mission/AbstractMission.java
IllianiBird: Ack, this is a whoopsie I originally had them initializing here and then I realized I could have them at the class level and forgot to move this out to the class constructor

PR #9393, opened 2026-06-19
Fix MHQCollapsiblePanel title centering, and re-land review fixes dropped from #9390

VicenteCartas:
## Summary
Fixes a layout bug where an `MHQCollapsiblePanel` created **with a title but no summary** renders its title **centered** instead of left-aligned. While here, this also re-lands several review-feedback fixes that were addressed on the original PR but never made it to `main`.

## Background
PR #9390 introduced `MHQCollapsiblePanel`, but only its **initial commit** was merged. The follow-up commits that addressed review feedback (localization, cursor, focus, static-analysis) landed after the merge and were left stranded. This PR brings them to `main` alongside the new centering fix.


MekHQ/src/mekhq/gui/baseComponents/MHQCollapsiblePanel.java
```diff
@@ -71,6 +73,8 @@ public class MHQCollapsiblePanel extends JPanel {
     public static final String EXPANDED_PROPERTY = "expanded";
     public static final String TOGGLE_ACTION = "toggle";
 
+    private static final String RESOURCE_BUNDLE = "mekhq.resources.GUI";
```
IllianiBird: You don't need to define a resource bundle for if the bundle is GUI. Instead of using getTextAt or getFormattedTextAt you just use getText and getFormattedText. That'll automatically use the GUI bundle

MekHQ/src/mekhq/gui/baseComponents/MHQCollapsiblePanel.java
```diff
@@ -135,7 +140,11 @@ public void setTitle(String title) {
 
     public void setSummary(String summary) {
         summaryLabel.setText(summary == null ? "" : summary);
```
IllianiBird: Here we could use StringUtility.isBlankOrNull (I think it's called). That'll catch Strings that are just white spaces, too. Which will address copilot's feedback too

IllianiBird (commented):
Some minor feedback. Nothing too awful

IllianiBird (approved):

PR #9439, opened 2026-06-22
Improved AbstractMissionTransition

IllianiBird:
- `AbstractMissionTransition` is now an actual `abtract` (whoops!)
- Some missing flattening methods have been brought over from a closed PR. The PR was closed following a change in design direction. Mostly this is just minor polish.
- Random Camouflage logic has been separated into a utility class
- I moved over some tests from the aforementioned closed PR.


MekHQ/unittests/mekhq/campaign/mission/MissionInheritanceFlatteningTest.java
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * <p>After the flattening, {@link Mission}, {@link Contract}, and {@link AtBContract} each extend
+ * {@link AbstractMissionTransition} directly. The most consequential behavioral change is that {@link AtBContract} is
+ * no longer an {@code instanceof Contract} and is no longer an {@code instanceof Mission}, and a {@link Contract} is no
+ * longer an {@code instanceof Mission}. A large amount of campaign code filters and dispatches on exactly these
+ * relationships, so these tests pin the relationships down to catch any accidental re-parenting in future work.</p>
+ *
+ * @author Claude (test author for PR #9417 review)
+ */
+class MissionInheritanceFlatteningTest {
```
psikomonkie: I think this test should be removed, this isn't the approach we're taking anymore - double check the other tests probably too - or maybe just this comment needs updated, because it looks like it might be fine ?

psikomonkie (approved):
I wouldn't worry too much about testing AbstractMissionTransition yet - from our discussion earlier, we'll be needing to remove the inheritance set up between AbstractMissionTransition and Contract to work towards splitting contract and mission properly.

IllianiBird:
Yeah, the tests were included only because they were already written and I might as well

PR #9442, opened 2026-06-22
Update Faction Standings Tests to Hopefully Fix Race Condition

IllianiBird:
# Requires https://github.com/MegaMek/megamek/pull/8373

MekHQ has an annoying Unit Test race condition where if another test initializes factions before a test that relies on faction data the stale data can persist. Polluting all future tests. Generally this isn't an issue unless you are specifically testing factional behavior.

This PR explicitly sets the Faction2 instance in the test, ensuring we have a 'pure' copy each time. I cannot be sure if this will fix the issue due to the nature of a race condition, however I'm hopeful.


MekHQ/unittests/mekhq/campaign/universe/factionStanding/FactionStandingsTest.java
```diff
@@ -73,6 +75,7 @@ void setUp() {
 
         assertDoesNotThrow(() -> {
             Factions.setInstance(Factions.loadDefault(true));
+            Factions2.setInstance(new Factions2(FACTIONS2_TEST_DIRECTORY));
             factions = Factions.getInstance();
```
psikomonkie: We should be testing the `getInstance()` method in MegaMek. Therefore, we shouldn't need to be explicitly testing it here too. 

We avoid this issue in `CombatTeamTest` via:
```
testFactions2 = new Factions2("testresources/data/universe/factions");
```
and
```
@SuppressWarnings("all") // get() without test; if it fails, test data is not loading; the test should fail
private Faction getFaction(String code) {
    return new Faction(testFactions2.getFaction(code).get());
}
```

We could do something similar here and on line 79 do:
`factions = new Factions2("testresources/data/universe/factions")` and then we won't need to worry about the Factions2 instance at all.

IllianiBird: Unfortunately this doesn't work here, as we need both the Faction2 _and_ Faction(1) instances to be initialized. Even if we're only directly using Faction2 in the tests the code we're testing relies on Faction(1), too

rjhancock (approved):

PR #9444, opened 2026-06-22
Simplify PersonnelTableModelColumn cell value extraction

hokvel:
Refactor the personnel table column structure to simplify the `getCellValue` method. This change delegates conditional column display logic to the view layer and cleans up data extraction by removing the view-related state flags from the cell evaluation.

Details:
- Add `MARKET_UNIT_ASSIGNMENT` and `SURNAME_GROUPED_BY_UNIT` columns
- Simplify `PersonnelTableModelColum.getCellValue`
- Update `isVisible` in `PersonnelTableModelColumn` to accept view flags and show new columns based on the view layer settings
- Use `MARKET_UNIT_ASSIGNEMNT` column in `PersonnelTableModel`
- Calculate `isLoadAssignmentFromMarket` instead of storing it as a field

### DIFF
--- a/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
+++ b/MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
@@ -103,6 +103,7 @@ public enum PersonnelTableModelColumn {
     PRE_NOMINAL("Column.PRE_NOMINAL.title", NaturalOrderComparator.INSTANCE),
     GIVEN_NAME("Column.GIVEN_NAME.title", NaturalOrderComparator.INSTANCE),
     SURNAME("Column.SURNAME.title", NaturalOrderComparator.INSTANCE),
+    SURNAME_GROUPED_BY_UNIT("Column.SURNAME.title", NaturalOrderComparator.INSTANCE),
     BLOODNAME("Column.BLOODNAME.title", NaturalOrderComparator.INSTANCE),
     POST_NOMINAL("Column.POST_NOMINAL.title", NaturalOrderComparator.INSTANCE),
     CALLSIGN("Column.CALLSIGN.title", NaturalOrderComparator.INSTANCE),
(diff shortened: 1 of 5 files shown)


MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
```diff
@@ -492,79 +493,78 @@ yield getAggregateSkillDisplay(person,
...
+                }
+
+                // Check for tech units
+                if (!person.getTechUnits().isEmpty()) {
+                    Unit refitUnit = person.getTechUnits()
+                                           .stream()
+                                           .filter(u -> u.isRefitting() && u.getRefit().getTech() == person)
+                                           .findFirst()
+                                           .orElse(null);
+                    String refitString = null != refitUnit ? "<b>Refitting</b> " + refitUnit.getName() : "";
```
IllianiBird, line 547: Bunch of player-facing hardcoded Strings. If we already have a RESOURCE_BUNDLE in this class can we move them into it, if not maybe a quick follow-up? Your call.

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
```diff
@@ -743,15 +737,15 @@ private static String getAggregateSkillDisplay(Person person, PersonnelRole prim
      */
     private boolean isNameRankOrStatusColumn() {
         return this == PERSON || this == FIRST_NAME || this == LAST_NAME ||
```
IllianiBird: I'd prefer if we added some newlines into this chain for ease of reading, something like...

```
return thing ||
        thing ||
        thing ||
        thing;
```

...makes chained conditionals a little easier to parse.

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -290,40 +282,39 @@ public Component getTableCellRendererComponent(JTable table, Object value, boole
...
                 case UNIT_ASSIGNMENT:
-                    if (loadAssignmentFromMarket) {
-                        final Entity en = personnelMarket.getAttachedEntity(person);
-                        setText((en != null) ? en.getDisplayName() : "-");
-                    } else {
-                        Unit u = person.getUnit();
-                        if ((u == null) && !person.getTechUnits().isEmpty()) {
-                            u = person.getTechUnits().getFirst();
-                        }
+                    Unit u = person.getUnit();
```
IllianiBird: Seeing that we're already editing this block, could we also nix the single letter/abbreviated variable names?

IllianiBird (changes requested):
The world's tiniest of nitpicks found here.

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
hokvel: Yes, eventually we will get there. I don't want to have too many changes in one PR

MekHQ/src/mekhq/gui/enums/PersonnelTableModelColumn.java
hokvel: Will update

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
hokvel: Yes, was trying to do it where I actually change the code. Here it was just indenting so I missed it, will update

IllianiBird (approved):

PR #9456, opened 2026-06-22
Refactor blob crew menu generation to use data-driven configuration

hokvel:
The previous implementation duplicated menu generation logic and explicitly checked campaign options for each personnel role. This change introduces an enum-based configuration for blob crews, streamlining menu creation, event handling, and campaign crew resets by iterating over role configuration.

Details:
- Add `BlobCrewConfig` enum to store i18n keys for `PersonnelRole` entries
- Replace individual `JMenu` fields with an `EnumMap`
- Update `buildBlobCrewPoolSubMenu` to accept a `BlobCrewConfig` argument
- Refactor related methods to loop over `BLOB_CREW_ROLES`


psikomonkie (approved):
Fantastic!!

PR #9467, opened 2026-06-23
Introduce Person.spendEdge and switch edge rerolls to use it

hokvel:
Previously, most skill checks were using Person.changeCurrentEdge which was not generating `PersonChangeEvent`. This change makes sure the event is always generated. Additionally, log a stack trace if there's no edge to spend.


IllianiBird (approved):
I wonder if there's merit in a follow-up where we add a boolean method that checks whether we have edge to spend, spends the edge, then reports back whether edge was spent. This would allow us to avoid the pattern where we need to check available edge and then spend the edge if there is edge available.

hokvel:
It depends on how many standalone `spendEdge` calls remain after legacy skill checks are migrated to the new API

PR #9500, opened 2026-06-26
Merged 'Vehicle & Vessel Crews' and 'Crew Requirements' Glossary Entries

IllianiBird:
As title


MekHQ/resources/mekhq/resources/GlossaryEntry.properties
```diff
@@ -2017,8 +2015,8 @@ VEHICLE_CREWS.definition=Some vehicles have crew requirements outside of drivers
   <br>- Vehicle Crew/Naval\
   <br>- Vehicle Crew/VTOL</blockquote>\
   <p>It is ultimately left to your discretion to decide what professions you use to fill these seats.</p>\
-  <p>Large Vessels, such as DropShips, may also have crew requirements. These can only be filled by characters with \
-  the Vessel Crewmember profession.</p>
+  <p>Large Vessels, such as DropShips, may also have 'other' crew requirements. These can only be filled by characters \
+  with the Vessel Crewmember profession.</p>
```
psikomonkie, line 2019: Ah. here's the text for that! Could we put this above the "Some vehicles have crew requirements outside of drivers and gunners"? With how it's worded, I didn't realize until here that the "Other" seats on Dropships/Jumpships needed vessel crew members, the text had implied I could use the professions above. 

I think it reads more naturally to have the specific case before the general case.

MekHQ/src/mekhq/campaign/utilities/glossary/GlossaryEntry.java
```diff
@@ -75,7 +75,6 @@ public enum GlossaryEntry {
     CONNECTIONS("CONNECTIONS", new Version("0.50.10")),
     CONTRACTS_AND_MISSIONS("CONTRACTS_AND_MISSIONS", new Version("0.50.06")),
     CONTRACT_VICTORY_POINTS("CONTRACT_VICTORY_POINTS", new Version("0.50.06")),
-    CREW_REQUIREMENTS("CREW_REQUIREMENTS", new Version("0.50.06")),
```
psikomonkie, line 78: Was this being used anywhere? It looks like it's still being used in `TutorialHyperlinkPanel.properties`, really sneaky for it to be using it in text so the IDE doesn't recognize it!

psikomonkie (approved):
Nice! Two comments to look at before merging; one might be a bug, one just feels better, so I'll leave the merging to you.

MekHQ/src/mekhq/campaign/utilities/glossary/GlossaryEntry.java
IllianiBird: You were right to ask. Fixed

PR #9504, opened 2026-06-26
Introduce ActionCheckRoll and promote ActionCheckResult to a class

hokvel:
In preparation for maintenance check migration to the new API, this refactor introduces support for normal, advantage, and disadvantage rolls and changes `ActionCheckResult` container to a class to control field access. Additionally individual dice results are not captured, enabling future UI improvements.

Details:
- Add `ActionCheckRoll` record to store individual dice and result
- Change `ActionCheckResult` from a record to a class
- Replace primitive roll integer in `ActionCheckResult` with `ActionCheckRoll`
- Update `ActionCheck.resolve` to use `ActionCheckRoll.perform` for evaluation
- Rename accessors to `getRollResult` and `getMarginOfSuccess`
- Update tests to reflect class changes and mock `Compute.d6` directly

### DIFF
--- a/MekHQ/src/mekhq/campaign/personnel/skills/ActionCheckResult.java
+++ b/MekHQ/src/mekhq/campaign/personnel/skills/ActionCheckResult.java
@@ -33,34 +33,53 @@
 
 package mekhq.campaign.personnel.skills;
 
-import mekhq.campaign.personnel.skills.enums.MarginOfSuccess;
-import mekhq.utilities.ReportingUtilities;
-
 import static mekhq.campaign.personnel.skills.enums.MarginOfSuccess.BARELY_MADE_IT;
 import static mekhq.campaign.personnel.skills.enums.MarginOfSuccess.getMarginOfSuccessObjectFromMarginValue;
 import static mekhq.utilities.MHQInternationalization.getTextAt;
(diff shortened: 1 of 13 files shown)


MekHQ/src/mekhq/campaign/personnel/skills/ActionCheckResult.java
```diff
@@ -33,35 +33,54 @@
...
  * @author Hokk
  * @since 0.51.01
  */
-public record ActionCheckResult(
-      int roll,
-      int marginOfSuccess,
-      boolean usedEdge,
-      String resultsText
-) {
+public class ActionCheckResult {
```
IllianiBird, line 49: What are the advantages of using a class vs a record here? Happy with the change, I just wanna get some context.

IllianiBird (commented):
Happy to approve, but before I do there is a minor question I'd like answered :)

MekHQ/src/mekhq/campaign/personnel/skills/ActionCheckResult.java
hokvel: I'm making `resultsText` private (since the intended way of getting reports is `getReport` now) and hiding the roll structure. As the class grows, we will need more control over encapsulation, and records don't have enough flexibility.

IllianiBird (approved):

PR #9510, opened 2026-06-26
Updated Alt Advanced Medical to Include New Campaign Options and Expanded Healing Systems

IllianiBird:
This PR adds a couple of changes I have wanted to make for a little while.

- The chance of a permanent injury has been increased to match ATOW (pg 193 6th printing). There is a campaign option to revert this change.
- Unassisted healing now _substantially_ increases healing time (x7) as per ATOW. So it takes a week to heal what would normally only take 1 day to heal.
- There is now a chance of an injury healing attempt also resulting in an additional injury, as per ATOW (pg 193 6th printing). Unlike ATOW this is not flavored as 'blood loss' but as a more generic and universally applicable 'medical complication'. As this is an injury it can kill the recipient. Albeit only if they already have 5 Hits worth of injuries, so shouldn't come up very often.
- There is a new campaign option to adjust healing times so players can make their campaigns as gritty or heroic as they see fit. Recommended values are 2.0 (realistic), 1.0 (current as of 51.0), 0.5 (the old 'kinder' mode). Kinder healing times has been removed.
- Numerous unit tests have been added, as alt advanced medical was wholly untested.
- Updated glossary


MekHQ/resources/mekhq/resources/GlossaryEntry.properties
```diff
@@ -185,8 +185,7 @@ ALT_ADVANCED_MEDICAL.definition=In 2009 <a href='GLOSSARY:ADVANCED_MEDICAL'>Adva
   <p>If the secondary location roll is a 6, the injury is more severe than normal and recovery time will be increased \
   by 100% before a second roll is made. If this second roll is also a 6, the severity is increased further and a final \
   roll is made. If this final roll is also a 6, then the injury is as severe as possible: if the primary location is a\
-  \ limb, the limb is completely severed (fatal, if the 'limb' was the head). Otherwise, recovery time is increased \
-  further (to the maximum of +300%).</p>\
+  \ limb, the limb is completely severed. Otherwise, recovery time is increased further (to the maximum of +300%).</p>\
```
psikomonkie, line 188: I liked how we didn't specify the numbers in the Campaign Options. I definitely understand wanting to include it in the glossary but I also think it leads to tracking values across different spots.

**For future similar glossary entries,** would it be possible to use a `{0}` to pass in a value derived from the actual calculation?

MekHQ/src/mekhq/gui/campaignOptions/contents/MedicalPage.java
```diff
@@ -128,25 +131,40 @@ class MedicalPage {
...
+        lblAlternativeAdvancedMedicalHealingTimeMultiplier = new CampaignOptionsLabel(
+              "AlternativeAdvancedMedicalHealingTimeMultiplier");
+        lblAlternativeAdvancedMedicalHealingTimeMultiplier.addMouseListener(createTipPanelUpdater(
+              "AlternativeAdvancedMedicalHealingTimeMultiplier"));
+        spnAlternativeAdvancedMedicalHealingTimeMultiplier = new CampaignOptionsSpinner(
+              "AlternativeAdvancedMedicalHealingTimeMultiplier",
+              1.0,
+              0.01,
+              10,
+              0.01);
```
psikomonkie: This spinner is newly exposed, ya? Can we add a metadata tag for `0.51.01` so it shows to players as a new option?

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedicalAlternate/HealingMarginOfSuccessEffects.java
psikomonkie: Please double check the copyrights before merging!

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedicalAlternate/AdvancedMedicalAlternateHealing.java
```diff
@@ -82,6 +81,9 @@ public class AdvancedMedicalAlternateHealing {
     private static final String RESOURCE_BUNDLE = "mekhq.resources.AdvancedMedicalAlternateHealing";
     private static final int PROSTHETIC_PENALTY = 4; // Interstellar Operations page 70
 
+    // ATOW pg193 6th printing
```
psikomonkie: We should chat sometime about a standard way to reference books and add it to the wiki. You ever tried searching the code base by a book/page reference? Can be kinda tricky, I'd love to see us use one pattern across the project.

psikomonkie (commented):
Nice! When I initially saw the new penalties I was shocked and had to go double check the rule book. All the more reason to keep a skilled team of doctors on hand!

4 comments: 
- Copyright _needs_ fixed before approval
- Metadata for the new spinner _should be_ fixed before merging/approval
- Glossary entry is a question  _for future changes_
- Book references is a Discord discussion reminder

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedicalAlternate/AdvancedMedicalAlternateHealing.java
IllianiBird: I concur

MekHQ/resources/mekhq/resources/GlossaryEntry.properties
IllianiBird: This is a different system, not related to 'less permanent injuries', however I do agree this is something we should explore in a future PR.

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedicalAlternate/AdvancedMedicalAlternateHealing.java
```diff
@@ -332,54 +241,86 @@ private static int getMarginOfSuccessForUnassistedHealing(Person patient, List<T
...
+                  healingSPAOptions.hasTraumaSurgeon(), healingSPAOptions.hasProthesisTechnician());
+            miscPenalty += healingSPAOptions.hasHypochondriac() ? 1 : 0;
 
-                    boolean useEdge = campaign.getCampaignOptions().isUseEdge();
-                    useEdge = useEdge && doctorOptions.booleanOption(EDGE_MEDICAL);
-                    int marginOfSuccess = getMarginOfSuccessForAssistedHealing(
-                          doctor, campaign, modifiers, miscPenalty, useEdge);
+            boolean useEdge = campaign.getCampaignOptions().isUseEdge();
+            useEdge = useEdge && healingSPAOptions.hasMedicalEdge();
+            int marginOfSuccess = getMarginOfSuccessForHealing(doctor,
```
VicenteCartas, line 309: **AI Review, Vicente didn't have time to check this PR.**

This refactor appears to change assisted healing cadence substantially. Previously, the Surgery check was inside `if (injury.getTime() <= 0)`, so an assisted injury progressed each day and only rolled when recovery was due. Here the check runs before the timer condition for every non-permanent injury. A negative margin then reverses that day's decrement, and a sufficiently bad result can spend Medical Edge even though no healing outcome is applied until the timer reaches zero.

Was the daily assisted roll intentional? The method documentation still says the check is resolved once the timer reaches zero. If not, the assisted check probably needs to remain inside the `time <= 0` branch while preserving the new weekly unassisted behavior.

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedicalAlternate/AdvancedMedicalAlternateHealing.java
```diff
@@ -393,22 +334,88 @@ private static void performAssistedHealingCheck(Campaign campaign, Person patien
...
+        ActionCheckResult actionCheckResult;
+        AttributeCheck attributeCheck = doctor.checkAttribute(BODY)
+                                              .withMiscModifier(miscPenalty)
+                                              .withExternalModifiers(modifiers);
+        String resourceBundleKey = "AdvancedMedicalAlternateHealing.naturalHealing.normal";
+        String reportText = getTextAt(RESOURCE_BUNDLE, resourceBundleKey);
+        actionCheckResult = attributeCheck.resolve(false, reportText);
+
+        // Edge
+        if (actionCheckResult.getMarginOfSuccess() <= PERMANENT_INJURY_THRESHOLD &&
```
VicenteCartas: **AI Review, Vicente didn't have time to check this PR.**

The Edge threshold does not account for `useAlternativeAdvancedMedicalFewerPermanentInjuries`. With that option enabled, margins from -3 through -5 map to `RECOVERY_DELAYED`, not a permanent injury, but this condition still spends Edge before the campaign option is applied. The replacement roll can also be worse: a -3 result that would have been a delay can reroll to -6 and become a permanent injury plus a medical complication.

Could the code determine whether the current result is actually permanent under the active campaign options before spending Edge, and reroll only in that case?

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptionsUnmarshaller.java
```diff
@@ -393,8 +393,10 @@ private static void parseNodeName(Version version, String nodeName, CampaignOpti
             case "tougherHealing" -> campaignOptions.setTougherHealing(parseBoolean(nodeContents));
             case "useAlternativeAdvancedMedical" ->
                   campaignOptions.setUseAlternativeAdvancedMedical(parseBoolean(nodeContents));
-            case "useKinderAlternativeAdvancedMedical" ->
-                  campaignOptions.setUseKinderAlternativeAdvancedMedical(parseBoolean(nodeContents));
+            case "useAlternativeAdvancedMedicalFewerPermanentInjuries" ->
```
VicenteCartas, line 396: **AI Review, Vicente didn't have time to check this PR.**

Could we retain a compatibility case for the old `useKinderAlternativeAdvancedMedical` XML tag? Existing saves with that option set to `true` now fall through to the unmarshaller's unknown-entry warning and load the new multiplier at its `1.0` default. The PR description identifies `0.5` as the equivalent of the old kinder mode, so this changes established campaigns back to full recovery times after loading.

A legacy case that maps `true` to `setAlternativeAdvancedMedicalHealingTimeMultiplier(0.5)` (and leaves `false` at the default) should preserve the old behavior.

MekHQ/src/mekhq/campaign/personnel/medical/advancedMedical/InjuryUtil.java
```diff
@@ -124,8 +126,13 @@ public static void resolveAfterCombat(Campaign c, Person p, int hits) {
      */
     public static void resolveCombatDamage(Campaign campaign, Person person, int hits) {
         if (campaign.getCampaignOptions().isUseAlternativeAdvancedMedical()) {
-            resolveCombatDamageUsingAlternateModel(campaign, person, hits,
-                  campaign.getCampaignOptions().isUseKinderAlternativeAdvancedMedical(), campaign.getLocalDate());
+            double healingTimeMultiplier = campaign.getCampaignOptions()
+                                                 .getAlternativeAdvancedMedicalHealingTimeMultiplier();
```
VicenteCartas: **AI Review, Vicente didn't have time to check this PR.**

Is the multiplier intended to apply only to combat-generated injuries? The Campaign Options tooltip says it multiplies **all** healing times, but I can only find it applied in this combat path and in replacement-surgery recovery. Other Alternate Advanced Medical injuries are created through `InjuryType.newInjury()` without consulting the multiplier, including diseases, postpartum recovery, discontinuation syndrome, blood loss, implant recovery, and the new medical-complication injury.

Could the multiplier be applied centrally for Alternate Advanced Medical injuries, or otherwise passed through all of those creation paths? As written, the same setting produces different healing-time behavior depending on where the injury originated.

VicenteCartas (commented):
**AI Review, Vicente didn't have time to check this PR.**

I left four inline comments focused on behavioral compatibility and healing mechanics. The focused medical test class and PR CI pass, but the identified edge cases are not covered by the current tests.

IllianiBird:
All feedback resolved

VicenteCartas (approved):

PR #9516, opened 2026-06-27
Refactor PersonnelTableModel renderers

hokvel:
- Extract personal state logic from `Renderer` into a separate method
- Remove dead code from `VisualRenderer`
- Introduce a shortcut method for fatigue calculations
- Introduce helper methods in `MHQOptions`
- Return `setRowHeight` statement that was lost in the personnel table view refactoring

### DIFF
--- a/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
+++ b/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
@@ -45,10 +45,10 @@ import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableCellRenderer;
 
 import megamek.common.annotations.Nullable;
-import megamek.common.units.Entity;
 import megamek.common.units.Jumpship;
 import megamek.common.units.SmallCraft;
 import megamek.common.units.UnitType;
+import mekhq.MHQOptions;
 import mekhq.MekHQ;
 import mekhq.campaign.Campaign;
 import mekhq.campaign.campaignOptions.CampaignOptions;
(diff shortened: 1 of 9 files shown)


MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -176,87 +177,63 @@ public Component getTableCellRendererComponent(JTable table, Object value, boole
...
+                cellColors = (cellColors == null) ? mhqOptions.getPregnantColors() : cellColors;
+            }
+            if (campaignOptions.isUseFatigue() && (getEffectiveFatigue(person, campaign) >= 5)) {
+                colorReasonKeys.add("colorReason.personnel.fatigued");
+                cellColors = (cellColors == null) ? mhqOptions.getFatiguedColors() : cellColors;
+            }
+            if (person.hasNonProstheticPermanentInjuries(campaignOptions.isUseAlternativeAdvancedMedical())) {
+                colorReasonKeys.add("colorReason.personnel.healedInjuries");
+                cellColors = (cellColors == null) ? mhqOptions.getHealedInjuriesColors() : cellColors;
+            }
```
IllianiBird, line 233: Do we want to condense this to using switch pattern-matching? Given that conditional B is overwriting A, so long as the switch executes in the same order it should produce the same result.

IllianiBird (commented):
Flagging a possible improvement, otherwise this is approve-worthy

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
hokvel: The problem is that we need visit every block for one thing, but stop early for another. So there's a compromise to be made. I will take another look at it. Usually, I avoid pass-through switch statements as they're more prone to errors.

IllianiBird: Entirely your call, just lemme know what your decision is. :)

hokvel: Let's keep it for now

IllianiBird (approved):

PR #9518, opened 2026-06-27
Fixed Intermittently Failing Tests

IllianiBird:
We were using needsFixing which ignores permanent injuries. If the injury a character gets becomes permanent it would then be falsely flagged as an uninjured character. Now we directly test whether the character has any injuries and also what their total hits are (test dependent).

### DIFF
--- a/MekHQ/unittests/mekhq/campaign/randomEvents/prisoners/EventEffectsManagerTest.java
+++ b/MekHQ/unittests/mekhq/campaign/randomEvents/prisoners/EventEffectsManagerTest.java
@@ -72,6 +72,7 @@ import mekhq.campaign.randomEvents.randomEventsSystem.PrisonerEventData;
 import mekhq.campaign.randomEvents.randomEventsSystem.PrisonerResponseEntry;
 import mekhq.campaign.stratCon.StratConCampaignState;
 import mekhq.campaign.universe.Faction;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Interesting! Q's before you merge but I think it's a me issue. 

 I still don't quite follow, if advanced medical is off, which is what this is testing, shouldn't `needsFixing()` act the same as `getHits() > 0`? 

Similarly, if these are non-advanced medical tests, are permanent injuries a thing? 


EDIT: Oh, if you're ever bored this'd be a great candidate for some `@BeforeEach` to reduce duplicate code.

IllianiBird:
> I still don't quite follow, if advanced medical is off, which is what this is testing, shouldn't `needsFixing()` act the same as `getHits() > 0`?

For non-AM/AAM campaigns needsFixing == getHits. I opted to change to getHits here as it's more of a direct access.

> Similarly, if these are non-advanced medical tests, are permanent injuries a thing?

They aren't, but I went in and updated all calls to `needsFixing` to something that will give more direct results.

> EDIT: Oh, if you're ever bored this'd be a great candidate for some @BeforeEach to reduce duplicate code.

I'm not that bored, but I'll see what I can do

IllianiBird:
Clarifying my last reply, these are tests for GREG _not_ AM/AAM explicitly. We just need to test how GREG interacts with AM

psikomonkie:
The question came up because one of the spots that changed `needsFixing()` to `getHits()` is named `testEventEffectInjuryPercent_NoAdvancedMedical`, which is why the change seemed strange.

IllianiBird:
Makes sense, but yeah, 'getHits' is fundamentally going to return the same value as if you used `needsFixing` but without extra conditionals that might pollute the return value if they're ever changed.

PR #9521, opened 2026-06-27
Consolidate UI updates in PersonnelTable.Renderer

hokvel:
- Stop calling `DefaultTableRenderer.getTableCellRendererComponent` and instead manage selection in the renderer
- Stop calling `MekHqTableCellRenderer.setupTableColors` and apply `MekHqTableCellRenderer.setupTigerStripes` only when needed
- Optimize `ImageIcon` retrieval in `PERSON_GRAPHICAL` column

### DIFF
--- a/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
+++ b/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
@@ -35,7 +35,6 @@ package mekhq.gui.model;
 import static mekhq.campaign.personnel.turnoverAndRetention.Fatigue.getEffectiveFatigue;
 
 import java.awt.Component;
-import java.awt.Image;
 import java.util.ArrayList;
 import java.util.List;
 import javax.swing.ImageIcon;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -237,19 +240,19 @@ private ComponentColors populatePersonalStateFlags(Person person, List<String> c
                   UIManager.getColor("Table.background")) : cellColors;
         }
 
-        private Image getImage(Person person, PersonnelTableModelColumn personnelColumn) {
+        private ImageIcon getImage(Person person, PersonnelTableModelColumn personnelColumn) {
```
IllianiBird, line 243: To confirm, we're not scaling the return value of getImage? If so we probably need to pass in dimensions. Trying to scale an ImageIcon gives very poor results.

hokvel: It doesn't change the logic, dimensions are hardcoded inside as previously. It prevents us from doing icon -> image -> icon in one case.

IllianiBird (approved):

PR #9523, opened 2026-06-27
Optimize PersonnelTable rendering performance

hokvel:
This PR focuses on minimizing object allocations and caching frequently rendered UI elements. This change introduces caches for portraits, entity images, and formation icons, reduces font updates during cell rendering, and adds an instrumentation utility for monitoring performance in development.

Details:
- Add `getEntityImage()` to `Unit` to simplify image caching
- Introduce caches in `PersonnelTableModel` to reduce memory overhead
- Optimize `getTableCellRendererComponent` by reducing font updates and reusing personal state flags
- Update `PersonnelTab` to use `refreshPersonnelList` and remove redundant `personnelListScheduler.schedule()` call
- Add `InstrumentedTableCellRenderer` to measure rendering performance

### DIFF
--- a/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
+++ b/MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
@@ -33,22 +33,31 @@
 package mekhq.gui.model;
 
 import static mekhq.campaign.personnel.turnoverAndRetention.Fatigue.getEffectiveFatigue;
+import static mekhq.gui.enums.PersonnelTableModelColumn.FORCE_GRAPHICAL;
+import static mekhq.gui.enums.PersonnelTableModelColumn.PERSON_GRAPHICAL;
+import static mekhq.gui.enums.PersonnelTableModelColumn.UNIT_ASSIGNMENT_GRAPHICAL;
 
 import java.awt.Component;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
 import javax.swing.ImageIcon;
 import javax.swing.JTable;
 import javax.swing.UIManager;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableCellRenderer;
 
+import io.sentry.util.Objects;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
```diff
@@ -236,23 +252,37 @@ private ComponentColors populatePersonalStateFlags(Person person, List<String> c
...
-                    yield (unit == null) ? null : new ImageIcon(unit.getImage(this));
+                    if (unit == null) {
+                        yield  null;
+                    }
+                    EntityImage entityImage = unit.getEntityImage();
+                    long cacheKey = Objects.hash(entityImage.getBase(), entityImage.getCamouflage());
+                    // key by base image and camouflage because EntityImage doesn't have an identity function
+                    // since we use WeakHashMap underneath, memory leaks won't be a problem
+                    yield entityImageCache.computeIfAbsent(cacheKey,
+                          key -> new ImageIcon(entityImage.loadPreviewImage(true)));
```
IllianiBird: It's my general preference that any lambda more than a couple lines long should be broken out into a utility method. It makes for far more readable code.

IllianiBird (changes requested):

MekHQ/src/mekhq/gui/model/PersonnelTableModel.java
hokvel: Which lambda do you mean? I only see `key -> new ImageIcon(entityImage.loadPreviewImage(true)` here

IllianiBird: Line 265 -> 277, and to a lesser extent 280 -> 285

hokvel: But it's a case block. I can change it to the old "case x:  break;" syntax

hokvel: Updated to if-elses

IllianiBird (approved):

PR #9549, opened 2026-06-29
Fixed Infantry Deploying Without Survival Gear

IllianiBird:
The infantry kit logic was incorrectly housed inside a conditional clause, causing it to not run. Now it does.


mmetke-qumulo (approved):
LGTM!

Sleet01 (approved):
LGTM

PR #9552, opened 2026-06-30
Converted Accountant from Record to Utility Class to Allow For Better Compatibility with Contract Overhaul & Multi-Locational Campaigns

IllianiBird:
The goal of this PR was to update the `Accountant` record to be more universally usable. Prior to this PR Accountant was strongly tied to `Campaign` and was wholly isolated from the new multi-locational stuff. The ongoing contract overhaul also needs to be able to evaluate the costs and values of portions of the campaign, so as to support multiple concurrent contracts and split forces.

The goal here was to make as few edits as possible, to the underlying logic. As well as to avoid needing to upheave large portions of the codebase. With this in mind I created an alternate constructor that just takes `campaign`. This allows us to make our changes without it affecting the rest of the codebase.


IllianiBird:
Reviewing alternative solutions

PR #9565, opened 2026-07-01
Refactored Accountant to Better Support Multi-Locational Forces and Concurrent Contracts

IllianiBird:
This PR creates static versions of all public methods in `Accountant` as well as doing a general refactoring pass to reduce the unnecessary handing off of `Campaign` objects.

These static methods allow a developer to access the calculations in `Accountant` without needing to instantiate a fresh `Accountant` class each time. This makes the calculations far more flexible.

## AI Disclosure
I used Claude to test new and prior functionality. I also used the same to expand test coverage of `Accountant`


MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -71,67 +73,231 @@
...
         return salaries;
     }
 
     public Money getMaintenanceCosts() {
-        if (getCampaignOptions().isPayForMaintain()) {
-            return getHangar().getUnitsStream()
-                         .filter(u -> u.requiresMaintenance() && (null != u.getTech()))
-                         .map(Unit::getMaintenanceCost)
-                         .reduce(Money.zero(), Money::plus);
+        return getMaintenanceTotal(getHangar().getUnits(), getCampaignOptions().isPayForMaintain());
```
psikomonkie: `Campaign.getHangar()` gets the campaign (main force's) local hangar - I think we want to use `Campaign.getAllUnits())` here.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -71,67 +73,231 @@
...
+                total = total.plus(unit.getMaintenanceCost());
+            }
+        }
+
+        return total;
     }
 
     public Money getWeeklyMaintenanceCosts() {
-        return getHangar().getUnitsStream().map(Unit::getWeeklyMaintenanceCost).reduce(Money.zero(), Money::plus);
+        return getWeeklyMaintenanceTotal(getHangar().getUnits());
```
psikomonkie: Same here, `Campaign.getHangar()` gets the campaign (main force's) local hangar - I think we want to use `Campaign.getAllUnits())` here.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -333,33 +539,164 @@ public Money getPeacetimeCost() {
...
-        return getHangar().getUnitCosts(u -> !u.isMothballed(), Unit::getSparePartsCost);
+        return getSparePartsTotal(getHangar().getUnits());
     }
 
     public Money getMonthlyFuel() {
+        return getFuelTotal(getHangar().getUnits());
+    }
+
+    public Money getMonthlyAmmo() {
+        return getAmmoTotal(getHangar().getUnits());
```
psikomonkie: These three too: `Campaign.getHangar()` gets the campaign (main force's) local hangar - I think we want to use `Campaign.getAllUnits())` here.

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -380,20 +717,40 @@ public Money getMonthlyFuel() {
...
+
+        return total;
     }
 
     /**
      * Calculates the total contract value of all qualifying units assigned to combat-role standard forces, applying
      * category-specific percentage multipliers and (optionally) diminishing returns.
      *
-     * <p>This method iterates over {@code campaign().getAllForces()} and includes only forces that are both:</p>
+     * <p>This method iterates over {@code campaign().getAllFormations()} and includes only forces that are both:</p>
```
psikomonkie, line 756: Good catch!!

MekHQ/src/mekhq/campaign/finances/Accountant.java
```diff
@@ -512,32 +899,59 @@ public Money getForceValue(boolean useDiminishingContractPay, boolean excludeInf
     }
 
     public Money getTotalEquipmentValue() {
-        Money unitsSellValue = getHangar().getUnitCosts(Unit::getSellValue);
-        return campaign().getWarehouse()
-                     .streamSpareParts()
-                     .map(Part::getActualValue)
-                     .reduce(unitsSellValue, Money::plus);
+        return getTotalEquipmentValue(getHangar().getUnits(), campaign().getWarehouse());
```
psikomonkie: Here as well: `Campaign.getHangar()` gets the campaign (main force's) local hangar - I think we want to use `Campaign.getAllUnits())` here.

Same with getWarehouse - we actually don't have a method for `getAllParts`, though it should mirror  `getAllUnits`. Something like:
```
    /**
     * @return All player's parts in {@code campaign}, not just the ones located with the main force.
     */
    public Collection<Part> getAllParts() {
        List<Part> parts = new ArrayList<>();
        for (AbstractLocation location : getCampaignLocationManager().getLocations()) {
            Set<Part> found = location.fetchPartsAtLocation();
            parts.addAll(found);
        }
        return parts;
    }
```

EDIT: Oh, or, bother, is warehouse one of those annoying classes that does some kind of logic inside it for equipment value? If it gets gross I'm okay with you slapping a TODO on it in the code and entering a bug on GitHub that parts outside the main warehouse don't contribute to total equipment value.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -6915,6 +6917,10 @@ public int getTemporaryMedicPool() {
         return humanResources.getTemporaryMedicPool();
     }
 
+    public Map<PersonnelRole, Integer> getTempCrewMapCopy() {
+        return Map.copyOf(humanResources.getTempPersonnelRoleMap());
+    }
+
```
psikomonkie: Do we need the method here or can we just use `HumanResource`'s `getTempPersonnelMap` that returns a copy of the map?

MekHQ/unittests/mekhq/campaign/finances/AccountantTest.java
```diff
@@ -73,7 +86,6 @@ public class AccountantTest {
 
     @Test
     void testGetMonthlyFoodAndHousingExpenses_WhenFoodAndHousingDisabled() {
-        // Setup
```
psikomonkie, line 76: In the future, should we be removing these from tests we otherwise aren't modifying? I personally like using them in test classes, so I worry this'll lead to competing styles in the repo changing over each other.

psikomonkie (changes requested):
I LOVE the changes to `Person`, love seeing `Campaign` get removed from methods!

Couple of changes but nothing serious! 

Really it's a consequence of `Campaign` filling a bit of a dual meaning unit it's split into `Force` and `Campaign`. `getHangar` and `getWarehouse` return the campaign's local hangar and warehouse, and in some of these methods I think getting _everything_, so `getAllUnits` would be more appropriate. 

I think this is great though, and should be good to go once you've updated the `getHangar().getUnits()`->`getAllUnits` (and double checked your other usages) - other comments are just questions and thoughts. 

Thanks!

MekHQ/src/mekhq/campaign/Campaign.java
IllianiBird: We could just do that, yeah

IllianiBird: We didn't

MekHQ/unittests/mekhq/campaign/finances/AccountantTest.java
IllianiBird: We shouldn't

psikomonkie (approved):
Excellent, amazing! Thanks for getting those couple of changes in! This puts finances in a much better stage for future additions.

PR #9566, opened 2026-07-02
Search only by UI text in Campaign Options and add search highlighting

hokvel:
This commit updates the Campaign Options search functionality to rely exclusively on the actual text displayed in the user interface, rather than internal route IDs and paths. It also adds visual highlighting for matching search terms directly within the supported UI components.

Details:
- Add `CampaignOptionsHighlighter` to extract UI text, and apply search term highlighting
- Modify `CampaignOptionsPane` to track the current search term and apply highlights to the active content host via `highlightSearchTerm`
- Update `CampaignOptionsPane` to use `extractAllText` for generating the section search text instead of relying on the page panel text
- Remove the `searchableText` field and its related evaluation logic from `CampaignOptionsRoute` to restrict matches to UI text only

It is an experimental feature.

### DIFF
--- a/MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
+++ b/MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
@@ -80,6 +80,7 @@ class CampaignOptionsNavigationPanel extends JPanel {
 
     private final List<CampaignOptionsRoute> routes;
     private final Consumer<CampaignOptionsRoute> routeSelectionListener;
+    private final Consumer<String> searchTermUpdateListener;
     private final JTextField filterField;
     private final JLabel filterStatusLabel;
     private final JTree navigationTree;
@@ -88,10 +89,12 @@ class CampaignOptionsNavigationPanel extends JPanel {
     private @Nullable Runnable searchIndexInitializer;
 
     CampaignOptionsNavigationPanel(@Nonnull List<CampaignOptionsRoute> routes,
-          @Nonnull Consumer<CampaignOptionsRoute> routeSelectionListener) {
+          @Nonnull Consumer<CampaignOptionsRoute> routeSelectionListener,
+          Consumer<String> searchTermUpdateListener) {
         super(new BorderLayout(0, CONTROL_GAP));
         this.routes = routes;
         this.routeSelectionListener = routeSelectionListener;
+        this.searchTermUpdateListener = searchTermUpdateListener;
 
         setName("campaignOptionsNavigationPanel");
         // Match the content panel's scroll-pane border instead of a TitledBorder. A TitledBorder reserves vertical
@@ -314,6 +317,7 @@ class CampaignOptionsNavigationPanel extends JPanel {
         if (currentRoute != null && findNavigationNode(root, currentRoute) == null) {
             navigationTree.clearSelection();
         }
+        searchTermUpdateListener.accept(normalizedFilter);
     }
 
     private List<CampaignOptionsRoute> getMatchingRoutes(String normalizedFilter) {
--- a/MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsRoute.java
+++ b/MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsRoute.java
@@ -50,12 +50,7 @@ class CampaignOptionsRoute {
     private final List<String> path;
     private final List<String> titleResourceNames;
     private final boolean showHelpPanel;
-    private final String searchableText;
     private String sectionSearchText = "";
-
-    CampaignOptionsRoute(@Nonnull String id, @Nonnull List<String> path, @Nonnull List<String> titleResourceNames) {
-        this(id, path, titleResourceNames, true);
-    }
 
     CampaignOptionsRoute(@Nonnull String id, @Nonnull List<String> path, @Nonnull List<String> titleResourceNames,
           boolean showHelpPanel) {
@@ -63,8 +58,6 @@ class CampaignOptionsRoute {
         this.path = List.copyOf(path);
         this.titleResourceNames = List.copyOf(titleResourceNames);
         this.showHelpPanel = showHelpPanel;
-        this.searchableText = normalizeSearchText(String.join(" ", path) + " " + id + " "
-                                                         + String.join(" ", titleResourceNames));
     }
 
     @Nonnull String getId() {
@@ -107,7 +100,7 @@ class CampaignOptionsRoute {
         }
 
         for (String token : normalizedFilter.split("\\s+")) {
-            if (!token.isBlank() && !searchableText.contains(token) && !sectionSearchText.contains(token)) {
+            if (!token.isBlank() && !sectionSearchText.contains(token)) {
                 return false;
             }
         }
--- a/MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
+++ b/MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright (C) 2026 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MekHQ.
+ *
+ * MekHQ is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License (GPL),
+ * version 3 or (at your option) any later version,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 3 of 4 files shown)


MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JSpinner;
+import javax.swing.JTextField;
+import javax.swing.text.BadLocationException;
+import javax.swing.text.DefaultHighlighter;
+import javax.swing.text.Highlighter;
+
+import megamek.logging.MMLogger;
+import mekhq.MekHQ;
```
VicenteCartas, line 52: Ideally we want to move these options system to megamek, so we can re-use it for all options dialogs in general. Importing into other mekhq classes will make this harder, I think we can get highlighting without these dependencies.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
```diff
@@ -314,6 +317,7 @@ private void buildNavigationTree(String filterText) {
         if (currentRoute != null && findNavigationNode(root, currentRoute) == null) {
             navigationTree.clearSelection();
         }
+        searchTermUpdateListener.accept(normalizedFilter);
```
VicenteCartas, line 320: This would need to be debounced so it doesn't run on every keystroke I think.

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+        String originalText = (String) comp.getClientProperty(ORIGINAL_TEXT_KEY);
+        return (originalText != null) ? originalText : currentText;
+    }
+
+    private static String getSpinnerText(JSpinner spinner) {
+        JComponent editor = spinner.getEditor();
+        if (editor instanceof JSpinner.DefaultEditor) {
+            return ((JSpinner.DefaultEditor) editor).getTextField().getText();
+        }
+        return spinner.getValue() != null ? spinner.getValue().toString() : null;
```
VicenteCartas, line 228: Why are we getting the value of the spinner for the highlight?

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+     * @param container The root container to start the traversal
+     */
+    public static List<String> extractAllText(Container container) {
+        List<String> extractedTexts = new ArrayList<>();
+        extractTextRecursive(container, extractedTexts);
+        return extractedTexts;
+    }
+
+    private static void updateHighlight(JComponent comp, String currentText, String searchTerm) {
+        if (currentText == null) {
```
VicenteCartas, line 95: For components that do not have a highlighter, could we instead of a custom class extending LayerUI that paints the highlight? This allows us to drop the mekhq dependencies, and also having to insert HTML into the UI.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
VicenteCartas: Something like this in the constructor:

filterDebounceTimer = new Timer(FILTER_DEBOUNCE_MS, e -> filterNavigation());
filterDebounceTimer.setRepeats(false);

filterField.getDocument().addDocumentListener(new DocumentListener() {
    @Override
    public void insertUpdate(DocumentEvent documentEvent) {
        filterDebounceTimer.restart();
    }

    @Override
    public void removeUpdate(DocumentEvent documentEvent) {
        filterDebounceTimer.restart();
    }

    @Override
    public void changedUpdate(DocumentEvent documentEvent) {
        filterDebounceTimer.restart();
    }
});

VicenteCartas (changes requested):

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        if (searchTerm == null || searchTerm.trim().isEmpty()) {
+            setText(comp, originalText);
+            return;
+        }
+
+        String highlightedText = applyHighlight(originalText, searchTerm);
+        setText(comp, highlightedText);
+    }
```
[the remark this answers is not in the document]
hokvel: All text is static there

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+                    extractTextRecursive(c, texts);
+                    yield null;
+                }
+                default -> null;
+            };
+            if (text != null && !text.trim().isEmpty()) {
+                texts.add(text);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
hokvel: There's logic preventing html matches in place.

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsRoute.java
```diff
@@ -107,7 +100,7 @@ boolean matches(@Nonnull String normalizedFilter) {
         }
 
         for (String token : normalizedFilter.split("\\s+")) {
-            if (!token.isBlank() && !searchableText.contains(token) && !sectionSearchText.contains(token)) {
+            if (!token.isBlank() && !sectionSearchText.contains(token)) {
                 return false;
             }
         }
```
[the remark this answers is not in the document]
hokvel: Intended.

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
hokvel: Yes, it's only needed to retrieve the highlight color. Will be trivial to make it more flexible if we decide to reuse it elsewhere

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
hokvel: Alternatively, we can crash if we encounter an unsupported component thus forcing new features to update this implementation accordingly. Given that this change is intended as an experiment, I would prefer to get more feedback before generalizing it and making reusable across the codebase (potentially unifying it with the Personnel Table highlighter)

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
```diff
@@ -0,0 +1,230 @@
...
+
+        int pos = 0;
+        while ((pos = text.indexOf(lowerSearch, pos)) >= 0) {
+            try {
+                highlighter.addHighlight(pos, pos + lowerSearch.length(), SPINNER_PAINTER);
+                pos += lowerSearch.length();
+            } catch (BadLocationException e) {
+                LOGGER.error("Error highlighting JSpinner", e);
+            }
+        }
```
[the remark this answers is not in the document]
hokvel: Intentional

MekHQ/src/mekhq/gui/campaignOptions/components/CampaignOptionsHighlighter.java
hokvel: I was thinking about allowing search by number. We can remove it

MekHQ/src/mekhq/gui/campaignOptions/CampaignOptionsNavigationPanel.java
hokvel: There's no change compared to the existing implementation that does the calculation every keystroke

hokvel: I thought it was intentional since the new campaign options are very lightweight and can actually afford to refresh results on every keystroke

VicenteCartas:
Testing more on my machine, the HTML insertion for the highlight seems to break the layout:

image

IllianiBird:
This has been resolved now (Alitar used this code as a basis). Closing

PR #9568, opened 2026-07-02
PR: Only Dispatch to Location on New Day

psikomonkie:
Includes #9565 

Removing more inconsistencies from transport. Education Travel was bespoke and I tried to work within its existing framework when doing the Location rework. 

Dispatching things was the Wild West. When did it happen? At some point! 

With this PR, any **non-GM** location changes happen during the new day processing instead of... Whenever. 

Additional changes I wanted to make relating to travel to and from local academies have been blocked while I was working on this - It's a pretty significant change to how/when Locations move.


MekHQ/src/mekhq/campaign/base/PlayerBase.java
```diff
@@ -137,6 +137,13 @@ public void writeToXML(PrintWriter pw, int indent) {
         MHQXMLUtility.writeSimpleXMLCloseTag(pw, --indent, "playerBase");
     }
 
+    @Override
+    public boolean writePendingTravelDestinationToXML(PrintWriter pw, int indent) {
+        MHQXMLUtility.writeSimpleXMLTag(pw, indent, "destinationType", "base");
```
IllianiBird: If `base` has special meaning is there value in adding it as a const somewhere? Maybe an enum if there are multiple special meaning strings.

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
```diff
@@ -1846,16 +1855,148 @@ private static void drainCampusPersons(Campaign campaign, AcademyCampusLocation
...
+        }
+    }
+
+    private static @Nullable ILocation resolvePendingDestination(Campaign campaign, Node routeNode) {
+        String type = getChildText(routeNode, "destinationType");
+        if (type == null) {
+            return null;
+        }
+        return switch (type) {
+            case "campaign" -> campaign;
```
IllianiBird: Building on my last point, I think we'd be better served using an enum to contain these Strings. Then we can just pass the enum into this and be done. And it means if we ever change what string means what we only need to update it in one place.

MekHQ/src/mekhq/campaign/location/LocationUtils.java
```diff
@@ -135,4 +149,57 @@ public static boolean isInTransit(@Nullable ILocation location) {
...
+     * Computes the total journey duration in days from a {@link JumpPath}, applying a minimum of
+     * 2 days.
+     *
+     * @param path         the route to calculate; must not be {@code null}
+     * @param date         the in-game date (used for pirate-point availability); must not be {@code null}
+     * @param startTransit the already-elapsed transit time at the origin, in days
+     * @return journey length in whole days, always at least 2
+     */
+    public static int computeJourneyDays(JumpPath path, LocalDate date, double startTransit) {
+        return Math.max(2, (int) Math.ceil(path.getTotalTime(date, startTransit, false)));
```
IllianiBird: This would read cleaner if we had a variable containing `(int) Math.ceil(path.getTotalTime(date, startTransit, false))` and then pass that variable into `Math.max(2, VARIABLE)`

IllianiBird (commented):
Gonna approve this as nothing marked is worth blocking the PR. I'll leave it to you to action the comments or just merge. Entirely your call :)

IllianiBird (approved):

MekHQ/src/mekhq/campaign/location/LocationUtils.java
psikomonkie: I agree, done!

MekHQ/src/mekhq/campaign/io/CampaignXmlParser.java
psikomonkie: Sure, done

PR #9569, opened 2026-07-02
Fix layout issues in the Command Center tab

hokvel:
Info panel was not filling horizontally resulting in width ratcheting. This change allows it (and its internal components) to retake horizontal space when it gets available.


IllianiBird (approved):
Man `gridBagConstraints.weightx` is such a trap. I've seen this issue crop up so many times now.

PR #9570, opened 2026-07-02
Separated Transport Cost Calculations from Campaign to Aid Contract Overhaul & Multi-Locational Forces

IllianiBird:
As per title


psikomonkie (approved):
Nice! Thanks!

PR #9598, opened 2026-07-07
Contract Overhaul Project: Completely Rewrote Employer, Enemy, and Contract Location Generation

IllianiBird:
# Introduction
The goal of the contract overhaul is take full advantage of the technological advances we have made in MekHQ over the past decade. Contract generation is _the_ first major system players are exposed to, early in their MekHQ careers, and one that is a frequent source of frustration. Despite AtB being nearly 20 years old we are still using it to power contract generation at all levels.

I have nothing against AtB and frankly MekHQ wouldn't be around if not for its integration. However, contract generation is showing its age.

With that in mind I have decided to take on the monumental task of completely rewriting every component of contract interactions. Starting with contract generation. This PR is the first major step in that journey.

# Rome Wasn't Built in a Single PR
A major undertaking of this level cannot be integrated in one fell apocalyptic PR. Such a PR would eclipse both the sun and anyones' reasonable ability to review it. Instead, I've opted to break the project up, where possible.

This PR focuses on the following:

- Employer Determination
- Target System Determination
- Enemy Determination

# How Things Used to Be
Under AtB employer, target system, and enemy were _largely_ determined independently of each other. We would take a look at the factions surrounding a player, pick one at random. Then we'd pick a random contract type. Finally we'd pick a random border world. All contracts followed this pattern and, while the player might realize it, it made the universe feel very static.

AtB's system also had a few core flaws.

1. If the employer didn't own any planets they couldn't generate contracts
2. If the enemy didn't own any planets they couldn't generate contracts. Some factions did had special handlers, though, Rebels and Pirates specifically.
3. If two factions were warring with each other, but didn't share borders, it was unlikely they would ever issue contracts against the other belligerent/s. This was most evident during the Periphery Uprisings and Clan Invasion. Where the player would need to preemptively migrate to the war front if they wanted to participate in the fighting.
4. While factional relationships were considered, they didn't have any significant impact on contract generation.

# Why Now?
This PR is being introduced as the vanguard of the new contract handling. It's being released first as I want to see how the new determination processes work 'in the real' so we can tackle any emergent issues before the bulk of the overhaul arrives.

## Legacy Handling
I have attempted to hook these new systems into the legacy AtB processes. The goal here is to ensure MekHQ continues to run as normal, just with the enhanced logic from this PR. While I made sure everything appeared to work correctly, the integration is mostly a bolt-on without any concern for longterm viability. All legacy AtB contract code will be removed through the Overhaul project, so I didn't take any time to make it pretty code.

# Employer Determination
image

If the player is not a mercenary campaign, then employer is simple: it's whatever faction the player is a part of. If the campaign is a mercenary campaign, but the player is taking advantage of the old AtB 'retainer' system, then the employer will be whoever the campaign is a retainer with. For all other campaigns we need a random employer.

For random employer we start by building a map of regional weights. This tells us how frequently an individual faction appears within the target area. For each faction found we check for 'early exit', this is a faction that we're going to skip.

Early exit occurs in the following circumstances:

1. Faction is an 'empty' faction (Undiscovered, Abandoned, None, or '???' - the default setting for a null faction)
2. The Faction isn't valid in the current date, this protects us against stale planetary data
3. The faction is the wrong type. Under the contract overhaul we will be searching for specific types of employer, such as Major Faction, Superpower, etc, following the rules in CamOps. In this PR we have everything set up to use that filtering, however no current call takes advantage of that.
4. The player is a mercenary campaign, but the prospective employer doesn't use mercenaries
5. The prospective employer is themselves the mercenary faction (mercenary employers are handled later)
6. The employer is the Republic of the Sphere and the current date is during the greater 'Fortress Republic' period. Of note is that we now differentiate between the wider 'Fortress Republic', which covered all Republic territories, and the more limited 'Terra only' that only covered Terra. Previously Fortress Republic was a single check that barred all Republic hiring after a certain period (without end).

If the faction isn't covered by an 'early exit' clause we next check whether it contains any other factions. For example, after their exile the Aurigan Restoration spent a period in exile in the Magistracy. This allows them to still issue contracts against the Directorate without needing to own planets themselves.

Finally, we end up with a weighted map with ineligible factions weened out. The contract employer is a random faction chosen from that weighted map.

It's worth noting that there are a couple of existing systems that can change employer. Such as mercenary sub-contracts, or ComStar/Word of Blake special handlers. This will continue to work as they do in 51.0, but have been wholly replaced by the WIP Contract Overhaul, just not in this PR.

# Contract Type Determination
The process of picking a contract type is largely unchanged by this PR, though the greater Overhaul does rewrite it to bring our contract picker in-line with CamOps. However, one special handler added was a 'landless' clause. If a faction does not hold any territory they are blocked from generating defensive contracts, such as Garrison Duty.

# Enemy Determination
image

Enemy determination is far more dynamic than under AtB and is largely based on a combination of employer and contract type. With different contract types resulting in a different picker 'profile', as shown below:

image

Each of the seven profiles follows a different process to pick the target faction. This allows us to better flavor the OpFor on the contract selected.

- **Pirates** If the employer is a Clan faction, the enemy will be the Bandit Caste. Otherwise the enemy faction will be Pirate.
- **Rebels** If enemy faction will always be Rebels
- **Raiders** The enemy has a 50/50 chance of being either Pirates or Rebels
- **At War** We draw up a list of all factions currently at war with the employer. The warring faction must own territory and not be covered by the Fortress Republic exclusion. A random faction is then picked from that list. If the list is empty we fallback to Default (covered below)
- **Occupying Power** First we check whether the employer is at war with anyone (as above). Then we check if any of the enemy factions has claimed territory from the employer within the past decade. We create a list of such factions and then randomly pick one. If no such factions exist we fallback to Default
- **Covert** This follows the same process as Default, excluding one step which will be covered under Default.

**Default** Is more complicated than the other profiles so will be receiving special attention.

Under default we start by counting how many times each other faction appears in the search radius. Then, if the employer isn't the Pirate or Bandit Caste faction, we add _all_ factions the employer is warring with (if any). This allows employers to target factions outside of the search radius.

For each potential enemy we filter out any factions considered 'empty' (see earlier) or covered by the Fortress Republic exclusion.

If the contract is considered **Covert** then allies can be added to the pool, should they appear within the search radius. This represents political black ops between 'allies'.  Otherwise allies are completely barred from being the target of a contract.

Factions that don't have a direct alliance record but share a common ally (e.g. two member states of the same superpower, each individually allied with it but not with each other) should still be treated as allies, unless factionHints directly contradicts that by recording them as at war with each other.

It's worth noting that a faction is never _normally_ its own enemy - unless factionHints explicitly records it at war with itself, which is how a civil war can be represented (e.g. the Ghost Bear Civil War). I'm not aware of us taking advantage of this special handler in any way, but I included it as some nice future proofing. And in the hope that future civil wars will be easier to handle than the FedCom Civil War was.

There are a couple of exceptions to the no blue-on-blue protections. The following factions can fight themselves:
- Pirates
- Bandit Caste
- Mercenaries

One we have our weighted map of possible target factions we then adjust the weightings, using the following rules:

1. If the employer is at war with the prospective target faction, we first enforce a minimum weight of 10. This is to ensure warring factions are eager to fight each other. Then we multiply the weight by 4.
2. If the two factions are rivals the weight is multiplied by 2.
3. If the contract is set during the Jihad and the target is Word of Blake then the weight is multiplied by 2.
4. If the employer is ComStar and the enemy is Clan, the weight is divided by 12. The pre-existing comment explains why: "This is pretty hacky, but ComStar does not have many targets and tends to fight the Clans too much between Tukayyid and the Jihad"

The final weight is then used to give us a random enemy faction. If no enemy faction is present (despite our best efforts) then we return the Rebel faction as a fallback. Realistically, though, this fallback should never be reached.

# Target System Determination
(General Logic)
image

(Special Faction Handlers)
image

As I touched upon earlier, under AtB target system was very random. So long as the target system contained the target faction then it was fair game. Something I wanted to do with the rewrite was to try and embed some flavor into the system.

Like picking target faction, I wanted to use a subtle system to help shape a players' narrative. Not in any obvious way, if I've done my job properly players will never know these systems are at play. However, by using a more robust framework we're able to create a more dynamic feeling universe.

With that in mind we start by checking against some factional special handlers. These special handlers are processed in order, so if one doesn't result in a valid target system we'll move onto the next.

1. If the **defender** is from the pirate faction we search for empty systems nearby. Places where pirates might holdout. If we find any we'll have the pirate contract appear there. If we don't, we'll favor systems bordering the periphery.
2. If the **attacker** is a pirate faction we check for nearby systems bordering the periphery and attack those.
3. If the **defender** is ComStar we only consider systems to be valid targets if they are explicitly owned by ComStar. Or if the system contains an A or B-rated HPG.
4. If the **attacker** is pirate (this is a fallback for step 2), mercenary, ComStar, or Word of Blake then we consider any system the target faction owns in the search radius to be equally valid. These factions can strike anywhere. We could have used a profile here, but I opted not to. This was done to keep these factions feeling unpredictable.
5. If the **defender** is rebel then we pick a valid target within the attackers' own territory using the **Interior Populated** profile (see below)

If none of these special handlers apply, or if they fall through, we revert to 'normal' picking. Similar to target faction selection we utilize a contract-based profile. This allows to to flavor target systems based on the type of contract.

image

Note: all profiles fall through to **Default** in the event the profile fails to find a valid target system.

- **Rear Area** Finds the defender's systems in range that are not on the shared border with the attacker: training cadres and standing retainers are stationed in the safe rear, not on a contested front-line world.
- **Interior Populated** Finds every system the defender holds in range, border or not: riots and internal-security work can flare up anywhere in the defender's space, with no particular relationship to the enemy's border.
- **Deep Raid** Here we utilize the existing code to find the systems bordering the attacker and defender. Then extend the depth the 'border' is considered. This allows us to make deep strikes into enemy territory.
- **Occupied Territory** Finds targets for a guerrilla campaign behind enemy lines. Preferred tier: defender-held systems in range that the attacker held 10 years ago - recently conquered worlds whose population plausibly still sympathizes with the attacker. Second tier: any defender system in range away from the shared border, since a guerrilla campaign on the contested front is just the regular war. Empty only when the defender holds nothing in range beyond the border itself.
- **Invasion** Identical to High Value, however if Invasion fails to find a valid system the invasion is 'called off'. This is to reduce the likelihood of factions 'invading' beyond their sphere of influence.
- **High Value** As Default, with an additional weighting described below.
- **Default** Finds the defender's systems on a genuine shared border with the attacker: the direct border between the two first and, when none exists, the border of whichever regional faction hosts the attacker as a "contained" opponent of the defender, as a proxy for the attacker's otherwise poorly-defined local presence. If default fails to find a valid target system it will just pick the defender's closest system. This ensures things like the Clan Invasion still function.

Once we have our list of valid target systems we do one final step:
- **Interior Populated** creates a weighted list of all systems with the weight influenced by population and HPG quality.
- **High Value** & **Invasion** does the same as above, but also includes Industry & Output in the weighting (to account for factory worlds).

With all that handled, we then pick a random system.


MekHQ/src/mekhq/MHQConstants.java
```diff
@@ -413,6 +412,11 @@ public final class MHQConstants extends SuiteConstants {
     public static final LocalDate BATTLE_OF_TUKAYYID = LocalDate.of(3052, 5, 21);
     public static final LocalDate COMSTAR_SCHISM = LocalDate.of(3052, 9, 21);
     public static final LocalDate OPERATION_SCOUR_ENDS = LocalDate.of(3078, 1, 1);
+    public static final LocalDate FORTRESS_REPUBLIC_START = LocalDate.of(3135, Month.NOVEMBER, 1);
+    public static final LocalDate FORTRESS_REPUBLIC_END = LocalDate.of(3049, Month.JANUARY, 1);
+    public static final LocalDate FORTRESS_REPUBLIC_TERRA_ONLY_END = LocalDate.of(3151, Month.JANUARY, 1);
+    public static final LocalDate OPERATION_BULLDOG_START = LocalDate.of(3059, Month.MAY, 20);
```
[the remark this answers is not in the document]
IllianiBird: Fixed

MekHQ/src/mekhq/campaign/mission/mission/contractGeneration/GlobalEmployerTableValue.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private final String lookupName;
+    private final String label;
+    private final String tooltip;
+    private final int lowerBand;
+    private final int upperBand;
+
+    private final String RESOURCE_BUNDLE = "mekhq.resources.GlobalEmployerTableValue";
+    private final static MMLogger LOGGER = MMLogger.create(GlobalEmployerTableValue.class);
+
```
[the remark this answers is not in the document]
IllianiBird: Fixed

IllianiBird:
Flipping this to live

Dark-Hobbit:
Two nice-to-haves from discussions on the Discord:
1. Reputation with a faction affecting its probability of being selected as an employer, meaning you get more contracts from factions with high reputation and fewer contracts from factions that hate you.
2. Let Clan factions fight IS bandits while in Inner Sphere space rather than limiting them to Bandit Caste.

MekHQ/src/mekhq/campaign/market/contractMarket/AbstractContractMarket.java
```diff
@@ -534,32 +534,30 @@ protected void rollTransportClause(AtBContract contract, int mod) {
...
+     * before, while the other profiles shape which factions the weighted pool prefers.
+     *
+     * @param contract the contract to select an enemy for; its type and employer must already be set
+     * @param campaign the active campaign
+     */
+    protected void setEnemyCode(AtBContract contract, Campaign campaign) {
+        EnemySelectionProfile profile = contract.getContractType().getEnemySelectionProfile();
+        Faction enemyFaction = RandomFactionGenerator.getInstance()
+                                     .getRandomEnemy(campaign.getCurrentLocation(), campaign.getLocalDate(),
+                                           contract.getEmployerFaction(), profile);
```
psikomonkie, line 551: I think we could avoid passing in `Campaign`, and I think it would be beneficial. Having parameters of `AtBContract, ILocation, LocalDate` would help with generating contracts from different origins.

MekHQ/src/mekhq/campaign/market/contractMarket/AtbMonthlyContractMarket.java
```diff
@@ -594,19 +589,25 @@ private static boolean doesComStarCoOpt(LocalDate today, AtBContract contract) {
...
-        contract.setName(generateDefaultName(employer, contract));
+        contract.setName(generateDefaultName(employer, contract, campaign));
 
         contract.clanTechSalvageOverride();
 
         return contract;
     }
 
-    static @org.jspecify.annotations.NonNull String generateDefaultName(String employer, AtBContract contract) {
+    static @org.jspecify.annotations.NonNull String generateDefaultName(String employer, AtBContract contract,
```
psikomonkie: I don't think this is the right `NonNull`, `jakarta.annotation.Nonnull` is the standard used.

MekHQ/src/mekhq/campaign/mission/newContract/MissionTargetFinder.java
```diff
@@ -0,0 +1,502 @@
...
+        // default chain below rather than failing contract generation.
+        List<PlanetarySystem> profileTargets = findProfileTargets(profile, attacker, defender, location, radius,
+              currentDate);
+        if (!profileTargets.isEmpty()) {
+            return profileTargets;
+        }
+
+        // INVASION found no shared border above, and it is the one profile that is a hard restriction rather than a
+        // preference: an invader can only supply and hold a conquest adjacent to its own territory, so there is no
+        // deep-placement fallback for it - no shared border means no viable invasion target at all.
```
psikomonkie: Why is having no adjacent / bordering systems owned a blocker for this? I feel like this is weird for the clan invasion, or other instances where systems were taken and held without adjacency - WoB?

MekHQ/src/mekhq/campaign/mission/newContract/targetFinder/MissionTargetFinder.java
```diff
@@ -0,0 +1,502 @@
...
+        return switch (profile) {
+            case REAR_AREA -> findRearAreaTargets(attacker, defender, location, radius, date);
+            case INTERIOR_POPULATED -> findAllDefenderTargets(defender, location, radius, date);
+            case DEEP_RAID -> borderTracker.getBorderSystems(attacker, defender, location, radius,
+                  deepRaidBorderSize(attacker, defender));
+            case OCCUPIED_TERRITORY -> findOccupiedTerritoryTargets(attacker, defender, location, radius, date);
+            case INVASION -> findSharedBorderTargets(attacker, defender, location, radius, date);
+            case HIGH_VALUE, DEFAULT -> Collections.emptyList();
+        };
+    }
```
psikomonkie: These should probably be part of the `MissionLocationProfile` enum, like the switch you did for `AtBContractType`.

MekHQ/src/mekhq/campaign/mission/newContract/PirateMissionTargetFinder.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * Catalyst Game Labs and the Catalyst Game Labs logo are trademarks of
+ * InMediaRes Productions, LLC.
+ *
+ * MechWarrior Copyright Microsoft Corporation. MekHQ was created under
+ * Microsoft's "Game Content Usage Rules"
+ * <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
+ * affiliated with Microsoft.
+ */
+package mekhq.campaign.mission.newContract;
```
psikomonkie: I've noticed a bunch of these `.*TargetFinder` classes. Do they need to be directly used by other classes in the `newContract` package? I don't think they do, so I think:
- Move all of the `.*TargetFinder` classes into a new package like `mekhq.campaign.mission.newContract.targetFinder`. You can keep the same access on the PirateMissionTargetFinder (package-private) etc. and that will make it so other packages are only able to utilize `MissionTargetFinder` directly, which seems the be the pattern used here.

MekHQ/src/mekhq/campaign/universe/FactionBorderTracker.java
```diff
@@ -302,6 +333,148 @@ public synchronized List<PlanetarySystem> getBorderSystems(Faction self, Faction
...
+            if ((radius < 0) || (system.getDistanceTo(origin) <= radius)) {
+                nearby.add(system);
+            }
+        }
+
+        lastSystemsNearOrigin = origin;
+        lastSystemsNearRadius = radius;
+        lastSystemsNearResult = nearby;
+        return nearby;
+    }
```
psikomonkie, line 460: Consider `Map` cache of some kind using a private record `systemsNearKey` that is composed of the `origin` and `radius` instead of only using the most recent. I think there's something similar used for the `Personnel` lookup cache.

MekHQ/src/mekhq/campaign/universe/RandomFactionGenerator.java
```diff
@@ -122,387 +165,599 @@ public void startup(Campaign c) {
...
+        }
+
+        return switch (profile) {
+            case PIRATES -> pirateFactionFor(employer);
+            case REBELS -> Factions.getInstance().getFaction(REBEL_FACTION_CODE);
+            case RAIDERS -> raiderEnemy(employer);
+            case AT_WAR -> atWarEnemy(location, date, employer);
+            case OCCUPYING_POWER -> occupyingPowerEnemy(location, date, employer);
+            case COVERT -> getRandomEnemy(true, location, date, employer);
+            case DEFAULT -> getRandomEnemy(false, location, date, employer);
```
psikomonkie: This should probably live in `EnemySelectionProfile` instead of being a switch.

psikomonkie (commented):
Wow, amazing! I'm excited for the new contracts. 

Couple of comments before I can can approve this:

- I think the standard `Nonnull` is the jakarta one, not jspecify, please change that before this is merged so we're only using one `Nonnull` annotation across the repo.
- Couple of switch statements that should probably be part of the enum they're switching over that should be updated.
- We should switch the package now if we think that'll be correct instead of waiting
- Removing as a parameter would be nice to have but isn't necessary.
- Other are questions that might just need an answer, or maybe they need a change too.

MekHQ/src/mekhq/campaign/mission/newContract/MissionTargetFinder.java
IllianiBird: Removed

psikomonkie (approved):
Amazing! This should be good once you resolve the errors & conflicts. Thanks!

MekHQ/src/mekhq/campaign/universe/FactionBorderTracker.java
IllianiBird: Done, great idea

MekHQ/src/mekhq/campaign/mission/newContract/PirateMissionTargetFinder.java
IllianiBird: Done

PR #9601, opened 2026-07-07
PR: Show location information in right panel for unit and person

psikomonkie:
Show Location information in right panel - Units and Persons will have their current, queued, and destination locations displayed. If the unit or person isn't queued for travel, queued won't be displayed, and if they aren't actively going to a destination, destination isn't displayed. Current location will always show.


IllianiBird (approved):
Approving pending the copilot feedback being closed

PR #9610, opened 2026-07-09
Fix #9560: Move A Time of War options out of Operations → Systems

### LINKED ISSUE #9560: Gripe about the "A Time of War Options" tab in MekHQ settings
Many of the settings in the tab "A Time of War Options" belong in the HR or Advancement tab, rather than in the "Systems" tab in the "Operations" folder. The reason I believe this is because these settings are generally about personnel or skills. A simple "based on A Time of War" like in the setting "Character Age Affects Skills" should be good enough if the issue is about attributing these settings to _A Time of War_.

I believe this will help avoid confusion, and increase the quality of MekHQ.

VicenteCartas:
## Summary

Per #9560, the "A Time of War Options" page sat under Operations → Systems even though its settings are about personnel, skills, and income. This relocates them next to their peers:

- **New page — Advancement → Skills → Attributes & Traits:** Starting Attributes Based on Profession, Random Attributes, Display All Attributes, Character Age Affects Skills, Randomize Traits, Soldiers Use Small Arms Only.
- **Operations → Finances → General** gains an "A Time of War Income" section: Monthly Reinvestment of Wealth, Monthly Connections Checks, Better Extra Income.
- The old A Time of War page and its route are removed; Systems keeps Reputation and Faction Standing.

ATOW attribution is preserved through the section summaries and the existing "Based on ATOW" tooltips.

## Notes

- Pure UI relocation — the backing settings still live in `RandomSkillPreferences` and `CampaignOptions`, so **saves and campaign presets are unaffected** (presets reload the moved toggles through the new page).
- Added a round-trip unit test for the new `AttributesAndTraitsOptionsModel`; the `SystemsOptionsModel` shed its now-unused `RandomSkillPreferences` dependency.

## Screenshots

image

image


IllianiBird (approved):

IllianiBird:
Make sure to include the line `fixes #CASE_NUMBER` in the description, so GitHub will also close the associated case/s

PR #9612, opened 2026-07-09
Contract Overhaul Project: Updated Command Rights Handling

IllianiBird:
Updated Command Rights handling, in an attempt to make each Command Right more distinctive and to try and reduce scenario sizes.

- **Integrated** all scenarios count as Turning Points (unchanged), removed integrated forces from scenario generation
- **House** 1-in-3 chance for a scenario to count as Turning Point (unchanged), all Turning Points have a House Officer present. Non-Turning Points don't
- **Liaison** 1-in-3 chance for a scenario to count as Turning Point (unchanged), removed extra Liaison units from scenarios
- **Independent** changed required Turning Points to 1, Turning Point scenarios only spawn if current CVP is below 1, at which point all scenarios are Turning Points.


binaryspica:
I think that the first independent is actually integrated command?

And how to determine that independent mission to be successful? Maybe using turning point as the checker for compulsory or not?

IllianiBird:
Correct, fixed

VicenteCartas (approved):

PR #9627, opened 2026-07-11
Fixed Ambushes Not Occurring For Already Deployed Forces

IllianiBird:
As per title


MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConRulesManager.java
```diff
@@ -303,11 +303,22 @@ public static void generateDailyScenariosForTrack(Campaign campaign, StratConCam
...
+                }
+
                 scenario = generateScenarioForExistingForces(scenarioCoords,
-                      track.getAssignedCoordForces().get(scenarioCoords),
+                      assignedForceIDs,
                       contract,
                       campaign,
-                      track);
+                      track,
+                      ambushTemplate,
```
VicenteCartas, line 320: Since this scenario later goes through the existing `finalizeBackingScenario`
call, should we also apply the same post-finalization state as the ambush path
in `deployForceToCoords`? That path forces Crisis and clears Turning Point,
while this one currently keeps the normal finalization result.

VicenteCartas (changes requested):

MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConRulesManager.java
IllianiBird: Good shout

IllianiBird:
Actioned all feedback

VicenteCartas (approved):

PR #9629, opened 2026-07-11
Contract Overhaul Project: Added the Ability to Zoom and Pan StratCon AO Map; Moved Sectors to Tabs; Added GUI Only Support for Multiple Concurrent Contracts

IllianiBird:
~~As per title. Nothing too fancy.~~ _It got fancy_

Updated the AO tab with new functionality:

- Added Pan and Zoom functionality
- Sectors are now displayed as tabs, not a list.
- Sector tab labels will be colored based on the status of the sector (green: all objectives done; yellow: some objectives outstanding; red: all objectives failed). These are default colors and can be changed by the player.
- Each Sector objectives list will only show the objectives for that sector
- Added GUI support for multiple concurrent StratCon contracts. This is GUI support only and does not mean MekHQ supports multiple concurrent contracts. That's coming later.

https://github.com/user-attachments/assets/6f81fa29-5ef6-4ffe-abb3-0dbf54a98673


MekHQ/src/mekhq/gui/StratConTab.java
```diff
@@ -308,38 +353,50 @@ public void updateCampaignState() {
...
+        if (endingDate != null && endingDate.isBefore(currentDate)) {
+            sb.append(getTextAt(RESOURCE_BUNDLE, "stratConTab.status.expired"));
         }
 
-        sb.append("<br/><b>Campaign Victory Points:</b> ").append(campaignState.getVictoryPoints())
-              .append("<br/><b>Support Points:</b> ").append(campaignState.getSupportPoints())
-              .append("<br/><b>Deployment Period:</b> ").append(currentTDI.track.getDeploymentTime())
-              .append(" days")
-              .append("</html>");
+        sb.append(getFormattedTextAt(RESOURCE_BUNDLE, "stratConTab.status.victoryPoints",
```
VicenteCartas: Could we include the required VP target here too?

The old objective display showed the contract-wide requirement from
`getRequiredVictoryPoints()`. I understand removing it from the sector
objective list now that each tab only shows that sector, but the replacement
status currently shows only the points we have, not how many are required.

Maybe `Campaign Victory Points:
current / required` would fit here?

MekHQ/src/mekhq/gui/StratConTab.java
```diff
@@ -370,207 +425,281 @@ private String buildShortStrategicObjectiveText(StratConCampaignState campaignSt
...
+        cmd.display(getCampaign(), currentContract.getStratConCampaignState(),
+              currentSectorTrack, getCampaign().isGM());
         cmd.setModalityType(ModalityType.APPLICATION_MODAL);
         cmd.setVisible(true);
     }
 
     @Subscribe
     public void handleNewDay(NewDayEvent ev) {
-        repopulateTrackList();
+        repopulateContractsAndSectors();
```
VicenteCartas, line 762: Should we also rebuild this on MissionNew/MissionChanged?

The contract-market flow calls `campaign.addMission` first, then
`acceptContract`, which creates the StratCon campaign state. This tab currently
rebuilds for new day, mission removal, and mission completion, but not when a
contract is added or its state is initialized.

As far as I can tell, a newly accepted concurrent contract can stay missing from
the selector until the next day or another unrelated rebuild happens.

VicenteCartas (changes requested):

MekHQ/src/mekhq/gui/StratConTab.java
IllianiBird: Done

MekHQ/src/mekhq/gui/StratConTab.java
IllianiBird: I'm not sure if this is correct - I wasn't able to verify - but adding the event here wouldn't hurt anything.

IllianiBird:
All feedback actioned

image

VicenteCartas (approved):

Issue #3705, opened 2023-07-01
RFE: Ability to mark ammo consumption on manual resolution


Skoraks:
### Description
It would be nice to have the ability to track ammo expenditure when using the "Resolve Manually" feature for scenarios. I primarily use MekHQ to track tabletop Mercenary campaigns and currently this requires some manual editing of the MUL if I want this extra detail.

A potential way to display this could be in a similar way to how armour and structure is handled currently, with a new Ammo section added, and number input boxes for each critical slot of ammo present on the mech, so you can edit the remaining shots value post game.

PR #9632, opened 2026-07-11
Add ability to edit ammo bin content for manual scenario resolution

Skoraks:
Adds additional button to "Edit ammo" alongside the existing "Edit unit" during manual scenario resolution. 

Allows ammo expenditure tracking to manual resolution, with the intent of supporting tabletop game input.

image

https://github.com/MegaMek/mekhq/issues/3705


IllianiBird (changes requested):
This should probably be moved to a similar location as megamek/client/ui/dialogs/UnitEditorDialog.java and hooked up so that it can be accessed in MegaMek. That way players can edit ammo in the megamek lobby as well as mhq hangar.

MekHQ/resources/mekhq/resources/EditAmmoDialog.properties
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2026 The MegaMek Team. All Rights Reserved.
+# Copyright (C) 2005-2025 The MegaMek Team. All Rights Reserved.
```
IllianiBird: This is incorrect, it should be `2026` as this is a brand new object

MekHQ/src/mekhq/gui/dialog/EditAmmoDialog.java
```diff
@@ -0,0 +1,371 @@
+/*
+ * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
```
IllianiBird: And here

MekHQ/src/mekhq/gui/dialog/EditAmmoDialog.java
```diff
@@ -0,0 +1,371 @@
...
+        scrollPane.setVerticalScrollBarPolicy(ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED);
+        scrollPane.setHorizontalScrollBarPolicy(ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
+        add(scrollPane, BorderLayout.CENTER);
+        
+        // Button panel (always visible, not scrolled)
+        add(buildButtonPanel(), BorderLayout.SOUTH);
+        
+        // Calculate preferred size with scrollbar limit
+        int calculatedHeight = Math.max(DIALOG_MIN_HEIGHT, 
+            Math.min(DIALOG_MAX_HEIGHT, ammoBindings.size() * DIALOG_ROW_HEIGHT + DIALOG_BUTTON_ROW_HEIGHT));
```
IllianiBird: This doesn't read cleanly, it'd be best to calculate your dynamic height and then pass it into calculatedHeight rather than trying to do everything on the one line. It also doesn't look like we're scaling for GUI

MekHQ/src/mekhq/gui/dialog/EditAmmoDialog.java
```diff
@@ -0,0 +1,371 @@
...
+            String weaponName = getAmmoWeaponName(ammo);
+            String location = entity.getLocationName(ammo.getLocation());
+            String displayText = String.format("%s (%s)", weaponName, location);
+            mainPanel.add(new JLabel(displayText), gbc);
+            
+            // Ammo type
+            gbc.gridx = 1;
+            String ammoName = ammo.getType().getName();
+            if (ammo.isHotLoaded()) {
+                ammoName = String.format("%s %s", ammoName, resourceMap.getString("label.hotLoaded"));
```
IllianiBird: This should be using our new i18n classes (see MHQInternationalization for example) not String.format()

MekHQ/src/mekhq/gui/dialog/EditAmmoDialog.java
```diff
@@ -0,0 +1,371 @@
...
+            
+            // Shots remaining spinner
+            gbc.gridx = 2;
+            int currentShots = ammo.getBaseShotsLeft();
+            int maxShots = calculateMaxShots(ammo, weaponName, currentShots);
+            
+            SpinnerNumberModel spinnerModel = new SpinnerNumberModel(currentShots, 0, maxShots, 1);
+            JSpinner spinner = new JSpinner(spinnerModel);
+            spinner.setName(String.format("ammoSpinner_%d", row));
+            spinner.setToolTipText(String.format(
```
IllianiBird: Same here

MekHQ/src/mekhq/gui/dialog/EditAmmoDialog.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Builds the button panel with standard buttons.
+     * @return configured button panel
+     */
+    private JPanel buildButtonPanel() {
+        JPanel buttonPanel = new JPanel(new FlowLayout());
+        
+        JButton setAllZeroButton = new JButton(resourceMap.getString("btn.setAllZero.text"));
```
IllianiBird: We shouldn't be using `getString` any more, this is something that would be better off using getTextAt as that's safer than getString

MekHQ/src/mekhq/gui/dialog/ResolveScenarioWizardDialog.java
```diff
@@ -402,10 +404,19 @@ private JPanel makeUnitStatusPanel() {
...
         pnlUnitStatus.add(new JLabel(resourceMap.getString("totaled")), gridBagConstraints);
 
+        gridBagConstraints.gridx = 2;
+        pnlUnitStatus.add(new JLabel("View Unit"), gridBagConstraints);
+        
+        gridBagConstraints.gridx = 3;
+        pnlUnitStatus.add(new JLabel("Edit Unit"), gridBagConstraints);
+        
+        gridBagConstraints.gridx = 4;
+        pnlUnitStatus.add(new JLabel("Edit Ammo"), gridBagConstraints);
```
IllianiBird: We shouldn't be adding new hardcoded player-facing Strings

IllianiBird (changes requested):
So I've done a full review, first I think this is a great add. However there are some significant departures from our currently preferred style:

- Your legal statements look like they were copy-pasted without having the years updated correctly
- We should be avoiding String.format() in favor of our i18n classes, generally this is getFormattedTextAt()
- Similarly, we should be using getText instead of resources.getString()
- Finally I spotted a handful of hardcoded player facing Strings that's a no-no nowadays.

IllianiBird:
When working in both mhq and mm you need to create two PRs. One in the mm repo and one in the mhq one (this one), you then make a comment in the description that they rely on each other.

IllianiBird:
Given that this hasn't seen movement for more than a month, I'm going to go ahead and close it

PR #9633, opened 2026-07-11
Close #9630: Added the Ability to Remove Individual Parts Without Flipping Unit to Strip in Repair Tab

### LINKED ISSUE #9630: [RFE] Strip part right click option
### Enhancement Type *

New Feature

Sometimes its easier to just yank a damaged part out and replace it and worry about repairing it later.  Under the current system, to do this, Once I identify I don't want to repair a part, I need to swap the chassis to salvage, filter and find the part again, order it removed, then swap the chassis back to repair, and finally que up the replacement job.  This makes players far more likely to not consider this solution as it is much more effort.  

This idea brought to you by recent conversations on the discord about aborting repair jobs and my ADD. 

### Use Cases or Rationale

A small but helpful change to the interface would be to add to the right click menu for a damaged part an option to "strip part".  Selecting this would have the exact same effect as the process above, but now in one click instead of many.  Allowing a player the equivalent shortcut of "yank it! we will deal with it later".  Then they would only have to que up the replacement action. 

Implementation:  the right click option would use the currently selected tech and time settings.  Ideally it would list the TN on the right click option.  If making the right click option have dynamic text like that is impractical or not viable, then it should pop up a confirmation dialog that provides the needed information "This action will take 180 minutes and have a TN of 5, are you sure you want to remove this part?"

### Attach Files

_No response_

### MekHQ Suite Version *

51.0

### Final Verification

IllianiBird:
Close #9630

Process temporarily flips the individual part to 'salvage' and then performs to strip task. This helps streamline the process of manually repairing components.

Contains AI tests

### DIFF
--- a/MekHQ/src/mekhq/gui/adapter/TaskTableMouseAdapter.java
+++ b/MekHQ/src/mekhq/gui/adapter/TaskTableMouseAdapter.java
@@ -51,18 +51,21 @@ import mekhq.campaign.Campaign;
 import mekhq.campaign.events.AcquisitionEvent;
 import mekhq.campaign.events.parts.PartChangedEvent;
 import mekhq.campaign.events.parts.PartModeChangedEvent;
+import mekhq.campaign.events.parts.PartWorkEvent;
 import mekhq.campaign.events.units.UnitChangedEvent;
 import mekhq.campaign.parts.Armor;
 import mekhq.campaign.parts.Part;
 import mekhq.campaign.parts.equipment.AmmoBin;
 import mekhq.campaign.parts.meks.MekLocation;
 import mekhq.campaign.parts.missing.MissingPart;
+import mekhq.campaign.personnel.Person;
 import mekhq.campaign.personnel.skills.SkillType;
 import mekhq.campaign.unit.Unit;
 import mekhq.campaign.work.IAcquisitionWork;
 import mekhq.campaign.work.IPartWork;
 import mekhq.campaign.work.WorkTime;
 import mekhq.gui.CampaignGUI;
+import mekhq.gui.ITechWorkPanel;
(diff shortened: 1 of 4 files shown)


MekHQ/src/mekhq/gui/adapter/TaskTableMouseAdapter.java
```diff
@@ -195,6 +212,116 @@ private void reportAndTriggerEvent(IPartWork partWork) {
...
+     * even though they cannot actually be stripped from here: they are surfaced (disabled, with an explanation) so the
+     * player understands why, rather than silently omitting the option. Replacement tasks (missing parts) and spare
+     * parts are excluded outright, as are units already in salvage mode (where removal is already the default
+     * task).</p>
+     *
+     * @param part the task to test
+     *
+     * @return {@code true} if a "Strip Part" entry should be shown for this task
+     */
+    static boolean isStripCandidate(Part part) {
```
NickAragua, line 228: Don't we usually want to have access keywords for these functions? e.g. public/private etc

NickAragua (changes requested):

MekHQ/src/mekhq/gui/adapter/TaskTableMouseAdapter.java
psikomonkie: No access modifier in Java is "package-private" which may be intentional - I know I've utilized it recently for some projects.

IllianiBird: Psi is correct, it's intentional - it's a private method that I elevated to package-private so it could be accessed by Unit Tests

NickAragua (approved):

PR #9635, opened 2026-07-12
PR: Extract Force and Detachment from Campaign

psikomonkie:
`Campaign` is a very large class - dare I say “bloated”. This PR begins the process of extracting a player’s forces’s specific information into its own dedicated class that `Campaign` can delegate to.  - `AbstractForce` and `PlayerForce`: “Player” specific information - list of all personnel, finances, faction, etc.   - Currently `PlayerForce` uses an interface `SingleDetachmentForce` to hold the intentionally stunted logic that only works if a force is on an active contract non a single spot, like `getWarehouse()` (there will be multiple for a force, so this method won’t work eventually!) - `Detachment` - Implements `IPlace` - The `Hangar`, `Personnel`, and `Warehouse` live under this node.  

Once this is merged, I can begin deprecating the passthrough methods in `Campaign` - I intend to go through the methods that pass through from `Campaign` to `AbstractForce`/`PlayerForce` and `Detachment`, mark them deprecated, and replace the usages of `campaign.foo()` with `campaign.getPlayerForce().foo()` / `campaign.getPlayerForce().getForceDetachment().foo()`. Then I can mark the deprecated methods as forRemoval, and we can formally remove them in a few versions.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -387,7 +387,7 @@ public class Campaign implements ITechManager, IPlace {
     private Systems systemsInstance;
     private final Map<String, PlanetarySystem> planetarySystemOverrides = new LinkedHashMap<>();
     private final CampaignLocationManager locationManager = new CampaignLocationManager();
-    private final ForceLocationManager forceLocationManager = new ForceLocationManager(this);
+    private final ForceLocationManager forceLocationManager = new ForceLocationManager(getPlayerForce());
     private boolean isAvoidingEmptySystems;
     private boolean isOverridingCommandCircuitRequirements;
 
@@ -1719,7 +1719,7 @@ public class Campaign implements ITechManager, IPlace {
 
     @Nonnull
     public ForceLocationManager getForceLocationManager() {
-        return forceLocationManager;
+        return getPlayerForce().getForceLocationManager();
     }
 
     /**
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -51,8 +51,6 @@ import static mekhq.campaign.force.Formation.FORMATION_NONE;
 import static mekhq.campaign.force.Formation.FORMATION_ORIGIN;
 import static mekhq.campaign.force.Formation.NO_ASSIGNED_SCENARIO;
 import static mekhq.campaign.force.FormationType.STANDARD;
-import static mekhq.campaign.market.contractMarket.ContractAutomation.performAutomatedActivation;
-import static mekhq.campaign.mission.RandomFactionCamouflage.pickRandomCamouflage;
 import static mekhq.campaign.parts.enums.PartQuality.QUALITY_A;
 import static mekhq.campaign.personnel.PersonnelOptions.ADMIN_INTERSTELLAR_NEGOTIATOR;
 import static mekhq.campaign.personnel.PersonnelOptions.ADMIN_LOGISTICIAN;
@@ -166,8 +164,8 @@ import mekhq.campaign.finances.enums.TransactionType;
 import mekhq.campaign.force.CombatTeam;
 import mekhq.campaign.force.Formation;
 import mekhq.campaign.force.FormationType;
+import mekhq.campaign.force.PlayerForce;
 import mekhq.campaign.icons.StandardFormationIcon;
-import mekhq.campaign.icons.UnitIcon;
 import mekhq.campaign.location.ILocation;
 import mekhq.campaign.location.IPlace;
 import mekhq.campaign.location.LocationNode;
@@ -224,7 +222,6 @@ import mekhq.campaign.personnel.medical.MASHCapacity;
 import mekhq.campaign.personnel.procreation.AbstractProcreation;
 import mekhq.campaign.personnel.ranks.AutoAssignRankForCompanyGenerator;
 import mekhq.campaign.personnel.ranks.RankSystem;
-import mekhq.campaign.personnel.ranks.RankValidator;
 import mekhq.campaign.personnel.skills.ActionCheckResult;
 import mekhq.campaign.personnel.skills.Appraisal;
 import mekhq.campaign.personnel.skills.Attributes;
@@ -246,7 +243,6 @@ import mekhq.campaign.unit.HangarStatistics;
 import mekhq.campaign.unit.TestUnit;
 import mekhq.campaign.unit.Unit;
 import mekhq.campaign.unit.UnitOrder;
-import mekhq.campaign.unit.UnitTechProgression;
 import mekhq.campaign.unit.enums.TransporterType;
 import mekhq.campaign.universe.*;
 import mekhq.campaign.universe.enums.HiringHallLevel;
@@ -297,20 +293,17 @@ public class Campaign implements ITechManager, IPlace {
     // OK now we have more, parts, personnel, forces, missions, and scenarios.
     // and more still - we're tracking DropShips and WarShips in a separate set so
     // that we can assign units to transports
-    private final Hangar units = new Hangar();
+    /** The player's active force: faction identity, finances, reputation, and the owned hangar/warehouse/personnel. */
+    @Nonnull
+    private final PlayerForce playerForce;
     CampaignTransporterMap shipTransporters = new CampaignTransporterMap(this, CampaignTransportType.SHIP_TRANSPORT);
     CampaignTransporterMap tacticalTransporters = new CampaignTransporterMap(this,
           CampaignTransportType.TACTICAL_TRANSPORT);
     CampaignTransporterMap towTransporters = new CampaignTransporterMap(this, CampaignTransportType.TOW_TRANSPORT);
-    private Warehouse parts = new Warehouse();
     private final TreeMap<Integer, Formation> formationIds = new TreeMap<>();
     private final TreeMap<Integer, Mission> missions = new TreeMap<>();
     private final TreeMap<Integer, Scenario> scenarios = new TreeMap<>();
     private final Map<UUID, List<Kill>> kills = new HashMap<>();
-
-    // The main force's per-part requested stock percentages. Each player base owns its own instance; see
-    // IPlace.getRequestedStockLevels().
-    private final RequestedStockLevels requestedStockLevels = new RequestedStockLevels();
 
     private transient final UnitNameTracker unitNameTracker = new UnitNameTracker();
 
@@ -326,7 +319,6 @@ public class Campaign implements ITechManager, IPlace {
 
     private GameOptions gameOptions;
 
-    private String name;
     private LocalDate currentDay;
     private LocalDate campaignStartDate;
 
@@ -336,11 +328,6 @@ public class Campaign implements ITechManager, IPlace {
     private Formation formations;
     private Hashtable<Integer, CombatTeam> combatTeams; // AtB
 
-    private Faction faction;
-    private megamek.common.enums.Faction techFaction;
-    private String retainerEmployerCode; // AtB
-    private LocalDate retainerStartDate; // AtB
-    private RankSystem rankSystem;
 
     private final ArrayList<String> currentReport;
     private transient String currentReportHTML;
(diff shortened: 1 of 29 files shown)


MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+
+    /**
+     * The {@link Detachment}s this force is currently split across. A {@link PlayerForce} tracks exactly one; the
+     * located resources (hangar, warehouse, personnel, location) are owned by the detachment(s), not the force.
+     */
+    public abstract Collection<Detachment> getDetachments();
+
+    /** All units across every {@link Detachment} of this force. Multi-detachment-safe. */
+    public Collection<Unit> allUnits() {
+        return getDetachments().stream()
```
IllianiBird: I think we should reconsider this. Having a stream that calls over streams is likely to end up more constantly than we expect.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+                     .flatMap(detachment -> detachment.getHangar().getUnits().stream())
+                     .collect(Collectors.toList());
+    }
+
+    /** All personnel across every {@link Detachment} of this force. Multi-detachment-safe. */
+    public Collection<Person> allPersonnel() {
+        return getDetachments().stream()
```
IllianiBird: Same here

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+        }
+
+        // We need to know the old rank system for personnel processing
+        final RankSystem oldRankSystem = getRankSystem();
+
+        // And with that, we can set the rank system
+        setRankSystemDirect(rankSystem);
+
+        // Finally, we fix all personnel ranks and ensure they are properly set
+        allPersonnel().stream()
```
IllianiBird, line 298: A stream is good here, I don't mind streaming personnel because players quite frequently end up with thousands of personnel entries. If we wanted to optimize this would could check the size of the personnel array and then decide whether we use a stream or not based on the size of the array. I won't force you to do this, but it may be worth sitting with a profiler and seeing if the approach is warranted with 500, 1000, 5000, and 10000 personnel.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+
+    public void setCamouflage(final Camouflage camouflage) {
+        this.camouflage = camouflage;
+    }
+
+    public PlayerColour getColour() {
+        return colour;
+    }
+
+    public void setColour(final PlayerColour colour) {
```
IllianiBird, line 319: PlayerColour needs the `@Nonnull` tag

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+
+    /**
+     * Whether the force's MASH theatre capacity covers its patients assigned to doctors. Off-contract or in transit
+     * there is no cap. Takes a {@link Campaign} for the contract/location context the force does not hold.
+     */
+    public boolean getMashTheatresWithinCapacity(Campaign campaign) {
+        return !campaign.isOnContractAndPlanetside()
+                     ||
+                     calculateMASHTheaterCapacity(campaign) >=
+                           getHumanResources().getPatientsAssignedToDoctors().size();
```
IllianiBird, line 539: This can be cleaned up by changing it to...

```
if (campaign.isOnContractAndPlanetside()) {
    return true;
}

int patiensAssignedToDoctors = getHumanResources().getPatientsAssignedToDoctors().size()
return calculateMASHTheaterCapacity(campaign) >= patiensAssignedToDoctors;
```

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+            }
+
+            combatTeams.remove(formationId);
+        }
+
+        return combatTeams;
+    }
+
+    public ArrayList<CombatTeam> getCombatTeamsAsList(Campaign campaign) {
+        return getCombatTeamsAsMap(campaign).values()
```
IllianiBird: I don't think we can reasonably expect there to be enough combat teams to warrant a stream here

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+                        unit.removeTech();
+                    }
+
+                    unit.setTech(formationTech);
+                } else {
+                    String cantTech = formationTech.getFullName() +
+                                            " cannot maintain " +
+                                            unit.getName() +
+                                            '\n' +
+                                            "You will need to assign a tech manually.";
```
IllianiBird, line 785: I know this isn't your code, but migrating this hardcoded string (and any others) would be a great move. And there's no better time to do it than a big refactor like this.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+        return (unit == null) ? null : getFormation(unit.getFormationId());
+    }
+
+    /**
+     * Resolves the {@link Formation} a person belongs to: the formation of their assigned unit if they crew one,
+     * otherwise — for techs — the formation they are the assigned tech of.
+     *
+     * @return the person's formation, or {@code null} if none applies
+     */
+    public @Nullable Formation getFormationFor(final Person person) {
```
IllianiBird, line 890: We're definitely not going to have enough formations to warrant a stream

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+        }
+    }
+
+    /**
+     * Removes {@code formation} from the TO&E: unassigns its units (clearing their scenario if it was deployed),
+     * removes it from any scenario it was deployed to and from its parent formation, and clears any StratCon track
+     * assignments. The {@link Campaign} is supplied as a parameter for the scenario, contract, and combat-team updates
+     * this drives.
+     */
+    public void removeFormation(Formation formation, Campaign campaign) {
```
IllianiBird, line 815: We're already editing these lines via migration, let's clean up the variable names to, like `fid` to `formationId`. I won't bring this up again and leave it to your judgment.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+        Vector<Unit> networkedUnits = new Vector<>();
+        for (Unit unit : allUnits()) {
+            if (null != unit.getEntity().getC3NetId() &&
+                      unit.getEntity().getC3NetId().equals(u.getEntity().getC3NetId())) {
+                networkedUnits.add(unit);
+            }
+        }
+        for (int pos = 0; pos < Entity.MAX_C3i_NODES; pos++) {
+            for (Unit nUnit : networkedUnits) {
+                if (nUnit.getEntity().hasNavalC3()) {
```
IllianiBird, line 982: This can trigger a NPE if `getEntity` returns `null` (and it absolutely can). There's a lot of risk of that happening in this method, the more I look at it.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+                    nUnit.getEntity().setNC3NextUUIDAsString(pos, null);
+                } else {
+                    nUnit.getEntity().setC3iNextUUIDAsString(pos, null);
```
IllianiBird, line 985: Actually, now that I think about it _why_ are we having units with `null` entities? Could we not treat entity in a similar way to singletons:

`nUnit.getEntity()` calls...

```
public @Nonnull Entity getEntity() {
    if (entity == null) {
        entity = new Entity(); // or whatever we call to create a new entity for a unit
    }

    return entity;
}
```

Maybe this isn't in the scope of this particular PR. Maybe it's worth exploring, though?

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+        refreshNetworks(game);
+        MekHQ.triggerEvent(new NetworkChangedEvent(networkedUnits));
+    }
+
+    /**
+     * Removes {@code removedUnits} from their shared C3i/Naval C3 network and rebuilds the remaining members' node
+     * UUIDs so the network stays contiguous, refreshing networks against the supplied {@link Game}. All removed units
+     * are assumed to share the first unit's network id.
+     */
+    public void removeUnitsFromNetwork(Vector<Unit> removedUnits, Game game) {
```
IllianiBird, line 998: Lots of `getEntity` calls here, which can result in NPEs

MekHQ/src/mekhq/campaign/force/AbstractForce.java
```diff
@@ -0,0 +1,1294 @@
...
+                }
+            }
+        }
+        refreshNetworks(game);
+    }
+
+    /**
+     * Adds {@code addedUnits} to the C3i/Naval C3 network identified by {@code networkID}, rebuilding every member's
+     * node UUIDs, and refreshes networks against the supplied {@link Game}.
+     */
```
IllianiBird, line 1042: And here, I'll not blather on about it any more though

MekHQ/src/mekhq/campaign/force/Detachment.java
```diff
@@ -0,0 +1,168 @@
...
+    /** Discriminator identifying a detachment as a serialized {@link ILocation} reference. */
+    public static final String LOCATION_REFERENCE_TYPE = "detachment";
+
+    private final DetachmentLocationManager detachmentLocationManager;
+
+    // Owned resources — this detachment is the IPlace that owns them.
+    private final Hangar units = new Hangar();
+    private final Personnel personnel = new Personnel();
+    private final RequestedStockLevels requestedStockLevels = new RequestedStockLevels();
+    private Warehouse parts = new Warehouse();
```
IllianiBird, line 80: This is perfect for my needs

MekHQ/src/mekhq/campaign/force/ForceOptions.java
```diff
@@ -0,0 +1,86 @@
...
+ * <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
+ * affiliated with Microsoft.
+ */
+package mekhq.campaign.force;
+
+import mekhq.campaign.campaignOptions.CampaignOptions;
+import mekhq.campaign.universe.Faction;
+
+/**
+ * Force-scoped options for an {@link AbstractForce}, modeled loosely on {@link CampaignOptions}.
```
IllianiBird, line 39: I recommend we add some text here explaining _what_ kind of options should be here instead of CampaignOptions

MekHQ/src/mekhq/campaign/Campaign.java
IllianiBird, line 1: Image

This is a thing of beauty

IllianiBird (approved):
I would _prefer_ we have Copilot review this before merging. I've also commented one some "it'd be nice if..." changes.

In both cases, this is approved and I leave merging to your best judgment. Great work, Psi, seriously.

MekHQ/src/mekhq/campaign/force/ForceOptions.java
psikomonkie: Will be expanded, this isn't implemented.

MekHQ/src/mekhq/campaign/force/AbstractForce.java
psikomonkie: I'm not gonna touch the existing MekHQ-C3 logic, I'll be here all week!

MekHQ/src/mekhq/campaign/force/AbstractForce.java
psikomonkie: Streams removed except from the personnel usage

MekHQ/src/mekhq/campaign/force/AbstractForce.java
psikomonkie: Sure, done

PR #9647, opened 2026-07-14
PR: Remove Usages of Campaign's Pass Through Messages

psikomonkie:
Went through all of the ~200 methods in `Campaign` that've been converted to pass throughs as part of the extraction of `Force` (#9641) and used IntelliJ's inline method refactor to replace all of them. Exciting!

Some of the mocks got absolutely clobbered by IntelliJ's inline method refactor, so I let Claude Opus 4.8 loose on the broken tests. I intervened as necessary (looking at you, `PrisonerRansomEventTest`)`.


IllianiBird (approved):
I went through this and nothing seemed glaringly wrong.

There were a few cases of ugly code: abbreviated variable names, or chained references thing().thing().thing().thing() not sure if we want to do anything about that. My real fear there is that it'll make the code harder to read. But that's also something we can clean up later if it becomes a problem. This PR touches too many places to get lost in the weeds.

Normally I'm a bit of a hard ass on this sort of thing. However here I set the bar as "doesn't this leave the codebase worse than it was before" and I don't believe it does.

psikomonkie:
> There were a few cases of ugly code: abbreviated variable names, or chained references thing().thing().thing().thing() not sure if we want to do anything about that. My real fear there is that it'll make the code harder to read. But that's also something we can clean up later if it becomes a problem. This PR touches too many places to get lost in the weeds.

I agree! Using `Campaign` unnecessarily is something everyone's done, and I think this is like ripping off a bandaid and it's showing all the ugly spots that we forced something into `Campaign`. 

My hope is that this, combined with deprecating and removing all these methods from `Campaign`, will start forcing us to be more deliberate with when we actually need to pass around `Campaign`. Because do we really need `Campaign`, or do we just need the right `Force`? 

Anyway, thanks!

IllianiBird:
I think the big next step is singleton Campaign. Then we can go straight in and start removing heaps of these params

PR #9649, opened 2026-07-14
PR: Make Campaign Options Staticly Available

psikomonkie:
`CampaignOptions` is now statically available via `Campaign.getCampaignOptions()`. `setCampaignOptions`, when called on Campaign, will set a static `CampaignOptions` instance that can be accessed from anywhere. It will throw an exception if access without being set first. If `CampaignOptions` is changed via `setCampaignOptions()` it will log a debug message.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -5603,12 +5603,16 @@ public class Campaign implements ITechManager {
 
     }
 
-    public static CampaignOptions getCampaignOptions() {
+    public static CampaignOptions campaignOptions() {
         if (campaignOptions == null) {
             throw new IllegalStateException(
                   "CampaignOptions accessed before it was set; a Campaign must be initialized first.");
         }
         return campaignOptions;
+    }
+
+    public CampaignOptions getCampaignOptions() {
+        return Campaign.campaignOptions();
     }
 
     public void setCampaignOptions(CampaignOptions options) {
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -375,7 +375,7 @@ public class Campaign implements ITechManager {
 
     private final List<String> customs;
 
-    private CampaignOptions campaignOptions;
+    private static CampaignOptions campaignOptions;
     private RandomSkillPreferences randomSkillPreferences = new RandomSkillPreferences();
     private CampaignGUI gui;
 
@@ -498,7 +498,7 @@ public class Campaign implements ITechManager {
         this.player = player;
         this.game.addPlayer(0, this.player);
         currentDay = date;
-        campaignOptions = campaignOpts;
+        setCampaignOptions(campaignOpts);
         this.gameOptions = gameOptions;
         game.setOptions(gameOptions);
         this.systemsInstance = systemsInstance;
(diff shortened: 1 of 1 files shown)


IllianiBird (approved):

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5603,22 +5603,43 @@ public void payPersonnel(TransactionType type, Money quantity, String descriptio
...
-    public CampaignOptions getCampaignOptions() {
+    public static CampaignOptions campaignOptions() {
+        if (campaignOptions == null) {
+            throw new IllegalStateException(
+                  "CampaignOptions accessed before it was set; a Campaign must be initialized first.");
+        }
         return campaignOptions;
     }
 
+    public CampaignOptions getCampaignOptions() {
```
VicenteCartas, line 5614: We already have the static getter, I don't think we need this method too.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -5603,22 +5603,43 @@ public void payPersonnel(TransactionType type, Money quantity, String descriptio
...
-        // Check if blob crew was disabled for each role
-        boolean infantryWasEnabled = campaignOptions.isUseBlobInfantry();
-        boolean baWasEnabled = campaignOptions.isUseBlobBattleArmor();
-        boolean vehicleGroundWasEnabled = campaignOptions.isUseBlobVehicleCrewGround();
-        boolean vehicleVTOLWasEnabled = campaignOptions.isUseBlobVehicleCrewVTOL();
-        boolean vehicleNavalWasEnabled = campaignOptions.isUseBlobVehicleCrewNaval();
-        boolean vesselPilotWasEnabled = campaignOptions.isUseBlobVesselPilot();
-        boolean vesselGunnerWasEnabled = campaignOptions.isUseBlobVesselGunner();
-        boolean vesselCrewWasEnabled = campaignOptions.isUseBlobVesselCrew();
+        // campaignOptions is a static/global reference shared by every Campaign, so log an actual swap.
```
VicenteCartas, line 5619: This method should be static, I think it makes more clear to callers that the campaign options are a static object instead of having to look inside this class to figure out.

VicenteCartas (changes requested):

MekHQ/src/mekhq/campaign/Campaign.java
VicenteCartas: Actually, I see that the playerForce requires the campaign options below, I think it should either be a static also in playerForce, or we don't duplicate given that you can now always get it through the campaign.

VicenteCartas: My idea is that if CampaignOptions is a static, we should treat it as a static everywhere, if we treat it as static in some places and instance in others, I think it can get confusing.

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: I don't want to do a repo-wide refactor (yet), I think that might conflict with y'alls work. I'll gladly throw a `@deprecated` on this one to make it clear we should use the static one.

MekHQ/src/mekhq/campaign/Campaign.java
psikomonkie: This method should have JavaDocs explaining this, I agree it's silly expecting a dev to know that w/o looking at the contents of this method.

Re: 
> My idea is that if CampaignOptions is a static, we should treat it as a static everywhere

I agree, I just don't want to refactor the whole repo _yet_. I'll do repo-wide inline at the start of the next dev cycle - removing the pass through methods from `Campaign` mid dev cycle really messed up a few PRs.

VicenteCartas (approved):

PR #9650, opened 2026-07-14
PR: Store Campaign Options in the `CampaignOptionsStore`

psikomonkie:
No more bespoke fields and methods for `CampaignOptions`. The correct usage is now to use the `get(CampaignOption campaignOption)` method. New campaign options should be added in `CampaignOption` and is defined using a string key for the xml, its default value, and its type. 

Compiler will properly check the type when using `get` based on the `CampaignOption` passed in. Enjoy!

A vast majority of this PR is deprecation notices and indicating the correct way to replace the method. I have not removed the usages of these passthrough methods.


MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
```diff
@@ -121,698 +118,48 @@ public static String getTechLevelName(final int techLevel) {
...
-    private boolean limitByYear;
-    private boolean disallowExtinctStuff;
-    private boolean allowClanPurchases;
-    private boolean allowISPurchases;
-    private boolean allowCanonOnly;
-    private boolean allowCanonRefitOnly;
-    private int techLevel;
-    private boolean variableTechLevel;
-    private boolean factionIntroDate;
     private boolean useAmmoByType; // Unofficial
```
IllianiBird: This might wanna be moved to the static

IllianiBird (approved):
Approved, but I did spot a handful of variables that are still in CampaignOptions. I commented on one of them, but GitHub was having a heart attack trying to display all your changes so you'll need to scan through CampaignOptions to spot there others - there's a handful of booleans, at least one String, some other non-list stuff.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOption.java
```diff
@@ -0,0 +1,1022 @@
...
+import mekhq.campaign.randomEvents.prisoners.PrisonerCaptureStyle;
+import mekhq.campaign.stratCon.StratConPlayType;
+import mekhq.gui.campaignOptions.enums.ProcurementPersonnelPick;
+import mekhq.service.mrms.MRMSOption;
+
+/**
+ * A typed key identifying a single campaign option.
+ *
+ * <p>Each constant carries the option's value type ({@link #type()}), its default value
+ * ({@link #defaultValue()}), and the XML tag under which it is persisted ({@link #xmlTag()}). Because the key itself is
```
VicenteCartas, line 66: Is this used during save/load right now? I couldn't find the marshaller or
unmarshaller consuming `xmlTag()`, so as far as I can tell it isn't actually
used for persistence (unless I'm missing something).

It also looks like a few of these differ from the existing on-disk tags, for
example `useOriginFactionForNames` here versus `useFactionForNames` in the
marshaller/unmarshaller. Is the plan to wire this into persistence later? If
not, maybe this should be called an internal key instead, and we should clarify
that adding a new option still requires updating the save/load code.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOption.java
```diff
@@ -0,0 +1,1022 @@
...
+          of(Boolean.class, false, "useRandomPrisonerDivorce");
+    public static final CampaignOption<Integer> RANDOM_DIVORCE_DICE_SIZE =
+          of(Integer.class, 900, "randomDivorceDiceSize");
+    public static final CampaignOption<Boolean> USE_MANUAL_PROCREATION =
+          of(Boolean.class, true, "useManualProcreation");
+    public static final CampaignOption<Boolean> USE_CLAN_PERSONNEL_PROCREATION =
+          of(Boolean.class, false, "useClanPersonnelProcreation");
+    public static final CampaignOption<Boolean> USE_PRISONER_PROCREATION =
+          of(Boolean.class, true, "usePrisonerProcreation");
+    public static final CampaignOption<Integer> MULTIPLE_PREGNANCY_OCCURRENCES =
```
VicenteCartas, line 505: Is `defaultValue()` intended to return the actual default for a new campaign?

For example, this declares `0`, but the `CampaignOptions` constructor immediately
sets it to `50`. There are a few similar cases where mutable options declare an
empty collection and the constructor fills it afterward.

Current behavior is preserved because `CampaignOptions` performs that extra
initialization, but callers using the new public `defaultValue()` API would get
a different result. Could we reconcile both values?

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
```diff
@@ -1050,258 +234,44 @@ public CampaignOptions() {
...
-        setUseRandomClanPersonnelDivorce(true);
-        setUseRandomPrisonerDivorce(false);
-        setRandomDivorceDiceSize(900);
-
-        // Procreation
-        setUseManualProcreation(true);
-        setUseClanPersonnelProcreation(false);
-        setUsePrisonerProcreation(true);
+
         setMultiplePregnancyOccurrences(50); // Hellin's Law is 89, but we make it more common, so it shows up more
```
VicenteCartas: Here is the 50.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptionsStore.java
```diff
@@ -0,0 +1,76 @@
...
+    <T> T get(final CampaignOption<T> option) {
+        return option.type().cast(values.get(option));
+    }
+
+    /**
+     * @param option the option to write
+     * @param value  the value to store; its type is enforced by the compiler against the key
+     */
+    <T> void set(final CampaignOption<T> option, final T value) {
+        values.put(option, value);
```
VicenteCartas, line 76: This accepts null, so this could happen:

campaignOptions.set(CampaignOption.CHECK_MAINTENANCE, null);

public boolean isCheckMaintenance() {
    return options.get(CampaignOption.CHECK_MAINTENANCE);
}

And that would give us a null pointer exception. Could we enforce a Objects.requireNonNull?

VicenteCartas (dismissed):

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOption.java
psikomonkie: Good catch, fixed.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptionsStore.java
psikomonkie: Yes, I can! I added an annotation and `Objects.NonNull` to these methods to avoid this.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOption.java
psikomonkie: Good catch, updated here.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
psikomonkie: I agree, done.

MekHQ/src/mekhq/campaign/campaignOptions/CampaignOptions.java
psikomonkie: Good catch, updated.

IllianiBird (approved):

PR #9670, opened 2026-07-20
Contract Overhaul Project: Added Alternate StratCon Sector Generation Methods; Updated StratCon Hex -> Scenario Terrain Integration; Updated StratCon GUI; Added New StratCon GM Tools

IllianiBird:
# Introduction
This PR overhauls StratCon sector generation. Pulling in more planetary data that previous, with a move towards creating believable landmasses and interesting terrain. The broader goal is to prepare StratCon for the integration of Persistent OpFor and Persistent Deployment. No ETA is available for either.

I'm going to be real here, this is the most complex PR I have ever made. It is, by all accounts, the product of nearly two week's labor and encompasses nearly everything I've learned in my three years on the project. I'm incredibly excited for players to get to experience the new sector generation tech and I'm very proud of what I've accomplished here.

# AI Usage
This PR involves a lot of systems I haven't really touched before. A heck of a lot of math and algorithms. To ensure everything was well tested and thoroughly exercised I used AI a lot more than I would normally. This was to ensure the PR offers as stable an experience as possible.

I want to be clear why I used AI here, as I think context matters. I know our AI policy allows me, as a developer, to use AI to whatever level I need (outside of AI art, not that I'd want to). However, I know that many players have concerns over AI use, so I like to take the time to confirm that the overwhelming majority of this PR is human written. Where AI was used it was used to test, correct, and polish.

# What Changes for Existing Campaigns
Three new campaign options have been added, alongside updated default values. These values will be propagated onto existing campaigns:

- **Used Advanced Terrain Generation** enabled by default
- **Sector Count Method** set to _Alternate-Regimental_ (as chosen by user poll)
- **Sector Size Multiplier** set to 1.0

Sector data is saved into the campaign file, so existing contracts will be unaffected by these changes. Only new contracts will be affected by the sector generation changes in this PR.

The scenario terrain generation changes are non-optional and affect all scenarios generated after this PR is merged. It is likely that prior generated scenarios will inherit the old generation terrain. I'm not too worried about that as it'll be a transitory thing.

All GM tools will be available for all campaigns immediately. Allowing players to use them to influence existing contracts.

# Class & Package Structure
This PR adds a **lot** of new classes, so I went ahead and updated the StratCon package structure to try and make things more usable. I had considered waiting to do that in a follow-up PR, but it worked out easier to do it now rather than updating it later and then needing to reconsider what access level each method would need.

The below shows the new structure.

image

# Bug Fix
Some weather and temperature changes were made to ensure they applied to StratCon low-altitude scenarios. Previously we were incorrectly discarding weather and temperature when generating low-alt scenarios. This got wrapped into the greater PR, because I was relying on it for logical testing. This will affect all future scenarios, regardless of campaign options.

# GUI Update
Before getting into the nitty gritty I wanted to point out that the AO tab has been updated again, to accommodate the changes made in this PR.

**Threat Level** is a new displayed value, it's the likelihood that a player deploying to the sector will spawn a scenario. This was previously hidden behind the 'Edit SP/CVP' button. I opted to make it player-facing as it's something players could reasonably intuit via their strategic staff.

image

**Use Alternate StratCon Fog of War Display** is a new Client Option. Normally, unscouted hexes have a blue hint. If this option is enabled (enabled by default) the hex is instead drawn at 50% opacity. This was done to better allow players to read the terrain prior to scouting. This will be very important once we introduce persistent StratCon deployment.

## Option Off
image

## Option On (new default)
image

# Sector Generation
The bulk of this PR concerns how sectors are generated, so that's what I'll start with.

## Class Overview
I wanted to start with a class overview, otherwise it'll be the first thing Psi asks me for when I request their review :D

I sure hope nobody is trying to read this on mobile, if you are, I'm sorry!

```mermaid
classDiagram
    direction TB

    class StratConContractInitializer {
        +initializeTrackState(SectorSpec, PlanetProfile) StratConTrackState
    }
    class StratConSectorPlanner {
        +generateSectorSpecs(int, StratConSectorCountMethod, int) List~SectorSpec~
    }
    class StratConSectorCountMethod {
        <<enumeration>>
        LEGACY
        ALTERNATE
        ALTERNATE_CONDENSED
        ALTERNATE_REGIMENTAL
    }
    class SectorSpec {
        <<record>>
        +int requiredLances
        +LatitudeBand latitudeBand
    }
    class PlanetProfile {
        <<record>>
        +from(contract, campaign) PlanetProfile
        +habitability() double
        +sizeFactor() double
    }
    class StratConSectorShape {
        +selectProfile() SectorShapeProfile
    }

    class ISectorGenerationStrategy {
        <<interface>>
        +initializeTrack(track, profile, band, allowCities)
        +regenerateTrack(track, profile, band, allowCities)
        +connectFacilitiesToRoads(track, facilityCoords)
    }
    class ImprovedStratConSectorGeneration
    class LegacyStratConSectorGeneration

    class StratConSectorGenerator {
        +generate(track, profile, band, allowCities)
    }
    class GenerationStage {
        <<enumeration>>
        BIOME
        OCEANS
        MOUNTAINS
        TERRAIN_FIELDS
        TERRAIN_FILL
        CITIES
        FARMLAND
    }
    class PipelineOrder {
        +enter(GenerationStage)
        +skip(GenerationStage)
    }

    class StratConBiomeManifest {
        +getMapTypesForTerrain(String) MapTypeList
        +getTempMap(String) TreeMap
    }
    class StratConHydrology {
        +selectProfile(int waterPercent) HydrologyProfile
    }
    class StratConOrogeny {
        +selectProfile(PlanetProfile) OrogenyProfile
    }
    class StratConUrban {
        +selectProfile(PlanetProfile) UrbanProfile
    }

    class StratConOceanPlacer {
        +placeOceans(track, type, targetHexes, terrain)$
    }
    class StratConMountainPlacer {
        +placeMountains(track, terrain, orogeny, gravity, habitability)$
    }
    class StratConTerrainFields {
        +compute(track, band, windDirection)$ StratConTerrainFields
    }
    class StratConTerrainFiller {
        +fill(track, biome, profile, fields)$
    }
    class StratConCityPlacer {
        +placeCities(track, profile, urban)$
    }
    class StratConFarmPlacer {
        +placeFarms(track, profile, urban)$
    }
    class StratConRoadPlacer {
        +recalculateRoads(track, extraEndpoints)$
    }

    class StratConTrackState {
        +setTerrainTile(coords, terrain)
        +getCities() Set~StratConCoords~
    }

    StratConContractInitializer ..> StratConSectorPlanner : how many
    StratConContractInitializer ..> PlanetProfile : planet data
    StratConContractInitializer ..> StratConSectorShape : proportions
    StratConContractInitializer ..> ISectorGenerationStrategy : fills each sector
    StratConSectorPlanner ..> SectorSpec : produces
    StratConSectorPlanner ..> StratConSectorCountMethod : campaign option

    ISectorGenerationStrategy <|.. ImprovedStratConSectorGeneration
    ISectorGenerationStrategy <|.. LegacyStratConSectorGeneration
    ImprovedStratConSectorGeneration ..> StratConSectorGenerator
    ImprovedStratConSectorGeneration ..> StratConRoadPlacer : roads once facilities exist

    StratConSectorGenerator *-- PipelineOrder : enforces order
    PipelineOrder ..> GenerationStage : declared prerequisites

    StratConSectorGenerator ..> StratConBiomeManifest : 1 select biome
    StratConSectorGenerator ..> StratConOceanPlacer : 2
    StratConSectorGenerator ..> StratConMountainPlacer : 3
    StratConSectorGenerator ..> StratConTerrainFields : 4
    StratConSectorGenerator ..> StratConTerrainFiller : 5
    StratConSectorGenerator ..> StratConCityPlacer : 6
    StratConSectorGenerator ..> StratConFarmPlacer : 6

    StratConOceanPlacer ..> StratConHydrology
    StratConMountainPlacer ..> StratConOrogeny
    StratConCityPlacer ..> StratConUrban
    StratConFarmPlacer ..> StratConUrban

    StratConSectorGenerator --> StratConTrackState : mutates in place
```

The $ suffixes mark static methods — every placer is a stateless utility rather than something the generator holds.

The 1–6 edge labels are the pipeline order, and it is enforced rather than conventional: several stages read what earlier ones wrote off the track instead of receiving it as an argument, so running them out of order would otherwise produce a quietly worse sector. StratConTerrainFields measures each hex's distance to water and relief, so it is meaningless before oceans and mountains exist; farmland radiates from cities, so placing it first yields none. PipelineOrder turns either mistake into an IllegalStateException on the first sector generated. Roads are deliberately not a pipeline stage: the network must also serve facilities, which are placed by the contract initializer after terrain is done, so the strategy lays roads as a separate step once facilities exist.

| Class | What it does |
| --- | --- |
| `StratConContractInitializer` | Entry point. Resolves the planet, asks the planner how many sectors, builds each track at its rolled dimensions, hands it to the strategy, then seeds facilities and objectives. Also hosts the GM operations. |
| `StratConSectorPlanner` | The whole-contract "how many, how big" step. Turns required combat teams into one `SectorSpec` per sector according to the chosen Sector Count Method, sharing every team evenly across the sectors. Condensed caps the count at ten and Regimental targets one sector per regiment, but both overflow into further sectors rather than let any single sector exceed the maximum size. |
| `StratConSectorCountMethod` | The Sector Count Method campaign option: Legacy, Alternate, Alternate-Condensed, or Alternate-Regimental. Alternate-Regimental is the default for new and existing campaigns. |
| `SectorSpec` | Describes one sector before any terrain exists: its share of the contract's teams, and its latitude. |
| `PlanetProfile` | Immutable snapshot of the destination planet — temperature, diameter, water, atmosphere, gravity, population. Every unknown gets a neutral Terra-like fallback, so generation never depends on complete universe data. |
| `StratConSectorShape` | Rolls the sector's proportions (square through corridor). Deliberately a plain weighted roll: area already follows the planet, so shape is just variety. |
| `ISectorGenerationStrategy` | The "how is a sector built" seam. Isolates terrain from sizing, scenario cadence and enemy generation. |
| `ImprovedStratConSectorGeneration` | Thin adapter onto the pipeline below. Also lays the road network as a separate step, once the contract initializer has placed facilities. |
| `LegacyStratConSectorGeneration` | The original random biome-stripe placer, unchanged. Still fully available, but no longer the default — Advanced Terrain Generation ships enabled. |
| `StratConSectorGenerator` | The pipeline. Picks the biome from temperature, then runs each placer in a strict order against one track. |
| `GenerationStage` | The seven pipeline stages and the stages each one requires to have run first. |
| `PipelineOrder` | Records progress through the pipeline and throws if a stage runs before its prerequisites, naming the missing one. Stages deliberately not run — cities and farmland under the Ares Conventions — are skipped rather than omitted, so what follows still counts them satisfied. |
| `StratConBiomeManifest` | The authored terrain catalogue: terrain types and categories, temperature-keyed biomes, battle-map pools, facility pools, tile sprites. No terrain name is hard-coded in the generation code. |
| `StratConHydrology` | Picks a water-layout profile, Gaussian-weighted on the planet's water coverage alone. |
| `StratConOrogeny` | Picks a mountain-building profile — multifactor: the geometric mean of Gaussian factors for gravity, temperature and water, times categorical multipliers for rocky, icy and airless worlds. |
| `StratConUrban` | Picks a settlement profile — geometric mean of factors for log₁₀ population, water, habitability and tech. |
| `StratConOceanPlacer` | Paints water in the profile's shape. Water-dominant profiles grow the *land* and flood the rest, so land reads as a coherent mass. No profile bisects the sector with a channel. |
| `StratConMountainPlacer` | Ranges, ridges, uplands, scattered peaks or a volcanic arc. Feature count scales with gravity; a per-profile fraction is volcanic; never overwrites ocean. |
| `StratConTerrainFields` | Computes three per-hex fields once water and relief exist: moisture, distance to relief, and rain shadow along the prevailing wind. |
| `StratConTerrainFiller` | Fills every remaining dry hex, weighting the biome by planetary bias plus those fields, then smoothing into coherent patches. |
| `StratConCityPlacer` | Cities as an overlay on base terrain; count scales with population, arrangement from the urban profile. Suppressed when both sides observe the Ares Conventions. |
| `StratConFarmPlacer` | Farmland radiating from each city and thinning with distance. Unlike cities it is a *base* terrain, and only replaces hexes the manifest marks arable. |
| `StratConRoadPlacer` | Minimum spanning tree (Prim's) over Dijkstra paths. Mountains are passable but costly; a road may bridge one ocean hex but never two consecutively; networks reaching the edge branch off-map. Runs after facilities are seeded, so the network serves them too. |
| `StratConTrackState` | The sector itself. Every stage mutates this one object, and it is also the save format (JAXB-bound). |

## The Sector Generation Journey
This is a damn complex PR, so I want to focus on how systems interact with each other, rather than trying to get into the nitty gritty. This journey can largely be summarized into n steps:

1. Pre-Terrain Generation
2. Terrain Generation
3. Post-Terrain Generation

### Pre-Terrain Generation
- The contract provides 'required combat teams' a value that is largely translated from 'required combat elements'. This value tells us roughly how many combat teams the player can expect to field in the contract.
- **How Many Sectors** Our next step depends on what sector generation campaign option the player has chosen:
  - Legacy: StratCon generates one sector for every three combat teams the contract requires, plus a smaller sector for any remainder. Sectors use the historical fixed size rather than being scaled to the force that must scout them. This is how StratCon behaved prior to 51.01.
  - Alternate:  StratCon generates roughly one sector for every nine combat teams the contract requires. This produces fewer, larger sectors than the legacy count for the same force size, with each sector scaled to the recon that must sweep it.
  - Alternate-Condensed: As Alternate, but StratCon will not generate more than ten sectors. A  force large enough to generate more instead receives ten proportionally larger sectors, preserving the total map area. These enlarged sectors require proportionally more combat teams to scout and hold. The one exception is a contract so large that ten sectors could not hold it without each fronting more than a regiment; past that point further sectors are generated rather than letting any one of them exceed the maximum sector size.
  - Alternate-Regimental: As Alternate, but StratCon generates roughly one sector for every forty-five combat teams the contract requires - a five-battalion regiment. Anything up to a regiment is fought over a single sector, scaled to the whole force, so this produces far fewer and far larger sectors than any other method. Where a regiment is more ground than one sector can hold - on a wet world, a large world, or with a raised size multiplier - the contract is split into further sectors rather than any of them being cut short, so the total map area is preserved either way.
- Combat team 'requirements' are shared evenly across the final sector count (excluding Legacy). 'Requirements' is used here to denote how many combat teams the player is implicitly required to 'deploy' to the sector to effectively resolve it. It's used to determine Sector size only. In the event of uneven combat teams the earliest sectors receive the overflow in order. i.e. Sector '0' will receive overflow first, then Sector '1', etc. All sectors are guaranteed at least one combat team.
- It is at this point we generate a random `LatitudeBand` for each sector. Nothing about the contract or planet influences latitude. Latitude is used later for terrain generation.

#### Per Sector Processing
- Sector temperature is set to planet temperature + latitude offset +/- 5 (Legacy uses equatorial temperature + a random value between -40 and +10). The variance is so that two sectors in the same latitude band can, potentially, land in different biomes. This isn't important now, but is very important later.
- We also determine whether cities are allowed in the sector. Sectors will not contain cities if both belligerents are Ares Conventions signatories during the years in which the Ares Conventions are active. GMs can still manually add cities, should they wish (see much later).
- Next we determine sector size.
  - Legacy: the sector equals 28 hexes per required combat team. This is unchanged.
  - Alternate: First we determine how many recon teams (patrols) the sector is balanced for. This is equal to the number of combat teams required by the sector divided by 3 (no rounding). A sector for four teams gets 1.33 recon teams' worth of hexes, not 1.
  - Second we determine how many playable hexes will be in the sector. Playable hexes are defined as any hex that isn't ocean (of any type). Playable hexes == recon teams x 36 x planet size factor x campaign multiplier. I picked 36 as it budgets a single Patrol scouting 3 hexes/week for 3 months. Players with better patrols and/or a higher tempo will fulfill this quicker than a player with poorer patrols and/or a lower tempo. Roughly, though, this should be balanced for general play. Players may struggle with very short 1-2 month contracts, but that's decidedly ok and represents the difficulty of such a short, intense, contract.
  - Next, the sector is grown to account for water. Water is largely influenced by the surface water of the planet. The intention here is that the playable area will always be roughly the same no matter how wet the world.
  - Sector size has a ceiling of 1024 hexes (roughly 45 combat teams). A large, wet, condensed sector with a doubled multiplier can otherwise ask for thousands, every one of which is repainted and walked by each placer. We had to pick a limit and 1024 hexes seemed reasonable.
  - After we have sector size we determine its shape. This just determines whether we have a square or rectangle sector and then what the dimensions are.
- Alternate-Condensed: If a contract would generate sectors greater than 10, those additional sectors are merged into sectors 1-10 (starting with 1, then 2, then 3, and so on). This works for both legacy and alternate sector count modes. If all sectors have reached the maximum size (and count) we start overflowing into new sectors, even if this would take us over 10.
- Alternative-Regimental: all sectors are merged into one giant sector where possible. If that sector reaches the maximum size, we overflow into a new sector.

### Terrain Generation (yippee)
- We convert the sector's temperature to Kelvin, this is used to pick a terrain profile based on terra. While it's true that now all planets are going to be earth-like we have to build off something. So why not our giant wet space rock? Either way, this basically dictates everything downstream.
- Next we pick the ocean terrain. For the most part this is just picking whether it's wet or icy. On top of that we pick a hydrology profile (I learned so many fancy words while researching this). This is gaussian-weighted based on the planet's water coverage. So a wet world will trend towards archipelago and a dry one towards inland. This largely determines the 'pattern' of water. Archipelago, for example, is characterized by multiple island chains. Inland, on the other hand might only have a couple of spots of water. While coastal has, well, a coastline. This is also the point where we roll water volume to determine how much water actually turns up. Finally, we place the water. Water dominant profiles 'grow' the dry land first and then flood the sector. While other profiles 'grow' the water directly.
- After that we pick a orogeny profile (told you, lots of fancy words)! This profile is the product of geometric average multiplied by categorical multipliers for rocky, icy and airless worlds. Orogeny dictates how mountainous the sector is and also the shape of those mountain ranges. Once we have a profile we place the mountains. Some mountains may generate as volcanos, based on the orogeny profile (which autocorrect keeps trying to change to urology).
- Once we have water and rocks we roll wind direction. This is used to fuel the next stages of the simulation and while wind changes we use this value as a general abstraction of history weather patterns. Because the Good Ideas Fairy had me in a chokehold and I thought it sounded cool.
- Now that we have water and relief, we give every hex a moisture value (closeness to ocean hexes), a distance-to-relief value, and a rain-shadow flag if the hex sits leeward along the wind direction. Then every empty hex draws terrain from the biome by weight, combining planet-wide bias (breathability, water, composition) with its own fields. We then smooth the results into coherent patches, so it doesn't look random.
- If cities are allow we pick an urban profile (this is a geometric average over log10 population, water, habitability, and tech). We place cities based on the profile, with some profiles preferring to cluster cities, others spreading them out. This is heavily influenced on population, so more densely populated planets will have densely populated cities. While planets with low population density will trend towards the more open profiles. Once the cities have been placed, we next place farmland radiating from it (only into hexes we've marked as 'arable').
- A note on urban locations: unlike older StratCon generation methods urban 'hexes' are not hexes at all, rather they are a characteristic that is applied to the hex it spawns in. This means that cities can appear in **any** non-ocean terrain.

### Post-Terrain Generation
- Once the sector has been generated we return to the original contract initializer code to place facilities and objectives.
- Facilities may occupy, at most, half of the sector's dry land
- If a facility is owned by a planetary owner then that facility will be hooked into the existing road network.
- Finally, we build the road network. The road network uses [Prim's minimum spanning tree](https://www.geeksforgeeks.org/dsa/prims-minimum-spanning-tree-mst-greedy-algo-5/) over [Dijkstra paths](https://www.geeksforgeeks.org/dsa/dijkstras-shortest-path-algorithm-greedy-algo-7/) to find the shortest paths connecting each city, farm, and facility owned by one of the planetary owners. We rate mountains as passable but expensive, and allow for bridges over one hex of water (at high cost). Every network that can reach the edge of the map will branch off-map (to give the impression of connecting to a larger planetary network.

# Scenario Generation
In addition to reworking sector generation, this PR also updates the interaction between StratCon hexes and scenario terrain. The goal here is to improve player immersion by more tightly binding the scenario terrain they receive and the terrain in StratCon.

Unlike sector generation, this portion is **non-optional** and does not come with campaign options.

It's worth noting that 'fixed maps' (player authored maps, as opposed to generated maps) are excluded from most of the actions taken in Phases One and Two. This is done to preserve the integrity of player authored maps.

## Phase One
- Temperature now comes from the hex, not the sector. Hex temperature is the sector's temperature plus the hex's offset. So volcanic hexes are hotter than the sector average, while glaciers would be colder. This is skipped for space scenarios.
- The hex terrain dictates what mapgen profile we use. If there is no facility on the map we use just the hex terrain. If there is a facility we use a special facility pool for that hex type. This means that a Dr Evil Volcano Lair facility will spawn a facility map, on volcanic ground. In the event we have no facility pool for the hex terrain we fallback to the legacy StratCon temperature-banded facility mapgen system. Obviously we skip this for space. Space is big and largely devoid of terrain.
- Next we choose the actual board (including fixed map options, if appropriate).
- Finally we set light and weather.

## Phase Two
- Phase one was focused on the initial setup, two is where we edit the existing board to apply special modifiers to shape the board so that it more closely reflects the terrain in StratCon.
- First we apply terrain emphasis:
  - VEGETATION — dense woods, more and larger, some heavy; Swamp gets swamp params instead
  - BARREN — rough ground and sand
  - HILLS — elevation 60
  - MOUNTAIN — elevation 70, cliffs, and snow-capped peaks when the terrain is cold
  - AGRICULTURE — planted fields
  - NEUTRAL, LUNAR, VOLCANIC, URBAN, OCEAN — deliberately nothing, because they get their whole look from dedicated mapgen themes
- Then we apply tileset based on the terrain theme: mars, lunar, volcano, snow, jungle, tropical, swamp, desert.
- If the hex has a road we make sure the map has a road. This road won't match the exact direction see in StratCon, but the board will have a road. This is the best we can do without some serious adjustments and wizardry.
- If the hex borders ocean, we increase minimum water spots so that a coastal hex spawns scenarios near water.
- One massive change has been made to urban 'hexes'. In 'old' StratCon we used specific urban tiles. Now an urban sprite can appear on any hex. So you can have a volcano city, or a forest city, or whatever. This is the point in generation where we add the urban parameters to the board. This is based on the sector's urban profile:
  - TOWN below the first threshold, then HUB, then METRO, with block count, 2–10 stories, 40–100% built-up density, and some paving around the buildings. All urban hexes also have roads (due to automatic connection to the road network).

# GM Tools
One of the things I wanted to do, while working on this rewrite is to go in and provide human GMs more tools to sculpt their StratCon experience. The goal is to put power back into the hands of human GMs.

## GM Toolbelt
The following GM tools have been added to the StratCon AO tab. These are all available as buttons beneath the sector map panel.

### Scout Sector
Marks every hex in the sector revealed. A blunt "I've seen it all" for when a GM doesn't want to play out reconnaissance.

### Reset Sector Fog
Clears every revealed hex (excluding ocean, as that's always scouted).

### Regenerate Sector
Completely rebuilds the sector's terrain, including latitude, temperature, and all the good stuff. This allows GMs to quickly 'roll up' a sector that matches their designs.

### Toggle Hidden Objects
GM reveals (or hides) any hidden scenarios or facilities. Essentially the equivalent of the old 'Reveal Sector' and 'Hide Sector' GM options.

### Edit SP
Sets the contract's Support Points.

### Edit CVP
Sets the contract's Campaign Victory Points.

### Change Terrain
Opens the terrain painter pallet.

image

This allows players to paint terrain directly to the sector. In this way a GM can completely hand-craft a sector in whatever fashion they wish. Three brush sizes exist, and painting can be performed by clicking on the target hex or left-click dragging.

## Resize Sector
Allows the GM to grow or shrink the sector to a minimum of 4 x 4.

image

When shrinking we try to move scenarios and facilities around to make sure they fit, if this isn't possible the action is blocked and the payer informed what steps they need to take. Player forces deployed to hexes that are being culled either travel with the target scenario or facility, or they are immediately recalled.

When growing a sector we copy the terrain from its already filled neighbor. This creates very ugly sectors. The intention is that players will use the terrain painter to 'pretty up' their new, bigger, sector.

## Hex Context Menu
The hex context menu (the right-click menu) has also had some changes:

### Add Facility
The previously existing Add Facility option is largely unchanged. The only change of note is that adding a facility owned by one of the planetary owners, will immediately cause it to be added to the road network.

### Remove Facility
As above, this is largely unchanged. Removing a facility owned by one of the planetary owners will cause the road network to be recalculated.

### Switch Owner
Unchanged. Does not update the road network.

### Remove Scenario
Unchanged

### Add City
Adds a city to the chosen hex, automatically connecting it to the road network.

### Remove City
The opposite of the above, also automatically updates the road network.


IllianiBird:
All copilot feedback is legitimate, as is some CodeQL

IllianiBird:
Flipping this to live. Good luck to whoever reviews it. May the odds be forever in your favor.

BLR-IIC:
I look forward to playing around with this.

IMO use as much AI as you need.  I don't think it's any different from asking my wife to provide input on stuff.  It's ultimately a human's responsibility to make sure it works the way everyone expects it to  =)

MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConContractInitializer.java
```diff
@@ -313,82 +393,671 @@ public static int getScenarioOdds(StratConContractDefinition contractDefinition)
...
+
+            StratConFacility facility = track.getFacility(source);
+            track.removeFacility(source);
+            track.addFacility(destination, facility);
+            track.moveObjective(source, destination);
+            moveAssignedForces(track, source, destination);
+        }
+
+        for (StratConCoords source : displacedScenarios) {
+            StratConCoords destination = getUnoccupiedCoords(track);
```
VicenteCartas: P1: Please preserve facility/scenario co-location when relocating occupants. Facility scenarios are created on the same coordinate as their facility, and completion/capture/destruction later resolves the facility via `track.getFacility(scenario.getCoords())`. Here facilities are moved first; `getUnoccupiedCoords(track)` then excludes the moved facility's destination, so a scenario that shared `source` is necessarily moved elsewhere. The same two-pass pattern exists in `relocateOccupantsOffOcean`. After a GM resize or regeneration affects a facility battle hex, resolving that battle can no longer capture, destroy, or switch ownership of the facility. Could the co-located facility and scenario be relocated as one occupant in both paths, with their objective and assigned forces moved once?

VicenteCartas (commented):
One P1 state-integrity issue found in the new sector relocation tools.

VicenteCartas (approved):
Left a comment, all else looks good.

MekHQ/resources/mekhq/resources/AtBStratCon.properties
```diff
@@ -138,14 +138,49 @@ batchallBreach.button.cancel=(Cancel) On second thoughts...
...
-stratConTab.manageCampaignState.text=Manage SP/CVP
-stratConTab.status.noContract=No active contract is selected, the contract has not started, or you are not in the target system.
-stratConTab.status.notStarted=Contract has not started.
-stratConTab.status.header=<html><b>{0}:</b> {1}<br/><i>{2}</i>
-stratConTab.status.expired=<br/>Contract term has expired!
+stratConTab.gmTools.title=GM Tools
+stratConTab.scoutSector.text=Scout Sector
+stratConTab.scoutSector.tooltip=Reveals every hex in the currently selected sector, as if fully scouted.
+stratConTab.resetSectorFog.text=Reset Sector Fog
+stratConTab.resetSectorFog.tooltip=Un-reveals every hex in the currently selected sector (open water stays visible), so scouting can be re-tested.
```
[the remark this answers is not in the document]
IllianiBird: Incorrect

MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConContractInitializer.java
IllianiBird: Fixed

PR #9671, opened 2026-07-20
Order contracts from newer to oldest

### LINKED ISSUE #9656: [RFE]Change the order of the mission to be show the most recent contract instead of the oldest contract
### Enhancement Type *

For now you can always see the oldest contract on the Briefing Room - Current Mission when you do not have any active missions and the order is fixed to most old, but this is weird. Isn't it should be the most recent contract instead? Else all you got is always the oldest one no matter how many decades your forces have fought.

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

0.51.00

### Final Verification

- [ ] I've discussed or asked about this enhancement on MegaMek Discord

VicenteCartas:
Fixes #9656


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1355,14 +1355,17 @@ public Collection<Mission> getMissions() {
...
-                                   .thenComparing(m -> (m instanceof Contract) ?
-                                                             ((Contract) m).getStartDate() :
-                                                             LocalDate.now()))
+                     .sorted(Comparator.comparing((Mission mission) -> mission.getStatus().isCompleted())
+                                   .thenComparingLong(mission -> {
+                                       long startDay = (mission instanceof Contract) ?
+                                                                     ((Contract) mission).getStartDate().toEpochDay() :
+                                                                     LocalDate.now().toEpochDay();
+                                       return mission.getStatus().isCompleted() ? -startDay : startDay;
+                                   }))
```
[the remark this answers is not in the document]
VicenteCartas: Correct, fixed.

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1355,14 +1355,17 @@ public Collection<Mission> getMissions() {
     }
 
     /**
-     * @return missions List sorted with complete missions at the bottom
+     * @return missions sorted with active missions first, followed by completed contracts from most to least recent
```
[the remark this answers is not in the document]
VicenteCartas: Updated

MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -1355,14 +1355,22 @@ public Collection<Mission> getMissions() {
     }
 
     /**
-     * @return missions List sorted with complete missions at the bottom
+     * @return missions sorted with active missions from oldest to newest, followed by completed missions from newest to
+     *       oldest; active missions without a start date use the campaign date, while completed missions without one
+     *       sort last
      */
     public List<Mission> getSortedMissions() {
         return getMissions().stream()
```
IllianiBird: I don't think a stream is warranted here. Especially with the amount of additional logic that's being passed in. I'd prefer for us to use a traditional loop. It'll be cleaner and more performative.

Generally we should be avoiding streams for anything where we can't reasonably expect to be handling thousands of objects.

IllianiBird (changes requested):
One change I think we should make.

MekHQ/src/mekhq/campaign/Campaign.java
VicenteCartas: I'll change this.

VicenteCartas: Changed this and extracted the big lamba into a private method as I remembered you have given me this feedback in the past.

IllianiBird (approved):

PR #9681, opened 2026-07-22
Fix #5924: Fixed MekHQ & MegaMek's Independent Game Options Instances Becoming Desync'd

### LINKED ISSUE #5924: Princess using illegal ammo
### Brief Description *

Princess is using some clan ammos in 50.02. Noticed clan rocket pods and a clan TAG. Only carefully observed on aircraft (have not paid attention to turrets or ground units).

### 3. Steps to Reproduce

Enter combat with enemy aircraft while "Auto Configure Bot Munitions" is checks. (Have not seen this behavior with it unchecked).

### Attach Files

Example combat where her fighters hit me with Clan TAG. If you need my Campaign save, let me know. I frequently lurk in the MekHQ discord.

[Round-6-autosave_2025-01-28_21-15-39.sav.gz](https://github.com/user-attachments/files/18582388/Round-6-autosave_2025-01-28_21-15-39.sav.gz)

### Severity *

### MekHQ Suite Version *

50.02

### Operating System *

Win 11 Home

### Java Version *

Adoptium

### Final Verification

IllianiBird:
Fix #5924

Here's a fun one:

- `Campaign` has a copy of `GameOptions`
- `Game` has a copy of `GameOptions`

Both `GameOptions` objects share the same instance. Hurray

_Except_ if we apply a campaign preset. At that point we correctly replace `Campaign`'s `GameOptions` instance, but we don't hand the new one to `Game`. Why is this a problem? Generally it's not, except there is one tinsy winsy little issue: every update from that point on only lands on `Campaign`'s copy. Including the "sync the allowed year to the campaign date" step that runs when you pick your start date and again before every scenario.

That means the date in `Campaign` changes to match and so does `Campaign`'s `GameOptions` reference. But `Game` is still clutching the stale object it was constructed with, frozen at the default year (3067). So `Game` thinks the Jihad is about to kick off while `Campaign` (and the player) know they're in [insert year] (3019, in my case).

Thus resulting in my 3019 game being littered with future munitions, because most systems were checking against `Campaign`'s year. But autoammo was checking against Game's incorrect year. Whoops!

Now `setGameOptions` points `Game` at the new instance too, so the two can't diverge again. There's also a regression test to make sure we don't backslide.

This bug has been in the game since at least 2020.

Contains AI tests.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -7338,6 +7338,11 @@ public class Campaign implements ITechManager {
 
     public void setGameOptions(final GameOptions gameOptions) {
         this.gameOptions = gameOptions;
+        // Keep the Game's reference in sync: MegaMek code (e.g. TeamLoadOutGenerator during scenario setup) reads
+        // options through campaign.getGame().getOptions(). Without this, replacing the campaign's options (e.g. when
+        // applying a campaign preset) leaves the Game holding a stale GameOptions object, so later updates such as
+        // the ALLOWED_YEAR sync are never seen there and bot forces get munitions from the wrong era.
+        game.setOptions(gameOptions);
     }
 
     public void setGameOptions(final Vector<IBasicOption> options) {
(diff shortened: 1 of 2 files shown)


VicenteCartas (approved):

mmetke-qumulo (approved):
LGTM

PR #9703, opened 2026-07-29
Contract Overhaul Project: Migrated Scenario, Contract & Facility Data to JSON; Added Scenario Template, Facility, StratCon Contract, and Scenario Modifier Writer; Added New Developer Tool Option to Toolbar

IllianiBird:
# Requires https://github.com/MegaMek/mm-data/pull/492

## n.b. this PR looks far heavier than it actually is due to test data that needed to be added.

Working with the above cited data objects was a pain in the bum. Our existing template editor has been broken for a very long time, which means all templates were being written by hand. This was incredibly error prone. Not least of which because any time a scenario template was edited data had to restage taking up to 10 minutes a pop. Finally, there were several bugs in the translation from template to megamek scenario. All in all it was a pain.

All of the following moved from .xml to .json, with the old XML removed:

- Scenario templates — every template + ScenarioManifest.xml → .json.
- Scenario modifiers — all individual modifiers, the master modifiermanifest/usermodifiermanifest, and the six curated list files (airBattleModifiers, groundBattleModifiers, alliedFacilityModifiers, hostileFacilityModifiers, primaryPlayerForceModifiers, requiredHostileFacilityModifiers).
- Contract definitions — all definitions + ContractDefinitionManifest.
- StratCon facilities — all 22 facility definitions + facilitymanifest.
- 
New top-level Developer Tools menu. The scenario template editor moved here, joined by three new JSON editors and two modal sub-editors:

- Scenario Modifier Editor — scalar/enum/boolean fields; nested force definition, objectives, and linked modifiers are preserved untouched on save.
- Contract Definition Editor + Objective Parameter sub-editor.
- StratCon Facility Editor + Biome sub-editor.

A heap of i18n work was done in the affected areas, as well as a bunch of new (AI authored) tests.

Finally, a bunch of bugs were fixed.

This PR is necessary for a full review of all data points, which will fix numerous scenario issues.


IllianiBird:
n.b. the code this is mostly good, I do need to go back in and fix some conflict issues so please review with that in mind (you might see the occasional rogue xml still sticking about)

IllianiBird:
This is good to review again

HammerGS:
# Review of #9703 (Scenario/Contract/Facility JSON migration + editors + Developer Tools)

Checked out locally and reviewed the full Java diff (the 129 testresources JSON files were skimmed, not audited). The JSON layer itself is well built: the four `*Json` companion classes are consistent (field-based mapper, `FAIL_ON_UNKNOWN_PROPERTIES` off for forward compat, `ALLOW_YAML_COMMENTS` for the `#` license header), `MMDataLicenseHeader` carrying the copyright start-year forward is a nice touch, the campaign-save embedding deliberately stays XML (and has a test pinning that), and the `ScenarioObjective` generated-description work removes a whole class of hand-authored-text drift. Real bug fixes ride along too: the `FORCE_GENERATION_METHODS` array missing "None" (AIOOBE), and the copy constructor no longer materializing generated descriptions into stale overrides. Test coverage is extensive and behavior-focused.

## The one big question: legacy XML reading is documented but not implemented

Several Javadocs state that XML is still read for legacy/user files - `StratConContractDefinition.Deserialize` even says "The format is detected by content (a leading '{' is JSON) rather than file extension", and `AtBScenarioModifierSerializationTest`'s class doc says "legacy XML is still read for backward compatibility." **As far as I can trace, none of that exists.** `ScenarioTemplate.Deserialize(File)`, `AtBScenarioModifier.Deserialize`, `StratConFacility.deserialize`, `StratConContractDefinition.Deserialize`, and every manifest reader now call straight into Jackson; an XML file will throw a parse exception and come back `null`. The JAXB machinery is still on the models (kept for the save embedding), so a content-sniff fallback would be cheap to add.

Concretely, what breaks for users upgrading with existing custom content:

- [x] **User XML scenario templates no longer load anywhere.** The community shares these; anyone with a custom `.xml` template gets a logged parse error and a null template. `FileType.SCENARIO_TEMPLATE` filters the load dialog to `.json` only, so they can't even *select* their old files to discover this.
- [x] **User manifests are silently orphaned.** `STRAT_CON_USER_SCENARIO_MANIFEST`, `usermodifiermanifest`, and `userfacilitymanifest` paths all flipped from `.xml` to `.json`, so existing user manifest files just stop being found (a WARN in the log is the only symptom), and even a renamed manifest would then point at XML files that can't parse.
- [x] Please either (a) add the content-detection fallback the Javadoc already describes - sniff for `{`/`#` vs `<` and route to Jackson or the retained JAXB unmarshaller - or (b) delete the "XML is still read" claims from the four `*Json` class docs, `StratConContractDefinition.Deserialize`, and the `AtBScenarioModifierSerializationTest` doc, and note the user-data break loudly in the PR description/changelog. Option (a) seems clearly better given how cheap it is.

## Ordering / coupling

- [x] **Hard dependency on https://github.com/MegaMek/mm-data/pull/492 merging first.** All shipped-data paths in `MHQConstants` now point at `.json`; without the mm-data side staged, StratCon can't load a single modifier/template/facility at runtime. Worth also confirming the CI/staging pipeline picks up the mm-data branch for the nightly that carries this (per the ci.yml branch-name matching, renaming one half of a coupled pair silently breaks the build).
- [x] Anyone testing this locally needs mm-data#492 checked out and restaged - a stale `./data` copy will make this PR look broken when it isn't.

## Smaller items

- [x] `ScenarioObjective.getGeneratedDescription` splices the time limit with `sentence.substring(0, sentence.length() - 1) + timeLimit + "."` - this silently assumes every `objective.criterion.*` resource string ends in a period. A translated/edited property without one gets its last character eaten. Cheap guard: only strip if `sentence.endsWith(".")`.
- [x] `testresources/data/scenariomodifiers/EnemyAirSupport_test.xml` appears to be referenced by nothing after the migration - dead file unless it's being kept for a future XML-fallback test (which would be a good test to have if option (a) above is taken).
- [x] `FileDialogs.developerDataDirectory` hardcodes `../../mm-data/data/<subdir>` for dev checkouts. Reasonable, but it assumes the working directory is the MekHQ module dir and that the checkout layout matches yours; a one-line note in the docs (or deriving from a configurable property) would spare the next contributor some confusion. Also, `SCENARIO_MODIFIER_DIRECTORY` etc. are resolved once at class-load, so a user who creates the mm-data directory mid-session keeps the fallback - fine, just noting it's static.
- [x] The stale CodeQL comments about deprecated accessors (`getContractTypeName`, `getRoleCollections`, ...) are self-resolving - this PR removes those `@Deprecated` annotations - so those can be dismissed.
- [x] `BriefingTab` now appends battlefield control + an objectives list to the scenario-acceptance dialog body. If the acceptance dialog (or anything downstream of `description`) already renders objectives separately, they'll show twice - worth one manual check in-game.
- [x] Minor doc drift: `ScenarioTemplateLicenseTest`'s class Javadoc describes a "`_license` block" but the implementation (and the test's own assertions) use `#` comment-line headers.

## Verdict

The engineering is solid and the editor rewrite + dev tools are a clear improvement over hand-editing XML. The blocking question is the legacy-XML story: right now the code and its own documentation disagree, and the losers are users with custom templates. Resolve that (ideally with the cheap sniff-fallback) and confirm the mm-data merge/staging order, and this looks good to go.

Local test run: all new/affected suites (mission, atb, stratCon, scenarioTemplateEditor, developerTools, FileType) pass on Windows - BUILD SUCCESSFUL.

MekHQ/src/mekhq/campaign/mission/ScenarioTemplate.java
```diff
@@ -341,21 +334,12 @@ public static ScenarioTemplate Deserialize(String filePath) {
...
-
         try {
-            JAXBContext context = JAXBContext.newInstance(ScenarioTemplate.class);
-            Unmarshaller um = context.createUnmarshaller();
-            try (FileInputStream fileStream = new FileInputStream(inputFile)) {
-                Source inputSource = MHQXMLUtility.createSafeXmlSource(fileStream);
-                JAXBElement<ScenarioTemplate> templateElement = um.unmarshal(inputSource, ScenarioTemplate.class);
-                resultingTemplate = templateElement.getValue();
-            }
+            return ScenarioTemplateJson.fromFile(inputFile);
```
VicenteCartas, line 338: **Major:** This makes standalone template loading JSON-only. Existing user-authored `.xml` templates now fail and return `null`; the same migration also changes user-manifest paths to `.json`, while modifier, facility, and contract loaders become JSON-only as well. That orphans existing custom content despite the new JSON serializer documentation saying JAXB is retained for legacy/user files. Please add a content-detected XML fallback and continue discovering the existing XML user manifests, or provide an explicit migration path before merging.

MekHQ/src/mekhq/gui/scenarioTemplateEditor/ForceRosterEditor.java
```diff
@@ -0,0 +1,107 @@
...
+        // A brand-new force, or a rename, must not collide with a different existing force.
+        if ((!editingExisting || renamed) && roster.containsKey(newForceId)) {
+            return CommitResult.failure(getFormattedTextAt(RESOURCE_BUNDLE,
+                  "ForceRosterEditor.duplicateId",
+                  newForceId));
+        }
+
+        // On a rename, drop the entry under the old ID so it is not left orphaned in the map.
+        if (editingExisting && renamed) {
+            roster.remove(editedForceId);
```
VicenteCartas, line 109: **Major:** Renaming a force only moves its roster key; it does not update other forces' `syncedForceName` or `objectiveLinkedForces`, nor any objective's `associatedForceNames`. After renaming `Alpha` to `Bravo`, synchronized deployment still looks up the removed `Alpha` entry and falls back to a random edge, while objectives no longer match the renamed force. I reproduced the stale reference with a focused test. Please propagate the rename through all template-owned force-ID references (or reject renaming a referenced force) and add regression coverage.

VicenteCartas (commented):
I found two major issues: existing user XML content is no longer discoverable/loadable, and force renames leave runtime force-ID references stale. Details and reproduction impact are in the inline comments.

VicenteCartas (approved):
Seems to be solid overall, left some comments that need to be addressed and then can be merged.

PR #9707, opened 2026-07-29
Updated Gradle Actions to Make Staging Data Less Painful

IllianiBird:
# REQUIRES https://github.com/MegaMek/megameklab/pull/2278
# REQUIRES https://github.com/MegaMek/megamek/pull/8615
# REQUIRES https://github.com/MegaMek/mm-data/pull/493

# Disclaimer
I'm not Tex and I can't replace Tex. I'm a monkey with a keyboard being fact checked by Claude. It's possible this PR could cause everything to explode and need to be rolled back. I am fully open to this and happy for it to take place. At this venture all I can do is say "works on my machine" and hope it works on yours' too.

# Introduction
Moving to a single data directory was one of our best - and worst - ideas. Not because it was a bad move, it is objectively better than anything we had before. Tex was 100% right in making the switch and we've all benefited from the change.

It was 'bad' because of how painful it made working with data. No so much instantial changes where you change one thing and move on with your life, but iterative changes became a nightmare. Every little change resulted in 7-10 minutes of waiting for data to stage before you could even check if your fix/change worked. This quickly mounts up to hours of lost dev time over any given week.

# How Things Used to Work
Essentially we were building the entire data directory twice.

First we were prepping it for duplication and then we were duplicating it. This meant that we were building and then moving nearly a gig of data _twice_. Any time anything changes - even a single white space - we would need to copy and distribute this data block _twice_.

If you, like me, have a terrible computer the processing takes forever. In my case this was made worse by me having to work off a pen drive due to hdd constraints. 

# How Things Work Now
The immediate change is that rather than viewing the data directory as a single gig-sized whole, we break it up into separate 'boxes'. If something changes within a box - say, the boards directory - then only that box is rebuild.

So, in the example above, if you change a single white space in a directory we only need to restate that directory (or the 'box' it's in for small directories).

We also threw out the middleman step where we previously prepped data for staging and then staged it (the double-handling). Removing that means we are facing **significantly** reduced staging time by a magnitude.

# The Difference
## Before (full build)
image

## After (full build)
image

## After (partial build, after making a data change)
image

### DIFF
--- a/MekHQ/build.gradle
+++ b/MekHQ/build.gradle
@@ -147,12 +147,106 @@ tasks.withType(Checkstyle).configureEach {
     maxHeapSize = "4g"
 }
 
-tasks.register('stageDataFiles', Sync) {
-    dependsOn gradle.includedBuild('mm-data').task(':stageFiles')
+// Data is staged directly out of mm-data/data (no intermediate staging copy) and split
+// into several Sync tasks by how often the content changes. The large, rarely-edited
+// binary trees (images, boards, fonts) each own a disjoint subtree of data/, so editing
+// loose data (units, scenarios, universe) only re-copies the small text payload and
+// leaves ~800 MB of assets untouched (UP-TO-DATE). stageDataFiles aggregates them.
+def mmDataDir = "../../mm-data"
+def mhqDataDir = "${layout.projectDirectory.asFile}/data"
+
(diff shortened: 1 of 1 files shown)


NickAragua (approved):
Once it builds anyway, assuming it'll build once the other PRs are merged.

PR #9714, opened 2026-07-31
Add configurable maximum Edge

### LINKED ISSUE #9708: [RFE] Option to cap total Edge
### Enhancement Type *

I know Edge is changing in 0.51.01, thought this would be a good addition for those who want to curb the effects of edge but might not be ready to make the jump to the new system. 

All this RFE would do is prevent further total edge gain above a specified number for all characters. Set it to 3 for example, no campaign character ever goes over 3/3 edge except with GM intervention (add a point, you get 4/3).

### Use Cases or Rationale

Manually adjusting edge totals for all your Mekwarriors, for example, is tedious currently, and having a hard cap defined by the player would save some headaches. Even in the upcoming new system, this limit could prevent spending XP to raise total edge over the given value too.

### Attach Files

_No response_

### MekHQ Suite Version *

0.51.00 Milestone

### Final Verification

VicenteCartas:
## Summary
- add a campaign-level Maximum Edge option with a backward-compatible default of 10
- prevent XP purchases, awards, generated attributes, founder bonuses, and story-created characters from gaining permanent Edge above that maximum
- preserve unrestricted GM assignment and saved-character loading, and do not reduce existing Edge when the maximum is lowered
- update award handling by only adding the Edge actually gained to the character's current Edge

## Screenshot
image

Fixes #9708


IllianiBird (commented):
Edge is an attribute score, it has a limit based on the characters' phenotype (as per ATOW). That's already implemented.

I'm not seeing how this PR is interacting with that. :/

VicenteCartas:
The configured Maximum Edge is an additional campaign ceiling, not a replacement for the existing phenotype cap. `Person.gainEdge` computes the effective maximum by clamping the campaign value to `person.getAttributeCap(SkillAttribute.EDGE)`, and `setEdge` still delegates to `Attributes.setAttributeScore`, which enforces the phenotype cap. For example, a campaign maximum of 3 narrows a phenotype cap of 6 to 3, while a campaign maximum of 10 leaves the phenotype cap of 6 unchanged. GM assignment may exceed the campaign maximum, but not the character's phenotype cap.

IllianiBird:
> The configured Maximum Edge is an additional campaign ceiling, not a replacement for the existing phenotype cap. 

That's what I thought, might be worth including a note to that effect in the campaign option tooltip

VicenteCartas:
@IllianiBird Added to the option tooltip: “This can lower, but never raise, a character's phenotype-based Edge limit.”

IllianiBird (approved):

Issue #8631, opened 2026-01-02
Cybernetic Prosthetics Count as Permanent Injuries, Causing Extreme Turnover Penalties


HammerGS:
### Brief Description *

Campaign Type: Standard campaign with turnover enabled

Description

In MekHQ 0.50.11, personnel who receive cybernetic prosthetics (e.g., VDNI, heavy augmentation, or full-body augmentation) are treated as having permanent injuries for the purposes of monthly turnover rolls.

While this is technically consistent with current injury handling, it produces unintended gameplay results:

The more cybernetically augmented a character becomes, the more likely they are to leave the unit

Extremely expensive upgrades (hundreds of millions to trillions of C-Bills) can result in +16 or higher turnover penalties

These penalties exceed the maximum negative GM Individual Turnover Custom Modifier (-10), making them effectively unavoidable

This strongly discourages the use of cybernetics in long-running campaigns

This interaction was discussed on Discord with IllianiBird (MekHQ developer), who confirmed this is not an intended interaction.

Steps to Reproduce

1. Start a campaign in MekHQ 0.50.11
2. Enable monthly turnover rolls
3. Apply cybernetic prosthetics (VDNI, heavy augments, full-body augments) to personnel
4. Advance to the monthly turnover phase
5. Observe turnover modifiers and results

Actual Result

- Prosthetics are counted as permanent injuries
- Each prosthetic contributes to turnover penalties
- Heavily augmented personnel are highly likely to retire or leave
- GM turnover modifiers cannot reasonably offset the penalties
- Some units may also appear marked in red in the TOE due to injury status

Expected Result

- Cybernetic prosthetics should not drastically increase turnover likelihood
- Investment in advanced augmentation should not penalize personnel retention to this degree
- Turnover penalties should remain within a controllable and intuitive range

Impact

- Discourages use of cybernetic systems
- Produces counterintuitive campaign outcomes
- Breaks balance for cyber-heavy or high-tech units
- Forces reliance on workaround flags rather than intended mechanics

Known Workarounds

- Disable Founder Retirement
- Mark heavily augmented personnel as Founders

These workarounds functionally prevent turnover but are not ideal or intended long-term solutions.

Suggested Solutions

- Any of the following would resolve or mitigate the issue:
- Treat prosthetics as a non-injury injury for turnover calculations
- Add an “Exclude from Turnover” flag separate from the Founder flag
- Increase or remove the cap on the GM Individual Turnover Modifier
[...]

PR #9719, opened 2026-08-02
Added Chassis Familiarity System

IllianiBird:
# Requires https://github.com/MegaMek/megamek/pull/8631
# Requires https://github.com/MegaMek/mm-data/pull/502

Chassis Familiarity allows characters to build a progressive bond with their units. Characters more familiar with their particular chassis gain small bonuses allowing them to perform better than other characters.

The goal here is to help facilitate roleplay, by allowing characters to specialize in a particular chassis. One character might become known as an expert Warhammer pilot. While another kicks ass in a Lucifer.

The other goal is to try and reflect the kind of personal bond between man and machine that we see all the time in the novels.

image

This PR includes full GUI support. So players can easily see who has familiarity with what. As well as GM tools, so characters can GM edit, add, or remove Familiarity.

Chassis Familiarity works with all units except: Fighter Squadrons, Tele-operated Missiles, Handheld Weapons, Battlefield Support Assets, buildings of any kind (including Gun Emplacements).


HammerGS:
# Review of #9719 (Chassis Familiarity)

Checked out locally and read the full diff. The feature is cleanly architected: `FamiliarityMode` centralizes the thresholds/caps as data, the bonus plumbs through the existing `Skill.getFinalSkillValue` chain rather than being bolted on at call sites, the combat/tech split (piloting<->Maintenance, gunnery<->Repair) is consistent everywhere, save round-trip for the per-person familiarity map is in place, and the GM edit dialog + meter-bar tab are nice touches. All of Copilot's earlier catches (the gunnery/piloting `false` swaps, the level-bucket boundary, the MessageFormat arg count, the Maintenance NPE guard, the glossary typo) are verified fixed in the current head. CI is green. (I could not compile locally - my MegaMek checkout doesn't have #8631 - so this review is code-reading plus CI, not a local test run.)

## The big gap: no tests

- [x] **This PR adds zero unit tests for a system that is almost entirely pure, testable logic.** `FamiliarityMode.getFamiliarityLevel` bucket boundaries (a bug Copilot already caught once in exactly this method), the award math in `awardFamiliarity` (Iron Bond x1.5, Emotionally Unavailable x0, cap behavior, the capped-report edge), `TechChassisModifiers`' family/weight-class mapping, and the Person XML round-trip are all trivially unit-testable with no GUI or campaign scaffolding. Given the boundary bug that already slipped in once during this PR's life, a `FamiliarityModeTest` + `TechChassisModifiersTest` + a round-trip test feels like the minimum before merge.

## Likely bugs

- [x] **`TechChassisModifiers.specialistModifier` penalizes units that have no specialty choice at all.** When `specialtyKey` returns `null` (super-heavy Meks/vehicles, or any weight class outside the mapped ranges), `specialist.stringValue().equalsIgnoreCase(null)` is `false`, so the tech takes the +1 "wrong specialty" penalty for a unit family/weight nobody *can* specialize in. Should return `0` when `currentKey == null`.
- [x] **`Person.getAssignedChassis()` is dead code** - private, never called, and its Javadoc describes a "daily chassis-familiarity accrual" that doesn't exist in this PR (accrual is per-scenario in `ResolveScenarioTracker`). Leftover from an earlier design; delete it (or wire it up if daily accrual was intended).

## Should fix / please double-check

- [x] **`Person.getEntity()` was renamed to `getUnitFromEntity()` - the name is backwards.** The method gets the *Entity from the unit*, not the unit from the entity. `getAssignedEntity()` or `getEntityFromUnit()` would say what it does; as-is every future reader will trip over it. (The rename itself is fine to do without deprecation since it's MekHQ-internal, but the new name should be right the first time.)
- [x] **Multi-tech-assignment warning will spam.** `awardTechFamiliarity` posts the "{tech} has more than one tech assignment and Cannot earn Familiarity" report once per maintained unit, per scenario, forever. A tech with several assigned units in one battle produces that many identical lines, and the same lines again after every future scenario. Consider reporting once per tech per resolution (or only the first time), or demoting it to a tooltip on the Familiarity tab.
- [x] **`hasFreeLove`** in `ResolveScenarioTracker.awardFamiliarity` is a leftover name for what is now `FAMILIARITY_IRON_BOND` - rename to `hasIronBond` before it confuses someone.
- [x] **Emotionally Unavailable semantics**: the x0 multiplier applies only when current familiarity is 0, so such a character is locked at zero *forever* - but if a GM seeds any value via the edit dialog, they then accrue at full rate. If "can never form the bond" is the intent, the current-familiarity check is redundant (x0 always); if "slow to warm up" is the intent, this isn't that. Worth one sentence in the SPA description either way (and keep it in sync with mm-data#502).
- [x] **Artillery**: the callers in `Unit` now pass `familiarityBonusGunnery` into `S_ARTILLERY`'s `getFinalSkillValue`, but `Skill.getSPAModifiers` gates the bonus on `familyForSkill(name) != null` - if artillery isn't mapped to a family, the bonus is silently dropped and passing it is misleading; if artillery *should* benefit, the gate blocks it. Pick one deliberately.
- [x] **Commander artillery is a behavior change beyond the feature**: in the `isOnlyCommandersMatter` path, the commander's artillery skill is now applied where previously it never was. Looks like a genuine fix, but it changes crew stats for existing campaigns with that option and deserves a line in the PR description.
- [x] **Cross-repo ordering**: MegaMek/megamek#8631 must merge first (`Entity.isChassisFamiliarityEligible()`), then mm-data#502 (the ~25 new SPA definitions). Since MekHQ CI picks the MegaMek branch by matching head-ref name, the green CI here is building against the MM `familiarity` branch - fine until MM#8631 merges, at which point re-run CI against MM main before merging this.

## Minor / take or leave

- [x] Mode-switch edge: values above the new cap (e.g. 250 after switching Hard -> Normal) survive load (the XML clamp is to the global 300) and overflow the Familiarity tab's meter bar until the next gain self-corrects them down to the cap. Cosmetic, but a one-line clamp against the active mode at display time would tidy it.
- [x] `PersonnelTableMouseAdapter` has a trailing-whitespace-only line in the new Edit Familiarity block, and `Unit.java` has a stray double blank line after the commander gunnery assignment.
- [x] The `gmMenu` rename "Personality & Roleplay" -> "Roleplay" and the `spa.getName()` -> `spaName` cleanup in `addSPAToMenu` are unrelated riders - both harmless, just noting them for the changelog.
- [x] `EditFamiliarityDialog`'s add-spinner defaults to the cap (a new chassis starts at maximum familiarity unless the GM lowers it) - defaulting to 0 or 100 might match intent better.

## Verdict

Good feature, good structure, and the earlier review round was clearly acted on. Before merge I'd want: tests for the pure-logic core (the strongest ask, given a boundary bug already appeared once in this PR), the null-specialty penalty fixed, the dead `getAssignedChassis()` removed, and the `getUnitFromEntity` name corrected. The rest is polish and sequencing.

IllianiBird:
Updated to remove Affinity, Antipathy, and Unit Specialization SPAs/Flaws. This was done because the Familiarity system was a better representation of what those short-lived SPAs were attempting to do.

IllianiBird:
Based on player feedback Familiarity Gain as been added to the following events:

- Training
- Patrolling

MekHQ/src/mekhq/campaign/personnel/Person.java
```diff
@@ -6707,6 +6748,92 @@ public List<Unit> getTechUnits() {
...
+    /**
+     * Adds (or, with a negative amount, subtracts) familiarity for the given chassis, clamping the result to the range
+     * {@code 0..cap}. Blank chassis names and no-op amounts are ignored. The cap is supplied by the caller from the
+     * active {@link Familiarity}.
+     *
+     * @param chassis the base chassis name
+     * @param amount  the amount of familiarity to add
+     * @param cap     the maximum familiarity permitted under the active mode
+     */
+    public void addChassisFamiliarity(final String chassis, final int amount, final int cap) {
```
VicenteCartas, line 6782: **Major:** This clamps the character's total familiarity to whichever cap the caller supplies. Training passes a cap of 100 for the trainee's chassis and 50 for the educator's chassis, so a veteran at 180 receiving +2 is overwritten to 100, while 250 +1 is overwritten to 50. The commander can also receive a full-cap grant and then be reduced to 50 during trainee processing in the same session. This permanently destroys persisted progression. Please make a positive, cap-limited training gain preserve an existing value above that training cap, and add regression tests for both cases.

MekHQ/src/mekhq/campaign/personnel/education/TrainingCombatTeams.java
```diff
@@ -399,6 +426,26 @@ private static String processTrainingTime(Campaign campaign, Person educator, Pe
...
+              trainee.getUnit(),
+              familiarityCap,
+              familiarityProgressOwnChassisSpeed,
+              FamiliarityGainType.SINGLE);
+
+        int familiarityEducatorChassisSpeed = (int) round(familiarityProgressOwnChassisSpeed * 0.5);
+        int trainerFamiliarityCap = (int) round(familiarityCap * 0.5);
+        Familiarity.assignFamiliarity(campaign,
+              educator.getUnit(),
+              trainerFamiliarityCap,
```
VicenteCartas, line 461: **Major:** `performTraining` already iterates once per trainee, but both calls here use `assignFamiliarity`, which awards every crew member in the supplied unit. A K-person unit therefore gives each occupant K identical grants per session. This second call also rewards the educator's crew, whereas the glossary says the trainee should gain familiarity with the educator's chassis, capped at 50. Large crews, especially infantry, can reach the cap in one session while the documented cross-chassis reward never reaches the trainee. Please award an explicit person/chassis pair once per session: the trainee with their own chassis up to 100, and that same trainee with the educator's chassis up to 50. Multi-crew and separate-educator-unit tests would cover both failures.

MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConRulesManager.java
```diff
@@ -1710,11 +1712,15 @@ private static void scanNeighboringCoords(StratConCoords coords, int forceID, Ca
...
         CombatTeam combatTeam = campaign.getPlayerForce().getCombatTeamsAsMap(campaign).get(forceID);
-        if (combatTeam != null && combatTeam.getRole().isPatrol()) {
-            scanRangeIncrease++;
+        if (combatTeam != null) {
+            boolean isPatrol = combatTeam.getRole().isPatrol();
+            if (isPatrol) {
+                scanRangeIncrease++; // Determine scan range (this is the furthest a hex can be revealed)
+            }
+
+            Familiarity.assignFamiliarityToCombatTeam(campaign, combatTeam, FamiliarityGainType.D3);
```
VicenteCartas: **Major:** This D3 award is outside the `isPatrol` branch, so every combat role receives the Patrol-only gain on deployment. The same path is used when assigning a force to an ordinary scenario, which then grants the normal D6 again at resolution. The award also occurs before any revealed-coordinate checks, so another invocation can award it again. Please gate the award itself with `isPatrol` and cover Patrol, non-Patrol, and scenario-assignment paths.

MekHQ/src/mekhq/campaign/personnel/skills/Skill.java
```diff
@@ -579,60 +563,22 @@ public int getSPAModifiers(PersonnelOptions characterOptions, int reputation, in
             }
         }
 
-        // 'Mek / Vehicular / Flight Affinity & Antipathy (piloting only) and Unit Specialist (gunnery and piloting).
-        // The unit family is inferred from this skill's type; the weight class is that of the unit being crewed.
-        modifier += getUnitFamilyModifiers(characterOptions, name, unitWeightClass);
+        // Chassis Familiarity (combat): the caller supplies the bonus already computed for this skill's role (piloting
+        // or gunnery) from the crew member's familiarity and the active FamiliarityMode; it may be negative (Hard
+        // mode). It applies only to combat piloting/gunnery skills (those tied to a specialist family).
+        if ((familiarityBonus != 0) && (familyForSkill(name) != null)) {
```
VicenteCartas: **Major:** This gate discards familiarity modifiers for several unit families that the feature otherwise treats as eligible. `familyForSkill` recognizes only Mek, vehicle, aero, and conventional-fighter constants; Anti-Mek, Small Arms, Gunnery/Battle Armor, Gunnery/ProtoMek, Piloting/Spacecraft, Gunnery/Spacecraft, and Artillery all return `null`. Those characters accrue familiarity and the UI reports bonuses, but they receive neither Normal bonuses nor Hard penalties. A focused probe confirmed that both +1 and -1 become zero for all seven skill constants while the Mek control applies them. Please either map every eligible combat skill, including Artillery, or consistently narrow eligibility, UI, glossary, and documentation.

VicenteCartas (changes requested):
Reviewed at `d5fe5c893b2f36386c975e9baef69408cc279093`. I found four major correctness issues:

1. Training caps can overwrite and permanently reduce accumulated familiarity.
2. Training multiplies awards by crew size and rewards the educator's crew instead of giving the trainee familiarity with the educator's chassis.
3. The Patrol D3 award applies to every StratCon deployment and ordinary scenario assignment, before the normal D6 resolution award.
4. Eligible infantry, battle armor, ProtoMek, spacecraft, and Artillery skills accrue/display familiarity but discard its combat modifiers.

I did not find a critical issue. Each major finding is documented on the relevant changed line with its impact and required correction.

VicenteCartas (approved):

PR #9720, opened 2026-08-02
Fix MekHQ#9594 Destroyed BA Suits cannot be replaced

### LINKED ISSUE #9594: [Issue] Destroyed battle armour cannot be replaced, even if replacement suits in warehouse
### Brief Description *

Two troopers in my battle armour unit were destroyed in a battle, while I had no replacement suits on hand. After ordering the replacement suits and having them delivered, the repair screen still shows it as impossible, although it does not say that parts need to be ordered. If an attempt is made anyways, the replacement suit is destroyed because the target is "Automatic failure", meaning any roll is less than the number that would destroy the part. 

GM mode can be used to complete these repair tasks without issue, restoring the unit to undamaged status. 

This may have something to do with how, upon loading a campaign, if there are more soldiers assigned to a BA unit than functional suits, it will "remove" soldiers to match the unit with the amount of functional suits, but this does not appear to immediately remove the soldiers.

### 3. Steps to Reproduce

1. Load attached save
2. Navigate to repair bay
3. Verify that suits do not appear to be able to be replaced
4. Verify that attempting to do so causes automatic failure and replacement suit destruction
5. Verify that replacement suits do, in fact, exist in warehouse
6. Verify that GM mode can be used to both complete repair bay tasks, or GM restore unit to undamaged condition

To recreate issue:
1. Create MekHQ company with battle armour units
2. Have the unit be damaged in a battle via suits being destroyed
3. Save and load campaign
4. Verify that message appears indicating soldiers were automatically unassigned
5. Verify that soldiers appear to still be assigned
6. Remove soldiers manually
7. Attempt to repair suits using above steps

### Attach Files

[Battle armour can't be replaced-part1.zip](https://github.com/user-attachments/files/29711462/Battle.armour.can.t.be.replaced-part1.zip)

Image

### Severity *

### MekHQ Suite Version *

0.51.0

### Operating System *

Windows 11

### Java Version *

Adoptium 21.0.11

### Final Verification

agrazela14:
Fix #9594

Campaign.getTargetFor includes a check for if a technician has 0 repair time left || a part takes 0 minutes to repair, and automatically fails if either condition is true. 

This comes after a check that auto fails if the tech has 0 repair minutes remaining, so all it's doing is automatically failing to repair parts that take 0 minutes to repair.

This was changed at some point in the past from an automatic success to an automatic fail, but I don't believe either is necessary, so this PR removes the check.

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -6473,10 +6473,6 @@ public class Campaign implements ITechManager {
         }
 
         final int minutes = Math.min(partWork.getTimeLeft(), techTime);
-        if (!(partWork instanceof Refit) && minutes <= 0) {
-            LOGGER.error("Attempting to get the target number for a part with zero time left.");
-            return new TargetRoll(TargetRoll.AUTOMATIC_FAIL, "No part repair time remaining.");
-        }
 
         int helpMod;
         if ((partWork.getUnit() != null) && partWork.getUnit().isSelfCrewed()) {


IllianiBird (approved):
This check was for a loooong abandoned project that intended for each part to track how much time was spent on it. So that repairs would be limited by how long there is in the day and not just by how much tech time is available.

That project was dead before I even joined the team. The removal here makes sense.

PR #9744, opened 2026-08-07
Close #82: Added Warrior's Almanac Dialog to Alert Players of When Parts & Units Enter Prototype, Production, Common, and Extinct States

### LINKED ISSUE #82: RFE: Daily Log: Add Tech Introduction News Stories
Had been messing with the news.xml because I have been wanting a notification of when new technology (weapons, equip, and units) become available.  Because I start out in 3025 with my tech level set to intro and need to know when to turn it to standard as new tech becomes available.  And I like to keep it all in HQ without having to research off sarna and write down a list of the into/re-intro dates or check MML constantly so, Arlith informed me my news.xml would create duplicate data.  The HQ source code has this data already but doesnt println("") to inform the player.  So, I'm using the news.xml personally so that i can have this feature now but, wont submit the tech part when I finish including the wars and other events.  Which may lead to some confusion with how the 3rd Succession War supposedly ends sometime around the beginning of 3025 but, HQ will state at the top of the window 3rd SW until 3028 when the 4th begins and I will be including exact dates that may not line up with HQ.

IllianiBird:
# Requires https://github.com/MegaMek/megamek/pull/8672

Close #82 

This PR closes _the_ oldest issue report in the MekHQ queue. It adds a small report that informs players when units and parts change state. Such as entering production, or going extinct.

This dialog has an associated Client Option, if enabled (enabled by default), the report will fire automatically so the player can't miss it.

https://github.com/user-attachments/assets/59e97179-05aa-410e-b99d-440eda44ec0f

The almanac can be fetched at any time via the Reports taskbar menu.

image

Regardless of contents or Client Options, the Almanac will also appear as a hyperlink in the Daily Report.

image

Inner Sphere and Mixed events are hidden to Clan forces prior to the start of Invasion Stage One. Clan and Mixed events are hidden to Inner Sphere forces prior to Tukayyid.


HammerGS:
Had a look at why units aren't being picked up. I think it's a string mismatch in `buildAlmanacUnitsData`, and there's a wrinkle that explains why it reads as "nothing at all" rather than "some units missing".

## Root cause

`MekSummary` stores tech base as the **localized display string** used in the unit selector - "Inner Sphere", "Clan", "Mixed Inner Sphere", "Mixed Clan". It's set at `MekSummaryCache.java:952` via `ms.setTechBase(e.getTechBaseDescription())`, and `Entity.getTechBaseDescription()` (`Entity.java:1557`) builds those from `messages.properties`.

The Almanac compares that against `TechBase.toString()`, which is a different vocabulary - "IS", "Clan", "All", "Unknown".

Lining up the four calls in `DataLoadingDialog`:

| Almanac asks for | Compares to | Cache actually holds | Match? |
|---|---|---|---|
| `TechBase.IS` | `"IS"` | `"Inner Sphere"` | no |
| `TechBase.CLAN` | `"Clan"` | `"Clan"` | **yes** |
| `TechBase.ALL` | `"All"` | `"Mixed Inner Sphere"`, `"Mixed Clan"` | no |
| `TechBase.UNKNOWN` | `"Unknown"` | *(never produced)* | no |

Three of the four buckets are empty always. Only Clan matches, and only by coincidence - the word happens to be identical in both vocabularies.

**Why it looks like zero units:** take a Lyran merc campaign starting in 3025. The date is before Tukayyid and the force isn't Clan, so `WarriorsAlmanacDialog` takes the `hideClanAndMixed` branch and builds the Units section from just two columns, Inner Sphere and Unknown. Those are precisely the two buckets that never match, so the Units section never gets written at all. Parts are unaffected because `buildAlmanacPartsData` reads `PartsStore` directly and never goes near this comparison.

Quick way to confirm: start in **3055** instead, or as a Clan faction. Clan units should appear while IS and Mixed still don't.

## Suggested fix

Worth avoiding the tempting one-liner of passing `"Inner Sphere"` instead of `"IS"` - those strings come from `Messages.getString()`, so it would silently break again under a non-English locale (es gives "Esfera Interior"), and it can't express Mixed at all since that's two separate strings.

`MekSummary` already has `isClan()`, so the only missing piece is a mixed-tech flag. I've suggested adding that on MegaMek/megamek#8672 since that PR is already touching `MekSummary`. With it in place this side becomes:

```java
public static Map<Integer, WarriorsAlmanacData> buildAlmanacUnitsData(TechBase techBase) {
    ...
    if (resolveTechBase(summary) != techBase) {
        continue;
    }
    ...
}

private static TechBase resolveTechBase(MekSummary summary) {
    if (summary.isMixedTech()) {
        return TechBase.ALL;
    }
    return summary.isClan() ? TechBase.CLAN : TechBase.IS;
}
```

The `UNKNOWN` bucket can go entirely - no unit ever produces it, so `getUnitsAlmanacUnknown()` and its column are dead weight in every branch of `appendSections`.

## Smaller things

- `WarriorsAlmanacData.java` - `part.getTechBase();` is a bare statement with its result discarded, looks like a leftover.
- `WarriorsAlmanacDialog.java` - the loop variables in `appendSection` and `groupNamesByPhase` are named `AlmanacTechAdvancementPhase`, shadowing the type name in UpperCamelCase. `phase` would read better.
- `WarriorsAlmanacDialog` calls `campaign.getPartsAlmanacIS().get(gameYear)` with no null guard. Those maps are only populated in `DataLoadingDialog` progress 8, so any `Campaign` built outside that path (tests, for instance) will NPE.
- Parts and units with no date return `DATE_NONE` (-1), so every item gets filed into a year `-1` bucket that's never read. Harmless, but it's a wasted entry per item across the whole parts store and unit cache - a `!= DATE_NONE` skip would avoid it.
- `Campaign.java:1954` picked up a whitespace-only change to that TODO comment's indentation, and `DataLoadingDialog.java` has a few reformatted-but-otherwise-unchanged lines around the marriage/divorce/procreation block. Probably worth reverting those so the diff stays clean.

This is all from reading the source rather than running it - I haven't built either branch, so the 3055 check is the thing that would confirm it outright.

VicenteCartas:
**Major: Reintroduction events are omitted**

The almanac builders register prototype, production, common, and extinction dates for [parts](https://github.com/MegaMek/mekhq/blob/fe669c0180d648187bf5a2dd8a81e819d0dce099/MekHQ/src/mekhq/campaign/universe/WarriorsAlmanac/WarriorsAlmanacEntry.java#L102-L109) and [units](https://github.com/MegaMek/mekhq/blob/fe669c0180d648187bf5a2dd8a81e819d0dce099/MekHQ/src/mekhq/campaign/universe/WarriorsAlmanac/WarriorsAlmanacEntry.java#L135-L142), but never the reintroduction date; the [phase enum](https://github.com/MegaMek/mekhq/blob/fe669c0180d648187bf5a2dd8a81e819d0dce099/MekHQ/src/mekhq/campaign/universe/WarriorsAlmanac/AlmanacTechAdvancementPhase.java#L47-L50) cannot represent it either. This omits an entire lifecycle phase explicitly requested by #82.

For example, Arrow IV is reported Extinct in 2830, but its 3044 reintroduction is never shown because it has no later common date. Gauss Rifle's 3040 return is also omitted and it appears only as Common in 3045. Tandem-Charge ammunition similarly returns in 3062 without any almanac entry. The last almanac statement about some of these technologies therefore remains that they are extinct even after they become available again.

Could this add a reintroduction phase and register `getReintroductionDate(...)` for both maps? The parts API already exposes that date. For units, dependency PR MegaMek/megamek#8672 would also need to expose a reintroduction field on `MekSummary`.

VicenteCartas (changes requested):
Left a comment.

VicenteCartas (approved):

Issue #9751, opened 2026-08-09
NPE in PersonnelMarketMekHQ constructor: campaign read before setCampaign (Campaign Options apply + silent market loss on load)


HammerGS:
## Problem

`PersonnelMarketMekHQ`'s constructor asks the campaign a question before it has a campaign. Line 111 reads `getCampaign().getCampaignOptions().get(CampaignOption.USE_CHAOS_REPUTATION)`, but both places that create the market do `new PersonnelMarketMekHQ()` first and `setCampaign(campaign)` after. During the constructor, `getCampaign()` is null, so it throws a NullPointerException every time.

The same root cause surfaces two different ways:

1. **Campaign Options dialog**: changing Personnel Market Style to "MekHQ" and clicking Apply throws an uncaught NPE on the EDT (`MarketsOptionsModel.applyTo`, line 126). The market swap fails, and every market option after that point in the apply chain is silently dropped for that click. The error popup returns on every Apply.
2. **Campaign load**: `NewPersonnelMarket.generatePersonnelMarketDataFromXML` (line 171) hits the same NPE, but its catch block swallows it, logs "Could not initialize Personnel Market", and falls back to a **disabled** market. A save using the MekHQ market style silently loses its personnel market on load.

The other market styles don't crash - they only call `getCampaign()` in methods that run after the campaign is attached.

## Example

Open any campaign. Campaign Options -> Markets -> set Personnel Market Style to "MekHQ" -> Apply:

```
12:09:15,037 ERROR [mekhq.MekHQ] {AWT-EventQueue-0}
mekhq.MekHQ.lambda$main$1(MekHQ.java:416) - Uncaught java.lang.NullPointerException detected.
java.lang.NullPointerException: Cannot invoke "mekhq.campaign.Campaign.getCampaignOptions()" because the return value of "mekhq.campaign.market.personnelMarket.markets.PersonnelMarketMekHQ.getCampaign()" is null
	at mekhq.campaign.market.personnelMarket.markets.PersonnelMarketMekHQ.<init>(PersonnelMarketMekHQ.java:111)
	at mekhq.gui.campaignOptions.contents.MarketsOptionsModel.applyTo(MarketsOptionsModel.java:126)
	at mekhq.gui.campaignOptions.contents.MarketsPages.applyCampaignOptionsToCampaign(MarketsPages.java:140)
[...]

PR #9752, opened 2026-08-09
Add Hangar/TO&E tractor-trailer setup with lobby visuals and full-train deployment (Fixes #9750)

### LINKED ISSUE #9750: Tractor/trailer setup: add Hangar tab menu, TO&E tow visuals, fix detach NPE, deploy full trains
## Problem

MekHQ can already hitch trailers to tractors, but only from the TO&E right-click menu, and that path is easy to miss and partly broken. The Hangar tab has no transport assignment options at all. A hitched trailer only gets a small "Towed by:" text line on its tree node — none of the at-a-glance connection graphics the MegaMek lobby has. The two tow detach menu items are cross-wired, and the tractor-side one throws an error instead of releasing trailers. When a scenario launches, only the first trailer in a train is towed.

## Example

Campaign with a Prime Mover (tractor) and two J-27 Ordnance Transports (trailers):

- Hangar tab, right-click a J-27: no option to assign it to a tractor (or to any transport).
- TO&E, right-click the J-27, "Assign Formation to a Tractor": works, and the node gains a "Towed by: Prime Mover" text line - but there is no branch mark or indent, so a hitched train reads as unrelated rows.
- TO&E, right-click the Prime Mover, "Unassign Formation from Tractor": throws a NullPointerException. It tries to unhitch the tractor from a tractor it does not have, instead of releasing its trailers.
- Chain both J-27s behind the Prime Mover and launch an AtB scenario. Answer Yes to the tow prompt: only the first J-27 arrives hitched. The second deploys loose.

The MegaMek lobby already solves the visibility half of this: towed units get a gray branch mark and sort under their tractor, so connected vs not is visible at a glance.

## Proposed fix

- **Hangar tab**: add the same assign-to-transport right-click options the TO&E has (ship / tactical / tow), reusing the existing menu classes.
- **TO&E visuals**: show tow links in the tree itself using the lobby's visual language (gray branch mark plus "towed by" text on the node), not just a tooltip.
- **Detach fix**: make the two tow detach items match their labels — tractor side releases its trailers, trailer side unhitches itself.
- **Full trains at launch**: replace the single-trailer handoff with MegaMek's `Client.sendBuildTrain(...)`, which validates and builds the whole train server-side.
- **Cleanup fix**: tow transporters are added to the campaign transporter map when a unit arrives but never removed when it leaves.

The campaign data model (`CampaignTransportType`, `TowTransportedUnitsSummary`) is NOT modified. Non-AtB `GameThread` scenarios apply no transport loading of any kind today; that is a separate issue.

## Implementation notes

- Hangar: wire `AssignForceToShipTransportMenu` / `AssignForceToTacticalTransportMenu` / `AssignForceToTowTransportMenu` into `UnitTableMouseAdapter`. They already take a `Campaign` and a set of `Unit`s.
- Detach fix: `TOEMouseAdapter.detachTrailerTransportMenuClass` calls `unassignTransportAction` on the tractor itself; `unloadFromTransport` returns null (no assignment) and the follow-up call NPEs. It should call `unassignFromTransportAction` like the ship/tactical equivalents. The two items also have swapped resource strings (`unassign.TOW_TRANSPORT` vs `unassignFrom.TOW_TRANSPORT`).
- Tree visuals: `ForceRenderer` already builds the "Towed by" tooltip; add the indicator to the node label, using the lobby's escaped-ASCII branch glyphs (see `LobbyMekCellFormatter.BRANCH_CORNER`).
- Full trains: `AtBGameThread` takes `.getFirst()` from the towed-unit list, so multi-trailer trains stored in the campaign never fully deploy. `Client.sendBuildTrain(tractorId, orderedTrailerIds)` replaces the one-hop `Utilities.towPlayerTrailers` path and gets server-side validation and rollback for free.
- Cleanup: the transporter-map removal path handles SHIP and TACTICAL but not TOW; the add path handles all three.
- Test: save/load a campaign with a two-trailer train; launch a scenario and verify both trailers arrive hitched.

## Background

Player request, 2026-08-09: wanted to set up tractors and trailers from the TO&E and Hangar tabs, matching what the MegaMek lobby can do. Investigation showed `TOW_TRANSPORT` already exists end-to-end in the TO&E; the gaps above are what remains of the request.

HammerGS:
## Summary

Players asked to set up tractors and trailers from the Hangar and TO&E tabs. The TO&E already had
tow assignment; this PR adds it to the Hangar tab, shows connections in the tree the way the
MegaMek lobby does, adds lobby-style Connect/Disconnect Train, and fixes three bugs found along
the way. About a third of the diff is tests and the shared menu class extracted from
`TOEMouseAdapter`.

## Changes Made

*Assign it*
- **Hangar tab**: the unit right-click menu now has the same assign/unassign transport actions
  (ship / tactical / tow) as the TO&E. Both tabs share one new builder,
  `TransportAssignmentMenus`, replacing six near-identical private methods in `TOEMouseAdapter`.
- **Connect Train**: multi-select one free tractor plus its trailers (or a whole convoy force)
  and connect the train in one action. Reuses the lobby's `TrainOrderDialog` for ordering, with
  the lobby's selection rules and an all-or-nothing weight/hitch check before anything connects.
- **Disconnect Train**: right-click any train member to dissolve the whole train, like the lobby.
  The single-link unassign items remain for breaking one hitch.

*See it*
- **TO&E tree**: carried or towed units get the lobby's gray branch mark, one level deeper per
  link - a Prime Mover's second trailer draws deeper than its first, so a train reads as a tree.
  Same escaped glyphs as `LobbyMekCellFormatter`.

*Deploy it*
- **Full trains**: at scenario launch, `AtBGameThread` now walks each train front-to-back and
  sends one `Client.sendBuildTrain(...)` per train. Before, only the first trailer was hitched
  (`.getFirst()`), a two-trailer train raised two prompts, and the second hop silently failed.
  One prompt per train now; the server validates every link and rolls back atomically.

*Fixes*
- **Detach NPE**: the tractor-side tow detach item called the wrong action helper and threw a
  NullPointerException instead of releasing trailers. The two tow detach items also had swapped
  labels. Both fixed in the shared builder, with a null guard.
- **Mothball**: `completeMothball` removed SHIP and TACTICAL from the campaign transporter map
  but not TOW, so a mothballed tractor still appeared in the assign menu. Both mothball and
  activation now loop all `CampaignTransportType` values.

## Files Changed

*New source*
- `MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java` - NEW: shared assign/unassign/connect/disconnect menu builder
- `MekHQ/unittests/mekhq/campaign/unit/TowTransportTest.java` - NEW: 12 tests

*Main edits*
- `MekHQ/src/mekhq/gui/adapter/TOEMouseAdapter.java` - both popups call the shared builder; ~127 lines of per-type helpers deleted
- `MekHQ/src/mekhq/gui/adapter/UnitTableMouseAdapter.java` - Hangar popup calls the shared builder
- `MekHQ/src/mekhq/gui/ForceRenderer.java` - branch glyphs for carried/towed units
- `MekHQ/src/mekhq/AtBGameThread.java` - train walk + single prompt per train
- `MekHQ/src/mekhq/Utilities.java` - `towPlayerTrailers` takes the ordered trailer list, sends `sendBuildTrain`
- `MekHQ/src/mekhq/campaign/unit/Unit.java` - mothball/activation transporter-map symmetry
- `MekHQ/resources/mekhq/resources/AssignForceToTransport.properties` - tow prompt covers trains; Connect/Disconnect Train strings

## Testing

- Unit tests: 12 tests in `TowTransportTest` covering train linking, capacity math, both detach
  directions (including the NPE path), Connect Train ordering and its atomic overweight refusal,
  Disconnect Train, XML round-trips with real BLK entities (new tow format and legacy pre-50.07
  attribute-less entries), and mothball/activation map bookkeeping. The mothball test fails with
  the `Unit.java` fix reverted. No serialization code changed; the round-trip tests are the first
  live save/load coverage for transport assignments.
- Full suite, `checkstyleMain`, and `javadoc` all pass.
- Manual testing: QA pass requested. Checklist for testers (Prime Mover + two J-27s covers most
  of it):
  - Assign from Hangar and from TO&E; both detach directions; Connect Train order dialog;
    Disconnect Train from a middle trailer
  - Two-trailer train deploys fully hitched into an AtB scenario with ONE prompt
  - Mothballed tractor disappears from assign menus; reappears on activation
  - Save/load with a train assigned (branch marks and assignments intact)

## Out of scope

- Non-AtB `GameThread` scenarios apply no transport loading of any kind (ship/tactical/tow
  alike); pre-existing gap for all transport types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,379 @@
...
+     */
+    public static void addTransportMenus(JFrame frame, JPopupMenu popup, Campaign campaign, List<Unit> units) {
+        Set<Unit> unitSet = new HashSet<>(units);
+        for (CampaignTransportType campaignTransportType : CampaignTransportType.values()) {
+            JMenu assignMenu = switch (campaignTransportType) {
+                case SHIP_TRANSPORT -> new AssignForceToShipTransportMenu(campaign, unitSet);
+                case TACTICAL_TRANSPORT -> new AssignForceToTacticalTransportMenu(campaign, unitSet);
+                case TOW_TRANSPORT -> new AssignForceToTowTransportMenu(campaign, unitSet);
+            };
+            JMenuHelpers.addMenuIfNonEmpty(popup, assignMenu);
```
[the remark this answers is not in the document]
HammerGS: Valid - `JMenuHelpers.addMenuIfNonEmpty` is `@Deprecated(since = "0.50.04")`. Fixed in 5f9c02375f.

Verified the designated replacement first: the deprecation note points to `JScrollablePopupMenu`, but that requires the popup itself to be that type - the TO&E and Hangar adapters hand this builder plain `JPopupMenu`s - and `JScrollablePopupMenu`'s constructors are themselves `@Deprecated(forRemoval = true)` since 0.51.0. So the fix inlines the six-line non-empty check + `MenuScroller` attachment as a private helper with identical behavior, removing the deprecated call without widening the diff into the adapters.

Note: `TOEMouseAdapter` still contains ~25 pre-existing calls to the same deprecated helper on lines this PR does not touch; migrating those belongs in a dedicated cleanup.

HammerGS:
Review feedback disposition:

- **Copilot review**: generated no inline comments (summary only) - nothing to address.
- **CodeQL** (deprecated `JMenuHelpers.addMenuIfNonEmpty`): valid, fixed in 5f9c02375f by inlining the non-empty/scroller behavior. The designated `JScrollablePopupMenu` replacement doesn't fit here (needs the popup itself to be that type, and its constructors are deprecated for removal). Details on the alert thread.

Compile, checkstyle, and `TowTransportTest` (12/12) re-verified after the fix.

HammerGS:
CI failure analysis for the `pr_ci` runs on 5f9c02375f - four failures, two causes:

**My three `TowTransportTest` XML round-trip tests (valid, fixed in ff51ce079f).** They passed locally but failed in CI because the read side went through `Unit.generateInstanceFromXML`, whose entity reload needs `MekSummaryCache` game data the CI test JVM doesn't have - the same trap that got `UnitIOTest.GenerateInstanceFromXMLTests` `@Disabled`. Fixed by extracting the `<transportAssignment>`/`<transportedUnit>` node parsing into package-visible helpers on `Unit` (pure code move, behavior unchanged) and pointing the tests at them. The tests still write real BLK entities through `writeToXML` and parse the written nodes with the same code the campaign loader runs, so the legacy-vs-new routing is now covered deterministically in any environment. Bonus: the legacy-format test now asserts against the attribute-less tactical entry the writer genuinely emits, rather than hand-edited XML.

**`StratConOceanPlacerTest.lakelands_scattersMultipleWaterBodies` (pre-existing flake, not addressed here).** Unseeded random ocean placement with a `>= 2` water-bodies assertion; it passed 5 consecutive local runs on this branch (two full suites + three targeted runs) and this PR touches nothing under StratCon. Worth a seeded RNG in a separate PR.

HammerGS:
QA request implemented in 90bafcd720: a hitched trailer now follows its tractor into the formation.

How it behaves:
- Hitch a trailer to a tractor that is in a formation (assign menu or Connect Train, either tab) and the trailer is added to that same formation automatically.
- A trailer the player already placed somewhere in the TO&E is left where it is - no surprise moves.
- If the tractor is not in any formation, nothing changes.

Example: the Prime Mover sits in Alpha Company / Transport Lance. Right-click a loose Trailer in the Hangar, assign it to the Prime Mover - the trailer appears under Transport Lance with no extra step.

Two new unit tests cover the add and both leave-alone cases (14 total in TowTransportTest). Checkstyle and javadoc green.

HammerGS:
## QA round 2 - both items fixed (`2f96fb4ec5`)

### The lockup

Found it. It is a real hang, and it was one click away.

**What happened:** the "Assign Formation to a Tractor" action hitches your unit onto the **last** unit of the tractor's train. If the unit you picked is already in that train, it *is* the last unit - so it gets hitched to itself. `TowTransportedUnitsSummary.getTractor()` then follows "towed by" around that loop forever. That runs on the Swing thread, so the window freezes, and because nothing throws, nothing reaches the log.

Example: Prime Mover tows a J-27. Right-click that same J-27, assign it to the Prime Mover again. The code walks Prime Mover -> J-27, reaches the end of the train, and hitches the J-27 to the J-27. Frozen.

The menu never hid the tractor a trailer was already hitched to, which is why the item "didn't disappear".

**Fix, in two layers:**

1. **Do not offer it.** The tow menu now drops any tractor whose train the selected unit already belongs to. Hitch your only J-27 to your only Prime Mover and the menu goes away, as you expected. The action also re-checks and shows a warning, in case a menu was built before an earlier click in the same selection changed things.
2. **Never hang again.** Every walk along a tow train now stops when it sees a unit twice: `getTractor()`, the trailer-weight walk next to it, the end-of-train walk in the assign action, and the shared `trainMembers()` helper `disconnectTrain` now uses. A looped hitch from an older save logs an error instead of locking the UI.

### Formation not updating

That was my guard, and it was the wrong call - it only moved a trailer that was in no formation. Now a hitched trailer **always** joins the tractor's formation, even if it was sitting in another one, and every trailer behind it comes along so the whole train stays together on the org chart. If the tractor is in no formation, nothing moves.

### Testing

17 tests in `TowTransportTest` (3 new): whole-train listing from any member, same-train rejection, and a self-towing unit that must not hang the walks. Checkstyle and javadoc clean. Also merged the latest main into the branch, which brings in IllianiBird's #9751 fix.

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3125,6 +3091,59 @@ public static Unit generateInstanceFromXML(final Node wn, final Version version,
...
+                  "campaignTransportType").getTextContent());
+        } else {
+            // Tactical transports were added before the campaignTransportType attribute
+            // was. Assume it's a tactical transport.
+            campaignTransportType = CampaignTransportType.TACTICAL_TRANSPORT;
+        }
+        UUID id = UUID.fromString(attributes.getNamedItem("id").getTextContent());
+
+        if (attributes.getNamedItem("transportedLocation") != null) {
+            int transportedLocationHash = Integer.parseInt(attributes.getNamedItem("transportedLocation")
```
IllianiBird: This should be using MathUtility.parseInt()

MekHQ/src/mekhq/campaign/unit/Unit.java
```diff
@@ -3125,6 +3091,59 @@ public static Unit generateInstanceFromXML(final Node wn, final Version version,
...
+     * Parses a {@code <transportedUnit>} save-file node onto the given unit, routing it to the
+     * summary matching its {@code campaignTransportType} attribute. Package-visible so the routing
+     * is unit-testable without loading a full entity.
+     */
+    static void parseTransportedUnitNode(Node transportedNode, Unit retVal) {
+        NamedNodeMap attributes = transportedNode.getAttributes();
+        CampaignTransportType campaignTransportType = CampaignTransportType.valueOf(attributes.getNamedItem(
+              "campaignTransportType").getTextContent());
+        retVal.addTransportedUnit(campaignTransportType,
+              new UnitRef(UUID.fromString(attributes.getNamedItem("id").getTextContent())));
```
IllianiBird: If "id" has special meaning it should be stored as a const.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
```diff
@@ -97,6 +101,67 @@ protected Set<TransporterType> filterTransporterTypeMenus(final Set<Unit> units)
...
+     * @param units      units being assigned a tractor
+     *
+     * @return tractors that can still take these units
+     */
+    @Override
+    protected Set<Unit> filterTransports(final Set<Unit> transports, final Set<Unit> units) {
+        Set<Unit> availableTransports = new HashSet<>();
+        for (Unit transport : transports) {
+            boolean alreadyInTrain = false;
+            for (Unit unit : units) {
```
IllianiBird: This is o(n), any time we have a for inside a for it's a red flag. Consider using unit.getTransportedUnits() to fetch units already transported instead of looping.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
```diff
@@ -97,6 +101,67 @@ protected Set<TransporterType> filterTransporterTypeMenus(final Set<Unit> units)
...
+
+        List<Unit> trailers = TransportAssignmentMenus.trailersBehind(transport);
+        if (trailers.isEmpty()) {
+            return MHQInternationalization.getFormattedTextAt(RESOURCE_BUNDLE,
+                  "AssignForceToTransportMenu.towTooltipEmpty.text", formationPath);
+        }
+
+        StringBuilder trainDescription = new StringBuilder(transport.getName());
+        for (Unit trailer : trailers) {
+            trainDescription.append(" -> ").append(trailer.getName());
```
IllianiBird, line 167: Player facing hardcoded string

(This is a joke, it's fine here :P)

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
```diff
@@ -109,6 +174,18 @@ protected Set<TransporterType> filterTransporterTypeMenus(final Set<Unit> units)
     protected void transportMenuAction(ActionEvent evt, TransporterType transporterType, Unit transport,
           Set<Unit> units) {
         for (Unit unit : units) {
+            // A unit already in this train would be hitched to itself, which loops every later
+            // walk along the train. The menu filters these out; a menu built before an earlier
+            // click in the same selection can still get here.
+            if (TransportAssignmentMenus.isInSameTrain(transport, unit)) {
+                JOptionPane.showMessageDialog(null, MHQInternationalization.getFormattedTextAt(
+                      "mekhq.resources.AssignForceToTransport",
```
IllianiBird: This should be `RESOURCE_BUNDLE`

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
```diff
@@ -109,6 +174,18 @@ protected Set<TransporterType> filterTransporterTypeMenus(final Set<Unit> units)
...
+                      "AssignForceToTransportMenu.warningUnitAlreadyInTrain.text",
+                      unit.getName(),
+                      transport.getName()), "Warning", JOptionPane.WARNING_MESSAGE);
```
IllianiBird: This looks like an actual hardcoded player-facing String

MekHQ/src/mekhq/gui/menus/AssignForceToTransportMenu.java
```diff
@@ -38,6 +38,7 @@
 import java.util.Set;
 import javax.swing.JMenuItem;
 
+import megamek.common.annotations.Nullable;
```
IllianiBird: This should be Jakarta (it's better)

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+
+    /**
+     * Menu item shown when every selected unit is part of a tow train: dissolves the whole train
+     * each selected unit belongs to, from lead tractor to last trailer, like the lobby's
+     * Disconnect Train. The single-link unassign items above only break one hitch; this one frees
+     * everything.
+     */
+    private static void addDisconnectTrainMenuItem(JPopupMenu popup, Campaign campaign, List<Unit> units) {
+        if (anyDeployed(units) ||
+                  !units.stream().allMatch(unit -> unit.hasTransportAssignment(TOW_TRANSPORT) ||
```
IllianiBird: How many units are we expecting to parse here? Generally streams are less performative than standard loops unless working with thousands of objects

I'd also really prefer we didn't put streams in conditional statements. It can get inefficient fast and is harder to read.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+    /**
+     * Lists the whole tow train the given unit belongs to, lead tractor first, then each trailer in
+     * hitch order. A unit that is in no train is returned on its own. Both walks stop as soon as a
+     * unit repeats, so a campaign that somehow ended up with a looped hitch cannot spin here.
+     *
+     * @param anyMember any unit of the train - tractor, middle trailer, or tail
+     *
+     * @return train members in order, lead tractor first; never empty
+     */
+    public static List<Unit> trainMembers(Unit anyMember) {
```
IllianiBird, line 177: Something about this method feels off, I feel like we need a more efficient way to fetch this data without two `while` statements.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+                if ((head != null) || unit.hasTransportedUnits(TOW_TRANSPORT)) {
+                    // The train is built from one unattached tractor; two heads or a tractor
+                    // already towing means the selection is not a connectable train.
+                    return;
+                }
+                head = unit;
+            } else {
+                // Selection contains a unit that is neither a tractor nor a trailer
+                return;
+            }
```
IllianiBird, line 271: Non-essential, but you could break out the contents of this loop into a utility method for cleaner code.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+                return false;
+            }
+            totalTrailerWeight += trailerEntity.getWeight();
+
+            // Every trailer except the last one is an attach point and needs a hitch of its own
+            boolean isLastTrailer = trailerIndex == (orderedTrailers.size() - 1);
+            if (!isLastTrailer &&
+                      trailerEntity.getTransports()
+                            .stream()
+                            .noneMatch(transporter -> transporter instanceof TankTrailerHitch)) {
```
IllianiBird: Not sure we should be using a stream here, feels really non-performative. Also a stream in a conditional is bad for readability and maintainability.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+     * Menu item shown when every selected unit is itself assigned to a transport of this type:
+     * unassigns the selected units from their transports.
+     */
+    private static void addUnassignSelfMenuItem(JPopupMenu popup, Campaign campaign, List<Unit> units,
+          CampaignTransportType campaignTransportType) {
+        if (anyDeployed(units) ||
+                  !units.stream().allMatch(unit -> unit.hasTransportAssignment(campaignTransportType))) {
+            return;
+        }
+        JMenuItem menuItem = new JMenuItem(MHQInternationalization.getTextAt(RESOURCE_BUNDLE,
```
IllianiBird: For readability, it can sometimes be good to define the key separate to the getTextX method when working with a dynamically defined reference key, for example:

```
String key = "TOEMouseAdapter.unassign." + campaignTransportType.name() + ".text"
JMenuItem menuItem = new JMenuItem(MHQInternationalization.getTextAt(RESOURCE_BUNDLE, key));
```

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+     * Menu item shown when every selected unit is carrying or towing units of this type: releases
+     * the selected transports' carried or towed units.
+     */
+    private static void addUnassignTransportedMenuItem(JPopupMenu popup, Campaign campaign, List<Unit> units,
+          CampaignTransportType campaignTransportType) {
+        if (anyDeployed(units) ||
+                  !units.stream().allMatch(unit -> unit.hasTransportedUnits(campaignTransportType))) {
+            return;
+        }
+        JMenuItem menuItem = new JMenuItem(MHQInternationalization.getTextAt(RESOURCE_BUNDLE,
```
IllianiBird: Same here

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
```diff
@@ -0,0 +1,476 @@
...
+     */
+    public static void unassignFromTransport(Campaign campaign, CampaignTransportType campaignTransportType,
+          Collection<Unit> units) {
+        Set<Unit> transportsToUpdate = new HashSet<>();
+        for (Unit transportedUnit : units) {
+            Unit oldTransport = transportedUnit.unloadFromTransport(campaignTransportType);
+            if (oldTransport != null) {
+                transportsToUpdate.add(oldTransport);
+            }
+            MekHQ.triggerEvent(new UnitChangedEvent(transportedUnit));
```
IllianiBird: Do we need both UnitChangedEvents? I worry it will create churn

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -541,37 +541,67 @@ public void run() {
...
+                            continue;
+                        }
+
+                        // Walk the train front to back. Each unit in the map only lists the trailer
+                        // directly behind it, so follow the links hop by hop, stopping at any gap.
+                        List<Integer> orderedTrailerIds = new ArrayList<>();
+                        Set<UUID> visitedUnits = new HashSet<>();
+                        visitedUnits.add(transportId);
+                        UUID currentUnitId = transportId;
+                        while (true) {
```
IllianiBird: This is really dangerous and can lead to an infinite loop. I know you have breaks built into the loop, but I suggest we reconsider whether this is the best pattern. If it is, then it needs some serious Unit Testing.

MekHQ/src/mekhq/AtBGameThread.java
```diff
@@ -541,37 +541,67 @@ public void run() {
...
+                        boolean towUnits = (JOptionPane.YES_OPTION ==
+                                                  JOptionPane.showConfirmDialog(null,
+                                                        MHQInternationalization.getFormattedTextAt(
+                                                              "mekhq.resources.AssignForceToTransport",
+                                                              "AtBGameThread.loadTransportDialog.TOW_TRANSPORT.text",
+                                                              transport.getName()),
+                                                        MHQInternationalization.getFormattedTextAt(
+                                                              "mekhq.resources.AssignForceToTransport",
+                                                              "AtBGameThread.loadTransportDialog.TOW_TRANSPORT.title"),
+                                                        JOptionPane.YES_NO_OPTION));
```
IllianiBird: We're defining the resource bundle twice in this dialog, probably best to create a variable and pass it around.

We're also using getFormattedTextAt and not passing in values, getFormattedTextAt is far more expensive than getTextAt so we should use getTextAt any time we're not needing to format the text.

IllianiBird (changes requested):
A few changes I'd like to see incorporated before we merge. A couple of red flags, in particular.

MekHQ/src/mekhq/campaign/unit/Unit.java
HammerGS: Done in `0ea7f4be66`.

Worth noting the history: this line is pre-existing code that this PR only moved. It used to sit inside `generateInstanceFromXML`'s `catch (Exception)`, so a bad value was caught - CodeQL only started flagging it because extracting the parse helper put it in a different method.

`MathUtility.parseInt(..., 0)` is better anyway. A hash that does not parse now leaves the assignment without a location, which is the same fallback the `transporterType` branch below already uses, instead of failing the whole unit load.

MekHQ/src/mekhq/campaign/unit/Unit.java
HammerGS: Done in `0ea7f4be66`. Added constants for all four attribute names on these nodes: `TRANSPORT_ATTRIBUTE_ID`, `_TYPE`, `_LOCATION`, `_TRANSPORTER_TYPE`.

`id` does carry two meanings, which is worth pinning down: on a `<transportAssignment>` node it is the transport's UUID, on a `<transportedUnit>` node it is the transported unit's UUID. Noted that on the constant.

Kept this to the read side. The writer builds the same names inline in string concatenation further up the file, but that is pre-existing and this PR does not touch it.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
HammerGS: Fixed in `0ea7f4be66`. You are right, and it was worse than it looks: `isInSameTrain` walks the train, so this was tractors x units x train length, re-walking the same trains for every tractor offered.

It now walks each selected unit's train once, collects the members into a Set, and tests each tractor with `contains`. That is units x train length + tractors.

`getTransportedUnits()` on its own would not cover it. A tow train is a linked list, so a tractor only lists the one trailer directly behind it, not the whole train, so the Set still has to come from a walk. Doing that walk once per unit instead of once per pair is the win.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
HammerGS: Ha. Left as is.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
HammerGS: Fixed in `0ea7f4be66`. Both warning dialogs in this method use `RESOURCE_BUNDLE` now.

MekHQ/src/mekhq/gui/menus/AssignForceToTowTransportMenu.java
HammerGS: Correct, that one is real. Fixed in `0ea7f4be66` with a new `AssignForceToTransportMenu.warning.title` key, read through a small `warningTitle()` helper since both dialogs in this method need it.

MekHQ/src/mekhq/gui/menus/AssignForceToTransportMenu.java
HammerGS: Fixed in `0ea7f4be66`, here and in the other two files where this PR added the import. Confirmed it is the house style before switching: 127 files in MekHQ already use `jakarta.annotation`.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Fixed in `0ea7f4be66`, and in the two other places this PR had the same shape (`addUnassignSelfMenuItem` and `addUnassignTransportedMenuItem`). All three are a plain loop with an early return now.

To answer the question directly: these are popup selections, so single digits, occasionally a formation's worth. Nowhere near stream territory.

This one is on me. The rule is already in our contributor notes and I wrote it three times anyway, so I have moved it to the top of both CLAUDE.md files where the hard rules live, including your point about conditionals specifically.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Partly fixed in `0ea7f4be66`, and worth explaining the half that stays.

The second walk cannot go away for this method's contract. A tow train is a singly-linked list: each unit records only the unit directly behind it, and there is no back-pointer. So listing the whole train starting from a middle trailer means finding the lead tractor first, then reading back down from it.

What I could fix is that most callers never needed the whole train. Split into `trainHead()` (walk forward) and `trailersBehind()` (walk back), with `trainMembers()` now just those two composed. Each method has one loop. The hot paths - the new menu tooltip and the tractor-side unassign - call `trailersBehind()` directly and skip the walk to the head entirely.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Done in `0ea7f4be66`. It is `parseConnectableTrain()` now, returning the one free tractor or null and filling the trailer list. Reads better, thanks.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Fixed in `0ea7f4be66`. Pulled out to a `hasTrailerHitch(Entity)` helper written as a plain loop. It scans one entity's transporters, so a handful of items at most.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Done in `0ea7f4be66`.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Done in `0ea7f4be66`.

MekHQ/src/mekhq/gui/menus/TransportAssignmentMenus.java
HammerGS: Verified, and you are half right. Fixed in `0ea7f4be66`.

Both sides genuinely change, so neither event can just be dropped: the trailer loses its assignment and its TO&E glyph, and the tractor loses a towed unit and gets its towing capacity back.

The churn is real though. Releasing a whole train puts every middle trailer in both sets - it is a released unit, and it is also the next trailer's old transport - so those units fired twice. The method now collects everything that changed into a Set and fires one event each.

MekHQ/src/mekhq/AtBGameThread.java
HammerGS: Fair call. Fixed in `0ea7f4be66`.

It did terminate: `visitedUnits.add()` returns false the second time it sees a unit. But you had to read all three breaks to know that, which is exactly the problem.

Extracted to `AtBGameThread.orderedTrainTrailerIds()`, now a `for` loop with the cycle guard in the loop condition, so termination is visible without reading the body:

```java
for (Unit towedUnit = nextTrailer(campaign, potentialTransports, leadTractorId);
      (towedUnit != null) && visitedUnits.add(towedUnit.getId());
      towedUnit = nextTrailer(campaign, potentialTransports, towedUnit.getId())) {
```

Unit tested it as you asked. New `MekHQ/unittests/mekhq/AtBGameThreadTest.java`, 4 tests: a full train, a gap where a unit is missing from the campaign, a hitch that loops back on the tractor, and a free tractor. The looped-hitch test asserts under a timeout, so a regression fails the build instead of hanging CI. The test class sits in package `mekhq` so the method can stay package-private.

MekHQ/src/mekhq/AtBGameThread.java
HammerGS: Both fixed in `0ea7f4be66`. The bundle is a `TRANSPORT_RESOURCE_BUNDLE` constant now, and the title uses `getTextAt` since it takes no format arguments.

Scoped it to the tow dialog, which is the block this PR rewrote. The DropShip, small craft, fighter and ground dialogs above repeat the same literal about ten more times, but they are pre-existing and untouched here. Happy to take them in a cleanup PR if you want them done.

IllianiBird (approved):

PR #9768, opened 2026-08-14
Close #977,  #1033, #3361, #4333, #5018, #5918, #6866, #6872, #7146, #7512, #8017, #8352, #8621, #8684, #9347, #9378, #9454, #9777: Contract Overhaul Project: Replaced AtB Contract Systems with Modern Contract Experience, Added New Contract Market

### LINKED ISSUE #977: RFE:  Have unit market and contract market available outside of ATB
Like the personnel market, can you make the unit market, and contract market available with ATB turned off?

### LINKED ISSUE #1033: Allow multiple contracts in the market to be selected at once
Can you guys adjust the contract market screen so that we can select multiple contracts entries at once using Ctrl+click and Shift+click functions? This would make it much easier to clean up unwanted contracts, especially after generating a large number of them.

### LINKED ISSUE #3361: RFE: Show the 'Contract Representative'/'House Liaison' in the Command Centre/Briefing Room/Negotiation screen with portraits and a 'lite' profile
### Environment
N/A

### Description
As an added 'fluff/flavour' component, it could be fun to show the contract representative or house liaison for a contract in the command centre and briefing room tabs, and the negotiation screen for new contracts with a civilian portrait. 

Portraits could be assigned to folders for each great house/potential employer, to give added flavour and narrative.

To support this, some kind of 'lite' personnel profile could be shown for them, or some canned dialogue/contract narrative could be generated from a file depending on the planet, contract/mission type, and the conditions of any allied and opposing forces.

Whilst adding narrative, this could also be used to highlight some of the game-impacting conditions a planet might have e.g. referencing planets with low or high gravity, high temperatures, thin atmosphere.

### Files
Wireframe/very basic sample text:

image

### LINKED ISSUE #4333: new contract start date does not account for transit
### Environment
win 11
java 17 (followed recommended link)
mmhq 49.20

### Description
started new game, took first contract, set travel, advanced day, month advanced, received notice that contract start date had been pushed back, enemy morale was set to low, and i received a shadowhawk as a special event.  start date continues to be pushed back 2 days at a time for a planet i was an estimated 80 days from.

### Files
[Galatia Mecenary Coalition27620830.cpnx.gz](https://github.com/user-attachments/files/16075902/Galatia.Mecenary.Coalition27620830.cpnx.gz)
[Galatia Mecenary Coalition27620911.cpnx.gz](https://github.com/user-attachments/files/16075903/Galatia.Mecenary.Coalition27620911.cpnx.gz)

### LINKED ISSUE #5018: [RFE] Expansion of Contracts V/ariety.

- [X] I'm reporting the RFE to the correct repository:

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

Sorry hope I'm doing this right.
I was wondering if more variety could be added to contract issuers and targets? along with major and minor powers issuing contracts. Could planets or large cooperation be added to issues contracts? Example, say a planet hires a garrison duty unit or pirate hunters or a corporation issue a raid on a rival corporation or planet? I'm sure some contracts might now work like a cooperation issuing a contract for planetary assault. but might add a bit more immersion to the game.
While I'm not a coder, I do have a extensive knowledge of the lore and could help with compiling stuff, but again. I have no coding abilities.
Hope this is enough to get the idea across.

### MekHQ Suite Version *

v0.50.0

### Attach Files

_No response_

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

### LINKED ISSUE #5918: [Issue] Date not correcting for travel time- non atb/stratcon turned on
### Brief Description *

For the last two contracts I manually did for a campaign when i put in the Location the program auto calculated travel time for the start date of the contract, on 3rd contract and it is using the date the contract is created. Stratcon/ATB is not enabled as it is a GM'd campaign/ RP campaign

### 3. Steps to Reproduce

Breifing room
Add Mission
New Contract
Location- For what i was doing Cussar from Ashio
Start date note updating

### Attach Files

[Shadow Dragons27281002.cpnx.gz](https://github.com/user-attachments/files/18579350/Shadow.Dragons27281002.cpnx.gz)

### Severity *

### MekHQ Suite Version *

0.50.02

### Operating System *

Windows 10 10.0 (amd64)

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.13

### Final Verification

### LINKED ISSUE #6866: [RFE] Nag for Subcontracts available
### Enhancement Type *

A Nag for Subcontracts available, when on Contract, I am not necessarily checking for nee contracts and can miss the notification.  A Nag from Command/Admin “hey boss, our employer likes the job we are doing and has an a new job for us, check out the Contracts Market for details.”  Would be awesome!

### Use Cases or Rationale

A Nag for Subcontracts available, when on Contract, I am not necessarily checking for nee contracts and can miss the notification.  A Nag from Command/Admin “hey boss, our employer likes the job we are doing and has an a new job for us, check out the Contracts Market for details.”  Would be awesome!

### Attach Files

None

### MekHQ Suite Version *

0.50.05

### Final Verification

### LINKED ISSUE #6872: [RFE] Allow GM Mode to change Command Rights on Contracts
### Enhancement Type *

Add functionality to Edit Contract to allow GM Mode to change Command Rights.

### Use Cases or Rationale

At the moment anyone playing as a Great House military force is stuck on all contracts being Integrated (boo hiss). Being able to change a contract at some point (either before or just after having accepted it) would be a great work around to let the Great House players play the game again.

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.05

### Final Verification

### LINKED ISSUE #7146: [RFE] AutoAward: Track successful negotiations for administrative staff
### Enhancement Type *

Tracking successful negotiations would be great (particularly if we could track big swings - renegotiating a 0% salvage to 100% for example)Tracking successful negotiations would be great (particularly if we could track big swings - renegotiating a 0% salvage to 100% for example)

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

50.06

### Final Verification

### LINKED ISSUE #7512: [RFE] Add campaign option for Max Battle Loss Compensation Percent
### Enhancement Type *

New Feature

We currently have the "Max Salvage Percent" campaign option for contract markets, so players can choose to limit the salvage terms of all generated contracts and avoid growing too fast. However, it is not uncommon the battle loss compensation close to 100%, which means you don't lose any money on taking losses and that salvage is still leading to consistent if slower unit growth.

I think adding a similar clamping option for the battle loss compensation could help balance things out, where fighting in battles wouldn't be as much of a growth boost as currently. Plus enourage retreating damaged units to save them, rather than fighting to the end in order to get more salvage while knowing that anything you lose is the same as being sold at 100% compensation.

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

0.50.06

### Final Verification

### LINKED ISSUE #8017: [RFE] Pirate looting
### Enhancement Type *

New Feature

(copied from discord)
So for Pirate StratCon games, I mentioned the idea of loot to Illiani a few weeks back, who suggested if I had an idea for it to submit an RFE.  I've pondered it and here is the idea I have come up with.  Posting here to get feedback before I submit it as an RFE.

The motivation here is that pirates don't raid to get into fights and take away combat salvage (they wount have such a bad rep if that was all they wanted).  To the contrary they much prefer minimal combat, instead they raid where they think they can find easy low risk targets with high value items, things like raw ores, jewelry, computer chips, laptops, or expensive industrial equipment, or even slaves (😬), etc.  I figure in a pirate game, the contract payment is actually an abstraction of any profits from fencing loot (as who would be paying a pirate to raid).  That said, I felt being able to collect actual loot once in a while would add some nice flavor to the game.  So I created a list of priorities:

- add loot in a way that makes sense, without breaking the economy balance
- use the existing code/mechanics as much as possible to minimize the effort to add the system to MHQ
- draw upon existing and established resources as much as possible

I referenced StratOps and TacOps extensively, and used them as a guide as much as possible, but there were very little RAW that I could find that even peripherally covered this.

When StratCon is activated **and** the player faction is Pirate, when generating base assault missions:
1.  calculate a random C-Bill value:  Loot value = monthly contract pay*(7%+X), where X is a random value between 0%-6%
2.  once the loot value has been determined, randomly determine the potential number of units.  My draft proposal, is an even 25% chance for each of four outcomes, 1 unit, 2 units, 3 units, and 3 units + check again
3.  once the total number of possible units has been determined in this way, MHQ selects a random unit from a civ/support RAT, assigns it to the base assault scenario using the existing reward system, then subtracts the units value from the total loot value.
4.  checks if the new loot value is negative, if it is it stops.  if the loot value is still positive, it checks to see if it still needs to generate another unit and repeats step 3.  this process continues until either the loot value becomes negative, or all units as defined by step 2 are generated.
5.  if after all units are generated, loot value remains positive, add that value as a C-Bill reward to the scenario.

The idea is that this can simulate the various different pieces of support equipment that a pirate group might find while raiding that they might actually decide to keep instead of just selling.  With the increase in need and usefulness of such equipment, and pirate games having more limited options to gain them as compared to other games, this gives them a way to acquire said support in a way that makes sense for that style of game.

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MekHQ Suite Version *

11/06 nightly

### Final Verification

### LINKED ISSUE #8352: OpFor and Employer show as same
### Brief Description *

Image

The employer and the OpFor both show as Star League, the background image however is SL/RWR.  Based on comments from Illiani, this appears to just be a display error.

### 3. Steps to Reproduce

I don't know.

### Attach Files

[Employer-Employee Files.zip](https://github.com/user-attachments/files/23757945/Employer-Employee.Files.zip)

### Severity *

### MekHQ Suite Version *

11/25 nightly

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

### LINKED ISSUE #8621: [Issue] Star League Attacking Hegemony in Contracts
### Brief Description *

As described in title, faction diplomacy not being taken into account generating contracts, getting contracts for the Hegemony against the Combine etc during the Star League era and no Star League contracts as a Star League faction campaign.

### 3. Steps to Reproduce

Wait until month roll over after contract complete, contracts will generate for hegemony against other star league members when based on Terra.

### Attach Files

[House Pendragon-Cameron26831203_Careful-part1.zip](https://github.com/user-attachments/files/24398697/House.Pendragon-Cameron26831203_Careful-part1.zip)

### Severity *

### MekHQ Suite Version *

50.11

### Operating System *

Windows 10

### Java Version *

 Java Vendor: Eclipse Adoptium     Java Version: 17.0.15

### Final Verification

### LINKED ISSUE #8684: [Issue][50.11] New transit cost calculations not playing well with un/mothballing
### Brief Description *

Since 50.11, compared to 50.10, drop ship transport handles mothballed units as cargo (it might have done so before, but not in 50.10 AFAICT). This creates a few issues in the way it is currently implemented:

1) The handy jump fee summary **doesn't show what's the cargo capacity of a single cargo drop ship** (unlike what it shows for personnel and mek bays), making it kind of hard to optimize
2) It becomes kind of **hard to guess the number of drop ships** required before mothballing. Workaround is to use GM de/mothballing, though.
3) **Nothing prevents de-mothballing while in transit** (in fact the new free spaceport mothballing would do so). That would seem impossible to do if there are insufficient bays. Or it should increase travel costs for extra bays for further jumps, but that seems needlessly complex.
4) Contract transport costs probably do not factor mothballing.

IDEAS FOR FIXES
A) Easy fix: Disallow de-mothballing in transit. It makes for a trade-off: you can't hit the ground running, but it's usually cheaper.
B) Adds a simple mention of the baseline dropship cargo allowance in the jump fee summary.
C) Possibly add a second computation ("cargo assuming all mothballed units") to the jump fees summary 

UNREQUITED COMMENT
I'm not too sure it makes sense to carry mothballed meks as cargo to start with (while travels can be very long, why do bays exist if not to handle and carry them safely?), but I s'pose it was debated and decided upon.
Disallowing this would prolly anger some, but it allays the above issues...

SOTA
I've seen discussions pertaining to mothballing, but not that feature. I have NOT checked Discord because I don't.

### 3. Steps to Reproduce

1. mothball all units
2. check jump fee summary
3. start transit
4. un-mothball unit
5. watch your drop ship rental fleet auto-expand

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.11

### Operating System *

win10

### Java Version *

17.0.11

### Final Verification

### LINKED ISSUE #9347: [Issue] Support Units Still Affect Pay
### Brief Description *

Units marked support in the TO&E are factored into contract pay. Though they do affect required forces.

### 3. Steps to Reproduce

In this save there are two pirate hunting contracts available. One was generated while my entire TO&E was combat. It requires 74 combat units and pays 57m.
The other was generated while only my 3 "always a combat team" companies were active and all other combat units were marked as support. It requires 24 combat units and pays 52m. 

If I go into the TO&E and mark all the combat units as combat, neither changes. If I then goto assignments in the Briefing Room and change everything but the mech and armor companies to "Reserve" both contracts automatically adjust down to ~20ish million.

### Attach Files

[Knight's Errant25800401_Keeper-part1.zip](https://github.com/user-attachments/files/28924366/Knight.s.Errant25800401_Keeper-part1.zip)

### Severity *

### MekHQ Suite Version *

51

### Operating System *

No

### Java Version *

Black, 2 creams, 3 sweetners

### Final Verification

### LINKED ISSUE #9378: [Issue] Using the planet link in the Contract Market no longer correctly takes you to the renamed Navigation screen.
### Brief Description *

When opening the Contract Market and inspecting a contract, clicking on the name of the target plant no longer correctly takes you to the Navigation tab. It still correctly selects the planet when you manually navigate to the Navigation tab, but expected functionality is that clicking the planet link immediately takes you to said planet in the Navigation view. 

### 3. Steps to Reproduce

1. Open any save.
2. Open contract market - select a contract, or generate one.
3. Click the planet link in the contract details. Note that nothing appears to happen.
4. Manually navigate to the Navigation tab. Note selected planet is correctly highlighted. 

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

MekHQ v0.51.01-nightly-2026-06-16

### Operating System *

Windows 11 10.0 (amd64)

### Java Version *

    Java Vendor: Eclipse Adoptium     Java Version: 25.0.1

### Final Verification

### LINKED ISSUE #9454: [Issue] False Start at contract start
### Brief Description *

After accepting a contract, MekHQ continued through all of the starting contract dialogs like normal, after going through the dialogs I immediately selected "Advance Multiple Days" however after advancing days, we did not move, there was no notification that jump drives were charging and no other reports associated with movement. Going to "Navigation" and attempting to start transit or recalculate jump path and start transit did not change the outcome

Note: This is a new start campaign. 

This problem did resolve on save reload. 

### 3. Steps to Reproduce

1. Accept new contract
2. Go through all dialogs, and have units mothballed for travel
3. At conclusion of all dialogs selected "General" tab to see how many days travel would take
4. Advance multiple days for the duration of travel

At conclusion --> we were still at Galatea.

### Attach Files

[Kansas Irregulars30440712_Transportation_no-go-part1.zip](https://github.com/user-attachments/files/29222289/Kansas.Irregulars30440712_Transportation_no-go-part1.zip)

### Severity *

### MekHQ Suite Version *

0.51.01 Nightly Active Build

### Operating System *

Win 11

### Java Version *

JDK 21

### Final Verification

### LINKED ISSUE #9777: [RFE] In mission Rental of Support facilities
### Enhancement Type *

New Feature

Can we get a way to contract support facilities, Hospital Beds / Kitchens / Holding Cells in mission?

### Use Cases or Rationale

Currently we can only request additional facilities when initially negotiating a contract.  The problem becomes sometimes you need more than you thought.  These in-mission facilities should cost more than the initial contract negotiated ones, or maybe cost Support points or something along those lines

### Attach Files

_No response_

### MekHQ Suite Version *

0.51.0

### Final Verification

IllianiBird:
Close #977
Close #1033
Close #3361
Close #4333
Close #5018
Close #5918
Close #6866
Close #6872
Close #7146
Close #7512
Close #8017
Close #8352
Close #8621
Close #8684
Close #9347
Close #9378
Close #9454
Close #9777

# Requires https://github.com/MegaMek/mm-data/pull/527
# Requires #9796

I will confess to being rather exhausted after this PR, so the description won't be as verbose as normal.

This PR performs the following tasks:
- Retire & remove AbstractMissionTransition and its inheritors (Mission, Contract, and AtBContract).
- It introduces the new 'mission style object' class: AbstractContract
- It removes the old AtB Contract Market
- It adds a new bespoke contract market built around Hot Spots Draconis Reach's contract generation system.
- It removes the old AtB Create and Edit Contract dialogs
- It adds a new dual-purpose Create/Edit Dialog
- It performs widespread refactoring to avoid for the removal of AtBContract and its related systems.
- It adds around 10k lines of new tests.

This PR contains AI tests and AI was used during quality assurance to ensure that as few bugs as possible will be introduced. However, this PR represents a paradigm shift from AtB to Modern and it is not possible to make an omelet without cracking a few eggs.


HammerGS:
## Review 1/5 — Save/load migration (`ContractXmlCodec`, `LegacyContractConverter`, `CampaignXmlParser`)

Reviewed against `main` @ `abe68d4`. `:MekHQ:compileJava` passes clean. Unit tests skipped per your note.

Doing this in sections; this is the persistence layer only. Ordered roughly by severity.

---

### Data loss / NPE risk

- [x] **`ContractXmlCodec.java:552-563` — `opposingCommander` is written but never read back.**
  `writeEnemyData` emits it at line 187, `EnemyDataBuilder.opposingCommander` is declared at line 544, but `createEnemyBinders()` has no entry for it. Every save/load silently nulls the enemy commander, and `EnemyData.opposingCommander` is not `@Nullable`.
  *Fix:* add `binders.put("opposingCommander", (b, n, c, v) -> b.opposingCommander = parseWrappedPerson(n, c, v));`. Note `parseEnemyData` (line 566) currently passes `null, null` for campaign/version — those need threading through from the `enemyData` reader for the binder to work.

- [x] **`ContractXmlCodec.java:370-374` — player negotiator is lost on every load.**
  `<contracts>` is written inside `<info>` (`Campaign.java:5263`, block closes at 5279), but personnel are written later at `Campaign.java:5305`. So when `playerNegotiatorId` resolves against `getHumanResources().getPerson(...)` the roster is still empty and the result is always `null`. Same applies to `ContractMarket` contracts written at `Campaign.java:5264`.
  *Fix:* stash the raw UUID during parse and resolve it in a post-load pass, same pattern as the new `relinkLegacyMissions` below.

- [x] **`LegacyContractConverter.java:299` — `placeholderPerson` is `@Nullable` but feeds non-nullable record slots.**
  Its result goes straight into `EmployerData` (240-242) and `EnemyData` (243-244) at lines 222-224 with no null check. If `newPerson` throws, three "never null" fields become null.
  *Fix:* either fail the conversion or guarantee a non-null fallback.

- [x] **`LegacyContractConverter.java:322` — `systemIdOrPlaceholder` can return `null`.**
  It returns the original (possibly null) `systemId` when `getCurrentSystem()` is null, and line 238 passes it into `SystemsTargetData` unchecked.

### Conversion drops data that IS present in legacy saves

Legacy tags confirmed against `AbstractMissionTransition.writeToXML` on `main`. The tag names the converter switches on all match — but these four are in the old XML and are being thrown away:

- [x] **`LegacyContractConverter.java:222-224` — `clanOpponent` and `employerLiaison` are discarded.**
  Legacy saves store both as full `<person>` nodes. The converter ignores them and generates *random* placeholder NPCs instead, so on migration every contract's enemy commander and liaison are replaced by strangers.
  *Fix:* parse those two nodes into `Person` and fall back to the placeholder only when absent.

- [x] **`LegacyContractConverter.java:244` — `batchallAccepted` hardcoded to `true`.**
  The legacy tag exists. A player who refused a Batchall has that silently reverted.
  *Fix:* add `case "batchallAccepted" -> batchallAccepted = Boolean.parseBoolean(value);` and pass it through.

- [x] **`LegacyContractConverter.java:238` — `planetName` is available but `planetId` is set to `null`.**

- [x] **`LegacyContractConverter.java:200-205` — `parentContractId` is dropped.**
  Subcontract parent/child linkage is lost with no warning. If that's intentional, worth a `LOGGER.warn` so it shows up in a migration log rather than being invisible.

### Correctness / polish

- [x] **`CampaignXmlParser.java:1630` — `relinkLegacyMissions` miscounts.**
  `int total = relinked;` seeds the total from the tracker's *successes*, not its *attempts*, so the `total - relinked` "had no match" figure in the log at 1641 is wrong whenever the tracker fails to resolve something.
  *Fix:* have `relinkLegacyMissionIds` return attempts and successes, or count them separately.

- [x] **`ContractEmergencyExtension.java:54` — resource bundle filename typo.**
  Loads `mekhq.resources.ContractEmergencyExtension`, but the file added is `CotractEmergencyExtension.properties` (missing the `n`). Every string in that feature will fail to resolve.
  *Fix:* rename the properties file.

- [x] **`Campaign.java:5357` and `5368` — dead `TODO : AbstractContractMarket` comment blocks.**
  They reference a class this PR deletes; the commented-out `getContractMarket().writeToXML(pw, indent)` is superseded by line 5264.

- [x] **`ContractXmlCodec.java:44`, `PlayerForce.java:43`, `LegacyContractConverter.java:42` — `jakarta.annotation.Nullable`.**
  Style guide calls for `megamek.common.annotations.Nullable`.

### Checked and OK (no action)

- `ContractObjectiveType.parseFromString` handles both enum names and legacy ordinals — legacy `contractType` converts correctly.
- `Camouflage(category, filename)` is null-tolerant, so the partial-camo case at `LegacyContractConverter.java:331` is safe.
- `ContractMarket.MARKET_TAG` reusing the legacy `"contractMarket"` name is safe — the new one is scoped under `<info>` and legacy top-level nodes now fall through unhandled.

---

Sections 2-5 (domain model, generation, deletions/call-site migrations, GUI) to follow.

HammerGS:
## Review 2/5 — Contract domain model (`AbstractContract`, `contractData/*`)

Two of these are, I think, campaign-breaking. Ordered by severity.

---

### Blockers

- [x] **`ContractHistoryData.java:73-96` — `getActiveAndStarted` and `getActiveAndNotYetStarted` are inverted.**

  ```java
  getActiveAndStarted(currentDate):        if (!startDate.isBefore(currentDate))   // startDate >= today  -> FUTURE
  getActiveAndNotYetStarted(currentDate):  if (startDate.isBefore(currentDate))    // startDate <  today  -> ALREADY STARTED
  ```

  Each method returns what the other one promises. Compare `main`: `getActiveContracts(false)` resolved to `isActiveOn(today, false)`, i.e. `startDate <= today <= endDate`, and `getFutureContracts()` was `startDate.isAfter(today)`.

  Effect: from the **day after** a contract starts, `Campaign.getActiveContracts()` (line 1370) stops returning it. That feeds monthly payroll, morale checks, faction standing, StratCon new-day, and `hasActiveContract()` — the campaign starts behaving as if it were between contracts while a contract is running. `getFutureContracts()` (line 1382) meanwhile returns contracts already in progress, which feeds `AbstractLocation:231,280` travel logic.

  *Fix:* swap the two conditions, and restore the end-date bound that `getActiveAndStarted` dropped — it currently never checks `endDate` at all, so expired contracts stay "active" indefinitely.

- [x] **`AbstractContract.java:709-723` — salvage accumulators discard their result.**

  ```java
  public void changeSalvagedByEmployerValue(Money delta) {
      salvagedByEmployerValue.plus(delta);   // Money is immutable; return value dropped
  }
  ```

  Same at 721-723. `Money.plus` returns a new instance (`Money.java:106`), so both totals stay at `Money.zero()` forever. Four live callers in `CamOpsSalvageUtilities.java:227,248,261,277`.

  Effect: the salvage gauge in `MissionViewPanel:281-282,339-356` always reads 0%, and the salvage-rights cap is never reached, so employer salvage split silently never applies.

  *Fix:* `salvagedByEmployerValue = salvagedByEmployerValue.plus(delta);`

### NPE risk

- [x] **`ContractHistoryData.java:78,93` — `contract.getStartDate()` dereferenced without a null check.**
  `getMissionSortKey` (line 115) explicitly handles a null start date, so null is a known state for these objects; these two loops will NPE on the same data.

- [x] **`ContractScheduleData.java` `isActiveOn` — NPEs on a null `startDate`/`endDate`.**
  Reached from `AbstractContract:311` and then from 12 call sites including `MissionViewPanel:136,371,416`, `BriefingTab:2612-2613`, `TrainingCombatTeams:145`. Unsigned market offers and legacy-converted contracts can both carry null dates (`LegacyContractConverter:118-119` only sets them if the legacy tags were present).

- [x] **`AbstractContract.java:725` — `getMonthsLeft` NPEs when `endDate` is null.**
  Callers: `Campaign:7633`, `CampaignNewDayManager:1344`, `MHQMorale:631`.

- [x] **`ContractScheduleData` copy constructor — `ChronoUnit.MONTHS.between(...)` on possibly-null dates.**
  `AbstractContract.updateScheduleData(null, null)` (line 306) falls through to the existing values, which may themselves be null.

  Also worth a second look: this constructor **silently recomputes `lengthInMonths` from the dates**. So `updateScheduleData(newStart, null)` changes the contract's agreed length as a side effect of moving the start date. If that is intended it needs a comment; if not it's a live bug.

### Correctness / polish

- [x] **`AbstractContract.java:230` — `getDescription()` is annotated `@Nonnull` but `description` (line 80) has no initializer.**
  Null until `setDescription` runs. Either initialise to `""` or drop the annotation.

- [x] **`AbstractContract.java:756` — `getMaximumSupportPoints()` hardcodes `scale * 3`.**
  The javadoc says it mirrors the cap in `SupportPointNegotiation`. Two copies of the same rule will drift; pull the `3` into a shared constant and reference it from both.

- [x] **`AbstractContract.java:538-544` — `getSponsorFactionCode()` / `getSponsorFaction()` silently mean the *enemy's* sponsor.**
  `EmployerData` also has a `sponsorFactionCode`, so an unqualified `contract.getSponsorFactionCode()` reads as though it could be either. Suggest `getEnemySponsorFactionCode()`.
  Also `getSponsorFaction()` passes a `@Nullable` code straight into `Factions.getFaction(...)`.

- [x] **`AbstractContract.java` — `jakarta.annotation.Nullable` / `Nonnull` again; style guide wants `megamek.common.annotations.Nullable`.**

- [x] **`AbstractContract.java:62` — wildcard import `contractData.*` replaced eight explicit imports.**
  Style guide allows wildcards only above 10 members. Same pattern in `ContractXmlCodec:62` and `LegacyContractConverter:60`.

### Checked and OK (no action)

- `ContractFinanceData` copy constructor null-means-keep semantics are consistent with the three `updateXPay` helpers.
- `changeMorale(int delta)` clamps correctly and handles the `@Nullable` return from `parseFromLevel`.
- `getSortedMissions` ordering matches its javadoc.

---

Next: deletions and call-site migrations (`CombatTeam`, `MissionViewPanel`, `BriefingTab`, `RetirementDefectionTracker`, `StratConRulesManager`).

HammerGS:
## Review 3/5 — Deletions and call-site migrations

Focus here was on logic lost in the AtB retirement rather than relocated. Most of the ~60 removed `instanceof AtBContract` guards are correct now that everything is an `AbstractContract` — I spot-checked the risky ones and they're fine (e.g. the `FacilityRemains`/`FacilityRemoved` cases in `ResolveScenarioTracker` correctly kept the `instanceof AtBScenario` half and only swapped the contract check for a null check).

These are the ones that need attention.

---

### Blockers

- [x] **`RetirementDefectionTracker.java:995` — `resolveAllContracts()` no longer clears anything.**

  The public `resolveContract(@Nullable Mission)` overload was deleted. It used to do:
  ```java
  if (mission == null) {
      unresolvedPersonnel.keySet().forEach(this::resolveContract);
      unresolvedPersonnel.clear();
  }
  ```
  Now `resolveContract(null)` binds to the surviving `private void resolveContract(UUID)` at line 999, where `unresolvedPersonnel.get(null)` is null and `rollRequired.remove(null)` is a no-op.

  Effect: `payouts.clear()` still runs, but `unresolvedPersonnel` and `rollRequired` are never cleared. `isOutstanding(...)` (line 985) keeps returning `true` after everything has been paid out, so the turnover roll stays flagged as required and the player gets re-prompted indefinitely. The javadoc at 990-992 still describes the old behaviour.

  *Fix:* restore the null branch inside `resolveAllContracts()`.

- [x] **`AtBScenarioFactory.java` — `scenarioMap` is now permanently empty.**

  The `static {}` block that called `registerScenario(...)` 29 times is gone along with the `atb/scenario` package, but `registerScenario` (line 94) survives with zero callers. So `getScenarios(type)` (line 62) always returns `null` and `createScenario(...)` (line 65) always returns `null`.

  Two knock-ons:
  1. `createScenariosForNewWeek` still runs every week from `CampaignNewDayManager.java:2254`, building combat-team tables and iterating contracts before calling a factory that can only return null.
  2. `Scenario.java:1190-1195` — every legacy AtB scenario in an old save now fails to load and logs `"Unable to load an old AtBScenario of battle type {}"`. That includes the ones `LegacyContractConverter.convertScenarios` (line 291-294) is explicitly trying to carry over, so converted contracts come back with an empty scenario list.

  *Fix:* if legacy AtB scenario generation is intentionally retired, drop the `createScenariosForNewWeek` call and `AtBScenarioFactory`/`AtBScenarioEnabled`/`IAtBScenario` with it, and have the converter skip scenarios deliberately instead of failing per-scenario. If it isn't retired, the registrations need to come back.

### Silent rules change

- [ ] **`RetirementDefectionTracker.getTargetNumbers` — the shares modifier was removed with no replacement.**

  The whole `if (campaignOptions.isUseShareSystem())` block went, including `targetNumber.addModifier(-max(0, ((contract.getSharesPercent() / 10) - 2)), resources.getString("shares.text"))`. With the share system on, turnover rolls are now strictly harsher than before and nothing else compensates.

  `shares.text` is left orphaned at `RetirementDefectionTracker.properties:55`, and the `import static java.lang.Math.max` was dropped with it, which confirms nothing replaced it.

  *Fix:* if the new contract model has no shares equivalent this needs to be a stated design decision in the PR body — right now it reads as collateral damage from the type migration.

### Dead code left behind

- [x] **`gui/dialog/CustomizeAtBContractDialog.java` — zero references.**
  Its siblings (`CustomizeMissionDialog`, `NewAtBContractDialog`, `NewContractDialog`, `MissionTypeDialog`, `ContractMarketDialog`) were all deleted; this one was missed. It's also the only remaining importer of `mekhq.campaign.mission.AtBContract`.

- [x] **`AtBScenarioEnabled` is orphaned** — the only thing that reads the annotation is the now-unreachable check inside `registerScenario`.

- [x] The legacy `Mission` / `Contract` / `AtBContract` hierarchy is now effectively vestigial (`import mekhq.campaign.mission.Contract;` has 0 users, `Mission` has 1). If it's staged for a follow-up PR that's fine — worth saying so in the description so reviewers don't chase it.

### Style

- [x] **9 inline fully-qualified `@jakarta.annotation.Nullable` annotations** — the style guide requires an import rather than an FQN mid-code:
  - `Campaign.java:1324`
  - `CombatTeam.java:89, 149, 153, 477`
  - `ChaosContractMarketAvailability.java:432, 438`
  - `ResolveScenarioTracker.java:1683, 1687`

### Minor / FYI

- [x] Daily-report logs persisted in old saves contain `MISSION:<int>` hyperlinks. Post-migration those fail `UUID.fromString` at `ReportHyperlinkListener.java:102`. It's caught and logged (line 104), so it's a dead link rather than a crash — just noting it as expected fallout.

- [x] `AtBScenarioFactory.createScenario` line 74 uses `new Random()` per call rather than the shared RNG. Pre-existing, but the method is being touched.

---

Next: contract generation (`ChaosContract*`), then the new market GUI.

HammerGS:
## Review 4/5 — Contract generation (`ChaosContract*`, `AbstractContractGeneration`, acceptance)

The first item here looks like it crashes on accepting any market contract, so it's worth checking before anything else.

---

### Blocker

- [x] **`AbstractContractGeneration.java:170` — generated contracts get `setStatus(null)`, and nothing sets a status on acceptance.**

  `ContractAcceptance.accept` (line 82) calls `campaign.addMission(contract)` at line 96, and `Campaign.addMission` just puts it straight into `contractHistory` — neither sets a `MissionStatus`. I grepped every `setStatus` call site: only `NewContractFactory.java:101` (the GM create path) assigns `ACTIVE`. `ContractAutomation.performContractStart` doesn't either.

  So an accepted market contract lands in `contractHistory` with a null status, and every filter in `ContractHistoryData` dereferences it unguarded:

  ```
  ContractHistoryData.java:49   !contract.getStatus().isActive()      // getCompleted
  ContractHistoryData.java:63    contract.getStatus().isActive()      // getActiveIncludingNotYetStarted
  ContractHistoryData.java:77    contract.getStatus().isActive()      // getActiveAndStarted
  ContractHistoryData.java:92    contract.getStatus().isActive()      // getActiveAndNotYetStarted
  ContractHistoryData.java:110   mission.getStatus().isCompleted()    // getSortedMissions
  ContractHistoryData.java:119,122                                    // getMissionSortKey
  ```

  `Campaign.getActiveContracts()` and `getSortedContracts()` are called from the Briefing Room refresh and all over the new-day loop, so this should NPE almost immediately after accepting — `accept` even fires `MissionChangedEvent` at line 123 on its way out.

  *Fix:* set `MissionStatus.ACTIVE` in `ContractAcceptance.accept` before `addMission`. Worth also guarding the `ContractHistoryData` filters, since `ContractXmlCodec.writeContract:119` already treats a null status as a legitimate persisted state for market offers.

  *Related:* `setStatus(null)` at generation time is deliberate (an unaccepted offer isn't a mission yet), but then `AbstractContract.getStatus()` should be `@Nullable` so the compiler pushes callers to handle it.

### NPE risk

- [x] **`ChaosContractDeterminationDifficulty.java:88` — `contract.getStartDate().getYear()` is unguarded.**
  Safe on the generation path (Step 6 `determineSchedule` sets the schedule at line 335 before Step 10 computes difficulty at 158), but the method is `public` and takes any `AbstractContract`. A legacy-converted contract can reach it with a null start date. Worth a guard returning `UNKNOWN_DIFFICULTY`.

- [x] **`ChaosContractMarketAvailability.java:262` and `:364` — `campaign.getCurrentSystem()` dereferenced without a null check** in `isUninhabitedSystem` and the skill-check report header.

### Logic

- [x] **`ChaosContractMarketAvailability.java:292-294` — the monthly refresh clears `TOURNAMENT` offers too.**
  The loop runs over `ContractSearchType.values()`, so `market.getContracts(TOURNAMENT)` is cleared every month, while `rollMonthlyOffers` deliberately skips TOURNAMENT (line 229) so `offerCount` returns 0. Nothing anywhere populates the tournament map, yet it's a visible tab in `ChaosContractMarketDialog:507,509`.
  *Fix:* if tournaments are a later feature, restrict the clear loop to `REPORTED_TYPES` so the map isn't wiped by a system that doesn't own it; otherwise it needs a generator.

- [x] **`ChaosContractMarketAvailability.java:299` — the summary report counts rolled offers, not generated ones.**
  `totalOffers += count` uses the roll result, but the loop above only stores what `generateOffers` actually returned, and that "may be shorter than count or empty" per its own javadoc at 402-403. The player can be told "3 Mercenary contracts available" and find one. The javadoc at 282-283 says this is intended — I'd push back: the daily report is the player-facing number and should match the market.

- [x] **`ChaosContractDeterminationDifficulty.java:101` — unit mismatch when `playerPower == 0`.**
  ```java
  double percentDifference = (playerPower != 0 ? difference / playerPower : difference) * 100;
  ```
  The fallback branch feeds a raw BV difference into something scaled and labelled as a percentage, so it always clamps to difficulty 10. That's probably the intended outcome (no player units = hardest), but it should be an explicit early return rather than falling out of the arithmetic.

### Style

- [x] `ChaosContractMarketAvailability.java:432, 438` — inline `@jakarta.annotation.Nullable` (also flagged in section 3).
- [x] `AbstractContractGeneration.java:81-84` — `createContract` takes 10 positional parameters, four of them booleans (`overridingCommandCircuitRequirements`, `isGM`, `provingGround`, plus the modifier int). The style guide asks for a parameter object at 4+ params / 3+ boolean flags; a `ContractGenerationContext` record would also remove the risk of transposing the three booleans at the `generateOne` call site (line 447-456).

### Checked and OK (no action)

- `averageBattleValue` guards its divisor (line 236-238).
- `rollOffer` clamps the success target so a 0 result stays reachable, and `baseOfferSlots`/`baseSuccessTarget` are exhaustive switches with no default — a new `HiringHallLevel` will fail the build rather than silently defaulting. Good.
- Generation ordering is sound: employer -> objective -> enemy -> location -> ratings -> schedule -> terms -> difficulty, with null-returns on unplaceable employer (line 91) and target (line 116).

---

Last section: the new market GUI (`ChaosContractMarketDialog`, `ContractEditorDialog`, `ContractNegotiationDialog`, `ContractDossierPanel`).

HammerGS:
## Review 5/5 — New market GUI (`markets/contractMarket/*`)

The new GUI is in good shape — i18n discipline is consistent (no `String.format` on resource strings anywhere; `getTextAt`/`getFormattedTextAt` throughout) and the formatting helpers guard their edge cases properly. These are the exceptions.

---

### MekHQ conventions

- [x] **`ContractNegotiationDialog.java:612` — `JOptionPane.showMessageDialog`.**
  MekHQ's rule is `ImmersiveDialogSimple` for user-facing dialogs. There is no `ImmersiveDialog*` usage anywhere in the new market GUI, so this is the one spot to convert (`LOGISTICS` or `COMMAND` speaker would suit a terms explainer).

- [x] **`ContractNegotiationDialog.java:411, 469` — hardcoded `"−"` button labels, and it's U+2212 MINUS SIGN, not ASCII.**
  Two rules at once: user-facing string not in a bundle, and non-ASCII in source. The paired `"+"` at lines 413 and 475 is also hardcoded (ASCII, so lower risk).
  *Fix:* move all four to `ChaosContractMarketDialog.properties` and use ASCII `-` / `+`.

- [x] **Non-ASCII in comments and javadoc:**
  - `ContractEditorDialog.java:190` — em dash
  - `MHQMorale.java:329, 350, 410` — curly quotes, `≤`, `≥`

- [x] **`ChaosContractMarketDialog.properties:18, 63, 210, 246, 269` — middle dot `·` (U+00B7) as a separator.**
  Only 4 of the 265 property files in the repo contain any non-ASCII, so this isn't established practice here. A plain `-` or `|` avoids any encoding question.

### Correctness

- [x] **`ContractEditorDialog.java:691` — `setStatus(enumValue(statusCombo, contract.getStatus()))` can write null straight back.**
  Falls out of the null-status issue in section 4; once acceptance assigns `ACTIVE` this becomes moot, but the fallback should be a concrete default rather than the current (possibly null) value.

- [x] **`ContractEditorDialog.java:700-702` — schedule length can be saved inconsistent with the dates.**
  It uses the canonical 3-arg `ContractScheduleData` constructor with `intValue(lengthSpinner)`, bypassing the copy constructor's date-derived recomputation. A GM can save start/end dates that don't match `lengthInMonths`, and `getMonthsLeft` reads the dates while pay calculations read the length. Worth either validating on save or deriving one from the other.

- [x] **`ContractEditorDialog.java:725` — edit mode carries `enemy.opposingCommander()` through by reference.**
  Correct as written, but see section 1: that field is null on any contract that has been through a save/load, so the commander editor will be blank on a reloaded campaign. Fixing the codec binder fixes this too.

### Minor

- [x] **`ContractMarketFormatting.java:87-88` — the two skull PNGs are re-read from disk and re-scaled on every `difficultySkulls` call**, i.e. once per contract card (`ContractCardPanel.java:205`). Hoisting the scaled `ImageIcon`s into statics keyed by size would avoid the repeat disk hits when the market lists a dozen offers.

### Checked and OK (no action)

- `ContractProfitEstimate.estimate` — the unpaid-transit model is sound and `getTotalMonthlyPay` correctly scales by contract length.
- `difficultySkulls` and `assessmentDifficulty` apply the same `< 1 -> DEFAULT_DIFFICULTY` guard, so the skulls and the assessment word can't disagree.
- `generateAndApplyNpcs` (line 771+) null-guards employer faction, enemy faction and current system before generating.
- HTML string concatenation in the panels is structural markup wrapping i18n content, which is the existing pattern.

---

## Wrap-up

That completes the pass: 277 files, +12,645/−15,768, reviewed against `main` @ `abe68d4`. `:MekHQ:compileJava` passes. Unit tests skipped per your note.

**The seven I'd fix before anything else:**

1. `ContractHistoryData.java:73-96` — `getActiveAndStarted` / `getActiveAndNotYetStarted` are inverted, and the former never checks `endDate`. Breaks `getActiveContracts()` campaign-wide. *(§2)*
2. `AbstractContractGeneration.java:170` + `ContractAcceptance.accept` — accepted contracts carry a null `MissionStatus` into `contractHistory`, which every `ContractHistoryData` filter dereferences. Likely NPEs on accept. *(§4)*
3. `AbstractContract.java:709-723` — salvage accumulators discard `Money.plus`, so salvage tracking is permanently zero. *(§2)*
4. `RetirementDefectionTracker.java:995` — `resolveAllContracts()` no longer clears `unresolvedPersonnel`/`rollRequired`. *(§3)*
5. `AtBScenarioFactory` — scenario registry is permanently empty; legacy AtB scenarios in old saves all fail to load, including ones the converter tries to carry over. *(§3)*
6. `ContractXmlCodec.java:552-563` — `opposingCommander` written but never read; lost on every save/load. *(§1)*
7. `Campaign.java:5263` vs `5305` — `<contracts>` is written inside `<info>`, before `<personnel>`, so the player negotiator never resolves on load. *(§1)*

**Two that are silent rules changes rather than defects**, and should be called out in the PR description either way: the shares modifier removed from turnover target numbers *(§3)*, and legacy `clanOpponent` / `employerLiaison` / `batchallAccepted` discarded during conversion in favour of random placeholders *(§1)*.

Happy to re-review any section once you've had a pass at these.

IllianiBird:
Hi Future Illiani, it's Past Illiani, make sure (when you review Claude's feedback) that you also check: https://github.com/MegaMek/mekhq/pull/9764

HammerGS:
## Final pass — re-reviewed at `b45f897`

All seven blockers are fixed, and fixed properly — the deferred `pendingPlayerNegotiatorId` and the `hasStatus`/`isActive`/`isCompleted` helpers solve the class of problem rather than the reported symptom. `:MekHQ:compileJava` passes, and non-ASCII is clear across every PR-owned file. Sweeping the fix commits turned up the following.

### Regressions

- [x] **`Campaign.java:7689-7729` — `completeMission` lost the salvage overage repayment and the end-of-contract share payout.**
  Both blocks are in `main:7901-7930` with no replacement. The salvage one matters twice: it was the only consumer of the salvage accumulators you just fixed, so those totals now feed nothing.
  *Fix:* port both against `AbstractContract` — a `ContractSettlement` helper would avoid growing `Campaign` further.

- [x] **`AbstractContract.java:806-813` — `getMonthsLeft` dropped the partial-month round-up (`main:377-379`).**
  Contract ends 15 June, completed 20 March: was 3 months of escrow, now 2. Also hits `CampaignNewDayManager:1355` and `MHQMorale:632`.
  *Fix:* restore the day-of-month increment. Separately, the new `max(0, ...)` clamp plus the deleted overage block leaves `Campaign:7714`'s negative branch unreachable — confirm that is intended.

- [x] **Six campaign options no longer have a read site, but still have controls on `ContractMarketPage`.**
  `getContractMaxSalvagePercentage`, `isClampReputationPayMultiplier`, `isContractMarketReportRefresh` and `isUseBolsterContractSkill` died with the AtB market — remove the controls. `isOverageRepaymentInFinalPayment` and `getContractNegotiationXP` are behaviour losses; the latter means negotiators no longer earn XP, since nothing in the new acceptance path calls `awardXP`.

- [x] **`RetirementDefectionTracker` — the turnover shares modifier is still missing (`main:500-532`).**
  The "Shares fix" commit restored the monthly payout, which is a different rule. Share-heavy contracts no longer suppress turnover.
  *Fix:* restore `-max(0, (sharesPercent / 10) - 2)`, or call it out as an intentional rules change.

### NPE risk

- [x] **`ChaosContractMarketDialog.java:182` — `campaign.getCurrentSystem().getName(currentDate)` is unguarded.**
  `Campaign.getCurrentSystem():700-703` returns null; `ChaosContractMarketAvailability` guards it in three places and this one was missed.
  *Fix:* guard it, and add `@Nullable` to `getCurrentSystem()` so any others surface.

### Logic

- [x] **`ChaosContractMarketAvailability.java:337` — "no contracts available" can print with offers sitting in the market.**
  `generatePityContracts` (line 317) adds Proving Ground offers but is excluded from `totalOffers`.
  *Fix:* count them in the empty test; they can stay out of the per-type breakdown.

- [x] **`PityContracts.java:87` — returns the rolled count, not the added count.** `generateProvingGroundOffer` can return null (guarded at line 82), while the javadoc at 67 promises adds.

- [x] **`Finances.java:598` — `payoutShares` dropped main's `isUseStratCon()` gate (`main:594-596`).** Monthly shares now pay in non-StratCon campaigns where they previously did not. Intended?

### Style

- [x] `Finances.java:580-596` — `payoutShares` was inserted between the `payTaxes` javadoc and `payTaxes`, so it carries two doc blocks and `payTaxes` (line 645) has none.
- [x] `ContractNegotiationDialog.java:616` — still `JOptionPane`; convention is `ImmersiveDialogSimple`.
- [x] `ChaosContractMarketAvailability.java:448, 454` — inline `@jakarta.annotation.Nullable`; add the import.
- [x] `ContractXmlCodec.java:529-590` — 50 binder lambdas use `(b, n, c, v)`; the reader table just above uses `(contract, node, campaign, version)` and reads far better.

### Checked and OK (no action)

- `ContractEditorDialog.saveAction` — the nullable `automaticToggle` boxes are only dereferenced inside `if (createMode)`, `planetAutomatic` is null-checked, and the derived-value ordering (scale before victory points, target before schedule, terms before finance) is right. No NPE on save.
- `PityContracts:103` is safe — `getCompletedContracts()` filters on `hasStatus`.
- `payOutSharesToPersonnel` early-returning after the debit is pre-existing (`main:640`), not from this PR.

HammerGS (approved):

Issue #9340, opened 2026-06-14
[RFE] Add search function to Glossary


heckindoc:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

As we continue to add entries to the glossary, we risk information becoming harder to find. Adding a search function or filter to the glossary will make it faster and easier to find the information that players need, and will further increase the value of adding information to the glossary. 

### MekHQ Suite Version *

0.51.01

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

PR #9781, opened 2026-08-16
feat: adds search to glossary for glossary and documentation entries

firefly2442:
feat: adds search to glossary for glossary and documentation entries, addresses #9340

This adds a search box to the glossary page.  You can search for one or more words.  If the items you search for don't appear in any results, it shows a message to the user detailing as such.  

This was AI generated by Claude Sonnet 5 and adjusted/tweaked.

The search only appears on the main glossary page, it does not appear when a user deep links and jumps straight to a page.  There may be some usability improvements to be made here such as adding search or consolidating the display and searching on a single dialog.  Something to consider.

Further improvements could include searching and highlighting the relevant text in the documents or PDFs themselves or even jumping to that section in the content.

Other improvements might also include semantic search to allow the user to search for similar words and return documents based on a score given the semantic similarity of the query.  This would require some work to chunk and turn the text content into vector embeddings for each document and then an embedding model to run on each search query.  This is probably overkill given the relatively small number of documentation materials so far.  However, this could be considered as it grows.

image

image

image

### DIFF
--- a/MekHQ/resources/mekhq/resources/NewGlossaryDialog.properties
+++ b/MekHQ/resources/mekhq/resources/NewGlossaryDialog.properties
@@ -37,3 +37,6 @@ GlossaryDialog.aboutPane=<h1 style="text-align:center;">MekHQ Glossary</h1>This 
 GlossaryDialog.button.close=Close
 GlossaryDialog.button.closeTab.single=Close Tab
 GlossaryDialog.button.closeTab.all=Close All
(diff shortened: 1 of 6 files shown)


IllianiBird:
Due to our policy on [AI Usage](https://github.com/MegaMek/megamek/wiki/Guidelines-for-Developer%E2%80%90Led-AI-Tool-Usage-in-MegaMek) we are unable to accept AI authored work from non-developers. I apologize for the inconvenience.

PR #9806, opened 2026-08-22
Fix #9804: Missions vanish from the briefing room after reloading a save

### LINKED ISSUE #9804: Missions cannot be found in the briefing room after reloading a save
### What went wrong

After saving and reloading the campaign i cannot find a mission i finished in the briefing room window, tried picking new contracts and after saving and reloading those new missions also cannot be found

### What you expected instead

Missions shouldn't get deleted

### Steps to reproduce

1.Accept any contract
2.Save the campaign
3.Load the campaign
4.Check the briefing room window

### Attach your files

[Patchwork Paladins28430620_Reach-part1.zip](https://github.com/user-attachments/files/31330367/Patchwork.Paladins28430620_Reach-part1.zip)

### Relevant log lines

```text

```

### MegaMek Suite Version

v0.51.01-nightly-2026-08-22

### Operating System

Windows 10

### Java Version

21.0.12

### Before you submit

- [ ] I have asked about this on the MegaMek Discord (optional)

HammerGS:
## What this changes

Contracts survive a save and reload again. Before this you could accept a contract, save, load, and find the Briefing Room empty.

Fixes #9804

There were two maps named `contractHistory`. The campaign stored contracts in one and saved from the other:

- `Campaign.contractHistory` held every contract and was never written to the save.
- `PlayerForce.contractHistory` was the map written to the save, but its `addContract()` had no callers, so it was always empty.

That broke both ends. The writer emitted no `<contracts>` block at all. The loader read `<contracts>` back into the PlayerForce map, which nothing gameplay-facing reads - so even a save that did contain contracts would still show an empty Briefing Room.

**The fix.** Contract history is global, so `Campaign.contractHistory` is now the only copy. `ContractHistoryData` owns its own save and load. The `PlayerForce` map, its accessors and both codec methods are gone, and every call site uses the campaign map. `PlayerForce` keeps the contract *market*, which really is force-level.

Loading goes through `Campaign.importMission()` rather than filling the map directly. That is deliberate. `importMission()` is the only thing that registers a contract's scenarios with the campaign and re-hooks StratCon's backing scenario pointers. A bare map put would restore contracts with no scenarios and dead StratCon tracks.

**A second fix, found in testing.** Once contracts actually loaded, deploying a force to a StratCon scenario threw a NullPointerException. `StratConCampaignState.contract` is an `@XmlTransient` back-pointer, so it is deliberately not serialized, and only `LegacyContractConverter` restored it - pre-migration saves only. `restoreTransientStratconInformation()` now re-points the campaign state at its own contract alongside the backing scenario pointers it already restored. It runs from `importMission()`, so both load paths are covered.

This was unreachable before rather than absent: with no contracts loading at all, you could never get to the StratCon deploy UI from a modern save.

**Existing saves.** Anyone who saved on an affected nightly has already lost that history. The contracts are not in the file, so this cannot bring them back. It stops further loss.

## Testing

`ContractHistoryDataTest`: 26/26 pass, seven of them new. They cover the save/load round trip, scenario registration, the StratCon back-pointer, full-replace-on-load, and the missing-status fallback.

Two were verified by reverting the fix:

- Point the save back at an always-empty second map and `aFullCampaignSaveCarriesTheContractHistory` fails. Nothing else does. That is the point of it: a codec-level round-trip test still passes with the bug present, because the bug was in the wiring, not the serializer. This test drives the real `Campaign.writeToXML`.
- Swap the loader's `importMission()` for a bare map put and `aRestoredContractsScenariosAreRegisteredWithTheCampaign` fails. Nothing else does.
- Drop the back-pointer restore and `restoredStratConStatePointsBackAtItsOwnContract` fails. Nothing else does.

Full suite: 14,536 tests, 0 failures. `checkstyleMain` and `javadoc` both run and clean, with no warnings in any file touched here.

Checked against the save attached to the issue: 0 `<contracts>` blocks and 0 `<contract>` elements, but 4 `contractMarket` references, so the file itself is intact.

## What is not proven yet

In-game testing has started and is not finished. It already caught the StratCon NPE fixed in the second commit; that specific path has not yet been re-tested in game since the fix. The remaining passes worth doing are accept, save, reload, check the Briefing Room, and deploy a force to a StratCon scenario.

No migration was attempted for saves that have already lost their history.

---
- [x] Tests added or updated, if this implements a rule or changes game state

### DIFF
--- a/MekHQ/src/mekhq/campaign/Campaign.java
+++ b/MekHQ/src/mekhq/campaign/Campaign.java
@@ -3671,7 +3671,7 @@ public class Campaign implements ITechManager {
         MHQXMLUtility.writeSimpleXMLTag(writer, indent, "colour", getPlayerForce().getColour().name());
         getPlayerForce().getUnitIcon().writeToXML(writer, indent);
         MHQXMLUtility.writeSimpleXMLTag(writer, indent, "lastFormationId", playerForce.getLastFormationId());
-        getPlayerForce().writeContractsToXML(writer, indent, this);
+        contractHistory.writeToXML(writer, indent, this);
         getPlayerForce().getContractMarket().writeToXML(writer, indent, this);
         MHQXMLUtility.writeSimpleXMLTag(writer, indent, "lastMissionId", lastMissionId);
         MHQXMLUtility.writeSimpleXMLTag(writer, indent, "lastScenarioId", lastScenarioId);
--- a/MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConContractInitializer.java
+++ b/MekHQ/src/mekhq/campaign/digitalGM/stratCon/StratConContractInitializer.java
@@ -1376,6 +1376,11 @@ public class StratConContractInitializer {
         // the campaign for this contract
         StratConCampaignState campaignState = mission.getStratConCampaignState();
         if (campaignState != null) {
+            // The campaign state's contract is an @XmlTransient back-pointer, so it comes back null from the save.
+            // Restore it before anything reads it: deploying a force to a scenario goes through
+            // StratConCampaignState.getContract(), which would otherwise hand a null contract to scenario generation.
+            campaignState.setContract(mission);
+
             for (StratConTrackState track : campaignState.getTracks()) {
                 for (StratConScenario scenario : track.getScenarios().values()) {
                     Scenario campaignScenario = campaign.getScenario(scenario.getBackingScenarioID());
--- a/MekHQ/src/mekhq/campaign/force/PlayerForce.java
+++ b/MekHQ/src/mekhq/campaign/force/PlayerForce.java
@@ -32,30 +32,17 @@
  */
 package mekhq.campaign.force;
 
-import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.UUID;
 
-import jakarta.annotation.Nullable;
-import megamek.Version;
-import mekhq.campaign.Campaign;
 import mekhq.campaign.campaignOptions.CampaignOptions;
 import mekhq.campaign.finances.Finances;
-import mekhq.campaign.mission.contract.AbstractContract;
 import mekhq.campaign.mission.contract.ContractMarket;
-import mekhq.campaign.mission.contract.contractData.MissionStatus;
-import mekhq.campaign.mission.contract.io.ContractXmlCodec;
 import mekhq.campaign.personnel.ranks.RankSystem;
 import mekhq.campaign.reputation.camOpsReputation.ForceReputationController;
 import mekhq.campaign.universe.Faction;
 import mekhq.campaign.universe.factionStanding.FactionStandings;
-import mekhq.utilities.MHQXMLUtility;
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
 
 /**
  * The human player's active force: the single {@link AbstractForce} a {@link mekhq.campaign.Campaign} is played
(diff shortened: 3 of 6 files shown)


IllianiBird (approved):

HammerGS:
In-game testing found a second bug, fixed in be12be583a. Writing it up here since the cause is not obvious from the diff.

## The crash

Deploying a force to a StratCon scenario:

```
12:06:16,226 ERROR [mekhq.MekHQ] {AWT-EventQueue-0}
java.lang.NullPointerException: Cannot invoke
"mekhq.campaign.mission.contract.AbstractContract.getEnemyFactionCode()" because "contract" is null
    at mekhq.campaign.mission.scenarios.AtBDynamicScenarioFactory.generateForce(AtBDynamicScenarioFactory.java:492)
    at mekhq.campaign.mission.scenarios.AtBDynamicScenarioFactory.generateForces(AtBDynamicScenarioFactory.java:367)
    at mekhq.campaign.mission.scenarios.AtBDynamicScenarioFactory.finalizeScenario(AtBDynamicScenarioFactory.java:283)
    at mekhq.campaign.digitalGM.stratCon.strategy.StratConOpForGenerationStrategy.generateOpFor(StratConOpForGenerationStrategy.java:52)
    at mekhq.campaign.digitalGM.stratCon.StratConRulesManager.assignForceToScenario(StratConRulesManager.java:1381)
    at mekhq.campaign.digitalGM.stratCon.strategy.StratConForceDeploymentStrategy.assignForceToScenario(StratConForceDeploymentStrategy.java:61)
    at mekhq.gui.stratCon.TrackForceAssignmentUI.actionPerformed(TrackForceAssignmentUI.java:173)
```

## What was wrong

StratCon keeps a shortcut back to its own contract, and that shortcut is deliberately not saved to disk:

```java
@XmlTransient
private AbstractContract contract;
```

Marking it `@XmlTransient` is correct - you do not want the whole contract serialized twice. But it means the pointer is `null` on the way back in, and something has to re-attach it on load. Only `LegacyContractConverter` did, at line 291, which covers pre-migration saves alone. Nothing restored it for a contract loaded through the modern path.

Tracing the stack backwards, the null starts at the top and is passed all the way down. `TrackForceAssignmentUI:173` reads `currentCampaignState.getContract()`, hands it to `assignForceToScenario`, which threads it through `generateOpFor` -> `finalizeScenario` -> `generateForces` -> `generateForce`, where scenario generation finally asks it who the enemy is.

Concretely: load a campaign, open StratCon, click a hex, deploy a lance. The UI asks the campaign state for the contract so generation knows which faction to draw an OpFor from, gets `null`, and dies on `contract.getEnemyFactionCode()`.

## Is this new?

The gap is pre-existing. It became reachable because of this PR.

Before this PR no contract loaded at all on the modern path, so there was no contract in the Briefing Room, no StratCon tab to open, and no way to reach the deploy button. The null was already there; nothing could touch it. That is why it belongs in this PR rather than a separate issue - the contract fix is incomplete without it.

## The fix

One line, in the method that already exists to restore StratCon's transient pointers:

```java
StratConCampaignState campaignState = mission.getStratConCampaignState();
if (campaignState != null) {
    campaignState.setContract(mission);   // added
    for (StratConTrackState track : campaignState.getTracks()) {
        // ...existing backing-scenario restore...
```

It sits next to the backing scenario restore it belongs with, and it runs from `Campaign.importMission()`, so the modern and legacy load paths are both covered. `LegacyContractConverter:291` is now redundant but harmless, so it is left alone rather than widening the diff.

## Verification

New test `restoredStratConStatePointsBackAtItsOwnContract`. Drop the one added line and it fails on its own; nothing else does.

`ContractHistoryDataTest` 26/26. Full suite 14,536 tests, 0 failures. `checkstyleMain` and `javadoc` both clean.

Still to do in game: re-run the same force deployment on the updated branch, and an accept -> save -> reload -> Briefing Room pass. Neither is proven yet, and "What is not proven yet" in the description says so.

PR #9809, opened 2026-08-22
Fix #9756: Resolve warehouse matching for Retractable Blades

### LINKED ISSUE #9756: [Issue] Retractable Blade in storage is not recognised by repair/salvage
### Brief Description *

While having retractable blades equipped, when trying to fix them, the repair/salvage tab allows you to order replacements as with any other part. However, when they are in storage they are not recognised and the mek cannot be fixed.

### 3. Steps to Reproduce

1. Have a mek with retractable blades equipped
2, Break the blade, in my case this was via punching with TacOps.
3. Back in the MekHQ, go to repair tab and observe the issue.

### Attach Files

[logs.zip](https://github.com/user-attachments/files/30908479/logs.zip)

[Hammerdown Company30260119_Trinket-part1.zip](https://github.com/user-attachments/files/30908484/Hammerdown.Company30260119_Trinket-part1.zip)

[customs.zip](https://github.com/user-attachments/files/30908527/customs.zip)

### Severity *

### MekHQ Suite Version *

51.00

### Operating System *

Windows 10

### Java Version *

Eclipse Adoptium 21

### Final Verification

MakerYuichi:
## Summary
Fixes #9756: Retractable Blades in storage are not recognized by the repair/salvage module.

## Root Cause
Retractable Blades are tracked as variable-tonnage assets under the physical weapon umbrella (`MiscType.F_CLUB`), dynamically scaling their weight based on the frame class of their parent chassis (1 ton per 20 tons of 'Mech tonnage). A damaged blade on an 80-ton Assault 'Mech (like the `Charger Pugilist CGR-PG` from the reporter's campaign save file) evaluates to an active mounted weight of 4.0 tons.

However, loose replacement blades bought on the market and stored in a warehouse default to an unmounted initialization base size of 1.0 ton because they have no parent unit context. Because `EquipmentPart.java:isSamePartType(Part part)` enforces a strict equality check (`getTonnage() == part.getTonnage()`), the 4.0-ton required profile fails to match the 1.0-ton loose warehouse item. This leaves the user interface reporting `0 available` in stock.

## Fix
Refactored `EquipmentPart.java:isSamePartType(Part part)` to bypass strict tonnage equality when checking `MiscTypeFlag.S_RETRACTABLE_BLADE` if either item is at its unmounted warehouse baseline size (`1.0` ton). This safely allows the repair panel to look up and bind generic warehouse stock to broken configurations, which are then smoothly scaled to their target installation weights during assembly.

## Testing
- Verified clean compilation across the multi-repository composite workspace with `./gradlew build -x test`.
- Loaded the issue save file and manually verified via GUI that the Repair Bay task screen correctly surfaces warehouse stock counts and cleanly processes installation actions without altering other physical weapon behaviors.
- Added a robust unit test variant utilizing a strongly-typed Mockito `spy()` inside the official test suite framework to prove compliance.

## What is not proven yet
While this has been verified to work perfectly for the standard Inner Sphere Retractable Blade on the provided 80-ton Assault 'Mech, it hasn't been extensively tested across more niche configurations, such as:
- Custom Quad 'Mechs or Support Vehicles modified to mount Retractable Blades.
- Unorthodox Mixed-Tech or Clan-base chassis modifications using this hardware.

However, because the code bypass is strictly gated behind the `S_RETRACTABLE_BLADE` flag, the blast radius is heavily isolated, meaning standard weapons (like PPCs or Lasers) face zero risk of regression.

- [ ] Javadoc literals use `{@code true}` / `{@code null}` rather than bare or quoted text
- [x] Dev team only: if AI tools were used, the AI Assisted Development label is applied, and I can explain and have verified the result


IllianiBird (changes requested):
I don't know that this is the right process. Take a look at `unitTonnageMatters` in Part.java

HammerGS:
I noticed you checked - Dev team only: if AI tools were used, the AI Assisted Development label is applied, and I can explain and have verified the result

Did you use AI tools in the coding?

MakerYuichi:
> I noticed you checked - Dev team only: if AI tools were used, the AI Assisted Development label is applied, and I can explain and have verified the result
> 
> Did you use AI tools in the coding?

To clarify, i used an AI assistant as an advanced search tool to help locate the specific file paths and track down where the tonnage variables were defined across the different repos. The actual logic implementation was manually done, compiled locally and verified by running the test suite and loading the reporter's campaign save file.

IllianiBird:
Before I can approve this I need to know why we can't just use unitTonnageMatters?

MakerYuichi:
For this, I actually spent quite a bit of time tracing down `unitTonnageMatters` during my initial exploration but ran into a hard structural limitation with how MekHQ tracks warehouse stock. 

When a part is mounted on a 'Mech, its parent unit tonnage tracks correctly (e.g., 80 tons for the Charger). However, when a spare replacement part sits loose in a warehouse storage bin, it doesn't belong to any chassis yet, so its parent unit tonnage profile defaults to `0`.

If we force `unitTonnageMatters = true` for Retractable Blades, the core engine's `isSamePartType` method falls straight down to evaluate this line:
`&& (!isUnitTonnageMatters() || getUnitTonnage() == part.getUnitTonnage())`

This forces the engine to evaluate `80 == 0` when comparing a broken 80-ton blade on a 'Mech against the unmounted warehouse stock box. Because `80 == 0` is false, the lookup fails instantly, and the Repair Bay layout drops right back to showing `0 available in stock` even if your inventory is full of them.

To get around this warehouse `0` tonnage trap while still protecting against the 1.0-ton light-frame wildcard edge case, I've pushed a refined commit that explicitly gates the 1.0-ton fallback check behind an unassigned unit validation loop. This gives us the exact safety tracking profile we need without hitting the hardcoded database wall!

IllianiBird:
I'll need to investigate this tomorrow. Unit tonnage matters was introduced precisely for this kind of case, and so I'd need to identify why it's working for other parts but not for this.

IllianiBird:
Investigating

HammerGS:
This is a Claude review, run by me (@HammerGS) against the current head of this branch. Everything below was checked by executing the real classes, not by reading alone. @IllianiBird, this answers your `unitTonnageMatters` question.

## Short answer: `unitTonnageMatters` is the right mechanism, and it is not blocked

The reason it works for other parts and not for this one is not in `Part.java` at all. It is in how `PartsStore` builds the two kinds of part.

A jump jet is stocked with one entry per **unit** tonnage (`PartsStore` line 198):

```java
for (int ton = 10; ton <= 100; ton += 5) {
    parts.add(new JumpJet(ton, equipmentType, -1, false, campaign));
}
```

`JumpJet` then sets `unitTonnageMatters = true` in its own constructor (line 60) and derives **both** its weight and its sticker price from `getUnitTonnage()`, with no entity needed. A loose jump jet in the warehouse is self-describing: it knows it is "the jump jet for a 55-ton 'Mech", so matching works with no special cases.

A retractable blade goes down the variable-tonnage branch instead (`PartsStore` line 264), which stocks one entry per **item** tonnage and hardcodes unit tonnage to zero (line 270):

```java
equipmentPart = new EquipmentPart(0, equipmentType, -1, 1.0, false, campaign);
equipmentPart.setEquipTonnage(ton);
```

That part still ends up with `unitTonnageMatters = true`, because `EquipmentPart`'s constructor sets it from `type.isVariableCost()` and the blade is `COST_VARIABLE`. So the part claims its unit tonnage identifies it while carrying no unit tonnage. That contradiction is the bug, and it is a stocking choice, not a structural limit.

## The `80 == 0` blocker is removable

I built the broken blade on the reporter's 80-ton Charger and ran the real matching methods against two candidate warehouse blades:

```
required (broken blade, 80t Mek):  4.5 tons, unitTonnage=80, unitTonnageMatters=true

A) warehouse stock as PartsStore builds it today (unitTonnage 0):
     isSamePartType          = false
     isAcceptableReplacement = false
B) identical part, stocked by unit tonnage (80):
     isSamePartType          = TRUE
     isAcceptableReplacement = false
```

Populating the unit tonnage clears the `80 == 0` comparison immediately. It is not a wall.

## But that only fixes the count, not the repair

Row B is the important one. Those are two different code paths:

- `EquipmentPart.isSamePartType` feeds the "N in stock" text.
- `MissingEquipmentPart.checkAttributesMatch` (line 248) decides whether the repair can happen at all. It compares **sticker price**, and only falls back to unit tonnage when the price is zero (line 268). Here it is 55,000 against 50,000, so the fallback never fires.

That 5,000 C-bill gap is the real defect, and it shows up most clearly as the same part object pricing differently depending on whether a unit is attached:

```
missing.getNewPart()   before setUnit(unit)  ->  55,000    (MekHQ's own formula)
                       after  setUnit(unit)  ->  50,000    (MegaMek's formula)
```

MekHQ keeps a private copy of the pricing rule for unit-less parts, in `EquipmentPart.resolveVariableCost` line 578:

```java
varCost = Money.of((1 + getTonnage()) * 10000);
```

The rule is 10,000 per ton *of blade*, plus a flat 10,000 for the retraction mechanism. MegaMek passes in blade-only tons, so for an 80-tonner that is `(1 + 4) * 10000 = 50,000` and the `+1` is the mechanism. MekHQ passes in the item's total weight, 4.5 tons, which already includes the half-ton mechanism, so the mechanism is billed twice.

It is off by exactly 5,000 at every tonnage, and the blade is the only club with that fixed half-ton, which is why every other physical weapon is fine:

```
Retractable Blade  unit=20t   tons=1.5   mounted=20,000   loose=25,000   MISMATCH
Retractable Blade  unit=55t   tons=3.5   mounted=40,000   loose=45,000   MISMATCH
Retractable Blade  unit=80t   tons=4.5   mounted=50,000   loose=55,000   MISMATCH
Retractable Blade  unit=100t  tons=5.5   mounted=60,000   loose=65,000   MISMATCH
Hatchet / Sword / Mace / Talons, every tonnage                           MATCH
```

## On the change as it stands

The premise in the PR description does not match the code. A blade on an 80-ton 'Mech weighs **4.5** tons, not 4.0 — half a ton of mechanism plus one ton per twenty tons of 'Mech. And loose blades are not pinned at 1.0 ton: `PartsStore` already stocks them from 1.0 to 5.5 in half-ton steps, so the 4.5-ton entry the Charger needs is already in the store and the weights already agree.

Because of that, the new condition `getTonnage() == 1.0 && getUnitTonnage() == 0` is never true for a real blade, so `isSamePartType` behaves exactly as it did before the change.

The test passes for two reasons that will not survive contact with real data:

- `doReturn(4.0).when(mountedBlade).getTonnage()` and `doReturn(1.0).when(warehouseBlade).getTonnage()` stub the exact values the new branch keys on, so the test asserts that the code does what it was told to do.
- `mock(EquipmentType.class)` makes `isVariableCost()` return Mockito's default `false`, which leaves `unitTonnageMatters` false and quietly disables the untouched final clause of `isSamePartType` — the clause that fails in production. Against a real `EquipmentType` this test would fail.

## What a correct fix looks like

Two halves, both needed:

1. Correct the mechanism double-count in `resolveVariableCost` so a loose blade prices the same as a mounted one. `(0.5 + getTonnage()) * 10000` gives 20,000 / 40,000 / 50,000 / 60,000, matching MegaMek exactly at every tonnage.
2. Give warehouse blades a real unit tonnage so `unitTonnageMatters` can do the job it was added for, following the `JumpJet` pattern rather than working around it.

The regression test should build on a real `EquipmentType` and assert that a loose 4.5-ton blade prices at 50,000 and is accepted as a replacement for a broken blade on an 80-ton 'Mech. That is the assertion that would have caught this.

## Separate finding, not for this PR

The same run turned up a second, pre-existing bug: loose **Lance** parts price at 0 C-bills at every tonnage, because MekHQ's lance branch in `resolveVariableCost` uses `getUnitTonnage() * 150` and store parts carry unit tonnage 0. Lance replacement is likely broken the same way. Worth its own issue.

IllianiBird:
Ha, funnily enough I literally just landed on the same conclusions as Claude, independently

IllianiBird:
After spending about an hour investigating the issue, and then having my findings echoed in a Claude review, I'm not able to approve this PR.

The PR, while working, only fixes a symptom and not the underlying bugged behavior. This creates a really fragile state where the issue is likely to crop up again. Especially as if any other physical weapon matches the same bug criteria it will fail in the same way.

If you'd like to take another bash at this, I'm happy to review, but as things currently stand this PR isn't something we can merge. :(

Issue #5837, opened 2025-01-20
[RFE] UX: Add message when trying to move/adjust deployed units in TO&E


firefly2442:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

It would be nice to display a message to the user that they cannot move or reorder units that are deployed in the TO&E.  I spent a little bit dragging-and-dropping just to realize the unit was deployed to an upcoming mission.  Would be a simple UX (user experience) helper.

### MekHQ Suite Version *

Nightly (ea016342ab9f24495d82167fb57dff5fec7208b9)

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

PR #9811, opened 2026-08-22
feat: adds a warning message to users moving deployed units in TO&E

### LINKED ISSUE #111: RFE: Add Entity History Tracking
I'd like to see the unit history for specific mechs, tanks, etc... track these details.

firefly2442:
feat: adds a warning message to users that they cannot move units or formations that are already deployed, addresses #5837

This helps with usability and assists users who may be unaware of what the highlighting means.

image

image

## What this changes

Addresses feature request #5837

## Testing

Ran unit tests and basic manual smoke-test.

## Questions/Considerations

* Should text strings be placed in a `.properties` file?  I leaned towards no since I didn't think it would be reused.
* Should a different dialog be used?  Elsewhere, I saw `showMessageDialog` used for warnings and such.

## What is not proven yet

---

No AI usage.

### DIFF
--- a/MekHQ/src/mekhq/gui/handler/TOETransferHandler.java
+++ b/MekHQ/src/mekhq/gui/handler/TOETransferHandler.java
@@ -32,6 +32,8 @@
  */
 package mekhq.gui.handler;
 
+import static mekhq.utilities.MHQInternationalization.getFormattedTextAt;
+
 import java.awt.datatransfer.DataFlavor;
 import java.awt.datatransfer.StringSelection;
 import java.awt.datatransfer.Transferable;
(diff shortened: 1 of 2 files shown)


MekHQ/src/mekhq/gui/handler/TOETransferHandler.java
```diff
@@ -83,15 +84,36 @@ public void exportDone(JComponent c, Transferable t, int action) {
...
 
+    /**
+     * A method that displays a dialog for units or formations that cannot be moved in the
+     * TO&E because they are already deployed.
+     *
+     */
+    private void showCannotMoveDialog(JComponent parent, String unit_formation_string) {
+        JOptionPane.showMessageDialog(
+            parent,
+            unit_formation_string + " cannot be moved because it is currently deployed to a mission.",
```
IllianiBird: Player facing hardcoded strings aren't something we allow. Though you may still see it in older code. Please take a look at getFormattedText and getFormattedTextAt

IllianiBird (changes requested):

MekHQ/src/mekhq/gui/handler/TOETransferHandler.java
firefly2442: Added new `.properties` resource and lookup via `getFormattedTextAt()`

IllianiBird (approved):

Issue #8820, opened 2026-03-21
[RFE] Improved Sensor quirk should act as Active Probe in Advanced Scouting


BlueThing00:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

<img width="472" height="52" alt="Image" src="https://github.com/user-attachments/assets/68e282fc-9c29-47bb-afce-8ed29ca6b7e4" />
The Improved Sensor quirk on a mech is supposed to act as an Active Probe.  It would be great if that functioned for that purpose when using Advanced Scouting in Stratcon.

<img width="1421" height="478" alt="Image" src="https://github.com/user-attachments/assets/f9b6be0e-90a6-4e60-8fe9-6778240059e6" />

### Use Cases or Rationale

1. This would make that quirk useful at the strategic layer, in line with how it is meant to work.

### MekHQ Suite Version *

0.50.11

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

IllianiBird:
I just checked the code, and this is how it already works. Closing :)

IllianiBird:
I also double-confirmed, the Glossary does say Improved Sensors count as Active Probe for the purposes of AdScout

PR #9830, opened 2026-08-26
Command Generator: build a starting force from the Force Generator ruleset

HammerGS:
**Depends on:** MegaMek/megamek#8820 and MegaMek/mm-data#534, in that order after mm-data. All three share the branch name `Implement-Force-Generator-in-MekHQ` and should merge together, this one last.

## Note for reviewers

This PR only compiles against MegaMek/megamek#8820. MekHQ's CI checks out a MegaMek branch **with the same name as this one** (`Implement-Force-Generator-in-MekHQ`) and builds against it; if that branch is not found it falls back to MegaMek `main` without saying so, and the build fails on missing API. So:

- Merge order is mm-data #534, then MegaMek #8820, then this one.
- Please do not delete the MegaMek branch until this PR has merged, or its CI turns red for a reason that has nothing to do with the code here.
- The `pr_ci` job also runs MegaMek's own test suite, so a MegaMek test failure shows up here too. The one that did (a filename-case bug in `FormationNamingConventionTest`, invisible on Windows) is fixed on the MegaMek branch.

## What this changes

Replaces the Company Generator with a Command Generator that builds a starting force from the Force Generator's ruleset rather than a flat list. Starting a merc campaign now produces a command organised the way its faction organises: named formations, a table of organisation you edit before committing, support personnel and transport sized to what the force holds, and senior appointments filled from the people generated.

The Command Designer holds Accept & Build to what you previewed: Generate captures the settings with the model, and if a tab has moved by the time you accept, you choose between building as previewed, regenerating, or cancelling. The eight temporary-crew toggles are offered on the Personnel & Officers tab and written to the campaign before the first unit is crewed, which is when the crew assembler reads them. Administrators are generated as one role at the CamOps headcount, following main's consolidation of the four specialities.

Depends on MegaMek/megamek#8820 and MegaMek/mm-data#534; the three branches share a name and must land together.

## Testing

- 35 new test classes across command generation, support personnel, rank assignment, senior appointments and temporary crew. `CommandGeneratorTemporaryCrewTest` proves the designer's choices reach the campaign and that every toggle maps to a role the crew assembler recognises.
- Full `:MekHQ:test` green, checkstyle and javadoc clean. `test` now depends on `stageDataFiles`.
- Playtested: a full generation with support staff filed into the TOE, the settings-changed prompt, the Command Designer's three tabs after their re-layout, and the Finances-tab race that used to throw mid-generation.

## What is not proven yet

- The HR-strain top-up now creates the consolidated `ADMINISTRATOR` role; its tests run with strain off, so whether the strain modifier reaches zero in a live campaign has only been checked by eye.
- The capability-vehicle crewing path (recovery vehicles and field hospitals crewed from generated staff) needs CamOps Salvage and MASH Theatres on; the playtest campaign had both off.
- `Phenotype.isTrueborn()` is deprecated for removal with no replacement named; three call sites here depend on the maintainers' answer.

---

- [ ] This PR is focused on one issue or RFE. Large refactoring or accessibility work is in its own PR

The first box is left unticked deliberately: 194 files replacing a subsystem. The logical chunks are the engine under `campaign/universe/commandGeneration`, the dialog and tabs under `gui/commandGeneration`, and the removal of the old Company Generator. Happy to split it if that would help review.


MekHQ/src/mekhq/campaign/universe/commandGeneration/CargoShipGenerator.java
```diff
@@ -0,0 +1,574 @@
...
+        FactionRecord factionRecord = (factionCode == null)
+              ? null
+              : RATGenerator.getInstance().getFaction(factionCode);
+        try {
+            return UnitTable.findTable(factionRecord, UnitType.JUMPSHIP, year, rating, null,
+                  ModelRecord.NETWORK_NONE, EnumSet.noneOf(EntityMovementMode.class),
+                  EnumSet.noneOf(MissionRole.class), 0);
+        } catch (Exception exception) {
+            LOGGER.warn(exception, "[CompanyGen][Cargo] could not build a JumpShip table for faction={}"
+                        + " year={} rating={}", factionCode, year, rating);
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CargoShipGenerator.java
```diff
@@ -0,0 +1,574 @@
...
+            LOGGER.debug("[CompanyGen][Cargo] no faction record for '{}'; drawing from the general table",
+                  factionCode);
+        }
+        try {
+            return UnitTable.findTable(factionRecord, UnitType.DROPSHIP, year, rating, null,
+                  ModelRecord.NETWORK_NONE, EnumSet.noneOf(EntityMovementMode.class),
+                  EnumSet.noneOf(MissionRole.class), 0);
+        } catch (Exception exception) {
+            LOGGER.warn(exception, "[CompanyGen][Cargo] could not build a DropShip table for faction={}"
+                        + " year={} rating={}", factionCode, year, rating);
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CargoShipGenerator.java
```diff
@@ -0,0 +1,574 @@
...
+    private static @Nullable Unit addCrewedShip(Campaign campaign, MekSummary hull) {
+        try {
+            PartQuality quality = campaign.getCampaignOptions().get(CampaignOption.USE_RANDOM_UNIT_QUALITIES)
+                  ? UnitOrder.getRandomUnitQuality(0)
+                  : PartQuality.QUALITY_D;
+            // allowNewPilots = true: the ship arrives crewed, like every other generated unit.
+            return campaign.addNewUnit(hull.loadEntity(), true, 0, quality);
+        } catch (Exception exception) {
+            LOGGER.error(exception, "[CompanyGen][Cargo] could not load cargo DropShip '{}' from {}",
+                  hull.getName(), hull.getSourceFile());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
```diff
@@ -0,0 +1,533 @@
...
+                for (int crewIndex = 0; crewIndex < crewSize; crewIndex++) {
+                    unit.addPilotOrSoldier(pool.get(startIndex + consumed + crewIndex));
+                }
+                consumed += crewSize;
+                campaign.getPlayerForce().addUnitToFormation(unit, parent.getId(), campaign);
+                LOGGER.info("[CompanyGen][SupportTOE]     {} vehicle '{}' unitId={} crewed with {}/{} staff",
+                      parent.getName(), vehicle.unitName(), unit.getId(), crewSize, crewNeeded);
+            } catch (Exception exception) {
+                LOGGER.error(exception, "Unable to load capability vehicle {}: {}", vehicle.unitName(),
+                      mekSummary.getSourceFile());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
```diff
@@ -0,0 +1,533 @@
...
+            Unit unit = campaign.addNewUnit(mekSummary.loadEntity(), false, 0);
+            unit.setFluffName(spec.professionLabel());
+            for (Person person : spec.crew()) {
+                ensureInfantrySkill(person);
+                unit.addPilotOrSoldier(person);
+            }
+            return unit;
+        } catch (Exception exception) {
+            LOGGER.error(exception, "Unable to load carrier unit {}: {}", spec.unitName(),
+                  mekSummary.getSourceFile());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportUnitGenerator.java
```diff
@@ -0,0 +1,387 @@
...
+            int matchingItems = 0;
+            for (MiscMounted item : entity.getMisc()) {
+                if (item.getType().hasFlag(flag)) {
+                    matchingItems++;
+                }
+            }
+            return matchingItems;
+        } catch (Exception exception) {
+            LOGGER.error(exception, "Unable to load entity {} to count equipment: {}", unitName,
+                  mekSummary.getSourceFile());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportUnitGenerator.java
```diff
@@ -0,0 +1,387 @@
...
+                PartQuality quality = useRandomQuality ? UnitOrder.getRandomUnitQuality(0) : PartQuality.QUALITY_D;
+                Unit unit = campaign.addNewUnit(mekSummary.loadEntity(), true, 0, quality);
+                if (unit != null) {
+                    if (autoAssignRanks) {
+                        AutomaticRankAssigner.assignRanks(campaign, unit, faction);
+                    }
+                    units.add(unit);
+                }
+            } catch (Exception exception) {
+                LOGGER.error(exception, "Unable to load entity {}: {}", unitName, mekSummary.getSourceFile());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1222 @@
...
+          @Nullable Ruleset.ProgressListener listener) {
+        if (listener != null) {
+            listener.updateProgress(0.0, "Preparing generation parameters...");
+        }
+        LOGGER.info("[CompanyGen][Pipeline]snapshot: faction={} year={} echelon={} unitType={} rating={} experience={} weightClass={} augmented={} sizeMod={} dropshipPct={} jumpshipPct={} cargoPct={} flags={} roles={}",
+              snap.getFaction(), snap.getYear(), snap.getEchelon(), snap.getUnitType(),
+              snap.getRating(), snap.getExperience(), snap.getWeightClass(),
+              snap.isAugmented(), snap.getSizeMod(),
+              snap.getDropshipPct(), snap.getJumpshipPct(), snap.getCargoPct(),
+              snap.getFlags(), snap.getRoles());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationIconBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+                // formations already render, distinct from Mek formations' chassis-plus-letter
+                // pair. Adding StratOps-style tank/boat silhouettes to mm-data would let us
+                // upgrade these to a paired icon later without changing this code.
+                case UnitType.TANK, UnitType.NAVAL -> appendSingle(icon, stratOps, weightFilename);
+                default -> appendSingle(icon, stratOps,
+                      MHQConstants.LAYERED_FORCE_ICON_BATTLEMEK_CENTER_FILENAME);
+            }
+        } catch (Exception exception) {
+            LOGGER.error(exception, "[CompanyGen][Icons] cannot resolve TYPE for '{}' (dominantType={}); falling back to BattleMek (Center)",
+                  formation.getName(), dominantType);
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
```diff
@@ -0,0 +1,585 @@
...
+    private @Nullable Tier resolveTier(FormationRequest request) {
+        if (request.echelon() == null) {
+            return null;
+        }
+        try {
+            return tierResolver.resolve(request.faction(), request.echelon());
+        } catch (Exception resolutionFailure) {
+            LOGGER.warn(resolutionFailure, "[CommandGen][Naming] could not resolve a naming rule for"
+                        + " faction {} echelon {}; keeping the ruleset's own names",
+                  request.faction(), request.echelon());
```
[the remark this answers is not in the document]
HammerGS: Declining - false positive. MMLogger takes SLF4J-style '{}' placeholders, not printf specifiers, so a scan for '%' finds no format arguments. The placeholder count matches the argument count on this line; I checked each of the ten reports individually. The overload with a leading Throwable is error(Throwable, String, Object...), where the exception is the first argument and the message the second.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/MultiCrewAssembler.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+        return person;
+    }
+
+    /**
+     * Vessel crew slots only apply to large craft. SmallCraft, DropShips, JumpShips, WarShips, and
+     * SpaceStations all need them; everything else is driver+gunner (or solo pilot).
+     */
+    private static boolean needsVesselCrew(Entity entity) {
+        if (entity instanceof SmallCraft || entity instanceof Jumpship || entity instanceof SpaceStation) {
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 03e3854154. SpaceStation extends Jumpship, so the third check could never add anything. It is gone, along with the now-unused import.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1222 @@
...
+
+            // Use the canonical Campaign API instead of parent.addUnit(uuid). The bare addUnit(uuid)
+            // only updates the Formation's unit list; the Unit's formationId back-reference stays at
+            // FORMATION_NONE, so UnitTableModel's Formation column and any caller of
+            // Campaign.getFormation(unit.getFormationId()) sees nothing. addUnitToFormation sets
+            // unit.setFormationId(id), pushes the assignment through AssignmentLogger, and fires
+            // OrganizationChangedEvent — its subscribers (BriefingTab, TOETab) only do
+            // ActionScheduler.schedule() which wraps Timer.restart() and is thread-safe.
+            LOGGER.info("[CompanyGen][Leaf] BEFORE addUnitToFormation parent={} unit={}", parentInfo, unit.getId());
+            campaign.getPlayerForce().addUnitToFormation(unit, parent.getId(), campaign);
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 03e3854154. The walker hands a null parent to the leaf handler when the force is a single unit with nothing above it; the handler logged that case as 'null' and then dereferenced it anyway. Such a leaf is now filed under the root formation with a warning rather than left to throw.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1222 @@
...
+
+        Money minimumStartingFloat = Money.of(options.getMinimumStartingFloat());
+        Money loan = Money.zero();
+
+        if (options.isPayForSetup()) {
+            Money costs = Money.zero();
+            if (options.isPayForPersonnel()) {
+                Money hiringCosts = Money.zero();
+                for (Person person : generatedPersons) {
+                    hiringCosts = hiringCosts.plus(person.getSalary(campaign).multipliedBy(2));
```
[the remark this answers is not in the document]
HammerGS: Fixed in 03e3854154. getSalary(Campaign) is deprecated in favour of getSalary(CampaignOptions, boolean, LocalDate); the three values are read once outside the loop and passed in.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SeniorAppointmentAssigner.java
```diff
@@ -0,0 +1,561 @@
...
+     */
+    private static @Nullable Person bestCandidate(Campaign campaign, List<Person> pool,
+          Set<String> disciplineSkills) {
+        Person bestCandidate = null;
+        int bestScore = Integer.MIN_VALUE;
+        for (Person candidate : pool) {
+            int score = scoreFor(candidate, disciplineSkills);
+            boolean outscoresBest = (bestCandidate == null) || (score > bestScore);
+            boolean tiedButOutranksBest = (bestCandidate != null) && (score == bestScore)
+                  && candidate.outRanksUsingSkillTiebreaker(campaign, bestCandidate);
```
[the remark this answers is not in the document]
HammerGS: Fixed in 03e3854154. Moved to outRanksUsingSkillTiebreaker(CampaignOptions, boolean, LocalDate, Person), which the deprecation names as the replacement. The unit test that stubbed the old overload now stubs this one.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
```diff
@@ -0,0 +1,460 @@
...
+              administrators, astechs, medics, generated);
+    }
+
+    /**
+     * Generates {@code ceil(baselineDemand × coverage / 100)} Persons of the given role, applying
+     * the user-selected skill level and faction-appropriate rank to each. Returns the actual count
+     * created (zero if either baseline or coverage is non-positive).
+     */
+    private static int generateRole(Campaign campaign, CommandGenerationOptions options,
+          AbstractSkillGenerator skillGen, PersonnelRole role, int baselineDemand, int supportRank,
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 03e3854154. generateRole derives each person's rank from the skill level they rolled (RulesetRankAssigner.supportRankFor), so the supportRank passed in was never read. Removed from generateRole and from the astech/medic helpers that only forwarded it; it stays on the HR-strain top-up, which does use it.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
```diff
@@ -0,0 +1,460 @@
...
+              options.isMedicsAsPersonnel(),
+              options.getMedicSkillLevel(), supportRank, faction, targetRankSystem, rankValidator,
+              out, AssistantPool.MEDIC);
+    }
+
+    private enum AssistantPool { ASTECH, MEDIC }
+
+    private static int applyAssistant(Campaign campaign, AbstractSkillGenerator skillGen,
+          PersonnelRole role, int needed, boolean asPersonnel, SkillLevel skillLevel,
+          int supportRank, Faction faction, RankSystem targetRankSystem, RankValidator rankValidator,
```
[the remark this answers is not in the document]
HammerGS: Same fix as the generateRole report, 03e3854154: applyAssistant never read supportRank, so it and the two helpers that forwarded it no longer take it.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
```diff
@@ -0,0 +1,782 @@
...
+            forceDescriptorSnapshot = new ForceDescriptorSnapshot();
+        }
+        return forceDescriptorSnapshot;
+    }
+
+    public void setForceDescriptorSnapshot(final ForceDescriptorSnapshot forceDescriptorSnapshot) {
+        this.forceDescriptorSnapshot = forceDescriptorSnapshot;
+    }
+
+    public Map<PersonnelRole, Integer> getSupportPersonnelCoveragePercents() {
```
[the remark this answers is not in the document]
HammerGS: Declining. These maps are the options object's editable state: the setup tab and the tests write individual entries through the getter (getSupportPersonnelCoveragePercents().put(role, percent)) by design, and the sibling collections on this class follow the same pattern. Returning a copy would silently break those writes; wrapping as unmodifiable would move every caller to whole-map replacement for no gain. There is no invariant on the map contents that a caller could violate.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
```diff
@@ -0,0 +1,782 @@
...
+        return supportPersonnelCoveragePercents;
+    }
+
+    public void setSupportPersonnelCoveragePercents(final Map<PersonnelRole, Integer> supportPersonnelCoveragePercents) {
+        this.supportPersonnelCoveragePercents = supportPersonnelCoveragePercents;
+    }
+
+    public Map<PersonnelRole, SkillLevel> getSupportPersonnelSkillLevels() {
```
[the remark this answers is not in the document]
HammerGS: Declining. These maps are the options object's editable state: the setup tab and the tests write individual entries through the getter (getSupportPersonnelCoveragePercents().put(role, percent)) by design, and the sibling collections on this class follow the same pattern. Returning a copy would silently break those writes; wrapping as unmodifiable would move every caller to whole-map replacement for no gain. There is no invariant on the map contents that a caller could violate.

MekHQ/src/mekhq/gui/CommandCenterTab.java
```diff
@@ -1180,6 +1215,13 @@
         basicInfoScheduler.schedule();
     }
 
+    @Subscribe
+    public void handle(OrganizationChangedEvent evt) {
```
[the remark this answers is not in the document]
HammerGS: Declining. The parameter is required: the event bus selects this handler by the declared parameter type, so removing it would unsubscribe the tab from OrganizationChangedEvent. This is the same shape as every other @Subscribe handler in MekHQ.

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
```diff
@@ -86,6 +88,15 @@
                     countInjured++;
                 }
                 salary = salary.plus(person.getSalary(getCampaign()));
+                // Counted among the active only, so the figure describes the warriors currently
+                // serving rather than everyone the command has ever had on its books.
+                if (person.getPhenotype().isTrueborn()) {
```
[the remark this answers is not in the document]
HammerGS: Declining for now, and asking for guidance. isTrueborn() was deprecated for removal in the J21 language sweep without naming a replacement. The only accessor that returns the same value is isExternal() - every phenotype constant sets the two flags identically - but that name says nothing about trueborn descent to a reader, so swapping it in would trade a deprecation warning for code that misdescribes what it checks. If the intended replacement is isExternal() or !isNone() I will switch all three call sites; happy to be told which.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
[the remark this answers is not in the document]
HammerGS: Fixed in 03e3854154 (the earlier copy of this finding): a leaf with no parent formation is filed under the root formation with a warning instead of dereferencing null.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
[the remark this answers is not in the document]
HammerGS: Declining, as on the earlier copy: these maps are the options object's editable state and the setup tab and tests write entries through the getter by design. A copy would silently break those writes.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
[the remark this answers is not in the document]
HammerGS: Declining, as on the earlier copy: these maps are the options object's editable state and the setup tab and tests write entries through the getter by design. A copy would silently break those writes.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorWalker.java
```diff
@@ -0,0 +1,641 @@
...
+     * @return the number of leaves visited
+     */
+    private static int wrapLoosePlatoonsByUnitType(List<ForceDescriptor> platoons, String factionCode,
+          FormationNamer namer, FormationSink sink, @Nullable Object rootHandle) {
+        if (platoons.isEmpty()) {
+            return 0;
+        }
+        Map<Integer, List<ForceDescriptor>> byUnitType = new LinkedHashMap<>();
+        for (ForceDescriptor platoon : platoons) {
+            byUnitType.computeIfAbsent(platoon.getUnitType(), key -> new ArrayList<>()).add(platoon);
```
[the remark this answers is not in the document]
HammerGS: Declining. The parameter is the key Map.computeIfAbsent hands to its mapping function; the function's signature requires it whether or not the body reads it, and 'key' is the conventional name for it.

MekHQ/src/mekhq/gui/CommandCenterTab.java
```diff
@@ -1180,6 +1215,13 @@ public void handle(UnitEvent evt) {
...
```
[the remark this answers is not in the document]
HammerGS: Declining, as on the earlier copy: the event bus selects this @Subscribe handler by its parameter type, so the parameter has to be declared.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
```diff
@@ -0,0 +1,459 @@
...
```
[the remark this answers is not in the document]
HammerGS: Fixed in 03e3854154 (the earlier copy of this finding): supportRank is gone from generateRole and the assistant helpers, which derive each person's rank from their rolled skill level.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
```diff
@@ -0,0 +1,459 @@
...
+              administrators, astechs, medics, generated);
+    }
+
+    /**
+     * Generates {@code ceil(baselineDemand × coverage / 100)} Persons of the given role, applying
+     * the user-selected skill level and faction-appropriate rank to each. Returns the actual count
+     * created (zero if either baseline or coverage is non-positive).
+     */
+    private static int generateRole(Campaign campaign, CommandGenerationOptions options,
+          AbstractSkillGenerator skillGen, PersonnelRole role, int baselineDemand, int supportRank,
```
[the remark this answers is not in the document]
HammerGS: Same fix, 03e3854154.

MekHQ/src/mekhq/gui/CampaignGUI.java
```diff
@@ -126,8 +126,10 @@
 import mekhq.gui.baseComponents.roundedComponents.RoundedJButton;
 import mekhq.gui.baseComponents.roundedComponents.RoundedLineBorder;
 import mekhq.gui.baseComponents.roundedComponents.RoundedMMToggleButton;
+import mekhq.gui.campaignOptions.CampaignOptionsDialog;
+import mekhq.gui.commandGeneration.CommandGenerationDialog;
 import mekhq.gui.dialog.*;
-import mekhq.gui.dialog.CompanyGenerationDialog;
+import mekhq.gui.commandGeneration.CommandGenerationDialog;
 import mekhq.gui.dialog.glossary.GlossaryDialog;
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 146f9712b6. The line was duplicated by the merge from main; Java tolerates an identical single-type import twice, which is why it still compiled, but it was noise. Removed.

MekHQ/src/mekhq/MHQConstants.java
```diff
@@ -74,8 +74,7 @@ public final class MHQConstants extends SuiteConstants {
     public static final String HISTORICAL_DAILY_LOG = "historicalDailyLog";
     public static final int MAX_HISTORICAL_LOG_DAYS = 120; // max number of days that will be stored in the history,
     // also used as a limit in the UI
-    public static final String COMPANY_GENERATOR_STARTUP = "companyGeneratorStartup";
-    public static final String SHOW_COMPANY_GENERATOR = "showCompanyGenerator";
+    public static final String SHOW_COMMAND_GENERATOR = "showCompanyGenerator";
```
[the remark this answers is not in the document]
HammerGS: Fair, done in 146f9712b6. A comment now says the key is kept so an existing preferences file still applies; only the constant was renamed.

MekHQ/src/mekhq/gui/commandGeneration/components/CommandGenerationUtilities.java
```diff
@@ -0,0 +1,65 @@
...
+ * affiliated with Microsoft.
+ */
+package mekhq.gui.commandGeneration.components;
+
+/**
+ * Shared helpers for the Company Generation dialog's styled components.
+ *
+ * <p>This is the Company Generation counterpart to {@code CampaignOptionsUtilities}. The two packages
+ * share their visual conventions but each reads from its own resource bundle, so a key like
+ * {@code "lblSparesArmor.text"} lives next to the dialog that uses it instead of polluting the larger
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 146f9712b6. The components' javadoc now says Command Generation throughout.

MekHQ/src/mekhq/gui/commandGeneration/components/CommandGenerationStandardPanel.java
```diff
@@ -0,0 +1,92 @@
...
+import static mekhq.gui.commandGeneration.components.CommandGenerationUtilities.getCommandGenerationResourceBundle;
+import static mekhq.utilities.MHQInternationalization.getTextAt;
+
+import javax.swing.JPanel;
+
+import mekhq.gui.baseComponents.roundedComponents.RoundedLineBorder;
+
+/**
+ * A specialized {@link JPanel} used to group related controls inside Company Generation tabs.
+ *
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 146f9712b6. The components' javadoc now says Command Generation throughout.

MekHQ/src/mekhq/gui/commandGeneration/components/CommandGenerationLabel.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * <p>Text and tooltip come from {@code mekhq.resources.CommandGenerationDialog}, indexed by
+ * {@code "lbl" + name + ".text"} and {@code "lbl" + name + ".tooltip"}. Pass {@code noTooltip = true}
+ * to omit tooltip lookup entirely (useful for value-display labels whose text is set programmatically).</p>
+ */
+public class CommandGenerationLabel extends JLabel {
+
+    /**
+     * Constructs a label with both text and tooltip pulled from the Company Generation bundle.
+     *
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 146f9712b6. The components' javadoc now says Command Generation throughout.

MekHQ/src/mekhq/gui/commandGeneration/components/CommandGenerationCheckBox.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * <p>Text and tooltip come from the {@code mekhq.resources.CommandGenerationDialog} bundle, indexed
+ * by {@code "lbl" + name + ".text"} and {@code "lbl" + name + ".tooltip"}. The wrap-size and font-scale
+ * helpers are reused from the Campaign Options package — only the bundle is package-local.</p>
+ */
+public class CommandGenerationCheckBox extends JCheckBox {
+
+    /**
+     * Constructs a checkbox whose text and tooltip are read from the Company Generation bundle.
+     *
```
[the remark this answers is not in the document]
HammerGS: Correct, fixed in 146f9712b6. The components' javadoc now says Command Generation throughout.

MekHQ/src/mekhq/campaign/report/PersonnelReport.java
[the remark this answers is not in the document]
HammerGS: Resolved upstream: #9840 corrected isTrueborn() to return the isTrueborn field and removed the deprecation. That change is merged into this branch, so the three call sites stand as they are.

MekHQ/src/mekhq/campaign/market/PartsInUseManager.java
```diff
@@ -284,8 +284,11 @@ private void updatePartInUseData(PartInUse partInUse, Part incomingPart, boolean
                 return;
             }
 
-            // Ignore parts if they are from mothballed units and the flag is set
-            if (ignoreMothballedUnits && incomingPart.getUnit() != null && incomingPart.getUnit().isMothballed()) {
+            // Ignore parts if they are from mothballed units and the flag is set.
+            // Uses the unit captured above rather than re-reading getUnit(): the warehouse can be
+            // refreshed on the EDT while a generation worker is still attaching and detaching parts, and
+            // re-fetching left a window where the part was detached between the null check and the call.
+            if (ignoreMothballedUnits && unit.isMothballed()) {
```
IllianiBird, line 291: If this isn't covered by tests it needs to be. I'm very wary of us introducing an NPE here, as unit can and will == null

MekHQ/src/mekhq/campaign/personnel/enums/GeneticLegacyRole.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final MMLogger LOGGER = MMLogger.create(GeneticLegacyRole.class);
+
+    private static final String RESOURCE_BUNDLE = "mekhq.resources.GeneticLegacyRole";
+
+    /**
+     * @return the localized name of this role, for display
+     */
+    public String getLabel() {
+        return getTextAt(RESOURCE_BUNDLE, name() + ".label");
+    }
```
IllianiBird, line 70: This is the wrong approach. In this way we end up needing to refetch and reprocess the String every time it's called. This ends up being far heavier than most people realize. Instead, take a look at how `mekhq/campaign/mission/contract/contractGeneration/ChaosEmployerType.java` handles this same problem

MekHQ/src/mekhq/campaign/personnel/Bloodname.java
```diff
@@ -172,60 +191,29 @@ private int phenotypeMultiplier(Phenotype warriorType, int year) {
...
-                } else if (wn.getNodeName().equalsIgnoreCase("created")) {
-                    retVal.startDate = Integer.parseInt(wn.getTextContent().trim()) + 20;
-                }
-            } catch (Exception e) {
-                LOGGER.error("", e);
-            }
-        }
+    public boolean isInactive(int year) {
+        return (year < startDate) ||
+                     ((inactive > 0) && (inactive < year) && !((reactivated > 0) && (reactivated <= year)));
```
IllianiBird, line 196: This conditional is confusing. Please break it up into separate booleans and then pass those in.

MekHQ/src/mekhq/campaign/personnel/Bloodname.java
```diff
@@ -240,6 +228,30 @@ public static Bloodname loadFromXml(Node node) {
...
+     * wants the house behind it - its founder, the Clan it originates with, whether it is exclusive -
+     * has to look the record back up.</p>
+     *
+     * @param name the Bloodname to find, matched without regard to case
+     *
+     * @return the matching Bloodname, or {@code null} if the name is blank, the data has not loaded,
+     *       or no such Bloodname exists
+     */
+    public static @Nullable Bloodname getBloodname(final @Nullable String name) {
+        if ((bloodnames == null) || (name == null) || name.isBlank()) {
```
IllianiBird, line 244: Inconsistent bracketing of clauses. Either all should be wrapped in brackets or none should

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1262 @@
...
+        if (listener != null) {
+            listener.updateProgress(0.0, "Loading factions and rulesets...");
+        }
+        RulesetEngineBootstrap.ensureLoaded(snap.getYear());
+
+        // 2. Build a fresh ForceDescriptor from the snapshot. The Force Generator panel does this
+        // server-side via buildForceDescriptor(); we mirror its inputs here so we never depend on the
+        // panel being instantiated.
+        LOGGER.info("[CompanyGen][Pipeline]Stage 2: build root ForceDescriptor from snapshot");
+        ForceDescriptor fd = new ForceDescriptor();
```
IllianiBird, line 238: Let's use a descriptive variable name here, not abbreviated. There's quite a few other instances of this in this class.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1262 @@
...
+        LOGGER.info("[CompanyGen][Pipeline]  built fd: faction={} year={} echelon={} unitType={} rating={} weightClass={}",
+              fd.getFaction(), fd.getYear(), fd.getEchelon(), fd.getUnitType(),
+              fd.getRating(), fd.getWeightClass());
+
+        // 3. Run the engine. Null listener is safe per Ruleset.processRoot's internal guards.
+        LOGGER.info("[CompanyGen][Pipeline]Stage 3: Ruleset.processRoot()");
+        if (listener != null) {
+            listener.updateProgress(0.0, "Building force structure...");
+        }
+        long t0 = System.currentTimeMillis();
```
IllianiBird, line 271: `t0` is an ambiguous variable name

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
```diff
@@ -0,0 +1,1262 @@
...
+                newUnitValue = newUnitValue.plus(unit.getBuyCost());
+                pricedUnits++;
+            }
+        }
+
+        // Base cash: percentage of unit value, or the dice roll when randomized.
+        int percent = options.getStartingCashPercent();
+        Money startingCash;
+        if (options.isRandomizeStartingCash()) {
+            startingCash = Money.of(1_000_000)
```
IllianiBird, line 727: This should probably be a const

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorSnapshot.java
```diff
@@ -0,0 +1,395 @@
...
+ * Phase 1 carries the minimum the engine needs (faction, year, echelon, unit type, rating); subsequent
+ * phases extend the snapshot to cover roles, weight class, flags, augmented / size modifier, transport
+ * percentages. Default-on-missing semantics keep old presets parsing cleanly.</p>
+ */
+public final class ForceDescriptorSnapshot {
+
+    private String faction;
+    private int year;
+    /** ratgen echelon int from {@code data/forcegenerator/faction_rules/constants.txt}. */
+    private Integer echelon;
```
IllianiBird, line 62: Unless we're expecting these to return `null` we can probably get away with primitives, such as int

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorSnapshot.java
```diff
@@ -0,0 +1,395 @@
...
+            MHQXMLUtility.writeSimpleXMLTag(pw, indent + 1, "cargoPct", cargoPct);
+        }
+        MHQXMLUtility.writeSimpleXMLCloseTag(pw, indent, "forceDescriptorSnapshot");
+    }
+
+    /**
+     * Parses a {@code <forceDescriptorSnapshot>} element. Missing children leave defaults intact, so this
+     * is safe to call on old presets that lack the block.
+     */
+    public static ForceDescriptorSnapshot parseFromXML(Node element) {
```
IllianiBird, line 294: This is an older style of loader, we've since moved to a String-lookup, as it's faster. Take a look at `mekhq/campaign/mission/contract/io` for inspiration

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorWalker.java
```diff
@@ -0,0 +1,641 @@
...
+
+        List<FormationRequest> requests = new ArrayList<>(namedChildren.size());
+        for (ForceDescriptor child : namedChildren) {
+            requests.add(requestFor(child));
+        }
+        List<NamedFormation> names = namer.nameSiblings(requests, parentDesignator);
+
+        int leaves = 0;
+        int nameIndex = 0;
+        for (ForceDescriptor child : children) {
```
IllianiBird, line 323: We're looping through children three times, is it possible to merge these into a single pass?

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorWalker.java
```diff
@@ -0,0 +1,641 @@
...
+     *
+     * @param echelon     the {@code ForceDescriptor.echelon} value (boxed, may be null)
+     * <p>Package-private so {@link RulesetRankAssigner} can resolve the requested echelon the same
+     * way, rather than reading a Formation's own level back after MekHQ has recomputed it.</p>
+     *
+     * @param echelon     the {@code ForceDescriptor.echelon} value (boxed, may be null)
+     * @param factionCode the faction code on the descriptor (e.g. "FS", "CW", "WOB"); may be {@code null}
+     * @return the matching {@link FormationLevel}, or {@code null} if the echelon doesn't map
+     */
+    static FormationLevel mapEchelonToFormationLevel(Integer echelon, String factionCode) {
```
IllianiBird, line 564: Needs `@Nullable` tag

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationIconBuilder.java
```diff
@@ -0,0 +1,478 @@
...
+     *   <li>Clan: TEAM (Point) / STAR_OR_NOVA / BINARY_OR_TRINARY / CLUSTER / GALAXY map directly.
+     *       TOUMAN falls back to the GALAXY icon (no Touman shape ships in mm-data).</li>
+     *   <li>ComStar / WoB: TEAM (Level I) / LEVEL_II_OR_CHOIR through LEVEL_VI all map directly;
+     *       all five levels ship in mm-data.</li>
+     * </ul>
+     *
+     * <p>Returns {@code null} only for non-canonical levels (NONE, INVALID, REMOVE_OVERRIDE) so the
+     * caller can skip those nodes cleanly.</p>
+     */
+    private static String formationFilenameFor(FormationLevel level, Faction faction) {
```
IllianiBird, line 421: needs `@Nullable`. There are loads of instances of this in this class, and others. Will probably need a full sweep to ensure all null returning methods are marked accordingly

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
```diff
@@ -0,0 +1,585 @@
...
+ * cannot be qualified, and a name reaching a player with one indicates a ruleset data gap.</p>
+ */
+public class FormationNamer {
+
+    private static final MMLogger LOGGER = MMLogger.create(FormationNamer.class);
+
+    /** Spelled ordinals matching the ruleset engine's {@code {ordinal}} token values. */
+    private static final List<String> SPELLED_ORDINALS = List.of("First", "Second", "Third", "Fourth",
+          "Fifth", "Sixth", "Seventh", "Eighth", "Ninth", "Tenth");
+
```
IllianiBird, line 93: These shouldn't be hardcoded and should be in a resource bundle

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
```diff
@@ -0,0 +1,585 @@
...
+    private static final String[] ROMAN_NUMERALS = { "I", "II", "III", "IV", "V", "VI", "VII", "VIII",
```
IllianiBird, line 94: Arguably, these should too

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
```diff
@@ -0,0 +1,585 @@
...
+        int modHundred = number % 100;
+        String suffix;
+        if (modHundred >= 11 && modHundred <= 13) {
+            suffix = "th";
+        } else {
+            suffix = switch (number % 10) {
+                case 1 -> "st";
+                case 2 -> "nd";
+                case 3 -> "rd";
+                default -> "th";
```
IllianiBird, line 580: Annoying as it is, these should be in a resource bundle.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/PersonnelRoleResolver.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Returns the primary crew role for a unit, with an optional {@link Entity} to refine the lookup
+     * (e.g. distinguishing LAMs from ordinary Meks).
+     *
+     * @param unitType the {@link UnitType} integer of the unit
+     * @param entity the live entity, or {@code null} if not yet available
+     * @return the primary {@link PersonnelRole}; never {@code null}
+     */
+    public static PersonnelRole primaryRole(int unitType, Entity entity) {
```
IllianiBird, line 79: I feel like this is defined in the codebase already.

see /Volumes/STOP EJECTING ME/MegaMek/mekhq/MekHQ/src/mekhq/campaign/unit/Unit.java specifically getDriverRole and getGunnerRole

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
```diff
@@ -0,0 +1,460 @@
...
+              resolved != null && resolved.isMercenary());
+        return resolved;
+    }
+
+    /**
+     * Converts a {@link SkillLevel} (where {@code NONE = 0, ULTRA_GREEN = 1, …, LEGENDARY = 7}) to a
+     * {@link SkillType} {@code EXP_*} constant (where {@code EXP_ULTRA_GREEN = 0, …, EXP_LEGENDARY =
+     * 6}). The two enums use different baselines; this is the canonical mapping.
+     */
+    static int toExperienceLevel(SkillLevel skillLevel) {
```
IllianiBird, line 407: This already exists elsewhere in the codebase: mekhq/campaign/personnel/skills/SkillType.java

skillLevelFromExperienceLevel(int experienceLevel)

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
```diff
@@ -0,0 +1,948 @@
...
+        this.payForAmmunition = payForAmmunition;
+    }
+    // endregion Getters/Setters
+
+    /**
+     * Two CommandGenerationOptionss are equal when every setting matches, which is how the dialog tells whether the
+     * settings have moved since the model was last generated.
+     */
+    @Override
+    public boolean equals(Object object) {
```
IllianiBird, line 816: This is incredibly clunky and will be a pain to maintain. There has to be a better way

MekHQ/src/mekhq/campaign/universe/commandGeneration/EnhancedImagingAugmentor.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Rolls for each formation and implants every warrior of those that take it.
+     *
+     * @return how many warriors were implanted
+     */
+    private static int implantWholeFormations(Map<Integer, List<Person>> formations, double chance) {
+        int implanted = 0;
+        for (Map.Entry<Integer, List<Person>> formation : formations.entrySet()) {
+            if (Math.random() >= chance) {
```
IllianiBird, line 165: Single use Math.random. This should be cached and reused rather than reinitialized each pass.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ManeiDominiAugmentor.java
```diff
@@ -0,0 +1,294 @@
...
+ * Fits generated Word of Blake Shadow Division warriors with Manei Domini rank, class and cybernetics.
+ *
+ * <p>The availability rules themselves live in {@link ManeiDominiImplants}, so MegaMek owns them and
+ * anything generating a Manei Domini can apply the same chart. What stays here is what only a campaign
+ * has: the campaign's own switches, a {@link Person} to fit implants to, and the campaign rank a
+ * warrior's Manei Domini standing is derived from.</p>
+ *
+ * @see ManeiDominiImplants
+ */
+public final class ManeiDominiAugmentor {
```
IllianiBird, line 62: This is gifting the implants, but it doesn't look like it's hooking those implants up with their Advanced Alt Medical 'injury'. This will break if the user has AAM enabled (enabled by default) and is using this system. In the best case the new implants will be just ignored, in the worst case they'll be removed on new day.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
```diff
@@ -0,0 +1,533 @@
...
+
+    // Carrier units are looked up by their full name, which MekSummary builds as chassis + model (see
+    // mm-data/data/mekfiles/infantry/Support Units; filenames differ from these names). Every squad
+    // size is a model of one chassis - "Support Squad (3 person)" - so the carrier for a given number
+    // of people is named rather than searched for.
+    private static final String UNIT_NAME_FORMAT = "%s (%d person)";
+    private static final String IS_PLATOON_CHASSIS = "Support Platoon";
+    private static final String IS_SQUAD_CHASSIS = "Support Squad";
+    private static final String CLAN_POINT_CHASSIS = "Clan Support Point";
+    private static final String CLAN_SQUAD_CHASSIS = "Clan Support Squad";
```
IllianiBird, line 92: Are these player-facing hardcoded Strings? If so, they should be in a RESOURCE_BUNDLE

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
```diff
@@ -0,0 +1,533 @@
...
+              maintenance.size(), medical.size(), command.size(), useClanStructure);
+
+        // Within each section, staff are grouped by profession into companies (e.g. "MekTech Company");
+        // capability vehicles get their own company named for their function ("Recovery",
+        // "Field Hospital"). Command has no capability vehicles, so its label is unused.
+        organizeSection(campaign, supportCommand, label("maintenance"), maintenance, profile,
+              useClanStructure, maintenanceVehicles, label("recovery"));
+        organizeSection(campaign, supportCommand, label("medical"), medical, profile,
+              useClanStructure, medicalVehicles, label("fieldHospital"));
+        organizeSection(campaign, supportCommand, label("command"), command, profile,
```
IllianiBird, line 222: Hardcoded player-facing Strings?

MekHQ/src/mekhq/campaign/universe/enums/CapitalType.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>The values form a significance ranking - a national capital outranks a regional (province) capital, which
+ * outranks a district capital. {@link #significance()} exposes that ranking so a system that contains several planets
+ * can report the most important capital status among them.</p>
+ *
+ * <p>The stored labels ({@code "National Capital"}, {@code "Region Capital"}, {@code "District Capital"}) are the exact
+ * strings authored in the data files; {@link #fromLabel(String)} maps them (and the bare enum names) back to constants
+ * during deserialization.</p>
+ */
+public enum CapitalType {
```
IllianiBird, line 53: I didn't think the planetary data held information beyond faction capital?

MekHQ/src/mekhq/campaign/universe/enums/TechAssignmentSortFactor.java
```diff
@@ -0,0 +1,79 @@
...
+    UNIT_WEIGHT("TechAssignmentSortFactor.UNIT_WEIGHT.text",
+          "TechAssignmentSortFactor.UNIT_WEIGHT.toolTipText"),
+    PILOT_SKILL("TechAssignmentSortFactor.PILOT_SKILL.text",
+          "TechAssignmentSortFactor.PILOT_SKILL.toolTipText");
+
+    private final String displayName;
+    private final String toolTipText;
+
+    TechAssignmentSortFactor(final String displayNameKey, final String toolTipKey) {
+        final ResourceBundle resources = ResourceBundle.getBundle("mekhq.resources.Universe",
```
IllianiBird, line 61: This should be a static String not a ResourceBundle - see the many uses of RESOURCE_BUNDLE for example

IllianiBird (changes requested):
A few areas for improvement, but they're largely minor. Of particular issue are areas where this PR reinvents functionality that already exists. I'd recommend having Claude do a wiper pass to see where extant methods can be re-used here, instead of having the same functionality defined in two places.

Beyond that it's mainly just the usual suspects: hardcoded strings and variable names.

MekHQ/src/mekhq/campaign/market/PartsInUseManager.java
HammerGS: Added tests. `unit` is captured on the line above and the whole block is inside `if (unit != null)`, so this call cannot NPE - the change actually removed a re-fetch of `getUnit()`, which was the risky part: the warehouse can refresh on the EDT while a generation worker is attaching and detaching parts, so the second `getUnit()` could come back null after the first one had passed.

Three tests now cover it in `PartsInUseManagerTest.TestUpdatePartInUseData`: a part on a mothballed unit is ignored when the flag is set, the same part is counted when the flag is clear (so the first test can't pass vacuously), and a part with no unit at all is counted normally rather than failing.

MekHQ/src/mekhq/campaign/personnel/enums/GeneticLegacyRole.java
HammerGS: Fixed - `GeneticLegacyRole` now resolves its label once in the constructor into a final field, following `ChaosEmployerType`. `getLabel()` just returns it.

MekHQ/src/mekhq/campaign/personnel/Bloodname.java
HammerGS: Fixed. Split into three named booleans:

```java
boolean isBeforeItExisted = year < startDate;
boolean wasRetired = (inactive > 0) && (inactive < year);
boolean hasSinceReturned = (reactivated > 0) && (reactivated <= year);
return isBeforeItExisted || (wasRetired && !hasSinceReturned);
```

MekHQ/src/mekhq/campaign/personnel/Bloodname.java
HammerGS: Fixed - `name.isBlank()` is now bracketed like the two clauses beside it.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
HammerGS: Fixed - swept the whole class: `fd` to `forceDescriptor` (29), `snap` to `snapshot` (33), and the `t0` below.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
HammerGS: Fixed - now `processRootStartedAt`.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/CommandGenerator.java
HammerGS: Fixed - it's now `RANDOM_STARTING_CASH_PER_DIE`, the C-Bills each die is worth when starting cash is rolled rather than taken as a percentage.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorSnapshot.java
HammerGS: These need to stay boxed, sorry - `null` is a real third state here, not an unset int. It means "don't touch this at all, let the ruleset pick", and `CommandGenerator` reads it that way:

```java
if (snapshot.getEchelon() != null) {
    forceDescriptor.setEchelon(snapshot.getEchelon());
}
```

With an `int` we'd need a sentinel value to say the same thing, and any sentinel we picked would be a legal echelon, unit type or weight class somewhere. I have marked all six getters `@Nullable` and documented what the null means, so the contract is at least visible now.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorSnapshot.java
HammerGS: Rewritten to the reader-map shape from `mission/contract/io` - a static `Map<String, SnapshotElementReader>` built once, looked up by element name.

It also fixed two things I'd missed. The old `default ->` branch dropped unknown elements silently, and the `catch (NumberFormatException nfe)` swallowed bad values with no log at all, so a preset that half-loaded gave no clue why. Both now log - unknown elements at DEBUG, unparseable values at WARN naming the element and the text.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorWalker.java
HammerGS: Merged the first two - building `namedChildren` and building `requests` are now one pass.

The third has to stay separate: `namer.nameSiblings()` needs the complete sibling group before it can pick designators that don't collide, so nothing can be walked until every name is settled. I've put a comment there saying so.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/ForceDescriptorWalker.java
HammerGS: Fixed - `@Nullable` on the return and on both parameters. The javadoc also had `@param echelon` twice with a stray paragraph wedged between them, from a bad merge; that's tidied.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationIconBuilder.java
HammerGS: Done, and swept. I ran a check across every file this PR adds for methods that can `return null` without the annotation - nine of them, in six files: `FormationIconBuilder` (3), `ForceDescriptorWalker` (2), `SupportPersonnelToTOE`, `RulesetRankAssigner`, `SupportPersonnelAssigner`, `SupportPersonnelGenerator`. All marked.

The same check found ten more in files this PR only touches - `Campaign.getUnit()`, `PlanetarySystem.getPlanetById()`, `WarehouseTab.getSelectedTech()` and so on. Those are all pre-existing methods, so I've left them alone rather than widen this PR; happy to do them as a follow-up if you want it.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
HammerGS: Fixed - moved to a new `mekhq.resources.FormationNamer` bundle, read once when the class loads.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
HammerGS: Agreed, done - the Roman numerals are in the same bundle.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/FormationNamer.java
HammerGS: Done. The suffixes are keyed by last digit (`ordinalSuffix.0` through `.9`, where index 0 is the one every other digit takes), plus a separate `teensOrdinalSuffix` for 11/12/13, which English treats as an exception and other languages may not. Commented in the properties file so the shape is obvious to whoever translates it.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/PersonnelRoleResolver.java
HammerGS: You're right, and thank you - `PersonnelRoleResolver` is deleted. `MultiCrewAssembler` was its only caller and it always has a `Unit` with the entity already attached, so it now just asks `unit.getDriverRole()` and `unit.getGunnerRole()`. That's 126 lines gone.

One difference to flag: those methods return `null` for a unit type they don't recognise, where the old resolver fell back to MekWarrior. I've kept the fallback in a small helper, because a generated command with an uncrewed unit in it is harder to notice and harder to fix than one with a warrior in the wrong speciality - but it logs a warning now, which the old code never did.

The two remaining methods on the resolver (`vesselCrewRole()`, `navigatorRole()`) just returned an enum constant each, so those calls became `PersonnelRole.VESSEL_CREW` and `PersonnelRole.VESSEL_NAVIGATOR` directly.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/SupportPersonnelGenerator.java
HammerGS: Half right, and the half that's right is the important one. `skillLevelFromExperienceLevel(int)` is the opposite direction - int to `SkillLevel` - so it can't be called here, but there is an existing accessor going the way I needed, and I'd missed it.

`SkillLevel.getExperienceLevel()` counts None as 0 and Ultra-Green as 1; `SkillType` counts None as -1 and Ultra-Green as 0. They run in step after that, so the whole table collapses to:

```java
return skillLevel.getExperienceLevel() - 1;
```

None and null still fall back to Regular - the generator can't build a support person with no skill at all - and there's an existing test pinning that.

MekHQ/src/mekhq/campaign/universe/commandGeneration/CommandGenerationOptions.java
HammerGS: Agreed, and the failure mode is worse than the maintenance: add a setting, forget to add its line here, and the dialog silently stops noticing that setting changed. It then rebuilds without ever telling the player their model was stale.

Both `equals` and `hashCode` now walk the declared fields instead of listing them, so a new setting counts from the moment it's declared. That took the two methods from 187 lines to about 50. I didn't reach for commons-lang3's `reflectionEquals` since MekHQ doesn't currently depend on it and this didn't seem worth adding one for.

MekHQ/src/mekhq/campaign/universe/commandGeneration/EnhancedImagingAugmentor.java
HammerGS: Fixed - it's `Compute.randomFloat()` now. Worth noting it was one of only two `Math.random()` calls left in MekHQ.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ManeiDominiAugmentor.java
HammerGS: Good catch, and it was the worse of your two cases - they were being removed on new day, not ignored.

Most Manei Domini implants are fine: they're set once and nothing else writes to them. The two dermal ones are not. `checkForDermalEligibility` runs every new day when Advanced Alternate Medical is on, counts the person's dermal prosthetics, and writes the answer straight over `MD_DERMAL_ARMOR` and `MD_DERMAL_CAMO_ARMOR`. A warrior we'd fitted with dermal armour had no prosthetics behind it, so the count was zero and the implant was cleared the following morning.

Fixed by recording the prosthetics your rule set expects - `DERMAL_MYOMER_ARM_ARMOR` on both arms and `DERMAL_MYOMER_LEG_ARMOR` on both legs, at severity zero, the same marker-injury shape `AdvancedReplacementLimbDialog` uses. That satisfies the four-limb threshold, and it puts the implants in the medical record where the rest of the campaign can see them. Campaigns not using AAM are untouched - there the option is the only record anyone keeps, and adding injuries would be inventing surgery that never happened.

There are tests for both directions, including one that pins the old failure: an implant with no prosthetics behind it does not survive the daily pass.

One thing you should know about: only 12 of the 25 implants `ManeiDominiImplants` can issue have any `ProstheticType` representing them. The prosthetic-limb family, the gas effusers, suicide implants and TSM have none. That doesn't break anything today, since nothing re-derives those options - but it does mean an AAM campaign can't see most Manei Domini augmentation. Happy to raise that separately if it's worth modelling.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
HammerGS: Checked, and these aren't display strings - they're unit lookup keys. `topUnitName()` builds "Support Platoon (28 person)" and that string goes to `MekSummaryCache.getInstance().getMek(...)` to find the actual carrier in the unit files. Putting them in a bundle would break support carrier lookup outright, and silently.

I've left them as constants. If it'd read better I can rename them to something that says "lookup key" more loudly.

MekHQ/src/mekhq/campaign/universe/commandGeneration/SupportPersonnelToTOE.java
HammerGS: These already are - `label("maintenance")`, `label("medical")` and `label("command")` all go through `getTextAt(RESOURCE_BUNDLE, ...)` via the `label()` helper at the bottom of the class. The keys live in `SupportPersonnelToTOE.properties`.

MekHQ/src/mekhq/campaign/universe/enums/CapitalType.java
HammerGS: It does, and it's used fairly widely. Counting the current planetary system files: 153 systems carry `capitalType: Region Capital`, 133 carry `District Capital`, and 26 carry `National Capital`.

That's what the enum reads - a system's administrative standing within its owner, per era, which is separate from the faction-capital flag. The ranking is there so a system holding several planets can report the most important status among them.

MekHQ/src/mekhq/campaign/universe/enums/TechAssignmentSortFactor.java
HammerGS: Fixed - `private static final String RESOURCE_BUNDLE` plus `getTextAt`, so the bundle isn't re-resolved per constant. The now-unused `ResourceBundle` and `MekHQ` imports are gone with it.

IllianiBird (approved):

MekHQ/src/mekhq/campaign/universe/commandGeneration/ManeiDominiAugmentor.java
HammerGS: Raised as #9852 with the full audit.

Short version of the numbers: 25 implants total. 12 already have a `ProstheticType` naming them, 2 are the dermal pair this PR handles, and 11 have no route into AAM.

Four of the 11 look like they only need the option added to a type that already exists — `MD_TSM_IMPLANT` to `DERMAL_MYOMER_{ARM,LEG}_TRIPLE`, `MD_GAS_EFFUSER_PHEROMONE` to `PHEROMONE_EFFUSER` (already flagged `isWordOfBlakeOnly`), `MD_PL_TAIL` to `COSMETIC_TAIL_PROSTHETIC`, and `MD_PL_MASC` to `ELECTIVE_MYOMER_LEG`. The other seven need new types: the toxin effuser, the four prosthetic limb/wing options, the improved enhanced limbs, and the explosive suicide implants.

Worth knowing: four of those seven are Capellan rather than Word of Blake — the Thuggee Phansigars get the same options — so it is not purely a Manei Domini gap. And the suicide implants may not want to be a prosthetic at all, which is a call for you rather than me.

PR #9839, opened 2026-08-26
Fixed Failing Unit Tests

IllianiBird:
## What this changes

A handful of tests were broken, now they're not.

---


Sleet01 (approved):
LGTM

PR #9913, opened 2026-09-03
Fix #9896: Fixed Multiple Large Vessel Temp Crew Bugs

### LINKED ISSUE #9896: Dropship Temporary Crew does not get unassigned when assigning multiple crew at once
### What went wrong

My dropship was nonfunctional despite being fully crewed and assigned in the TOE. A user noticed that I had too many pilots assigned to the dropship. Clearing the dropship and reassigning the crew fixed the issue. The original crew was populated with Temp Crew automatically when the dropship was unmothballed. Assigning my crew manually somehow broke the number of Pilots assigned.

### What you expected instead

Assigning my own crew should not break my dropship. The number of pilots should not exceed the maximum value.

### Steps to reproduce

1. Select both Vessel Pilots in the Personnel tab (Ken'ichi Shacklock and Stephane Mary)
2. Right click and Assign to Unit both pilots to the Mule dropship (simultaneously)
3. Navigate to the Hangar and notice that the Mule has 4/3 pilots

### Attach your files

[The Raven Dreadnaughts30480830.cpnx.gz](https://github.com/user-attachments/files/31712246/The.Raven.Dreadnaughts30480830.cpnx.gz)

### Relevant log lines

```text

```

### MegaMek Suite Version

51.0

### Operating System

### Java Version

    Java Vendor: Eclipse Adoptium     Java Version: 21.0.11

### Before you submit

IllianiBird:
Fix #9896

## What this changes

Multi-crewed units were missing the mechanic that removed temp crew when replacing them with real personnel.

Furthermore, the intention for multi-crewed large vessels is that at least 1 real person must exist in each of the 3-4 roles (pilot, gunnery, tech, and navigator). This was also missing, creating a situation where a player could crew a dropship (for example) with a single pilot and then have no tech nor gunnery skills).

This PR fixes both and adds a regression test.

## Testing

- Manual testing
- AI regression test
- AI code review.

---


HammerGS (approved):

HammerGS:
The job failed in :MekHQ:test, and the actionable failure is in CampaignTest.TempCrewTests.testFillTempCrewPoolCalculatesNeedsFromUnitsWithCrew(PersonnelRole).

Root cause
The test expected the temp crew pool to be filled based on a unit’s missing crew slots, but the code returned 0 for these roles:

Vessel Pilot: expected 1, got 0
Vessel Gunner: expected 2, got 0
Vessel Crewmember: expected 9, got 0
That points to the crew-pool calculation not recognizing these vessel roles when deriving needs from units with crew.

Relevant failing area
The failure is reported at:

MekHQ/src/.../CampaignTest.java:987
The workflow is defined in:

.github/workflows/ci.yml
Likely fix
The implementation that populates temporary crew needs should include vessel crew roles in its role-matching / unit-crew aggregation logic. In practice, the bug is probably one of these:

The method only counts a subset of personnel roles and skips vessel-specific roles.
The unit crew requirement lookup is ignoring roles unless they are tied to a mech/standard unit type.
The pool-fill logic is filtering by the wrong role enum/category before summing deficits.

PR #9922, opened 2026-09-03
fix: explicitly call out selectedScenario as not being Null

firefly2442:
fix: explicitly call out `selectedScenario` as not being Null, prevent potential Null being passed to constructor

`selectedScenario` is checked to make sure it's not Null.  However, at the end, it's used as a parameter to a constructor.  If it shouldn't ever be null, it seems more clear to spell that out up-front with `requireNonNull`.

Found via static analysis tool [SpotBugs](https://github.com/spotbugs/spotbugs) 4.10.4.

## Testing

Ran Gradle tests, no smoketests run.

## What is not proven yet

---

No AI usage.

### DIFF
--- a/MekHQ/src/mekhq/gui/TOETab.java
+++ b/MekHQ/src/mekhq/gui/TOETab.java
@@ -248,7 +248,10 @@ public final class TOETab extends CampaignGuiTab {
      * @since 0.50.10
      */
     private void deployToRegularScenario(Scenario selectedScenario) {
-        Objects.requireNonNull(selectedScenario, "selectedScenario");
+        if (selectedScenario == null) {
+            LOGGER.warn("TOETab: Null selectedScenario in deployToRegularScenario()");
+            return;
+        }
 
         // Get available forces
         mekhq.campaign.Campaign campaign = getCampaign();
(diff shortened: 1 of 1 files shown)


MekHQ/src/mekhq/gui/TOETab.java
```diff
@@ -247,6 +248,8 @@ private void deployToStratCon(Scenario selectedScenario) {
      * @since 0.50.10
      */
     private void deployToRegularScenario(Scenario selectedScenario) {
+        Objects.requireNonNull(selectedScenario, "selectedScenario");
```
IllianiBird: This will throw an error and brick the process if we pass in a `null`. I'd far rather we logged the issue and provided a soft failure instead of hard.

firefly2442: That makes sense.  How about this defensive check at the beginning and logging a warning?

IllianiBird: That can work, but you should probably include...

```
 getCampaignGui().undeployForce(selectedFormation);
            selectedFormation.clearScenarioIds(getCampaign(), true);
```
            
            Otherwise, you'll end up with stale deployment data baked into the unit.

IllianiBird: I have no idea why that message is formatted weird

firefly2442: OK thanks

IllianiBird (approved):

PR #9951, opened 2026-09-05
fix: SpotBugs code logic cleanup

firefly2442:
This addresses a couple minor items from the static analysis tool SpotBugs 4.10.4.

* A logical AND check
* Some logging that will never be reached
* Cloning will always return a non null object so some code cleanup to simplify

## Testing

Gradle tests

## What is not proven yet

---

No AI used.


IllianiBird (approved):

IllianiBird:
Let's try and avoid multiple unrelated changes in individual PRs. Makes reviewing a tiny bit more troublesome that it needs to be :)

firefly2442:
Will do, thanks.

PR #9957, opened 2026-09-05
Added & Integrated Piloting/ProtoMek Skill from ATOW

IllianiBird:
# Requires https://github.com/MegaMek/mm-data/pull/549

## What this changes

At TW-scale ProtoMek pilots do not use a separate piloting skill, instead using their gunnery skill. This PR adds the initial work necessary to segregate gunnery and piloting for ProtoMeks.

Existing personnel will automatically be given the missing Piloting/ProtoMek skill for free. The level of this skill will be equal to the characters' current Gunnery/ProtoMek skill.

## Testing

- Manual testing

## What is not proven yet

- ~~MegaMek integration.~~ _Works out the box_

---


HammerGS:
## MegaMek review: does this need a MegaMek change?

Reviewed against MegaMek `main` at `204ea6125c` and this branch at `b1b86ec2c3`.

**Short answer: no MegaMek change is needed for this PR to work. One small MegaMek follow-up is recommended so the new skill cannot change results it should not touch.**

### Why it works as-is

- MekHQ already hands MegaMek a plain gunnery/piloting pair through `Crew.setGunnery` / `Crew.setPiloting` (`Unit.java:5847` and `:6003`). MegaMek accepts any 0-8 value and does not care which MekHQ skill it came from.
- ProtoMeks never roll PSRs in MegaMek (`ProtoMek.getBasePilotingRoll` returns CHECK_FALSE), matching TW p.41.
- BV already forces a ProtoMek's piloting to 5 (`BVCalculator.bvMultiplier:1319`), so BV is identical before and after.
- ProtoMek physical attacks use a fixed base of 4. ProtoMeks cannot be charged. Space sensor rolls are spaceborne-only. Lobby, tooltip and MUL show gunnery/piloting for every unit already, so nothing new appears.

### Where the new number does reach MegaMek

The PR flips `getDrivingSkillFor(ProtoMek)` from Gunnery/ProtoMek to Piloting/ProtoMek. Today the crew's piloting slot is a copy of gunnery, which is why the three sites below give the rules-correct answer by accident. After the migration the two skills start equal, then drift apart as one is trained and the other is not.

| MegaMek site | ProtoMek reaches it? | What the rules want |
|---|---|---|
| Grapple as attacker (`GrappleAttackAction:109`) | Yes, ProtoMeks may grapple | Gunnery (TO:AR p.89) |
| Break grapple as attacker (`BreakGrappleAttackAction:118`) | Yes | Gunnery (TO:AR p.91) |
| DFA piloting differential as target (`DfaAttackAction:358`) | Yes, ProtoMeks are legal DFA targets | Gunnery (TW p.41: they have only Gunnery) |

Example: a Roc pilot with Gunnery 2 and Piloting/ProtoMek 3. Her grapple base to-hit is 2 today and becomes 3 after this PR. A Piloting 4 MekWarrior DFA-ing her gets +2 today and +1 after.

### Recommendation

- **MegaMek follow-up (recommended):** substitute gunnery for a ProtoMek's piloting at those three sites. Cleanest shape is one helper on `Entity` that returns crew piloting, overridden in `ProtoMek` to return gunnery. That makes MegaMek correct for every client, not just MekHQ, and does not block this PR in either order. I can open that PR.
- **Alternative, MekHQ only:** keep sending gunnery as the crew piloting for ProtoMeks and use the new skill everywhere else. Zero MegaMek change, but `getDrivingSkillFor` also feeds the personnel table, crew list, CamOps rating and the force generator adapter, so it would need a second method.

### MekHQ-side things I noticed while tracing the handoff

- **Migration never runs for most ProtoMek pilots.** `resolveRolePerformability` is only called from the `!canPerformRole(secondaryRole)` branch (`CampaignXmlParser.java:2548`), and the primary-role `if` below it is now empty. A ProtoMek pilot with secondary role None passes the secondary check, so the gunnery-to-piloting copy never happens. `canDrive` now requires Piloting/ProtoMek, so those pilots cannot be assigned to their ProtoMeks after loading.
- **The helper's branches are crossed.** The primary-role failure clears the *secondary* role and reports `ineligibleForPrimaryRole`; the secondary-role failure clears the *primary* role. Both checks also pass `primary=false`, so a valid primary role fails the "cannot equal the primary role" rule every time. The old code set an invalid primary to Dependent, not None.
- `PersonnelMarketAtB.java:221` still adjusts only Gunnery/ProtoMek for ProtoMek pilots, so the market's piloting modifier never applies to the new skill.
- `AverageExperienceRating.java:266` now averages piloting for ProtoMek pilots. Worth confirming that is the intended CamOps reading, since the TW sheet has no piloting entry for them.
- Bloodname targets (`ForceHumanResources.java:2060`) and `canGun` keep using Gunnery/ProtoMek. That is the right skill for both.

### Rulebook basis

> ProtoMech pilots never need to make Piloting Skill Rolls, and so these pilots have only Gunnery Skill. (TW 11th printing, p.41)

> The base Target Number for a grapple attack is equal to the attacking unit's Piloting Skill Rating (for ProtoMechs, use their Gunnery Skill Rating). (TO:AR 6th printing, p.89)

> Anytime these rules call for a Piloting Skill Roll to be made by ProtoMechs, use the unit's Gunnery Skill Rating. (TO:AR 6th printing, p.91)

IllianiBird:
Flipping to live now we've confirmed no MegaMek action required

HammerGS (approved):

Issue #9937, opened 2026-09-04
Company Generator Incorrectly Crews Support Assets


IllianiBird:
### What went wrong

When using the new company generator some support assets (MASH and recovery vehicles) generate with the wrong crew composition.

<img width="594" height="215" alt="Image" src="https://github.com/user-attachments/assets/d5ea89f3-a332-4a40-8f0e-8b2b6e55d055" />

<img width="579" height="196" alt="Image" src="https://github.com/user-attachments/assets/eb7d181f-af2e-46d0-a907-35b4028d7e82" />

MekHQ recognizes that these units are incorrectly crewed, meaning they end up non-functional and the excess crew is ejected on load.

Furthermore, no support assets are crewed by techs:

<img width="1300" height="166" alt="Image" src="https://github.com/user-attachments/assets/b62fae75-2dd6-4173-ba18-14b2deccc9a0" />

### What you expected instead

- All units should be correctly crewed
- All units should have techs

On that last note, MekHQ already has a built-in tech crewing mechanic, so the bespoke one in company generator can probably be removed. The pre-existing one, which is well tested and works 100%, can just be run once after the player force is generated. See: `mekhq/campaign/utilities/AutomatedTechAssignments.java`

### Steps to reproduce

1. Launch new campaign
2. Generate force of any size
3. Observe crewing & tech assignment issues
4. When save is reloaded the excess crew are ejected

### Attach your files

[Defunct Defenders31510101_Thrust-part1.zip](https://github.com/user-attachments/files/31842393/Defunct.Defenders31510101_Thrust-part1.zip)

### Relevant log lines

```text

```

### MegaMek Suite Version

0.51.01

### Operating System

Mac OS X 26.6.2 (aarch64)

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

- [x] This report only contains one issue or request, and I haven't already reported it
- [x] I am on the right repository - MegaMek for battles, MegaMekLab for unit design, MekHQ for campaigns, mm-data for unit and planet data

- [x] I have asked about this on the MegaMek Discord (optional)

PR #9963, opened 2026-09-06
Use the shared tech assigner in the Company Generator

HammerGS:
## What this changes

The Company Generator had its own tech-to-unit assigner. It now uses MekHQ's, the one the new day and the Hangar's quick-assign button already run. Nothing changes for the player at default settings.

Asked for by @IllianiBird in #9937. That issue is closed by #9962, so this PR does not close it.

- The Setup tab's sort grid still works. Units are offered a tech in the player's chosen order.
- Generation still never claims a tech the campaign already had. Only the techs it just made are offered.
- Support assets get techs, deployed and refitting units are skipped, and a person marked "never assign maintenance automatically" is now left alone. The old assigner ignored that flag.
- Techs are load balanced. The old one filled the best tech first.
- One behaviour change: a tech now takes at most two units, where the old code budgeted 480 maintenance minutes per tech. No practical difference at the default 100% tech coverage. Below roughly 50% coverage, some units will be left unmaintained that the old code would have covered.
- 646 lines fewer.

Stacked on #9962. Please merge that one first.

## Testing

- `./gradlew :MekHQ:test :MekHQ:checkstyleMain :MekHQ:javadoc` green. 15,416 tests, 0 failures.
- 10 new tests in `TechAssignmentOrderTest`, 4 new in `AutomatedTechAssignmentsTest`.
- `constructor_withOrderSupplied_offersUnitsInThatOrderInstead` fails when the new comparator is ignored, so the seam is really covered.
- In game: new campaign, Company Generator, 1 Mek battalion, 1 BA company, 1 VTOL company, everything on defaults, CamOps Salvage and MASH Theatres on.
- All 70 maintainable units came out with a tech and none without. The 14 infantry units correctly have none.
- The two-unit cap was reached: 24 techs took two units each and 22 took one, covering every unit with 46 of the 70 techs available.
- The defaults set all three sort slots, so the run used the new ordering rather than the fallback.
- Saved and reloaded. Crew and techs both survived.

## What is not proven yet

- The aero bucket. The test force had no fighters or small craft, so only the Mek, vehicle and Battle Armor buckets ran in game.
- The fallback ordering, used when a caller supplies no comparator. Every caller outside generation takes that path, and they are unchanged, but the in-game run did not exercise it. Covered by the existing unit tests.
- Tech coverage below 50%, where the two-unit cap starts leaving units unmaintained. Reasoned about, not played.

---

- [x] Tests added or updated, if this implements a rule or changes game state


MekHQ/src/mekhq/campaign/utilities/AutomatedTechAssignments.java
```diff
@@ -122,6 +130,28 @@ public List<String> getReports() {
...
+     * available techs changes, so a caller that has its own idea of which units matter most - the Company Generator
+     * lets the player rank them by pilot rank, unit weight and pilot skill - gets that order honoured without
+     * maintaining a second assigner.</p>
+     *
+     * @param techs             all available personnel to consider for assignment
+     * @param units             all units that may require a tech assignment
+     * @param unitPriorityOrder the order to offer units a tech in, or {@code null} for battle value descending
+     *
+     * @author Illiani
+     * @since 0.50.11
```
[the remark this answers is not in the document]
HammerGS: Valid, fixed in 4f208e4505.

Both the new constructor and its `sortUnitBucket` helper carried the class's original `@since 0.50.11`, copied from the two-argument constructor they sit next to. They are new here, so both now read `@since 0.51.01` - the suite version this ships in, and the form the rest of MekHQ uses for current work (228 uses against 81 of `0.51.0`). `TechAssignmentOrder` keeps `0.51.0` to match its own package, where all 12 tags use that form.

I also dropped the `@author Illiani` tags from those two members while I was there. The class is Illiani's and the class-level tag still records that, but these two members are not, and this repo attributes per author rather than per file.

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/TechAssignmentOrder.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Builds the unit ordering the player asked for on the Setup tab.
+     *
+     * @param campaign the campaign the units belong to, used to read pilot skill levels
+     * @param options  the generation options holding the three sort slots and their directions
+     *
+     * @return the ordering, or {@code null} when no slot is set, so the caller keeps the shared assigner's default
+     *
+     * @author Illiani
```
IllianiBird: Don't sign things with other peoples' names. Also the version is wrong

IllianiBird (approved):

MekHQ/src/mekhq/campaign/universe/commandGeneration/ratgen/TechAssignmentOrder.java
HammerGS: Both valid, thank you - fixed across two commits.

**Signing with someone else's name.** Entirely my error: I copied the header pattern from neighbouring classes, which put your name on files you did not write. Removed in `e6faaea4aa` - five `@author` tags in total, across this PR and #9966. I then audited all eleven open branches in both repos for any `@author` line my work added, and there are none left.

The project's own guidance has been updated so this cannot recur: attribution is to The MegaMek Team, and new files get no `@author` tag at all. I left the pre-existing tags in files I was only editing rather than widening the diff.

**Version.** Correct - fixed in `914af90f0e`. `Version.properties` is `major=0 minor=51 patch=01`, formatted `%d.%02d.%02d`, so the tag should read `0.51.01`. Four tags in this PR said `0.51.0`. For what it is worth `main` currently carries 228 uses of `0.51.01` against 74 of `0.51.0`, so the malformed form has spread a fair way.

Compiles clean. No behaviour change in either commit - javadoc only.

PR #10018, opened 2026-09-11
Close #4047: Rebuild the interstellar navigation map

### LINKED ISSUE #4047: RFE: Map Improvements
This is CGL request for improvements to the universe map.

### Need to have: 

- [ ] Ability to enter in two planets without scrolling through 2000 in a pulldown menu. Autocomplete as you type the name is ideal.
- [ ] Ability to have a detailed output: time at each jump, transit time at each system
- [ ] Calculator that lets you change travel time by changing G. Example: if the trip is 7 days, but I want it to be 3 days, then it tells me what G is needed for that. Or, if the travel time is 17 days, but I want it done at 1.5 G, it would tell me what the new travel time is.
- [ ] A Short cut on the main menu to immediately open to the map tab.

Nice to have: 

- [ ] Ability to select alternative paths to the target world, or to use multiple waypoints. 
- [ ] Ability to account for command circuits, whole or partial, that may not be along the intended route, but provide sufficient net time benefit that it should be used 
- [ ] Ability to factor in pirate points, including the tool indicating the difficulty of the jump given the Piloting Skill and other modifiers that you can select or deselect. 
- [ ] Route planner that suggests a path given the intent to be at a given waypoint at a given date, or to plan in time-on-planet for a mission, or for R&R. 
- [ ] Tool that allows you to see what the impact is of detection ranges

VicenteCartas:
## What this changes

Rebuilds the interstellar map as an immersive navigation workspace with semantic zoom, intrinsic stars, retained territory rendering, clearer strategic markers, searchable route planning, editable waypoints, route alternatives, command-circuit assumptions, detailed itineraries, acceleration and schedule planning, reachability, distance measurement, and explicit route-risk analysis.

Route planning deliberately allows an explicitly selected empty destination or waypoint and reports the risk as a caution. The existing avoid-empty-systems option still excludes empty systems when they are merely intermediate candidates.

Adds dated capital and administrative hierarchy support to the map and system dossier. Region and district outlines reuse national contours to form closed shapes without duplicate cross-faction borders, while missing or conflicting administration remains unassigned.

Also corrects same-date planetary event loading so distinct fields are merged instead of later records replacing earlier ones. Against the current canon data, this revives 144 previously discarded field values: 123 capital types, 13 population values, four messages, two socio-industrial values, one faction, and one name. The reviewed National Capital results match the source data.

The renderer uses retained pannable layers, exposed-strip refresh, asynchronous cartography preparation, stable mode-transition surfaces, direct-snapshot active zoom, and spatially bounded HPG/system painting. Shared framed command buttons and map controls preserve keyboard, focus, tooltip, and accessible-description behavior.

Closes #4047.

Data dependency: MegaMek/mm-data#561 activates the existing dated administration records. The map remains functional without that data but omits unavailable administrative details and boundaries.

## Testing

- `:MekHQ:checkstyleMain` and `:MekHQ:checkstyleTest` passed on the final source/test tree. These checks include project formatting, spacing, and required source/test file headers.
- The complete `:MekHQ:test` suite ran 15,608 tests: 15,607 passed, 12 were skipped, and one known unrelated baseline test failed: `CommandGeneratorAugmentationRulesTest.theChosenRulesAreWrittenToTheCampaign` (expected `true`, was `false`). The same unchanged failure was reproduced independently before this publication run.
- All map, route-planning, administrative-boundary, marker, cache, gesture, dossier, and rendering regressions in the complete suite passed.
- `git diff --check upstream/main...HEAD` passed.
- The feature was exercised in live campaigns at normal and forced 200% scaling. Checks covered mode transitions and reversal, semantic zoom, pan/zoom continuity, rings and labels, marker clipping, HPG placement and viewport entry, administrative closure, route editing and analysis, dossier reveal and scrolling, and inspector collapse/restore.
- Three-run dense, detail, atlas, transition, zoom, and overlay-heavy benchmarks completed without visual artifacts or synchronous full renders during active zoom. No final measured interaction frame exceeded 33 ms.
### Performance harness

The deterministic live-Swing harness is disabled by default. On Windows PowerShell, enable it before launching MekHQ:

```powershell
$env:JAVA_TOOL_OPTIONS="-Dmekhq.map.renderProfiling=true -Dmekhq.map.renderBenchmark=true"
.\gradlew.bat run
```

With the interstellar map open:

- `Ctrl+Shift+B`: `pan-v2` warm-pan benchmark (5-second warmup, 30-second measurement)
- `Ctrl+Shift+T`: `transition-cold-v1` retained mode transition plus separate exact-cache regeneration
- `Ctrl+Shift+Z`: `zoom-v1` active cursor-anchored zoom plus separate settled regeneration

Each run restores its original camera or mode. Results are written as `Map render benchmark result:` entries in `MekHQ/logs/mekhq.log`.

## What is not proven yet

Full administrative presentation depends on MegaMek/mm-data#561 and should merge after that data PR. Missing administration is deliberately not inferred. The one unrelated full-suite baseline failure described above remains outside this PR.

The reusable immersive controls remain in `mekhq.gui.baseComponents`; a broader package/theme architecture and whole-application control migration are intentionally deferred to a focused follow-up PR.

---


MekHQ/src/mekhq/campaign/Campaign.java
```diff
@@ -4132,175 +4206,78 @@ public JumpPath calculateJumpPath(PlanetarySystem start, PlanetarySystem end, bo
...
+    private NavigationRouteAnalysis.Policy createNavigationPolicy(boolean skipAccessCheck,
+          boolean skipEmptySystemCheck, List<AbstractContract> activeContracts, FactionHints factionHints) {
+        class CampaignNavigationPolicy implements NavigationRouteAnalysis.Policy {
+            private final boolean escapingOutlawedOrigin;
 
-            if (bestMatch == null) {
-                break; // No path exists
+            private CampaignNavigationPolicy(boolean escapingOutlawedOrigin) {
+                this.escapingOutlawedOrigin = escapingOutlawedOrigin;
             }
```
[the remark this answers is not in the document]
VicenteCartas: Addressed in 4ef10265a4. The campaign policy no longer caches an `escapingOutlawedOrigin` value for a whole segment. `canTraverse` now passes each leg's actual origin and destination to `canEnterTargetSystem`, so the existing escape exception applies only while leaving an outlawed leg origin.

MekHQ/src/mekhq/campaign/NavigationRouteAnalysis.java
```diff
@@ -0,0 +1,355 @@
...
+            findings.add(new Finding(avoided
+                                           ? FindingKind.ABANDONED_DESTINATION_AVOIDED
+                                           : FindingKind.ABANDONED_DESTINATION_ALLOWED,
+                  avoided ? Severity.BLOCKED : Severity.CAUTION));
+        }
+        if (Double.isFinite(rechargeHours)) {
+            findings.add(new Finding(FindingKind.RECHARGE_DURATION, Severity.INFO));
+        } else {
+            findings.add(new Finding(FindingKind.RECHARGE_IMPOSSIBLE, Severity.BLOCKED));
+        }
```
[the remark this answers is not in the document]
VicenteCartas: Addressed in 4ef10265a4. Impossible recharge remains blocking for intermediate systems, but is now an informational finding at the final arrival. Reachability includes such a destination while deliberately not enqueueing it for another hop. `NavigationRouteAnalysisTest` covers both final arrival and non-expansion.

MekHQ/src/mekhq/campaign/NavigationRouteAnalysis.java
```diff
@@ -0,0 +1,355 @@
...
+            for (PlanetarySystem destination : neighbors) {
+                String destinationId = systemId(destination);
+                if (reachedHops.containsKey(destinationId)) {
+                    continue;
+                }
+
+                LegAssessment assessment = assessLeg(origin, destination, date, useCommandCircuit, segmentPolicy);
+                ReachabilityEntry entry = new ReachabilityEntry(destination, nextHop, assessment);
+                if (!segmentPolicy.isSystemAllowed(destination)
+                          || (assessment.severity() == Severity.BLOCKED)) {
```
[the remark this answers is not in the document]
VicenteCartas: Addressed by the same root fix in 4ef10265a4. The campaign policy no longer derives or retains segment-scoped escape state, so reachability's `canTraverse(origin, destination)` evaluates the actual BFS edge and cannot inherit the anchor's exemption.

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9306 @@
...
+        button.putClientProperty("navigationUtilityButton", Boolean.TRUE);
+    }
+
+    static JButton createNavigationUtilityIconButton(int symbolCodePoint, int buttonSize, String toolTipText,
+          String accessibleName, String accessibleDescription) {
+        JButton button = new JButton() {
+            @Override
+            protected void paintComponent(Graphics graphics) {
+                Graphics2D buttonGraphics = (Graphics2D) graphics.create();
+                paintNavigationUtilityButton(butto
```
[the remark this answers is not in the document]
VicenteCartas: Addressed in 4ef10265a4. Administrative classification now requires every evidence system to have exactly the cell owner and the same complete, nonblank administration path. Missing or contradictory evidence leaves the cell unassigned; the boundary test covers both evidence orders and missing data.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -334,12 +1407,14 @@
             getCampaign().addReport(GENERAL, jumpReport);
         }
 
-        getCampaign().getPlayerForce().getForceDetachment().getCurrentLocation().setJumpPath(panMap.getJumpPath());
+        currentLocation.setJumpPath(jumpPath);
```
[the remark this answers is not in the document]
VicenteCartas: Addressed in 5410b2801c. `beginTransit` now returns immediately when `currentLocation` is null, then consistently uses that proven non-null local for the active path, transit time, and final route assignment.

MekHQ/src/mekhq/gui/view/PlanetViewPanel.java
```diff
@@ -518,44 +558,281 @@
...
+        constraints.anchor = GridBagConstraints.NORTHWEST;
+        return constraints;
+    }
+
+    private String format(String key, Object... arguments) {
+        MessageFormat formatter = new MessageFormat(resourceMap.getString(key), MekHQ.getMHQOptions().getLocale());
+        return formatter.format(arguments);
+    }
+
+    private class RevealBandPanel extends JPanel {
```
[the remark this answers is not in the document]
VicenteCartas: Addressed in 5410b2801c. `RevealBandPanel` is now a static nested class, removing its unnecessary implicit reference to `PlanetViewPanel`; the stateful `DossierSection` remains a non-static subclass.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -416,27 +1627,143 @@
...
+        panSystem.repaint();
+    }
+
+    @Subscribe
+    public void handle(MissionEvent event) {
+        panMap.refreshPresentationData();
+    }
+
+    @Subscribe
+    public void handle(ScenarioEvent event) {
```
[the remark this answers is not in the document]
VicenteCartas: Keeping this parameter. This is a Guava EventBus `@Subscribe` method: subscriber methods must accept exactly one parameter, and its type defines the event subscription. Removing `ScenarioEvent event` would invalidate registration even though the payload value is not otherwise read.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -416,27 +1627,143 @@
...
+    }
+
+    @Subscribe
+    public void handle(LocationRemovedEvent event) {
+        panMap.refreshPresentationData();
+        panSystem.repaint();
+    }
+
+    @Subscribe
+    public void handle(MissionEvent event) {
```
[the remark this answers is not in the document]
VicenteCartas: Keeping this parameter. This is a Guava EventBus `@Subscribe` method: subscriber methods must accept exactly one parameter, and its type defines the event subscription. Removing `MissionEvent event` would invalidate registration even though the payload value is not otherwise read.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -416,27 +1627,143 @@
...
+    }
+
+    @Subscribe
+    public void handle(LocationAddedEvent event) {
+        panMap.refreshPresentationData();
+        panSystem.repaint();
+    }
+
+    @Subscribe
+    public void handle(LocationRemovedEvent event) {
```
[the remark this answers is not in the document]
VicenteCartas: Keeping this parameter. This is a Guava EventBus `@Subscribe` method: subscriber methods must accept exactly one parameter, and its type defines the event subscription. Removing `LocationRemovedEvent event` would invalidate registration even though the payload value is not otherwise read.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -416,27 +1627,143 @@
...
     }
 
     @Subscribe
     public void handle(OptionsChangedEvent ev) {
         panMap.repaint();
+        updateRouteStrip();
+    }
+
+    @Subscribe
+    public void handle(LocationAddedEvent event) {
```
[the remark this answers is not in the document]
VicenteCartas: Keeping this parameter. This is a Guava EventBus `@Subscribe` method: subscriber methods must accept exactly one parameter, and its type defines the event subscription. Removing `LocationAddedEvent event` would invalidate registration even though the payload value is not otherwise read.

MekHQ/src/mekhq/campaign/PiratePointAnalysis.java
```diff
@@ -0,0 +1,193 @@
...
+    public record Modifier(ModifierCategory category, int value, boolean enabled) {
+        public Modifier {
+            Objects.requireNonNull(category);
+        }
+    }
+
+    /** Explicit physical and check assumptions used for one analysis. */
+    public record Input(double standardDistanceKm, double pirateDistanceKm, double detectionRadiusKm,
+                        double accelerationG, int baseTargetNumber, List<Modifier> modifiers) {
+        public Input {
```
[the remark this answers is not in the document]
VicenteCartas: This is a false positive. The record compact constructor reassigns `modifiers = List.copyOf(modifiers)`, so the generated field stores an immutable defensive copy rather than the caller's list. `analysisInputDefensivelyCopiesModifierState` clears the source list after construction and verifies the stored state and analysis result remain unchanged.

MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
```diff
@@ -334,6 +334,25 @@ public Set<Faction> getFactionSet(LocalDate when) {
...
+            List<String> planetAdministration = planet.getAdministration(when);
+            if (planetAdministration.isEmpty()) {
+                continue;
+            }
+            if ((administration != null) && !administration.equals(planetAdministration)) {
+                return List.of();
+            }
+            administration = planetAdministration;
+        }
+        return (administration == null) ? List.of() : List.copyOf(administration);
```
IllianiBird, line 353: Consider `Collections.emptyList()` instead of `List.of()`

MekHQ/src/mekhq/campaign/JumpPathSchedule.java
```diff
@@ -0,0 +1,201 @@
...
+     * @param itinerary                  immutable route forecast
+     * @param requestedStops             explicit requested stops in route-planning order
+     * @param dwellHoursByRequestedStop  dwell hours aligned with {@code requestedStops}; missing values are zero
+     * @param mode                       whether {@code anchor} is departure or arrival
+     * @param anchor                     selected departure or arrive-by moment
+     * @param earliestFeasibleDeparture  earliest moment at which travel can begin
+     *
+     * @return immutable anchored schedule
+     */
+    public static Result calculate(Plan itinerary, List<PlanetarySystem> requestedStops,
```
IllianiBird, line 68: This is missing the destination planet, something needed for nuContract as contracts no longer always target the primary planet

MekHQ/src/mekhq/campaign/PiratePointAnalysis.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public static Facts analyze(Input input) {
+        Objects.requireNonNull(input);
+        ApproachFacts standardPoint = analyzeApproach(input.standardDistanceKm(), input.detectionRadiusKm(),
+              input.accelerationG());
+        ApproachFacts piratePoint = analyzeApproach(input.pirateDistanceKm(), input.detectionRadiusKm(),
+              input.accelerationG());
+
+        long modifierTotal = input.modifiers().stream()
```
IllianiBird, line 101: Is a stream really warranted here?

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
```diff
@@ -60,268 +125,1535 @@
...
+    private final List<JLabel> piratePointResultLabels = new ArrayList<>();
+    private ApproachLabels standardPointLabels;
+    private ApproachLabels piratePointLabels;
+    private JLabel piratePointTargetValue;
+    private JLabel piratePointOddsValue;
+    private JLabel piratePointTransitDifferenceLabel;
+    private JLabel piratePointTransitDifferenceValue;
+    private JLabel piratePointAdjustedArrivalValue;
 
     public JumpPathViewPanel(JumpPath p, Campaign c) {
```
IllianiBird, line 218: With he amount of changes you're making in this class, I think removing these single letter param names could be justified.

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
```diff
@@ -60,268 +125,1535 @@
...
+    private <T> void configureComboBox(JComboBox<T> comboBox, String labelKey, String tooltipKey, int width) {
+        comboBox.setToolTipText(resourceMap.getString(tooltipKey));
+        comboBox.getAccessibleContext().setAccessibleName(resourceMap.getString(labelKey));
+        comboBox.getAccessibleContext().setAccessibleDescription(comboBox.getToolTipText());
+        setPlannerControlWidth(comboBox, width);
+    }
+
+    private List<NavigationSource> navigationSources() {
+        List<NavigationSource> sources = new ArrayList<>();
+        for (var unit : campaign.getUnits()) {
```
IllianiBird, line 827: I greatly prefer strong typing

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
```diff
@@ -60,268 +125,1535 @@
...
+        private static String navigationStatusText(LegAssessment assessment, ResourceBundle resources) {
+          NavigationRouteAnalysis.FindingKind primaryFinding = assessment.findings().stream()
+              .filter(finding -> finding.severity() == Severity.BLOCKED)
+              .map(NavigationRouteAnalysis.Finding::kind)
+              .findFirst()
+              .orElseGet(() -> assessment.findings().stream()
+                  .filter(finding -> finding.severity() == Severity.CAUTION)
+                  .map(NavigationRouteAnalysis.Finding::kind)
+                  .findFirst()
+                  .orElse(null));
```
IllianiBird, line 1378: This is a very ugly block of code. If you're needing to do this much processing a stream is not the call

MekHQ/src/mekhq/gui/view/PlanetViewPanel.java
```diff
@@ -77,426 +94,449 @@
...
+        header.setLayout(new GridBagLayout());
+        header.setBorder(BorderFactory.createEmptyBorder(12, HORIZONTAL_PADDING, 10, HORIZONTAL_PADDING));
+        LocalDate currentDate = campaign.getLocalDate();
+        String printableSystemName = system.getPrintableName(currentDate);
+
+        JLabel eyebrow = new JLabel(resourceMap.getString("dossier.eyebrow.text"));
+        eyebrow.setForeground(DOSSIER_ACCENT);
+        eyebrow.setFont(eyebrow.getFont().deriveFont(Font.BOLD, eyebrow.getFont().getSize2D() * 0.85f));
+        GridBagConstraints constraints = createFullWidthConstraints(0);
+        constraints.insets = new Insets(0, 0, 3, 0);
```
IllianiBird, line 227: These inserts aren't scaled for gui

MekHQ/src/mekhq/gui/view/PlanetViewPanel.java
```diff
@@ -77,426 +94,449 @@
...
+        header.add(planetContext, constraints);
+        return header;
+    }
+
+    private JPanel createOperationalSummary(Planet planet, Set<InjuryType> activeDiseases) {
+        JPanel summary = createRevealBand(SUMMARY_REVEAL_DELAY_NS, SUMMARY_REVEAL_DURATION_NS);
+        summary.setLayout(new GridBagLayout());
+        summary.setBorder(BorderFactory.createCompoundBorder(
+              BorderFactory.createMatteBorder(1, 0, 0, 0, DOSSIER_DIVIDER),
+              BorderFactory.createEmptyBorder(9, HORIZONTAL_PADDING, 10, HORIZONTAL_PADDING)));
```
IllianiBird, line 259: Lots of values that aren't scaled in this class

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+        }
+
+        private static final List<MapLegendSection> MAP_LEGEND_SECTIONS = List.of(
+            new MapLegendSection("NAVIGATION", List.of(
+                new MapLegendEntry(MapLegendSymbol.SELECTED_SYSTEM, "Selected system",
+                    "An amber ring identifies the selected system at distant zoom; corner brackets replace it as navigation detail appears."),
+                new MapLegendEntry(MapLegendSymbol.HOVERED_SYSTEM, "Hovered system",
+                    "A cyan ring identifies the system under the pointer at distant zoom; corner brackets replace it closer in. Selected systems suppress hover."),
+                new MapLegendEntry(MapLegendSymbol.CURRENT_FLEET, "Current fleet",
+                    "An amber JumpShip above-right of a system marks the fleet. At distant zoom, an amber ring surrounding the system replaces the ship."),
```
IllianiBird, line 527: Hardcoded player facing strings

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+            contractEmployers = Set.copyOf(contractEmployers);
+            contractTargets = Set.copyOf(contractTargets);
+        }
+
+        static MapPresentationData empty() {
+            return new MapPresentationData(Map.of(), Map.of(), Set.of(), Set.of());
+        }
+    }
+
+        record SystemRenderData(Set<Faction> factions, List<Color> factionColors, List<Faction> capitalFactions,
```
IllianiBird, line 1330: With the tiny number of factions in any given system two streams are not warranted and will affect performance.

IllianiBird, line 1330: If you need to sort the values, consider translating them into a List and then sorting that. If we find ourselves needing to sort a Set we're not using the right class.

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+                            LOGGER.error("", ex);
+                        }
+                        conf.centerX = originalX;
+                        conf.centerY = originalY;
+                        clearRenderLayerCaches();
+                        g.dispose();
+                        mapPanel.repaint();
+                    });
+                    popup.add(item);
+                    JMenu menuGM = new JMenu("GM Mode");
```
IllianiBird, line 2564: Looks like we've picked up some hardcoded player-facing strings

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+                            InterstellarMapPanel.this.campaign.getPlayerForce()
+                                  .getDetachmentLocationManager()
+                                  .moveToPlanetarySystem(InterstellarMapPanel.this.campaign, selectedSystem);
+                                routePlanningHandler.clearPlannedRoute();
+                            center(selectedSystem);
+                        });
+                    }
+                    menuGM.add(item);
+                    /*
+                     * TODO: re-enable this later
```
IllianiBird, line 2580: Is this TODO still valid?

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+                          semanticZoomReference);
+
+                Arc2D.Double arc = new Arc2D.Double();
+
+                // Draw auras around a selected planet
+                if (selectedSystem != null) {
+                    final double x = map2scrX(selectedSystem.getX());
+                    final double y = map2scrY(selectedSystem.getY());
+                    // Contract Search Radius Aura
+                    if (!InterstellarMapPanel.this.campaign.getCampaignOptions().get(CampaignOption.CONTRACT_MARKET_METHOD).isNone()
```
IllianiBird, line 2810: If we're fetching campaign options a lot, cache it at the class level

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+          optHpgNetworkDetail = new ImmersiveComboBox<>(HpgNetworkDetail.values());
+        optHpgNetworkDetail.setSelectedItem(HpgNetworkDetail.CLASS_A_B);
+        optHpgNetworkDetail.setEnabled(false);
+          Dimension hpgDetailSize = new Dimension(UIUtil.scaleForGUI(148),
+              optHpgNetworkDetail.getPreferredSize().height);
+          optHpgNetworkDetail.setPreferredSize(hpgDetailSize);
+          optHpgNetworkDetail.setMinimumSize(hpgDetailSize);
+          optHpgNetworkDetail.setMaximumSize(hpgDetailSize);
+        optHpgNetworkDetail.setToolTipText(
+              "Maximum HPG detail. Wide zoom shows A only; medium zoom shows A-B; detailed zoom shows station badges.");
```
IllianiBird, line 3490: More hardcoded strings. This PR needs a full pass to remove them all, please

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+              ZOOM_BENCHMARK_PATH_ID, phase, benchmarkRun.width(), benchmarkRun.height(), benchmarkRun.date(),
+              benchmarkRun.mapMode(), benchmarkRun.territory(), benchmarkRun.hpgNetwork(),
+              benchmarkRun.operations(), benchmarkRun.reachability(), benchmarkRun.emptySystems(),
+              benchmarkRun.origin().centerX(), benchmarkRun.origin().centerY(), benchmarkRun.origin().scale(),
+              benchmarkRun.anchorX(), benchmarkRun.anchorY(), benchmarkRun.range().atlasScale(),
+              benchmarkRun.range().detailScale(), elapsedMillis, report);
+    }
+
+    private void cancelZoomBenchmark(String reason) {
+        ZoomBenchmarkRun benchmarkRun = zoomBenchmarkRun;
```
IllianiBird, line 4156: Is all this benchmarking code still needed?

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+    }
+
+    private @Nullable String getCurrentSystemId() {
+        PlanetarySystem currentSystem = campaign == null ? null : campaign.getCurrentSystem();
+        return currentSystem == null ? null : currentSystem.getId();
+    }
+
+    static JTabbedPane createMapLegendTabbedPane() {
+        int contentWidth = UIUtil.scaleForGUI(MAP_LEGEND_CONTENT_WIDTH);
+        List<MapLegendSection> orderedSections = MAP_LEGEND_SECTIONS.stream()
```
IllianiBird, line 4320: We definitely do not need an expensive stream to sort this

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
```diff
@@ -99,1004 +143,9313 @@
...
+        button.putClientProperty("navigationUtilityButton", Boolean.TRUE);
+    }
+
+    static JButton createNavigationUtilityIconButton(int symbolCodePoint, int buttonSize, String toolTipText,
+          String accessibleName, String accessibleDescription) {
+        JButton button = new JButton() {
+            @Override
+            protected void paintComponent(Graphics graphics) {
+                Graphics2D buttonGraphics = (Graphics2D) graphics.create();
+                paintNavigationUtilityButton(butto
```
IllianiBird, line 5521: single letter param name

IllianiBird, line 7974: Use `MERCENARY_FACTION_CODE` instead, please

IllianiBird, line 8512: This is another example where we really shouldn't be using a stream

IllianiBird, line 8527: This is the wrong way to fetch a string. We should always be using the i18n method getText, getTextAt, getFormattedText, and getFormattedTextAt

IllianiBird, line 8851: Is there any way we can make this n(1)?

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
IllianiBird, line 1: This class has a lot of problems: player facing hardcoded strings, incorrect RESOURCE_BUNDLE fetching, single letter param or variable names, and a **lot** of overuse of streams. Streams should be reserved for tasks where we can reasonably expect thousands of objects. They should also not be used when we need to do a lot of processing within the stream. I saw instances of us nesting a stream within a stream which is a big red flag.

Furthermore, this class is far too big. I'd really prefer if we broke out methods into focused utilities classes and had this class delegate to them. As it is, this class is too unwieldy and smells like a god class.

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -90,191 +124,1118 @@
...
+        refreshSystemView();
+        applyLayoutState();
+    }
+
+    private JPanel createNavigationHud() {
+        JPanel navigationHud = new JPanel(new GridBagLayout());
+        navigationHud.setBackground(ROUTE_STRIP_BACKGROUND);
+        navigationHud.setBorder(BorderFactory.createCompoundBorder(
+              BorderFactory.createMatteBorder(0, 0, 1, 0, ROUTE_STRIP_BORDER),
+              BorderFactory.createEmptyBorder(4, PADDING, 4, PADDING)));
```
IllianiBird, line 325: Needs scaling

MekHQ/src/mekhq/gui/MapTab.java
```diff
@@ -90,191 +124,1118 @@
...
 
-        JPanel panTopButtons = new JPanel(new GridBagLayout());
-        GridBagConstraints gridBagConstraints = new GridBagConstraints();
-        gridBagConstraints.gridx = 0;
-        gridBagConstraints.gridy = 0;
-        gridBagConstraints.weightx = 0.0;
-        gridBagConstraints.weighty = 0.0;
-        gridBagConstraints.anchor = GridBagConstraints.WEST;
-        panTopButtons.add(new JLabel(resourceMap.getString("lblFindPlanet.text")), gridBagConstraints);
+        JLabel searchLabel = new JLabel(resourceMap.getString("mapHud.systemSearch.text"));
```
IllianiBird, line 327: This is the wrong way to fetch strings

MekHQ/src/mekhq/gui/PlanetarySystemMapPanel.java
```diff
@@ -411,14 +423,9 @@ protected void paintComponent(Graphics g) {
...
 
-        btnBack = new JButton();
-        btnBack.setFocusable(false);
-        btnBack.setPreferredSize(new Dimension(36, 36));
-        btnBack.setMargin(new Insets(0, 0, 0, 0));
-        btnBack.setBorder(BorderFactory.createEmptyBorder());
-        btnBack.setToolTipText("Back to Interstellar Map");
-        btnBack.setBackground(Color.DARK_GRAY);
-        btnBack.setIcon(new ImageIcon("data/images/misc/back_button.png")); // TODO : Remove inline file path
+          btnBack = InterstellarMapPanel.createNavigationUtilityIconButton(0xE5C4, BACK_BUTTON_SIZE,
```
IllianiBird, line 426: Hardcoded player facing strings

MekHQ/src/mekhq/gui/RoutePlanningIntent.java
IllianiBird, line 1: I think this needs to be updated to consider the final stopping point if it's a planet, not a system

IllianiBird (changes requested):
While the overwhelming majority of this PR is perfect there are some significant compliance issues and performance concerns:

- I'm seeing a **lot** of hardcoded player facing strings.
- I'm also seeing a **lot** of fetching Strings using `resourceMap.getString` instead of the modern methods, such as getTextAt.
- There are numerous places where we're hardcoding UI object sizes instead of scaling them

Outside of compliance, I'm very concerned with the overuse of streams. We're using streams in some cases nested within other streams. And others where we can only realistically expect 2-3 objects in the array.

Streams are surprisingly expensive and should be reserved only for instances where we're realistically dealing with thousands of objects and **only** if we're just performing simple tasks. Any time you're performing multiple tasks on the stream, even with large array sizes, it should be a traditional loop.

HammerGS:
Deep review pass over the whole diff, plus a cross-check against MegaMek/mm-data#561. Everything below was verified against the branch at `5410b280`, either by running the real loader or by measuring the real data set. Where I could not reproduce a suspicion I left it out.

This is excellent work and none of it is a reason to hold the feature. Items 1, 2 and 4 do need a decision before merge.

---

## 1. The `Planet` event merge switches on 144 dormant records across the whole universe

`Planet.PlanetPostLoader` changed `planet.events.put(date, event)` to `planet.events.merge(date, event, copyDataFrom)`.

That change is necessary and correct. In the data set, `administration` very often sits in its own event that shares a date with a `faction` event on the same planet. Under `put` the second one wins outright and the first is thrown away. I counted it: **4,864 of the 8,692 administration blocks in mm-data#561 share a date with a later event**, so more than half of that PR is inert without this change.

The problem is that the same merge also revives every *other* field that a same-date pair used to discard. I scanned all 3,186 canon system files on `mm-data/main`: there are 9,751 duplicate-date event groups, and in **144** of them the earlier event carries a planet-level field the later one does not.

| Field | Records revived |
|---|---|
| `capitalType` | 123 (11 National, 57 Region, 55 District) |
| `population` | 13 |
| `message` | 4 |
| `socioIndustrial` | 2 |
| `faction` | 1 |
| `name` | 1 |

Verified by running `PlanetarySystemYamlIO.read` on the real files, once with `merge` and once with `put` restored:

```
Arcturus @3025-01-01   put: capitalType=NONE     merge: capitalType=NATIONAL
Addicks  @3025-01-01   put: capitalType=NONE     merge: capitalType=REGION
Addicks  @3025-01-01   put: admin=[Lockdale Province]
                       merge: admin=[Draconis March, Robinson Operational Area, Region 1]
```

The eleven revived National Capitals are Arcturus, Atreus, Bremen, New Avalon (x2), Rasalhague (x2), Sian, St. Ives and Tharkad (x2). Those all look right, which is the strongest argument that this is a real data-loss bug being fixed rather than a regression.

Why it matters here: this PR also adds capital markers to the map and a capital row to the dossier. So 123 worlds change their capital status the moment this merges, **with or without mm-data#561**, because the fix is in MekHQ and not in the data. The PR body says the map "omits unavailable administrative details" without that data. Capitals are not gated the same way.

Proposed fix: keep the merge. Add it to "What this changes" as a universe-wide data correction in its own right, and have an owner eyeball the 123 capital revivals before merge. A unit test asserting that a same-date `capitalType` plus `faction` pair both survive would lock the behaviour in.

## 2. Administration paths of different depth are treated as a conflict

`classifyAdministrativeKey` requires every evidence system in a cell to have an *identical* path. `buildAdministrativeBoundaries` then emits a DISTRICT boundary whenever two same-region neighbours have different paths.

A shorter path that is a prefix of a longer one does not mean "a different district". It means no district is recorded on that world.

Real example, read through the loader with mm-data#561 data at 3067:

```
Skye           admin=[Skye Province]
Zebebelgenubi  admin=[Skye Province, Freedom Theater]
```

They are 6.1 ly apart. `TERRITORY_HEX_SIZE` is 30 ly, so they share a cell, the paths disagree, `classifyAdministrativeKey` returns null, and the cell holding Skye drops out of the Skye Province outline entirely. Where two such worlds sit in adjacent cells instead you get the opposite symptom: a district line drawn between them that is not a district line.

Across the map I count 124 same-faction, same-region prefix pairs within 30 ly at 3067. Also 17 at 3145, 14 at 2765, 7 at 3025.

Proposed fix: treat a prefix relationship as unknown depth rather than disagreement.

- In `classifyAdministrativeKey`, reduce disagreeing paths to their common prefix instead of returning null, and return null only when the first elements differ.
- In `buildAdministrativeBoundaries`, skip the DISTRICT case when one path is a proper prefix of the other.

## 3. `AccessStatus.BLOCKED` can never be produced

`RouteAlternativesPlanner.createCourse` hardcodes `AccessStatus.CLEAR`, and nothing else constructs a `Course`. Courses come out of `findSegment`, which already drops any leg failing `canTraverse` or `isSystemAllowed`, so CLEAR is a tautology.

`JumpPathViewPanel` reads it at line 433 and renders `course.access.blocked.text` ("ACCESS BLOCKED") plus the accent colour. Neither can ever appear.

Proposed fix: pick one. Either delete `AccessStatus`, the record field and the two resource strings, or set it honestly. `planFastestSegmentWithFallback` already knows when it fell back off the strict policy, so that is the natural place to mark a course BLOCKED.

## 4. "Avoid empty systems" no longer stops you plotting to an empty system

The old `calculateJumpPath` returned an empty path and raised `unableToEnterSystem.abandoned` when the destination had zero population and the option was on. That block is gone and both strings were deleted from `Campaign.properties`. The replacement `isRequestedDestinationAllowed` returns `true` unconditionally.

On the map itself the player still gets feedback, because `assessLeg` raises `ABANDONED_DESTINATION_ALLOWED` at CAUTION. Other callers of `calculateJumpPath` now just get a route.

Is this deliberate? If it is, please say so in the PR body, because it changes what an existing campaign option does. If it is not, restore the destination gate behind `isRequestedDestinationAllowed`.

## 5. `calculateJumpPath` lost its A* heuristic

The A* search (heuristic = straight-line distance to the goal) is now `RouteAlternativesPlanner.findSegment`, which is plain Dijkstra with no heuristic at all.

I measured the expansion count on the real 3,186-system coordinate set, same 30 ly neighbour rule, uniform recharge cost:

| Route | Dijkstra pops | A* pops |
|---|---|---|
| Terra to Tharkad | 1103 | 13 |
| Terra to Luthien | 1369 | 15 |
| Terra to Sian | 698 | 11 |
| New Avalon to Atreus | 1813 | 21 |

Roughly 60x to 90x more node expansions on the path the player hits every time they plot a course. On top of that, each relaxation copies the entire path into a fresh `ArrayList` and builds a new full-path `String` key, and every expansion calls `getNearbySystems`, which sorts by distance, after which `findSegment` sorts the same list again by id.

Proposed fix: put an admissible heuristic back, in the same unit as the edge cost. Edge cost is recharge days, so `ceil(distanceTo(goal) / MAX_JUMP_RADIUS) * minimumRechargeDays` works and never overestimates. Then drop one of the two sorts, and consider a cheaper node key than a growing concatenated string.

## 6. The route planner never sets a destination planet, so the planet-aware code is unreachable

`RoutePlanningIntent.derive` does `derivedPath.setTargetPlanet(segment.getTargetPlanet())`. `segment` comes from `planFastestSegment`, which builds a bare `JumpPath` and never sets a target planet, so that value is always `null`. `adopt` then faithfully preserves the null.

Consequence: `JumpPathItinerary`'s target-planet arrival transit and `JumpPathViewPanel`'s destination-planet naming only ever fire for contract-generated paths. `ContractUtilities` and `AbstractContractGeneration` are the only writers in the tree. A route planned on the map always measures arrival to the primary world, even after the player picks a planet in the system view.

This is the mechanism behind the existing note on `RoutePlanningIntent`. Fix: carry the selected planet into the intent and set it on the final segment, and the two call sites start working.

## Smaller items

- `AdministrativeKey.region()` calls `path.getFirst()`, but the compact constructor does not reject an empty path. One guard makes the record safe on its own.
- `JumpPathItinerary.calculate` decides "no recharge here" with `system.equals(firstSystem)` / `equals(lastSystem)` rather than by index, so a route that revisits its origin or destination as an intermediate hop recharges free there. It mirrors `JumpPath.getTotalRechargeTime`, so it is at least consistent, but index comparison says what is meant.
- Seven files change only by import order, plus a 57-line trailing-whitespace-only edit in `CommandGenerationOptions.hashCode`: `StratConRulesManager`, `FinancesTab`, `CommandGenerationDialog`, `CommandGenerationPane`, `ForceGeneratorTab`, `SetupTab`, `LanceAssignmentView`. That looks like `spotlessApply` fallout rather than intent, and it is what checkbox two is asking about. `./gradlew :MekHQ:spotlessCheck` passes on the branch, so reverting those seven is safe if you want the diff tighter.
- Indentation drifts between 4, 6, 8 and 10 spaces in several new blocks (`JumpPathItinerary.calculate`, the new `Campaign` navigation methods, parts of `InterstellarMapPanel`). Checkstyle does not catch it, but it makes the diff harder to read.
- `classifyTerritoryCell` falls back to every non-empty system within `TERRITORY_HEX_SIZE * 1.3` (39 ly) when the hex itself holds none. That is 2.6 times the hex radius, so an unpopulated hex can adopt the ownership and the administration of a system two hexes away. Worth a look on screen to confirm the bleed distance is what you intended.

## Things I checked and found clean

Listing these so nobody re-treads them.

- `PiratePointAnalysis` exposure maths. Both branches are right: `sqrt(R/2d)` below half the distance and `1 - sqrt((d-R)/2d)` above it match the symmetric accelerate and decelerate profile. `successful2d6Outcomes` is exact.
- `PlanetaryEvent.copyDataFrom` covers every field on the record, so the merge in item 1 loses nothing.
- The outlaw escape exception really is per-leg now. `canEnterTargetSystem` grants passage whenever the leg origin is outlawed, so chains through outlawed space still work without the old whole-route flag.
- `isUseCommandCircuit()` is equivalent to the inline `FactionStandingUtilities.isUseCommandCircuit` call it replaced.
- The background cartography thread only touches an immutable `TerritoryAtlas`. It never reads dated `Planet` state, so the shared mutable accumulator in `CurrentEvents` is not exposed to a race.
- `RenderLayerCache` and `PannableRenderLayerCache` are single-slot, so the alpha bits in the cache keys cannot grow the cache during an animation.
- `CircuitPlan` index conventions line up. `populateCircuitDepartures` uses the system index and `JumpPathItinerary.calculate` queries with the same index, so there is no off-by-one, and `withCoverage` only ever runs against `customCircuitPlan`, so switching away from WHOLE cannot silently clear coverage.
- The `findSegment` edge cost (recharge at the departure system, zero at the origin) sums over exactly the same set of systems as `JumpPath.getTotalRechargeTime`.

VicenteCartas:
Thanks for the thorough pass and the real-data measurements. I have worked through these items locally; the follow-up commits have **not been pushed yet**, so the PR head is still `5410b280`.

1. **Same-date planet events:** Kept the merge and added a loader regression proving that a same-date `capitalType` plus `faction` pair both survive. I also updated the PR description to disclose the universe-wide correction and the 144 revived field values. I am treating your review of the eleven revived National Capital results as the requested owner eyeball.
2. **Administration path depth:** `classifyAdministrativeKey` now reduces compatible evidence to its common prefix, while different first-level regions still conflict. Boundary generation skips district lines when one path is a proper prefix of the other. Both cases have regression coverage.
3. **Dead `AccessStatus`:** Removed the enum, `Course` field, unreachable blocked-course rendering, and its two resource strings. Planning failures remain represented by `PlanningStatus` and the route analysis.
4. **Empty destinations:** This is deliberate. Explicitly requested empty destinations and waypoints are honored with a caution, while empty systems encountered only as intermediate candidates remain excluded. The PR description now says so explicitly.
5. **A* regression:** Restored A* with objective-specific, unit-consistent admissible heuristics: `ceil(distance / 30 ly)` for fewest jumps, and minimum remaining recharge stops times the proven 10-hour lower bound for fastest routes, accounting for the zero-cost segment origin. Policies without a proven lower bound default to zero. The search now uses predecessor nodes instead of copying complete paths per relaxation, and production neighbor lookup no longer performs a distance sort before the planner's stable ID sort. Added a focused expansion regression.
6. **Destination planet:** Already fixed in the preceding local review commit. The selected planet is carried through `RoutePlanningIntent`, set on the final path, and preserved when adopting alternatives or editing waypoints.

Smaller items are also addressed: `AdministrativeKey` rejects empty paths; recharge endpoint exemptions now use route indexes in both `JumpPath` and `JumpPathItinerary`, with a repeated-endpoint regression; and the eight import/whitespace-only files now match the PR base byte-for-byte. I also normalized indentation in the newly touched blocks.

I left the 39 ly territory fallback unchanged. It is visual tuning rather than a demonstrated correctness defect, so I would prefer to assess it on screen rather than change the ownership envelope from the data audit alone.

Current local validation is clean for production and test-source compilation, diagnostics, localization-key checks, and the PR-base whitespace check. Per the project workflow, I am holding Checkstyle and unit tests for the final unchanged patch immediately before push.

MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
VicenteCartas: Updated. Just curious, why the preference?

MekHQ/src/mekhq/campaign/JumpPathSchedule.java
VicenteCartas: Nice catch! I have added support for a target planet.

MekHQ/src/mekhq/campaign/PiratePointAnalysis.java
VicenteCartas: Replaced it by a regular loop.

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
VicenteCartas: Done

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
VicenteCartas: Fixed

MekHQ/src/mekhq/gui/view/JumpPathViewPanel.java
VicenteCartas: Replaced the nested streams with a single-pass loop in LegAssessment.primaryFindingKind()

MekHQ/src/mekhq/gui/view/PlanetViewPanel.java
VicenteCartas: Fixed

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
VicenteCartas: I wanted it for the nightlies, in case people have issues they could run the benchmarks and give me data. But I have removed it and if there are issues I can always add it back again for a build.

IllianiBird (approved):

MekHQ/src/mekhq/campaign/universe/PlanetarySystem.java
IllianiBird: Answered in Discord

MekHQ/src/mekhq/gui/view/PlanetViewPanel.java
VicenteCartas: Fixed

MekHQ/src/mekhq/gui/InterstellarMapPanel.java
VicenteCartas: Fixed

MekHQ/src/mekhq/gui/RoutePlanningIntent.java
VicenteCartas: Fixed

PR #10049, opened 2026-09-13
Fix #10040: Fixed Role Validator to Include Role Names & Instructions to Resolve Issue

### LINKED ISSUE #10040: repeated error message about pesrson was found to be ineligible for their Primary role
### What went wrong

Error "_WARNING: [PERSON] was found to be ineligible for their Primary role. That role has been removed, and they were assigned the DEPENDENT role._" appears on every save/load on the the same campaign, even after progressing over time. Notably, this happens only to inactive personnel 

### What you expected instead

I would expect this pass to happen once, and not persist with the campaign for further saves down the line.

### Steps to reproduce

1. Load campaign.
2. Advance a day/week/month
3. save again
4. load new save, observe the message

### Attach your files

[DWC30640908.cpnx.gz](https://github.com/user-attachments/files/32147972/DWC30640908.cpnx.gz)

### Relevant log lines

```text

```

### MegaMek Suite Version

0.51.01-nightly-2026-09-12

### Operating System

Windows 11 10.0 (amd64)

### Java Version

Eclipse Adoptium 25.0.2

### Before you submit

IllianiBird:
Fix #10040

## What this changes

Fixed the role validator so that it now correctly reports the right profession and what next steps the player needs to take.

## Testing

- Manual testing

## What is not proven yet

- What the original profession was that set the character on this loop

---


HammerGS (approved):

IllianiBird:
Copilot is a non-issue and everything is working as intended

PR #10057, opened 2026-09-14
Contract Overhaul Project: Implemented Piracy Contracts

IllianiBird:
## What this changes

This PR implements piracy raid contracts so pirate campaigns are supported. It also refactors some file structure in the mission package to ensure we're taking advantage of AbstractContract's abstract nature.

Of special note I've opted to remove acts of piracy options from non-pirate campaigns. I know some players were looking forward to doing piracy against the Clans as a pirate to circumvent the Clan tech limitations. I opted to limit acts of piracy to pirate campaigns to retain their unique gameplay. I will be reviewing other ways to allow players to circumvent Clan tech limitations later today.

## Testing

- Manual testing

---


IllianiBird:
Tests haven't been updated or added at this time, so this should fail automated testing and that should not be reviewed.

IllianiBird:
Flipping to live

HammerGS (approved):

Issue #8962, opened 2026-05-13
[Issue] Text Color for Fixes unable to change


heckindoc:
### Brief Description *

Discussed briefly on Discord -- some text within MekHQ cannot have the color changed within settings and doesn't update when themes are changed. Example images are below, previous discussion from several months ago will be linked below that and the most recent conversation below that as well. 

<img width="699" height="581" alt="Image" src="https://github.com/user-attachments/assets/25274bbb-afda-48d0-bd43-5689a192558e" />
<img width="656" height="599" alt="Image" src="https://github.com/user-attachments/assets/4df73526-32c5-4d74-a968-a445354a633d" />

https://discord.com/channels/458705327911731231/458706545379639296/1486413330359779489

https://discord.com/channels/458705327911731231/475119845646532628/1504245859833876592

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MekHQ Suite Version *

51.0 Nightly 5/12

### Operating System *

Win 11

### Java Version *

JDK21

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have filled and provided all necessary information above
- [x] I have included any and all logs, custom units, and CPNX (save) files
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
Can you double-check something for me, go into MegaMek client settings and change the warning color. Lemme know if that also changes this (this code lives in MegaMek so is handled there).

heckindoc:
Can confirm that it changes it. That's not super intuitive though that it needs to be changed there, though I know it's likely easier said than done to make it more intuitive. 

<img width="363" height="265" alt="Image" src="https://github.com/user-attachments/assets/6615e3df-5c1c-4dab-934f-c12e1139a8b0" />
<img width="332" height="247" alt="Image" src="https://github.com/user-attachments/assets/0823636f-a08c-4cd6-aed1-38e9e680efc4" />
<img width="352" height="283" alt="Image" src="https://github.com/user-attachments/assets/922e1206-27a9-4ae9-ba6b-daf41c8f2894" />

PR #10069, opened 2026-09-15
Fix MegaMek/megamek#8989: PACAR auto resolve no longer sits at Preparing

### LINKED ISSUE #8989: PACAR Autoresolve won't start
### What went wrong

When using the auto resolve for a MekHQ scenario the PACAR window opens, click Ready and it gets stuck preparing. 

Image

### What you expected instead

The Auto resolve should start going through turns

### Steps to reproduce

1. Create a new veteran campaign
2. Create a new company
3. Use GM to create a contract
4. Use GM to create a new scenario
5. Use the auto resolve

### Attach your files

[Rust & Ruin30490307_Tip Toe-part1.zip](https://github.com/user-attachments/files/32263880/Rust.Ruin30490307_Tip.Toe-part1.zip)

### Relevant log lines

```text

```

### MegaMek Suite Version

0.51.01-nightly-2026-09-15

### Operating System

Windows 11 10.0 (amd64)

### Java Version

Eclipse Adoptium 21.0.12

### Before you submit

HammerGS:
## What this changes

The Princess auto resolve (PACAR) starts again instead of sitting at "Preparing...". Since MegaMek #8962 the server refuses a bot that asks to take a human's units, and MekHQ sent that request from the bot; it now sends it from the player's own client, which the server allows.

Fixes MegaMek/megamek#8989

## Testing

- Three new unit tests check the hand-over leaves on the player's client with only the player's units.
- With the sender switched back to the bot, the two sender tests fail.
- MekHQ compile, checkstyle and spotless green.
- Played the reporter's own save against current MegaMek main: all eight units handed over, scenario ran to victory.

## What is not proven yet

- The playtest ran before the rebase onto main; the two commits merged since touch only command generation.

---


IllianiBird (approved):

HammerGS:
The checkstyle failure was Spotless flagging NameAndPortraitGenerationPage, which this PR does not touch: #10062 changed that file on main after this branch was cut, so the ratchet compared the older copy. Merged main in (6b14ed1b8c) so the file matches; no change to the fix itself.

PR #10074, opened 2026-09-16
Contract Overhaul Project: Rewrote StratCon Force Deployment Dialog

IllianiBird:
## What this changes

This PR completely overhauls the supporting code and GUI of the StratCon deployment GUI. The goal is to modernize the experience and resolve multiple sources of frequent user contact.

image

## Testing

- Manual testing
- AI unit tests

---


HammerGS:
# Review of #10074 (StratCon deployment wizard rewrite)

Checked out the PR head (`9090b94`) locally, compiled main and test sources, and ran `DeploymentEvaluatorTest` plus `spotlessCheck` - all green on Windows. I read the new wizard, inspector, service and evaluator side by side with the deleted `StratConScenarioWizard` / `TrackForceAssignmentUI` to check that each old rule survived the port. The split into a GUI-free service plus a pure evaluator is a real improvement, the batchall / singles / official-challenge paths are ported faithfully, and the off-board plumbing (scenario field, XML, game-thread and auto-resolve entry points, client option) is complete end to end.

Everything below is what I found on top of that, ordered by severity. I have not playtested the branch, so the first section is "what the code does as written" - please treat each as something to confirm rather than a verdict.

## Likely bugs (please verify before merge)

- [x] **`isUnlocked` returns the opposite of its name, which disables Unstage and hides the off-board checkbox for every force the player can actually pick.** `StratConDeploymentWizard.java:916` returns `true` when the item *is* in the deployed (locked) lists. Two callers trust the name: line 700 arms the Unstage button with `isUnlocked(focused)`, so a force you just staged cannot be unstaged (button greyed) while an already-committed force shows an armed Unstage that does nothing; and line 757 gates the off-board checkbox with `&& isUnlocked(item)`, so the checkbox never appears for a board or newly staged force - only for locked ones, whose toggle `applyOffBoardSelections` (line 1086) never reads. Net effect: a player cannot undo a stage, and the off-board choice is only reachable through the client-option default. Fix: invert the body (or rename it `isLocked` and flip both call sites), and don't show the checkbox for locked forces at all since their state is already committed.

- [x] **The wizard is non-modal and reads the target hex at commit time, so clicking another hex on the map before Commit deploys to the wrong hex.** The old `TrackForceAssignmentUI` was modal. The new dialog is created with `super(null, false)` (line 179) and `deployPrimaryForces` calls `owner.getSelectedCoords()` / `owner.getCurrentTrack()` at commit (lines 1112-1113). Steps: right-click hex 0304 > Manage Deployment, stage a lance, click hex 0507 on the map behind the wizard, Commit. The lance lands on 0507. Advancing the day with the wizard open has the same shape (the `scenario` reference goes stale). Fix: capture coords, track and scenario into fields in `display()`, or make the dialog modal again.

- [x] **Reinforcements can be rolled and committed before the primary force is, and then get swept into the primary list.** On an unresolved scenario the Reinforce tab is open (only Auxiliaries/Utility are locked, line 1013), and `commit()` runs `processReinforcementBatch` (line 1048) *before* `deployPrimaryForces`. `commitReinforcementForces` calls `scenario.addForce(formation, templateId)`, which registers the force under a template; then `assignForceToScenario` -> `StratConRulesManager.commitPrimaryForces` -> `StratConScenario.commitPrimaryForces()` copies *every* templated force into `primaryForceIDs`. So reinforcements that already rolled and already cost support points become primary forces, are present when `generateOpFor` sizes the enemy, and sit in the delayed-arrival list at the same time. The old flow could not do this because reinforcements were only reachable at `PRIMARY_FORCES_COMMITTED`. Fix: lock the Reinforce tab while `assignToScenario` is true (mirrors the old gate), or commit the primary first and only then roll reinforcements.

- [x] **Leadership and defensive budgets are no longer enforced.** The old wizard disabled Commit when selected BV exceeded the leadership budget or the selected infantry count exceeded the defensive points. `refreshCommitEnabled` (line 1001) only checks that something is staged and a primary exists, and `stage()` never looks at the budget. `leadershipBattleValueRemaining` (line 987) simply goes negative and `minefieldsRemaining` floors at 0, so a Leadership 1 commander (500 BV) can stage three 400-BV units, and a 2-defensive-point scenario can take six utility units. Fix: refuse to stage an item that would exceed the remaining budget (and/or disarm Commit while over).

- [x] **A staged primary force's own units can be staged again as auxiliaries or utility units.** `loadEligibleAuxiliaryUnits` / `loadEligibleUtilityUnits` (lines 518-532) filter through `isForceEligible`, which only rejects units of a *deployed* formation - the staged primary isn't deployed until commit, so its units are offered on the unit pages. On commit the primary deploys first, then `addAuxiliaryUnits` -> `scenario.addUnit` -> `containsPlayerUnit`, which checks only the individual-unit list, so the unit appears to be added a second time as an individual with its BV charged to leadership. The old flow avoided this because leadership units were only picked after the primary was committed. Fix: exclude units belonging to any staged formation when loading the unit boards, and skip them in the commit as a guard.

- [x] **Auxiliary eligibility is computed against the scenario's generated primary, not the staged one.** `getEligibleLeadershipUnits` derives the primary unit type from `scenario.getPrimaryForceIDs()`. On an unresolved scenario those are the generator's suggestion (or empty), so the Auxiliaries board can be empty or filtered for the wrong unit type even though `refreshBudgetInputs` correctly derives the *budget* from the staged primary. Given this and the two items above, the simplest consistent rule is the old one: Auxiliaries/Utility (and Reinforce) only once the primary is committed on the scenario, i.e. lock those tabs while `assignToScenario` is true, and after a fresh primary commit re-open the wizard on the Reinforce page the way the old code did automatically.

## Please double-check (behaviour changes and accuracy)

- [x] **GM reinforcements now breach an accepted batchall.** Old `REINFORCE_GM` / `REINFORCE_GM_INSTANTLY` called `btnCommitClicked(0, true, ...)` with no breach processing; new lines 1193-1206 pass `brokeBatchallTerms` through to `processBatchallBreach`. Arguably the old behaviour was the bug (the warning dialog was shown either way), but it changes what a GM override costs, so worth a line in the PR body if intended.

- [x] **Displayed odds ignore the natural-2 auto-fail and the improved roll.** `successProbabilityTwoD6` (`DeploymentEvaluator.java:78`) returns 1.0 for TN <= 2 and counts a roll of 2 as a success at every TN, but `processReinforcementDeployment` treats a 2 as a critical failure regardless of TN, so every shown percentage is 1/36 (about 3%) too high and a TN of 2 shows 100% rather than 97%. Conversely, a force whose eligibility reads "Improved Reinforcement Roll" (best of two 2d6) is shown the plain single-roll odds. Fix: subtract the 1/36 in the evaluator and add a best-of-two flag for AUXILIARY eligibility (P = 1 - (1-p)^2).

- [x] **"Average arrival: turn N" is the best case, not an average.** `estimateReinforcementArrival` (line 726) mirrors only the non-delayed branch of `setDeploymentTurnsForReinforcements`; a DELAYED result uses 2-3x the arrival scale. Suggest "Arrival if the roll succeeds: turn N" (`AtBStratCon.properties:128`). It also includes transported units, which the real calculation skips.

- [x] **The off-board tooltip contradicts itself.** `GUI.properties:1062` says "Only artillery units are placed off-board; a force cannot be split between on- and off-board deployment" - but `applyPlayerOffBoardDeployment` does exactly that split (artillery off, everything else on). Drop the second clause.

- [x] **Unit test coverage stops at the arithmetic.** `DeploymentEvaluatorTest` is good, but nothing covers the pieces that actually change campaign state: the `offBoardForceIDs` XML round-trip in `AtBDynamicScenario`, and `StratConDeploymentService.commitReinforcementForces` (failed-roll bookkeeping, delayed vs instant lists). Both are testable without a GUI.

## Minor / take or leave

- [x] `processReinforcementBatch` (lines 1147-1155) recomputes the command liaison and target number inline; `reinforcementAdvisor` was built in `display()` for exactly that and is non-null here. The confirmation dialog wants the `TargetRoll`, so consider exposing that from the advisor rather than duplicating the lookup.
- [x] `AtBDynamicScenario.java:717` parses `offBoardForceIDs` with `Integer.parseInt`; a hand-edited or corrupted save throws `NumberFormatException` and aborts the scenario load. Log and skip the bad value instead. `setOffBoardForceIDs(null)` (line 216) would NPE later; a null guard or `Set.copyOf` is cheap.
- [x] Dead code left behind: `ScenarioWizardLanceRenderer` and `ScenarioWizardUnitRenderer` in `mekhq/gui/stratCon` no longer have any reference (only `ScenarioWizardLanceModel` is still used). Also unused: `invalidReinforcementType.error` (`AtBStratCon.properties:34`), `HudStyle.statTile(...)` and `HudStyle.INELIGIBLE`.
- [x] `DeploymentInspectorPanel.java:486-497` has an orphaned Javadoc ("Replaces the staged-forces tray...") immediately followed by a second Javadoc on `setStagedOffBoardForceIds`; lines 377-384 are indented one level too deep.
- [x] `HudButton` and the `HudModeSelector` cells are focusable and fire on Space/Enter but paint no focus indicator, so a keyboard user cannot see which control will fire.

## Checked and found clean

Listing these so nobody re-treads them.

- Reinforcement cost and modifier arithmetic (`chosen + 1` / `+ 2` base, `chosen * -2 / forceCount`) matches the old wizard exactly, including the per-force division.
- Leadership budget cap (`min(skill, 5) * 500`), official-challenge zeroing and the defensive-assignment gate match `getEligibleLeadershipUnits` and `commitPrimaryForces`; `defensivePoints = tactics * 2` matches `StratConRulesManager:2431`.
- Minefield count: `commitPrimaryForces` already sets it to `tactics * 2`, so only writing it from the Utility page (rather than on every commit as before) is safe.
- `finalizeForceDeployment` being skipped on a primary-only commit is fine: `assignForceToScenario` / `deployForceToCoords` already run `processForceDeployment`, and objective translation happens inside `generateOpFor`.
- Batchall warning loop, nag handling, singles-mode path, official-challenge notice and the "staying home" summary are ported faithfully.
- `commitReinforcementForces`' failed / delayed / instant handling and `SupportCarrierDeployment.staysHome` filtering are line-for-line with the old `btnCommitClicked`.
- Off-board: `entityHasArtillery` uses `CLASS_ARTILLERY`, `applyPlayerOffBoardDeployment` walks parent formations, `removeFormation` clears the ID, and both `AtBGameThread` and `StratConSetupForces` apply it.
- `MHQOptions` / `MHQOptionsModel` / `MHQAdvancedPage` wiring for the new client option is complete.
- The diff is clean: every changed file carries a real change, no whitespace-only edits.

The first section is the important one. The `isUnlocked` inversion and the non-modal hex capture are the two I would definitely want resolved (or shown to be misreadings) before merge; the tab-gating items three to six share one fix.

HammerGS (approved):

PR #10077, opened 2026-09-16
Fix #10063: Updated Shell Scripts for Linux to Correctly Pull User Directory and Execute

### LINKED ISSUE #10063: [Nightly 2026-09-15] Unable to start MekHQ
### What went wrong

Clicking on MekQH.sh opens file in vim, displaying the file contents, instead of running as executable.

### What you expected instead

Clicking on MekHQ.sh should open MekHQ as an application, as it does with the main release.

### Steps to reproduce

Double-click MekHQ.sh

### Attach your files

_No response_

### Relevant log lines

```text
No log file generated
```

### MegaMek Suite Version

0.51.01 Nightly 2026-09-5

### Operating System

CachyOS 7.2.4-3 (64-bit)

### Java Version

21.0.12

### Before you submit

- [ ] I have asked about this on the MegaMek Discord (optional)

IllianiBird:
Fix #10063

## What this changes

The old scripts assumed the process's working directory was already the install root. It isn't, when you double-click or "Run in Konsole" — the cwd is set elsewhere, and your Games/MegaMek Suite/… path made it worse: the unquoted space split into a nonexistent …/Games/MegaMek, so Konsole couldn't cd there, warned, and fell back to fish before the app ever launched. The new version derives its directory from $0, quotes everything (spaces are safe), and cds in — which also matters because MekHQ reads data/, logs/, and userdata/ relative to the working directory.

## Testing

- No testing has occurred 

## What is not proven yet

- Whether this works

---


IllianiBird:
Confirmed not working

PR #10081, opened 2026-09-16
First pass at adding game and thread cleanup; not entirely successful

Sleet01:
NOT FOR MERGING
@IllianiBird Apologies for the late update.

This PR is for tracking work related to debugging MegaMek/megamek#8306
My work consisted of making a couple small changes and then attempting to show drop in memory leaks between the stock code and the updated code; progress was small.

All files used in testing, as well as profiling data generated, are in the below gzipped tar file:
[Warheads_Wrath_saves_logs_jfr_heap_backups.tar.gz](https://github.com/user-attachments/files/32300801/Warheads_Wrath_saves_logs_jfr_heap_backups.tar.gz)


IllianiBird:
Champion. I should have a profiler soon, so once there is a break in my schedule I can take a proper look.

PR #10083, opened 2026-09-16
Fix #10082: Command Generator gave unit commanders a name from the wrong gender list

### LINKED ISSUE #10082: Females can generate with male given names and vice versa
### What went wrong

When generating a new force, sometimes female personnel is generated with a male given name, and male personnel is generated with a female given name.

For example a Bill Schindler (Female) and Teresa Corbucca (Male) were generated, but there were multiple other examples.

I did some searching and Bill does not occur in data\names\femaleGivenNames.csv nor Teresa in maleGivenNames.csv

Also, according to AI, there may be a race condition involved:

The populateNames() method runs in a separate background thread (Random Name Generator name initializer) to prevent the game from freezing during loading. However, the variables femaleGivenNames, maleGivenNames, and surnames are defined as static, whereas the generate method is not. If the game already attempts to generate characters while the background thread is still busy filling the maps (since initialized = true is only set at the very end), a timing error can occur. Although there is an if (initialized) check inside the methods, the initialized variable does not provide full thread-safe protection during the writing of the data itself. If another thread accesses the object references at the exact moment the maps are being overwritten, the wrong map (for example, maleGivenNames instead of femaleGivenNames) could be read.

Update: I have also seen this with a command generated for the Federated Suns now. One thought. At the start of a new campaign,  there will already be personnel generated for recruitment. Could it be that this can lead to this issue?

### What you expected instead

Females should only get given names from the female list, and males from the male list.

### Steps to reproduce

Start a new campaign. I used the FWL as the faction, I'm not sure if this matters.
Generate a new force.
Examine the names of the generated personnel.

### Attach your files

_No response_

### Relevant log lines

```text

```

### MegaMek Suite Version

v0.51.01

### Operating System

Windows 11

### Java Version

Eclipse Adoptium 21.0.10

### Before you submit

HammerGS:
## What this changes

In the Command Generator, a unit commander now gets the gender that matches the name it inherits from MegaMek's force engine. Before, MekHQ rolled its own gender for that seat and kept it, so about half the commanders came out as a female Bill or a male Teresa.

Fixes #10082

## Testing

- New CrewDescriptorAdapterGenderTest covers both mismatch directions, a match, non-binary, and a genderless descriptor.
- Both CrewDescriptorAdapter test classes pass. Spotless check green.
- Played in game: generated a force and checked the log. 21 of 36 commanders were corrected, every one to the list the name belongs to.

## What is not proven yet

- Nothing beyond the above. Non-commanders keep their own MekHQ name and gender and were never affected.

---


IllianiBird (approved):
Oh this is amusing

Issue #10088, opened 2026-09-18
Size and select support vehicles by the force they support


HammerGS:
### What kind of request is this?

Improvement to an existing feature

## Problem

A command's free support vehicles are neither sized to the force nor chosen from the units available to it.

Recovery vehicles and logistics trucks use a flat count: one per formation base size, so four for any Inner Sphere command and ten for any Clan one. A lance and a battalion get the same four trucks. MASH trucks and canteens do scale, but MASH is sized by how many patients the command might have while being crewed from the medical section, which holds only doctors, so the count and the crew disagree.

Every capability also fields one hardcoded model. Every command in every era gets the same 50-ton BattleMech Recovery Vehicle and the same Flatbed Truck, while the Heavy BattleMech Recovery Vehicle, the Burro II, the GESV trailer and the rest are never used.

## Example

A Mek battalion, 36 Meks of roughly 1,980 tons:

- Logistics: the resupply code already knows this force needs about 63 tons of cargo hauled by its own convoy (`calculateTargetCargoTonnage` gives ~16 tons, and `CARGO_MULTIPLIER` of 4 applies for player convoys). A Flatbed Truck carries 6 tons, so that is about eleven trucks. The command is given four.
- Salvage: the opfor is built to an equal BV budget, so a company fights about a company. A quarter of twelve wrecks is three recoveries per scenario, which for a battalion is about nine recovery vehicles. The command is given four, all of them the lightest model.
- Medical: with three doctors and a two-truck target, the command gets one MASH truck crewed three of six, because the truck needs six crew and medics are a pool rather than people.

Capacity varies enormously between candidates, which is why fixed counts cannot work: a Skoda Growler carries 0.13 tons, a Flatbed Truck 6, a GESV Trailer 11.5, a Burro II 12.5. A MASH Truck (Small) carries one theatre; the GESV (MASH) carries six.

## Proposed fix

- Size each capability by what the force actually needs, computed once when the command is generated or when the option is switched on: cargo tonnage for logistics, expected wrecks for salvage, patients for medical, personnel for the commissary.
- Add up the real capacity of the units chosen rather than assuming a fixed capacity per vehicle.
- Select by capability rather than by name: the new `RECOVERY` role for salvage, `F_MASH` for medical, `F_FIELD_KITCHEN` for the commissary, and `cargo, support` with a minimum cargo capacity for logistics so utility runabouts are not fielded as a supply convoy.
[...]

PR #10090, opened 2026-09-18
Fix #10002: Open the Command Generator on a company

### LINKED ISSUE #10002: Command Generator Feedback
### What went wrong

- Default force size is set to Regiment, it should be Company
- Logistics (cargo vehicles) does not scale with the generated force size, resulting in 4 flatbeds being generated to support an entire regiment.

### What you expected instead

Cargo vehicles should probably be 1 lance per company

HammerGS:
## What this changes

The Command Generator now opens on a company, or a trinary for a Clan command, instead of whatever size the faction's ruleset gives as its default.

Those defaults are large: a regiment for the Draconis Combine, the Capellan Confederation and the Free Worlds League, a division for the Federated Suns and for a Lyran command after 3040, and a cluster for the Clans. A command is designed from the bottom up, so the dialog opened on a size most players immediately changed.

This only decides the opening selection. A player who picks another size keeps it, a faction that fields neither a company nor a trinary keeps its own default, and no ruleset data is changed. The selection also follows the faction if it is changed inside the dialog.

Logistics scaling, the other half of the issue, is not addressed here. It needs the support vehicle counts to be sized by the force, which is #10088.

Fixes #10002

## Testing

- `compileJava`, `compileTestJava`, `spotlessCheck`, `checkstyleMain` and `javadoc` are clean.
- The whole `commandGeneration` test package passes, over twenty classes with no failures.
- The echelon ladders are numbered alike in `data/forcegenerator/faction_rules/constants.txt`, which is what lets one preference serve both: COMPANY 4 and TRINARY 5 for the Inner Sphere and the Clans respectively.

## What is not proven yet

- Not yet seen on screen. The dialog opening on a company for an Inner Sphere campaign, on a trinary for a Clan one, and the selection following a faction changed inside the dialog, have all been read in the code but not watched in game.
- No unit test. The change sets a dialog's opening selection and touches no campaign state, so a test would only restate the constant.

---

- [x] Tests added or updated, if this implements a rule or changes campaign state - not applicable, this sets a dialog default and changes no campaign state


HammerGS:
Back to draft: this cannot work until MegaMek/megamek#9014 merges.

`setPreferredEchelon` stored the preference and nothing applied it. The Force Generator view fills its formation combo during construction, so the faction ruleset's default is already selected before MekHQ receives the view, and every later refresh kept that selection. This PR's call is made - it logs `formation combo opens on echelon 4 (clan=false)` - and the view discarded it, so the dropdown still opened on Regiment in testing.

MegaMek/megamek#9014 fixes that (filed as MegaMek/megamek#9013). Once it merges this PR works as written, with no changes needed here.

HammerGS:
Out of draft. Verified in game with MegaMek/megamek#9014 applied: the dropdown opens on a company, and the faction tooltip shows the accuracy caveat.

Still needs MegaMek/megamek#9014 to merge first, otherwise this is a no-op.

IllianiBird (approved):

